PR #7, opened 2019-08-29
Get the project to build again with latest ROCm stack

jzuniga-amd:
Do basic corrections to get the code to compile
    - remove non-existing rocblas-types references
    - correct signature of template functions
    - add .gitignore file
Update/correct CMake files to build the project


justxi:
Hi, I am glad to see that rocsolver is not dead. I build rocsolver master branch with the patches from 29th of july. On my system (GCC 9.1) there are a few further warnings, except the hardcoded path to rocBLAS it builds. Is it possible to use cmake to detect rocBLAS on the system and if that fails, to download it? (It would help to package rocsolver for e.g. Gentoo, because it does not except downloads while creating the configuration files)

amdkila:
hi @justxi , that should be doable. I am not sure how to do that in CMake itself off the top of my head, but I know other projects have install scripts that download dependent packages if the user passes a flag. Maybe something similar can be done with rocSOLVER- I'll get back to you on that.

justxi:
You could do it like this:
https://github.com/ROCmSoftwarePlatform/rocRAND/blob/6a27e931c7c2a8197eadf22be495ed6e7eee6d72/cmake/Dependencies.cmake#L96

Try to find rocBLAS installed on the system and if not installed download it.

amdkila:
@justxi I haven't had the chance to add this to rocSOLVER yet as I've been busy with other tasks. Thanks for the link though- I'll get around to working on this during the weekend, if not early next week.

PR #11, opened 2019-09-16
split Jenkins job into 2

amdkila:
- Testing this in CI- if it works, wget race condition should be avoided


amdkila:
Not needed- issue is networking-related, and there is a workaround

PR #32, opened 2019-11-27
fix bit manipulation on a 16-bit float

scchan:
This is causing  compilation error with a newer compiler.

### DIFF
--- a/clients/common/arg_check.cpp
+++ b/clients/common/arg_check.cpp
@@ -4,6 +4,7 @@
 
 #include "arg_check.h"
 #include "rocblas.h"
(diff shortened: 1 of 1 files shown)


jzuniga-amd:
We are refactoring rocSOLVER so that it can re-use rocBLAS common code including the clients/gtest infrastructure... it is likely that arg_check.cpp won't be used anymore. I suggest we keep this PR open until the refactoring is completed. Then we can decide where to apply the modification, if still required.

jzuniga-amd:
change outdated after library refactoring

PR #41, opened 2020-01-05
rocSOLVER - Changes to support multiple ROCM installation

pruthvistony:
- Adding rocm_set_soversion to set versioning to the lib so file.

Signed-off-by: Pruthvi Madugundu <mpruthvi@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ get_os_id(OS_ID)
 message (STATUS "OS detected is ${OS_ID}")
 
 # Versioning via rocm-cmake
-set ( VERSION_STRING "2.7.0" )
+set ( VERSION_STRING "2.7" )
 rocm_setup_version( VERSION ${VERSION_STRING} )
 
 # Append our library helper cmake path and the cmake path for hip (for convenience)
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -95,6 +95,7 @@ target_include_directories( rocsolver
 
 set_target_properties( rocsolver PROPERTIES VERSION ${rocsolver_VERSION} SOVERSION ${rocsolver_SOVERSION} CXX_EXTENSIONS NO )
 set_target_properties( rocsolver PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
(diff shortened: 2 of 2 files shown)


library/src/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ target_include_directories( rocsolver
 
 set_target_properties( rocsolver PROPERTIES VERSION ${rocsolver_VERSION} SOVERSION ${rocsolver_SOVERSION} CXX_EXTENSIONS NO )
 set_target_properties( rocsolver PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
+rocm_set_soversion(rocsolver ${VERSION_STRING})
```
jzuniga-amd: Just a little modif here:
what we need is to set the .so version; and if using rocm_set_soversion, let's remove it from the set_target_properies statement, so:

set_target_properties( rocsolver PROPERTIES CXX_EXTENSIONS NO )
set_target_properties( rocsolver PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
rocm_set_soversion(rocsolver ${rocsolver_SOVERSION})

pruthvistony: The variable - rocsolver_SOVERSION value is 0. 
For rocm_set_soversion function need to send major.minor version string.

jzuniga-amd: Oh, right! It should be 
set( rocsolver_SOVERSION 0.1 )
in library/CMakeLists.txt

CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ get_os_id(OS_ID)
 message (STATUS "OS detected is ${OS_ID}")
 
 # Versioning via rocm-cmake
-set ( VERSION_STRING "2.7.0" )
```
saadrahim: Library versioning should be maintained by project owner @jzuniga-amd

library/src/CMakeLists.txt
saadrahim: SOVERSION should be 0.1 or 1.0 not 2.7. I prefer 1.0 but @jzuniga-amd  should decide

saadrahim:
Implemented in a separate PR

PR #56, opened 2020-03-05
Change hip target to be compatible with hip-clang

amdkila:


clients/benchmarks/CMakeLists.txt
```diff
@@ -123,7 +123,7 @@ endif()
 target_link_libraries( rocsolver-bench PRIVATE roc::rocblas ) #${ROCBLAS_LIBRARY})
 
 
-get_target_property( HIPHCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
+get_target_property( HIPLOCATION hip::device IMPORTED_LOCATION_RELEASE )
```
aaronenyeshi: I think this one might not work, the get_target_property didn't work for rocBLAS. I think you might just have to remove this line, and use hip::device when linking below.

library/src/CMakeLists.txt
```diff
@@ -86,10 +86,10 @@ add_library( rocsolver
 add_library( roc::rocsolver ALIAS rocsolver )
 
 # Remove this check when we no longer build with older rocm stack(ie < 1.8.2)
-if(TARGET hip::device)
-target_link_libraries( rocsolver PRIVATE hip::device )
+if(CMAKE_CXX_COMPILER MATHCES ".*/hipcc$")
```
aaronenyeshi: there is a typo here: MATCHES

PR #70, opened 2020-04-24
Centos fixes

TorreZuk:
* hipcc centos adjustments


TorreZuk:
This builds in CentosOS container w/o tensile.  Eiden do you want to add in the scl enable fixes now?

tfalders:
There appears to be a packaging failure on centos.

BTW, why are we switching CI to use rocblas master branch?

TorreZuk:
We will build hcc for a while, and even after 3.5 you may want to build with hcc locally to investigate performance differences.

PR #74, opened 2020-04-27
Single threaded compile in install script

eidenyoshida:
Test to see if CI errors reduce


eidenyoshida:
I've built a few times and CentOS build has not failed as of yet. Can we put this in for now until a more permanent fix can be implemented?

PR #77, opened 2020-04-28
find_package OpenMP

TorreZuk:
* try and fix centos failures


TorreZuk:
Test fails with current dockers

PR #79, opened 2020-04-28
use cmake not cmake3

TorreZuk:
* test only for build cmake3 used in docker and other places


eidenyoshida:
Do we know if they did the same thing in SLES

TorreZuk:
The switch statement in install to use cmake3 was only for centos/rhel I don't know who added that.  I guess git blame may show.  Is SLES failing too?

TorreZuk:
Test over for now

PR #80, opened 2020-04-29
Complex support 1

tfalders:
Added complex support for reflector functions (larf, larfg, larft, larfb) and QR factorization functions (geqr2, geqrf, gelq2, gelqf).

### DIFF
--- a/rocsolver/library/include/rocsolver-functions.h
+++ b/rocsolver/library/include/rocsolver-functions.h
@@ -318,6 +318,26 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_dlarf(rocblas_handle handle,
                                                 double* A, 
                                                 const rocblas_int lda);
 
+ROCSOLVER_EXPORT rocblas_status rocsolver_clarf(rocblas_handle handle, 
+                                                const rocblas_side side, 
+                                                const rocblas_int m,
+                                                const rocblas_int n, 
+                                                rocblas_float_complex* x, 
+                                                const rocblas_int incx, 
+                                                const rocblas_float_complex* alpha,
+                                                rocblas_float_complex* A, 
+                                                const rocblas_int lda);
+
+ROCSOLVER_EXPORT rocblas_status rocsolver_zlarf(rocblas_handle handle, 
+                                                const rocblas_side side, 
+                                                const rocblas_int m,
+                                                const rocblas_int n, 
+                                                rocblas_double_complex* x, 
+                                                const rocblas_int incx, 
+                                                const rocblas_double_complex* alpha,
+                                                rocblas_double_complex* A, 
+                                                const rocblas_int lda);
+
 
 /*! \brief LARFB applies a block reflector H to a general m-by-n matrix A.
 
--- a/rocsolver/library/include/rocsolver-functions.h
+++ b/rocsolver/library/include/rocsolver-functions.h
@@ -162,6 +162,20 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_dlarfg(rocblas_handle handle,
                                                  const rocblas_int incx, 
                                                  double *tau);
 
+ROCSOLVER_EXPORT rocblas_status rocsolver_clarfg(rocblas_handle handle, 
+                                                 const rocblas_int n, 
+                                                 rocblas_float_complex *alpha,
+                                                 rocblas_float_complex *x, 
+                                                 const rocblas_int incx, 
+                                                 rocblas_float_complex *tau);
+
+ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfg(rocblas_handle handle, 
+                                                 const rocblas_int n, 
+                                                 rocblas_double_complex *alpha,
+                                                 rocblas_double_complex *x, 
+                                                 const rocblas_int incx, 
+                                                 rocblas_double_complex *tau);
+
 
 /*! \brief LARFT Generates the triangular factor T of a block reflector H of order n.
 
--- a/rocsolver/library/include/rocsolver-functions.h
+++ b/rocsolver/library/include/rocsolver-functions.h
@@ -368,9 +368,9 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarf(rocblas_handle handle,
     the values of side and trans:
 
         H  * A  (No transpose from the left)
-        H' * A  (Transpose from the left)
+        H' * A  (Transpose or conjugate transpose from the left)
         A * H   (No transpose from the right), and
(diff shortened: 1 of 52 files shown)


jzuniga-amd:
Please complete the table in docs/source/userguide_intro.rst

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -304,6 +340,26 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_dlarf(rocblas_handle handle,
...
+                                                const rocblas_int m,
+                                                const rocblas_int n, 
+                                                rocblas_double_complex* x, 
+                                                const rocblas_int incx, 
+                                                const rocblas_double_complex* alpha,
+                                                rocblas_double_complex* A, 
+                                                const rocblas_int lda);
+
 
 /*! \brief LARFB applies a block reflector H to a general m-by-n matrix A.
```
jzuniga-amd, line 365: We need to uptade description here to include transpose-conjugate.
H' * A  (Transpose from the left) --> (transpose or transpose-conjugate from the left)
[in] trans: rocsolver_operation.
Specifies whether the block reflector or its transpose is to be applied. --> Specifies whether the block reflector or its transpose or transpose conjugate is to be applied

rocsolver/library/src/auxiliary/rocauxiliary_larfg.cpp
```diff
@@ -42,7 +45,8 @@ rocblas_status rocsolver_larfg_impl(rocblas_handle handle, const rocblas_int n,
                                       tau,
                                       strideP, 
                                       batch_count,
-                                      (T*)norms);
+                                      (T*)norms,
+                                      (T*)work);
 
     hipFree(norms);
```
jzuniga-amd, line 51: you must hipfree(work) as well...!!

jzuniga-amd:
Please remove this line:
bool t = rocblas_is_managing_device_memory(handle);
in larfg.cpp...
It was for debugging only; is not needed.

rocsolver/library/src/auxiliary/rocauxiliary_larfg.cpp
tfalders: Oops! I must have forgotten to transfer that over from the template function when I rebased. I'll be sure to fix that!

rocsolver/library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -65,32 +102,17 @@ rocblas_status rocsolver_larfg_template(rocblas_handle handle, const rocblas_int
     //if n==1 return tau=0
     dim3 gridReset(1, batch_count, 1);
     dim3 threads(1, 1, 1); 
-    if (n == 1) {
+    if (n == 1 && !COMPLEX) {
```
jzuniga-amd, line 105: This should apply for complex case too!
Also, please restore the pointer mode to old_mode before return (rocblas_set_pointer_mode(handle,old_mode);)… I forgot to add this in the refactoring.

rocsolver/library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -65,32 +102,17 @@ rocblas_status rocsolver_larfg_template(rocblas_handle handle, const rocblas_int
...
-    
-    //compute norm of x
-    for (int b=0;b<batch_count;++b) {
-        xp = load_ptr_batch<T>(xx,b,shiftx,stridex);
-        rocblas_nrm2(handle, n - 1, xp, incx, (norms + b));
     }
+    
+    //compute squared norm of x
+    rocblasCall_dot<COMPLEX,T>(handle, n-1, x, shiftx, incx, stridex,
+                               x, shiftx, incx, stridex, batch_count, norms, work);
```
jzuniga-amd, line 112: Good idea to use dot here!!

rocsolver/library/src/auxiliary/rocauxiliary_larfg.hpp
tfalders: I added an ignore for the complex case because the original lapack ensures that beta is real, so you can get a non-trivial reflection even if n == 1. I wasn't sure if it was worth writing a separate kernel for the n == 1 complex case, but I can add it in if desired.

jzuniga-amd: yes, you are right!

tfalders: The larfg template appears to already restore the pointer mode before returning. Am I looking at the right file?

jzuniga-amd: I mean before returning inside the if (n == 1 && !COMPLEX)

rocsolver/library/src/lapack/roclapack_geqr2.hpp
```diff
@@ -48,10 +48,14 @@ rocblas_status rocsolver_geqr2_template(rocblas_handle handle, const rocblas_int
                                  A, shiftA + idx2D(min(j+1,m-1),j,lda), //vector x to work on
                                  1, strideA,                            //inc of x    
                                  (ipiv + j), strideP,                   //tau
-                                 batch_count, diag);
+                                 batch_count, diag, work);
```
jzuniga-amd, line 51: The size of work is only calculated taking larf requirements into account!
Are we sure that is going to be enough for the larfg call as well?
If size_1 is what the call to larf requires, and size_2 is what larfg requires, then the size of work must be max(size_1,size_2)

rocsolver/library/src/lapack/roclapack_gelq2.hpp
```diff
@@ -39,16 +39,21 @@ rocblas_status rocsolver_gelq2_template(rocblas_handle handle, const rocblas_int
...
+
         // generate Householder reflector to work on row j
         rocsolver_larfg_template(handle,
                                  n - j,                                 //order of reflector
                                  A, shiftA + idx2D(j,j,lda),            //value of alpha
                                  A, shiftA + idx2D(j,min(j+1,n-1),lda), //vector x to work on
                                  lda, strideA,                          //inc of x    
                                  (ipiv + j), strideP,                   //tau
-                                 batch_count, diag);
+                                 batch_count, diag, work);
```
jzuniga-amd, line 56: The size of work is only calculated taking larf requirements into account!
Are we sure that is going to be enough for the larfg call as well?
If size_1 is what the call to larf requires, and size_2 is what larfg requires, then the size of work must be max(size_1,size_2)

PR #83, opened 2020-05-01
Fix for CentOS 8.1

tfalders:
Ticket ROCMOPS-883


saadrahim:
Looks good to me but hope this hack can be removed with proper cmake find_packages.

tfalders:
Just to be safe, I'll rename the variable to ROCM_OPENMP_PATH.

PR #90, opened 2020-05-12
Added geqrf_ptr_batched

tfalders:
Added a new version of geqrf_batched where tau is an array of pointers. This matches the cuBLAS signature and will allow hipBLAS to remain lightweight.


rocsolver/clients/benchmarks/client.cpp
```diff
@@ -432,6 +433,16 @@ int main(int argc, char *argv[])
     else if (precision == 'z')
       testing_geqr2_geqrf_strided_batched<rocblas_double_complex,double,1>(argus);
   } 
+  else if (function == "geqrf_ptr_batched") {
+    if (precision == 's')
+      testing_geqrf_ptr_batched<float,float>(argus);
+    else if (precision == 'd')
+      testing_geqrf_ptr_batched<double,double>(argus);
+    if (precision == 'c')
```
doctorcolinsmith, line 441: Missing `else` on this case?  I would not expected `precision` to be mutated in the previous calls.

rocsolver/clients/gtest/geqrf_ptr_batched_gtest.cpp
```diff
@@ -0,0 +1,148 @@
+/* ************************************************************************
+ * Copyright 2018 Advanced Micro Devices, Inc.
```
doctorcolinsmith, line 2: Copyright notice is outdated.

tfalders: This seems to be a problem across all test files. We may wish to ask Juan to update them during his test refactor.

PR #92, opened 2020-05-14
Add -DROCM_USE_FLOAT16 so that it can build after upcoming rocBLAS changes

leekillough:
This change needs to go in before the upcoming rocBLAS hotfix, so that `_Float16` is still used by rocBLAS when it's included by rocSOLVER.


TorreZuk:
I presume this would have to also go into Solver staging if get's into 3.5 as a hot-fix.

leekillough:
> I presume this would have to also go into Solver staging if get's into 3.5 as a hot-fix.

I didn't see an up-to-date `master` branch for Solver, so I thought it was still "WIP". But if you are specifying the commit tag for a 3.5 staging area, then yes, it will need to be moved up, or else it will get errors building rocSOLVER, errors about `rocblas_half` not being an arithmetic type.

PR #94, opened 2020-05-18
Centos8 additions to install script and CI

eidenyoshida:
SWDEV-236880

Fix for installing dependencies from install.sh on Centos8. Also adds centos8 to CI

### DIFF
--- a/docker/dockerfile-build-centos8
+++ b/docker/dockerfile-build-centos8
@@ -1,0 +1,52 @@
+# Parameters related to building rocsolver
+ARG base_image
+
+FROM ${base_image}
+LABEL maintainer="rocsolver.maintainer@amd.com"
+
+ARG user_uid
+
+# Install dependent packages
+RUN yum install -y --nogpgcheck\
+    sudo \
+    ca-certificates \
+    git \
+    cmake3 \
+    make \
+    gcc-c++ \
+    gcc-gfortran \
+    pkgconfig \
+    python27 \
+    python3 \
+    python3-devel \
+    python3-pip \
+    python3-pytest \
+    python3-setuptools \
+    python3-virtualenv \
+    python3-pyyaml \
+    llvm-devel \
+    llvm-static \
+    libcxx-devel \
+    boost-devel \
+    zlib-devel \
+    libgomp \
+    numactl-libs \
+    rpm-build
+
+RUN yum remove -y rocblas *gtest*
+
+# Grant any member of sudo group password-less sudo privileges
+RUN useradd --create-home -u ${user_uid} -o -G video --shell /bin/bash jenkins && \
+    echo '%video ALL=(ALL) NOPASSWD:ALL' | tee /etc/sudoers.d/sudo-nopasswd && \
+    chmod 400 /etc/sudoers.d/sudo-nopasswd
+
+ARG ROCBLAS_SRC_ROOT=/usr/local/src/rocBLAS
+
+# Clone rocblas repo
+# Build client dependencies and install into /usr/local (LAPACK & GTEST)
+RUN mkdir -p ${ROCBLAS_SRC_ROOT} && cd ${ROCBLAS_SRC_ROOT} && \
+    git clone -b develop --depth=1 https://github.com/ROCmSoftwarePlatform/rocBLAS . && \
+    mkdir -p build/deps && cd build/deps && \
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -179,13 +182,15 @@ install_packages( )
...
     library_dependencies_ubuntu+=( "rocm-dev" )
     library_dependencies_centos+=( "rocm-dev" )
+    library_dependencies_centos_8+=( "rocm-dev" )
     library_dependencies_fedora+=( "rocm-dev" )
     library_dependencies_sles+=( "rocm-dev" )
   fi
 
   # dependencies to build the client
   local client_dependencies_ubuntu=( "gfortran" "libomp-dev" "libboost-program-options-dev")
   local client_dependencies_centos=( "devtoolset-7-gcc-gfortran" "libgomp" "boost-devel" )
```
saadrahim, line 192: Centos7, best not leave defaults

install.sh
```diff
@@ -169,6 +169,9 @@ install_packages( )
                                       "python34" "PyYAML" "python3*-PyYAML" "python3*-distutils-extra"
                                       "gcc-c++" "llvm7.0-devel" "llvm7.0-static"
                                       "zlib-devel" "wget" )
+  local library_dependencies_centos_8=( "epel-release"
```
saadrahim, line 172: Why are we using cmake3? Default may be new enough.

docker/dockerfile-build-centos8
```diff
@@ -0,0 +1,52 @@
...
+    rpm-build
+
+RUN yum remove -y rocblas *gtest*
+
+# Grant any member of sudo group password-less sudo privileges
+RUN useradd --create-home -u ${user_uid} -o -G video --shell /bin/bash jenkins && \
+    echo '%video ALL=(ALL) NOPASSWD:ALL' | tee /etc/sudoers.d/sudo-nopasswd && \
+    chmod 400 /etc/sudoers.d/sudo-nopasswd
+
+ARG ROCBLAS_SRC_ROOT=/usr/local/src/rocBLAS
```
saadrahim, line 43: Do we need this rocblas install?

install.sh
eidenyoshida: STG1 docker container from devops doesn't have any cmake installed that I can tell.
yum install cmake and cmake3 installs the same version I believe in centos8. You can call cmake3 or cmake after installing the package and it will point to the same cmake (3.11)

docker/dockerfile-build-centos8
eidenyoshida: This rocblas download is to just install gtest I believe

install.sh
eidenyoshida: I will fix this shortly

saadrahim:
> 
> 
> Note by making existing centos 7 guarded whereas before it was all pre-existing you may be breaking Centos6 which is used by manylinux2010. You usually can't drop a default without breaking things.
> Centos6 dependences were the same as Centos7 with the exception that llvm7.0-devel llvm7.0-static were excluded as not available.

Centos6 is an unsupported OS. We never supported it in the first place. If support is required, the manylinux team can issue a PR.

eidenyoshida:
Shall I merge? If centos6 is desired later on it seems like a fairly simple fix as Saad said.

saadrahim:
As much as I get annoyed, maybe there should be a default path. But if we remove a default path for all our install.sh scripts, it may be better. Users of unsupported platforms will need to request us to support them or work independently. If Torre agrees, we should remove defaults going forward.

TorreZuk:
I agree it would make sense to make all platform constants OS version specific.   As for Centos6 support I was commenting as we did get the external PR had to revise and test and they are still waiting for rocBLAS 3.5 to hopefully make the manylinux2010 group's life simpler with Centos6.   In this case the default could likely have been renamed _6_7 and worked as it was with version >= 8 being the future case support, but you can decide you won't support Centos6 except on a case by case basis.  This is not my project so leave it to Juan to approve or not, but the PR looks fine.

docker/dockerfile-build-centos8
amdkila: We're installing dependencies for gtest and more importantly, lapack. We could switch this with libgtest-dev and lapack-devel if we want

eidenyoshida:
I'm going to merge as this fix is fairly urgent to unblock devops. We can implement other specific fixes later

PR #101, opened 2020-05-25
Make hip-clang default

amdkila:


eidenyoshida:
Are we targeting for develop or master-rocm-3.5?

PR #106, opened 2020-06-02
update install script

jzuniga-amd:


saadrahim:
@jzuniga-amd You need to keep the hcc option available for a little longer. Devops still needs to update their infrastructure. Please change default to hipcc but wait a few releases to remove this completely.

amdkila:
So, the hcc mode is still available, but I'm concerned about removing --hip-clang. If we kept --hip-clang, then I think this PR should be fine

jzuniga-amd:
I am not disabling HCC, on the contrary... with the current script where HIPCC was set as the default, there is NO WAY to build the library with HCC. I am providing this option with the flag --hcc.

Why is the flag --hip-clang still needed?

TorreZuk:
I think they are just saying if everything else still accepts --hip-clang you don't want to be the odd man out, otherwise you need to special case the rocSolver build

jzuniga-amd:
mmm good point! I don't know, Akila could confirm, but I will leave -h = --hip-clang just in case.

PR #108, opened 2020-06-03
Add optimization for LU factorization (small sizes)

jzuniga-amd:


tfalders:
It looks like batched getf2 and getrf still call some rocblas functions in for loops. Is it possible yet to use batch functionality instead?

rocsolver/library/src/lapack/roclapack_getf2.hpp
```diff
@@ -71,7 +342,7 @@ template <typename T, typename U>
 rocblas_status rocsolver_getf2_template(rocblas_handle handle, const rocblas_int m,
                                         const rocblas_int n, U A, const rocblas_int shiftA, const rocblas_int lda, 
                                         const rocblas_stride strideA, rocblas_int *ipiv, const rocblas_int shiftP, 
-                                        const rocblas_stride strideP, rocblas_int* info, const rocblas_int batch_count,
+                                        const rocblas_stride strideP, rocblas_int* info, const rocblas_int batch_count, const rocblas_int pivot,
```
tfalders, line 345: Maybe we can allow ipiv to be nullptr when pivot == 0?

jzuniga-amd: Yes, this is what cuBLAS does. They use ipiv = null to disable pivoting...
I was not sure about this. At the end I did not like the idea of using a nullptr to enable functionality or to chose between different algorithms. We accept nullptr as result of memory allocation with no size, for example, and would be bad to accept ipiv=null when it is result of memory allocations and not a user decision to disable pivoting... in addition, no pivoting is only valid for small sizes, so we would need additional checks. 

From the linear algebra perspective, if the matrix has non-zero dimensions m-by-n, then there will be min(m,n) pivots in the U factor no matter how you chose them (sequentially or as the maximum value in the column)… if the user disables pivoting, ipiv should exist anyways and be populated with the sequence 1,2,...min(m,n)

We can always revisit this in the future if the final user request it.

jzuniga-amd:
> It looks like batched getf2 and getrf still call some rocblas functions in for loops. Is it possible yet to use batch functionality instead?

I think Torre has already enabled batched versions of iamax and trsm, but I did not have the time to test them. We can change this for 3.7. It will help, but won't improve performance as much as the small-sizes algorithm.

cgmb:
The addition of optional parameters in `rocsolver-functions.h` breaks C compatibility. That's a C++ feature made possible by overloading.

cgmb:
I did a bit more research into this problem. Here's what I found:
- Everything about this works fine in C++.
- `rocsolver.so` contains these functions as if they were valid C functions, but the pivot parameter is mandatory.
- This header is not valid in C due to the default value. A workaround is to delete `= 0` from the header. Or, to preserve the C++ behaviour, the `= 0` could be hidden behind an `#ifdef __cplusplus`  check. In either case, the parameter is mandatory in C.

The pivot is not a parameter in the LAPACK interface, so I don't think it should be included in ours. I think if we're going to add an optional pivot parameter, we should be defining a function with a new name that takes the pivot, and leaving the original function interface unchanged. However, I have no idea what such a function should be named.

There's also a question as to what our binary compatibility guarantees are. Even in C++, I think change could result in undefined behaviour for programs built against an older version of `rocsolver.so` that are used with this newer version.

PR #116, opened 2020-06-25
Fix for geqrf, getrf, getrs, and getri test matrix initialization

tfalders:
Test matrices for geqrf, getrf, getrs, and getri are now initialized properly when evaluating performance.

### DIFF
--- a/rocsolver/clients/include/testing_getri.hpp
+++ b/rocsolver/clients/include/testing_getri.hpp
@@ -176,9 +176,6 @@ void getri_getError(const rocblas_handle handle,
 
     // CPU lapack
     for (rocblas_int b = 0; b < bc; ++b) {
-        if (hInfo[b][0] != 0)
-            return;  // error encountered - unlucky pick of random numbers? no use to continue
-        
         cblas_getri<T>(n, hA[b], lda, hIpiv[b], hW.data(), &sizeW);
     }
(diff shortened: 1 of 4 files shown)


rocsolver/clients/include/testing_getri.hpp
```diff
@@ -153,6 +176,9 @@ void getri_getError(const rocblas_handle handle,
 
     // CPU lapack
     for (rocblas_int b = 0; b < bc; ++b) {
+        if (hInfo[b][0] != 0)
+            return;  // error encountered - unlucky pick of random numbers? no use to continue
```
jzuniga-amd, line 180: In theory this info test should be obsolete as we are scaling the matrix to avoid singularity...
I think it could be removed.

rocsolver/clients/include/testing_getrs.hpp
```diff
@@ -158,6 +185,9 @@ void getrs_getError(const rocblas_handle handle,
 
     // CPU lapack
     for (rocblas_int b = 0; b < bc; ++b) {
+        if (hInfo[b][0] != 0)
```
jzuniga-amd, line 188: In theory this info test should be obsolete as we are scaling the matrix to avoid singularity...
I think it could be removed.

PR #118, opened 2020-07-06
rocBLAS fixes

tfalders:
Updating rocSOLVER for changes made in rocBLAS. Fixes broken gemm template calls and missing msgpack installation.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -151,6 +151,21 @@ install_zypper_packages( )
     done
 }
 
+install_msgpack_from_source( )
+{
+    if [[ ! -d "${build_dir}/deps/msgpack-c" ]]; then
+      pushd .
+      mkdir -p ${build_dir}/deps
+      cd ${build_dir}/deps
+      git clone -b cpp-3.0.1 https://github.com/msgpack/msgpack-c.git
+      cd msgpack-c
+      ${cmake_executable} .
+      make
+      elevate_if_not_root make install
+      popd
+    fi
+}
+
 # Take an array of packages as input, and delegate the work to the appropriate distro installer
 # prereq: ${ID} must be defined before calling
 # prereq: ${build_clients} must be defined before calling
@@ -189,6 +204,12 @@ install_packages( )
   library_dependencies_centos_8+=( "rocm-dev" )
   library_dependencies_fedora+=( "rocm-dev" )
   library_dependencies_sles+=( "rocm-dev" )
+
+  case "${ID}" in
+    centos|rhel|sles|opensuse-leap)
+      install_msgpack_from_source
+      ;;
+  esac
 
   # dependencies to build the client
   local client_dependencies_ubuntu=( "gfortran" "libomp-dev" "libboost-program-options-dev")
--- a/install.sh
+++ b/install.sh
@@ -155,8 +155,8 @@ install_msgpack_from_source( )
 {
     if [[ ! -d "${build_dir}/deps/msgpack-c" ]]; then
       pushd .
-      mkdir -p ${build_dir}/deps
-      cd ${build_dir}/deps
+      mkdir -p "${build_dir}/deps"
+      cd "${build_dir}/deps"
       git clone -b cpp-3.0.1 https://github.com/msgpack/msgpack-c.git
       cd msgpack-c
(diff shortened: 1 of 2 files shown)


tfalders:
The updated gemm template on the rocBLAS side has apparently removed the STRIDED template argument, but I've kept it in our rocblasCall_gemm wrapper. Main reason is I'm not sure if the template arguments will be altered further, e.g. by removing BATCHED or reintroducing STRIDED.

cgmb:
Your changes fix the compile-time issues, but it appears that there's a few tests that fail now that didn't before. Is there still work to do on our end? Or is that a regression in rocBLAS? Or something else entirely?

install.sh
```diff
@@ -151,6 +151,21 @@ install_zypper_packages( )
     done
 }
 
+install_msgpack_from_source( )
+{
+    if [[ ! -d "${build_dir}/deps/msgpack-c" ]]; then
+      pushd .
+      mkdir -p ${build_dir}/deps
+      cd ${build_dir}/deps
```
cgmb, line 159: I would recommend always quoting arguments containing shell variables (if they're supposed to be a single argument). `$build_dir` hopefully doesn't contain spaces anyways, but it's a good habit to get into.
```suggestion
      mkdir -p "${build_dir}/deps"
      cd "${build_dir}/deps"
```

tfalders: Good catch. It was a straight copy, so I wasn't looking too closely at that kind of thing. I'll add them in.

tfalders:
> The updated gemm template on the rocBLAS side has apparently removed the STRIDED template argument, but I've kept it in our rocblasCall_gemm wrapper. Main reason is I'm not sure if the template arguments will be altered further, e.g. by removing BATCHED or reintroducing STRIDED, and therefore updating gemm calls across the library could be wasted effort.

I noticed this failure as well. It's in getrs, which doesn't call gemm, and should not be affected by this PR. A regression in rocBLAS is possible.

I'm also noticing that some functions are freezing (or taking an incredibly long time to execute) on my machine. I need to investigate more, but this should also not be affected by my PR.

jzuniga-amd:
> The updated gemm template on the rocBLAS side has apparently removed the STRIDED template argument, but I've kept it in our rocblasCall_gemm wrapper. Main reason is I'm not sure if the template arguments will be altered further, e.g. by removing BATCHED or reintroducing STRIDED, and therefore updating gemm calls across the library could be wasted effort.
> 
> I noticed this failure as well. It's in getrs, which doesn't call gemm, and should not be affected by this PR. A regression in rocBLAS is possible.
> I'm also noticing that some functions are freezing (or taking an incredibly long time to execute) on my machine. I need to investigate more, but this should also not be affected by my PR.

rocsolver_GETRS calls rocblas_TRSM which calls rocblas_GEMM multiple times

tfalders:
> > The updated gemm template on the rocBLAS side has apparently removed the STRIDED template argument, but I've kept it in our rocblasCall_gemm wrapper. Main reason is I'm not sure if the template arguments will be altered further, e.g. by removing BATCHED or reintroducing STRIDED, and therefore updating gemm calls across the library could be wasted effort.
> > I noticed this failure as well. It's in getrs, which doesn't call gemm, and should not be affected by this PR. A regression in rocBLAS is possible.
> > I'm also noticing that some functions are freezing (or taking an incredibly long time to execute) on my machine. I need to investigate more, but this should also not be affected by my PR.
> 
> rocsolver_GETRS calls rocblas_TRSM which calls rocblas_GEMM multiple times

In light of some of the "rocBLAS Extended Failures on CI" emails that have going around, there seems to be an issue with gemm. It looks like the other issue I've been having is definitely a Tensile issue.
Error

cgmb:
The patch looks good and I'm happy to +1 this PR if there's no more code changes on our end, but I want to be sure that there exists somebody who is working on fixing this issue (and knows we're affected) before calling our work here complete.

tfalders:
> Let's merge as hipBLAS is also failing because of rocSOLVER. I don't think we can do much with the getrs/trsm regression on our side...

I forgot to mention that Daine updated rocBLAS with an apparent fix. I'll merge now and hopefully the fix will work.

PR #120, opened 2020-07-09
Minor cleanup for docs, scripts and tests

cgmb:
These are a collection of minor improvements to the docs, scripts and tests that I made while familiarizing myself with the rocSOLVER library and associated tooling.

- Used [shellcheck](https://www.shellcheck.net/) to lint the install script and addressed a subset of the warnings it listed. I also fixed some inconsistent indenting. The most important change was to quote the `$build_dir` when removing the old build directory.

- Fixed the example code in the README, which didn't build as specified prior to this change. It appears the example was originally C code. I fixed all errors I encountered, and reformatted it to be a little more compact. Sprawl is fine for most code, but a README is often read in more limited environments (like Notepad or a mobile browser).

- Added a version header and footer to the `rocsolver-test` utility to help make that information more obvious in our logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,39 +1,39 @@
 # rocSOLVER
 
-rocSOLVER is a work-in-progress implementation of a subset of [LAPACK](http://www.netlib.org/lapack/explore-html/index.html) 
-functionality on the [ROCm platform](https://rocm.github.io). 
+rocSOLVER is a work-in-progress implementation of a subset of [LAPACK][1]
+functionality on the [ROCm platform][2].
 
 # Documentation
 
-For a detailed description of the rocSOLVER library, its implemented routines, the installation process and user guide, see the
-[rocSOLVER documentation](https://rocsolver.readthedocs.io/en/latest).
+For a detailed description of the rocSOLVER library, its implemented routines,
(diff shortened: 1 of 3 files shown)


cgmb:
By the way, I noticed that there's [an API function](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/f0f1d78a763c5fdccc5731c3282c8ff860c4cacd/rocsolver/library/include/rocsolver-functions.h#L19-L31) to get the rocsolver version, though it's not mentioned in the web docs. There's a similar function for rocBLAS. They're a little annoying to use, though, since there's no way to know what size of buffer is required to guarantee success. That's why I opted for using the macros directly. I'm not sure why we don't just provide a static version string, but perhaps we should consider adding another function that returns the required buffer size?

tfalders:
There appears to be a CI failure on centos that I'm not sure is a false failure, but also doesn't seem to have been introduced by your changes. We may want to take a closer look at that.

cgmb:
It built properly the first time!

I'm still not entirely sure why the CentOS 7 CI build goes back and forth on poisoned symbol errors. It's not a major issue, though, so perhaps that's a problem for another time.

PR #121, opened 2020-07-09
Fix ubuntu 18 packaging issue

amdkila:
- Brackets are needed to set dependency for packages on Ubuntu 18, but not Ubuntu 16, where it works with or without brackets


amdkila:
> I've poked at the occasional debian control file and occasionally daydreamed about simpler packaging scripts while reading the CPack documentation...
> 
> I'm definitely not an expert, but it looks good to me.

Thanks @cgmb for the approval...I've daydreamed about simpler packaging scripts a few times myself.

PR #123, opened 2020-07-20
getri improvements 2

tfalders:
Added getri_kernel for small size matrices and implemented blocked version of trtri for large size matrices.


rocsolver/library/src/lapack/roclapack_getri.hpp
```diff
@@ -12,88 +12,235 @@
...
+}
+
+template <typename T, typename U>
+rocblas_status getri_small_sizes(rocblas_handle handle, const rocblas_int n, U A, const rocblas_int shiftA, const rocblas_int lda,
+                                 const rocblas_stride strideA, rocblas_int *ipiv, const rocblas_int shiftP, const rocblas_stride strideP,
+                                 rocblas_int* info, const rocblas_int batch_count)
+{
+    #define runGetriSmall(DIM)                                                         \
+        hipLaunchKernelGGL((getri_kernel<DIM,T>), grid, block, 0, stream,              \
+                           A, shiftA, lda, strideA, ipiv, shiftP, strideP, info);
```
cgmb, line 138: On one hand, we probably could replace this macro with some clever use of `std::make_index_sequence`. On the other hand, I think it would be slower  to compile, and we would be leaning on the optimizer more. So, I think it's reasonable to leave it as a macro.

However, I would treat macros with care. There's a few things I would suggest doing:

1. Use `ALL_CAPS` for macros, especially if they take arguments. While `runGetriSmall` looks like a function, the macro abstraction is leaky. For example, calling `runGetriSmall(std::map<int,int>::size_type(10));` would be acceptable for a function, but the comma in the template type list breaks the macro invocation. The naming scheme helps make it more obvious that it's not a real function. I'm not fanatical about the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-ALL_CAPS), but this is quite a common convention.

2. This is mostly a matter of style, but if you're defining a function-like macro, don't include a semi-colon at the end. Leave it off to ensure the semi-colon is added as appropriate at the call site.

rocsolver/library/src/lapack/roclapack_getri.hpp
```diff
@@ -12,88 +12,235 @@
...
+        case 56: runGetriSmall(56); break;
+        case 57: runGetriSmall(57); break;
+        case 58: runGetriSmall(58); break;
+        case 59: runGetriSmall(59); break;
+        case 60: runGetriSmall(60); break;
+        case 61: runGetriSmall(61); break;
+        case 62: runGetriSmall(62); break;
+        case 63: runGetriSmall(63); break;
+        case 64: runGetriSmall(64); break;
+    }
```
cgmb, line 213: 1. What happens if we pass `65` by mistake? There should be a default case that either returns an error, calls `__builtin_trap()`, or calls `__builtin_unreachable()`, depending on the behaviour we want.

2. It's not that important, but we could probably shorten this with a clever usage of `__COUNTER__`.

rocsolver/library/src/include/rocblas.hpp
```diff
@@ -410,6 +410,76 @@ rocblas_status rocblasCall_gemm(rocblas_handle    handle,
...
+                            rocblas_stride    strideB,
+                            rocblas_int       batch_count,
+                            T*                work,
+                            T**               workArr)
+{
+    constexpr rocblas_int nb = ROCBLAS_TRMM_NB;
+    constexpr rocblas_stride strideW = 2 * ROCBLAS_TRMM_NB * ROCBLAS_TRMM_NB;
+
+    V AA = (V)workArr + batch_count;
+    V BB = (V)workArr + 2*batch_count;
```
cgmb, line 468: Is there a way to make what these two lines do more obvious at a glance? Or maybe add a brief comment here?

rocsolver/library/src/lapack/roclapack_getri.hpp
cgmb: On second thought, `std::make_index_sequence` probably is simpler than trying to make `__COUNTER__` work.

cgmb: See [my std::integer_sequence example](https://gist.github.com/cgmb/5a7ec238cc62d7b3f257c421a67ab7c5).

cgmb: You may want to check it provides the same performance, but I would expect so.

rocsolver/library/src/include/common_device.hpp
```diff
@@ -75,6 +75,15 @@ __global__ void get_array(T** out, T* in, rocblas_stride stride, rocblas_int bat
         out[b] = in + b*stride;
 }
 
+template <typename T, typename U>
+__global__ void shift_array(T** out, U in, rocblas_int shift, rocblas_int batch) 
+{
+    int b = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
```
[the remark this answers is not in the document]
tfalders: That depends on whether having over 2 billion matrices in a batch is something we want the library to support. It may be worth considering, but for now the assumption throughout the code seems to be that we don't.

rocsolver/library/src/lapack/roclapack_getri.hpp
cgmb: Semi-related: `hipLaunchKernelGGL` is a function-like macro. I wonder why they use the `do { ... } while(0)` trick for it, given that it's only one expression.

rocsolver/library/src/include/rocblas.hpp
```diff
@@ -410,6 +410,76 @@ rocblas_status rocblasCall_gemm(rocblas_handle    handle,
...
+                            rocblas_stride    strideB,
+                            rocblas_int       batch_count,
+                            T*                work,
+                            T**               workArr)
+{
+    constexpr rocblas_int nb = ROCBLAS_TRMM_NB;
+    constexpr rocblas_stride strideW = 2 * ROCBLAS_TRMM_NB * ROCBLAS_TRMM_NB;
+    return rocblas_trmm_template<BATCHED,nb,nb,T>(handle,side,uplo,transA,diag,m,n,cast2constType<T>(alpha),
+                                                  cast2constType<T>(A + offsetA),lda,strideA,
+                                                  B + offsetB,ldb,strideB,batch_count,work,strideW);
```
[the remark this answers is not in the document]
tfalders: Personally, I'd also prefer a shorter name, but it's not something I came up with and is outside the scope of this PR.

rocsolver/library/src/lapack/roclapack_trtri.hpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    // only non-unit upper triangular matrices currently supported
+    if (uplo != rocblas_fill_upper || diag != rocblas_diagonal_non_unit)
+        return rocblas_status_not_implemented;
+
+    // everything must be executed with scalars on the host
+    rocblas_pointer_mode old_mode;
+    rocblas_get_pointer_mode(handle,&old_mode);
+    rocblas_set_pointer_mode(handle,rocblas_pointer_mode_host);
```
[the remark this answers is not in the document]
tfalders: That would require a change across the entire library and is outside the scope of this PR. I think we'd also have to get the rocBLAS guys on board since we probably want to handle this consistently across both libraries and in customer usage.

rocsolver/library/src/include/rocblas.hpp
```diff
@@ -410,6 +410,76 @@ rocblas_status rocblasCall_gemm(rocblas_handle    handle,
...
+    constexpr rocblas_int nb = ROCBLAS_TRMM_NB;
+    constexpr rocblas_stride strideW = 2 * ROCBLAS_TRMM_NB * ROCBLAS_TRMM_NB;
+
+    V AA = (V)workArr + batch_count;
+    V BB = (V)workArr + 2*batch_count;
+    
+    hipStream_t stream;
+    rocblas_get_stream(handle, &stream);
+
+    rocblas_int blocks =  (batch_count - 1)/256 + 1;
```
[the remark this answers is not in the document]
tfalders: I think this occurs often enough in the library and is such an easy thing to replicate that replacing it with a function call would be more trouble than it's worth. Plus there's also the runtime cost of calling the function to consider.

rocsolver/library/src/lapack/roclapack_getri.hpp
tfalders: I was basing this code off of similar code for getf2. Whatever style we adopt here should also be ported over to getf2 for consistency.

rocsolver/library/src/include/rocblas.hpp
tfalders: I can probably do that, yeah. Basically, since trmm doesn't have offset/shift parameters, I had to manually do the shift in the batched case, and the shifted pointers all go into workArr. I'll probably ask the rocBLAS team if they will add shifting to trmm, because not having it is really inconvenient.

rocsolver/library/src/lapack/roclapack_getri.hpp
tfalders: Same comment as above. Additionally, this function is not exposed to the user, so passing 65 by mistake should only occur by developer error. I'm not sure what we want to do to catch developer errors like that, but it should also be ported to getf2.

rocsolver/library/src/lapack/roclapack_getri.hpp
tfalders: > Semi-related: `hipLaunchKernelGGL` is a function-like macro. I wonder why they use the `do { ... } while(0)` trick for it, given that it's only one expression.

My guess is that it's intended to create a local scope, so that variables in the macro don't interfere with variables outside the macro. Not my wheelhouse, though, so I could be wrong.

rocsolver/library/src/include/common_device.hpp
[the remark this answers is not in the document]
cgmb: kk. There's nothing to change here. I was just asking to learn more generally about our current error-handling strategy.

As you mention, throughout rocSOLVER there are number implementation limits or error conditions that we're not fully handling or don't seem to be documented. Fixing that is just work that needs to be prioritized like everything else. I wanted to find out if that had been recognized already and considered in the long-term plan.

rocsolver/library/src/lapack/roclapack_getri.hpp
cgmb: This isn't a super big deal, btw. It's pure style. I just wanted to submit these points for consideration and this code is simply a good example.

rocsolver/library/src/lapack/roclapack_getri.hpp
cgmb: There should be a default case just to make it clear to people reading the code that all cases were considered.

If you want to tell readers (and the optimizer) that the default case should never be reached, you can add `default: __builtin_unreachable();` to the end of the switch. In some code, that could even improve performance, though I doubt it does here. The catch is that you get undefined behaviour if your code is wrong and `__builtin_unreachable()` would be reached. It will have the same performance or better than the current code.

If you instead want to tell readers that the default case should never be reached and have it checked at runtime, you can add `default: assert(false);`. It will have a negligible performance impact (and it can be reduced to nothing in Release builds), but developers are fallible and it serves to both check our assumptions and communicate them.

cgmb: Don't worry about the `std::integer_sequence` stuff. It's cool that we could shorten this, but it's not important.

rocsolver/library/src/include/rocblas.hpp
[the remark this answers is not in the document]
cgmb: A inline function for this would have no runtime cost. It would compile down to the same machine code.

If I knew a good name for the operation, I would insist. This pattern doesn't work for unsigned integers, so you need to know the types to know that this calculation is correct. A named function, by contrast, could be defined to correctly handle whatever types it's overloaded for. Then if you know the function is correct for all the types it's defined for, all the calls to it would be more obviously correct.

That said, I'm not going to insist on defining a function which I don't have a good name for. So, this is just something to think about.

rocsolver/library/src/lapack/roclapack_trtri.hpp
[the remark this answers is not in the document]
cgmb: It would be good to get rocBLAS on board, but this code is entirely internal to rocSOLVER. There's nothing wrong with starting small, and refining a solution in our code before bringing it to a wider audience.

However, there's nothing wrong with this code. It was just a comment for consideration.

rocsolver/library/src/include/rocblas.hpp
tfalders: fixed

rocsolver/library/src/lapack/roclapack_getri.hpp
tfalders: Fixed for getri. I didn't want to change it for getf2 as well, since I think Juan is working on it. We'll have to ask him to make the change as well.

rocsolver/library/src/lapack/roclapack_getri.hpp
tfalders: Fixed for getri.

tfalders:
> If I'm understanding this right, the revised kernels calculate the same end results, just with a lot less syncing?

That seems about right. The main thing I wanted to accomplish with the revised kernels was to have each thread process an entry in the matrix product. The previous implementation based on vanilla LAPACK had each thread processing an entry in the operand and then distributing it to entries in the product. This lets me accumulate results in a local variable (e.g. yi or cij) instead of global GPU memory, and also seems to require fewer syncs.

jzuniga-amd:
Looks good. I think there is still some room to improve performance, but this is much better than the original functionality implementation. Let's merge and maybe revisit in the next sprint.

PR #124, opened 2020-07-24
Updating to match new rocblas exported path.

daineAMD:
Updating to work with rocblas PR [210](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/210)


tfalders:
It looks like the include for rocblas-version.h in rocsolver_gtest_main.cpp also needs to be updated.

daineAMD:
Everything seems to be passing except for what is an unrelated error. Debug build is failing on centos because an included boost file is using the C++ poisoned std::err, I see this was failing 3 days ago as well.

PR #126, opened 2020-07-27
Update Google gtest to 1.10.0

tfalders:
Updating rocSOLVER to match rocBLAS PR https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/208


tfalders:
Gtest version does not appear to be mentioned in the documentation. I'll wait for CI before merging the PR.

jzuniga-amd:
It will never pass CI unless you rebase on top of latest commit by Daine. It seems that latest- successful-rocblas-built is already picking up the new location of rocblas_exported_proto.hpp

tfalders:
Okay, I will merge it in now then. It looks like the prior build passed except for an intermittent failure.

PR #127, opened 2020-08-03
Getrf (LU factorization) optimizations

jzuniga-amd:
- Revisit small size optimized factorization
- Add panel optimized factorization (to work with mid-size matrices)
- Enable use of rocblas_iamax and rocblas_trsm template functions (the use of internal trsm and/or iamax did not seem necessary)
- Add no-pivoting versions to the public API
- Modify unit tests to test pivoting
- Other minor fixes and additions 

(Getri PR must be merged first, then conflicts must be resolved for this PR)


rocsolver/library/src/lapack/roclapack_getrf.cpp
```diff
@@ -23,39 +24,59 @@ rocblas_status rocsolver_getrf_impl(rocblas_handle handle, const rocblas_int m,
...
-    if (!scalars || (size_2 && !pivotGPU) || (size_3 && !iinfo))
+    hipMalloc(&pivot_val,size_2);
+    hipMalloc(&pivot_idx,size_3);
+    hipMalloc(&iinfo,size_4);
+    hipMalloc(&work,size_5);
+    if (!scalars || (size_2 && !pivot_val) || (size_3 && !pivot_idx) || (size_4 && !iinfo) || (size_5 && !work))
         return rocblas_status_memory_error;
 
+    // (CAUTION: THIS PART IS ACTUALLY ALLOCATED IN THE ROBLAS HANDLE)
+    rocblas_status perf_status = rocblasCall_trsm_mem<false,T,U>(handle,rocblas_side_left,m,n,batch_count,x_temp,x_temp_arr,invA,invA_arr);
```
tfalders, line 46: We should be able to save some memory depending on how trsm is called. It looks like trsm is always called with the number of rows = jb (<= GETRF_GETF2_SWITCHSIZE), so you can probably replace m with GETRF_GETF2_SWITCHSIZE.

rocsolver/library/src/lapack/roclapack_getrf.cpp
```diff
@@ -23,39 +24,59 @@ rocblas_status rocsolver_getrf_impl(rocblas_handle handle, const rocblas_int m,
...
+                                                    invA,
+                                                    invA_arr,
+                                                    optim_mem);
 
     hipFree(scalars);
-    hipFree(pivotGPU);
+    hipFree(pivot_val);
+    hipFree(pivot_idx);
     hipFree(iinfo);
+    hipFree(work);
```
tfalders, line 79: Do we not need to free x_temp, x_temp_arr, invA, invA_arr as well?

rocsolver/library/src/lapack/roclapack_getrf_batched.cpp
```diff
@@ -22,39 +23,59 @@ rocblas_status rocsolver_getrf_batched_impl(rocblas_handle handle, rocblas_int m
...
-    if (!scalars || (size_2 && !pivotGPU) || (size_3 && !iinfo))
+    hipMalloc(&pivot_val,size_2);
+    hipMalloc(&pivot_idx,size_3);
+    hipMalloc(&iinfo,size_4);
+    hipMalloc(&work,size_5);
+    if (!scalars || (size_2 && !pivot_val) || (size_3 && !pivot_idx) || (size_4 && !iinfo) || (size_5 && !work))
         return rocblas_status_memory_error;
 
+    // (CAUTION: THIS PART IS ACTUALLY ALLOCATED IN THE ROBLAS HANDLE)
+    rocblas_status perf_status = rocblasCall_trsm_mem<true,T,U>(handle,rocblas_side_left,m,n,batch_count,x_temp,x_temp_arr,invA,invA_arr);
```
tfalders, line 45: Same comments as non-batched.

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
```diff
@@ -19,18 +20,29 @@ rocblas_status rocsolver_getrf_strided_batched_impl(rocblas_handle handle, const
...
-    if (!scalars || (size_2 && !pivotGPU) || (size_3 && !iinfo))
+    hipMalloc(&pivot_val,size_2);
+    hipMalloc(&pivot_idx,size_3);
+    hipMalloc(&iinfo,size_4);
+    hipMalloc(&work,size_5);
+    if (!scalars || (size_2 && !pivot_val) || (size_3 && !pivot_idx) || (size_4 && !iinfo) || (size_5 && !work))
         return rocblas_status_memory_error;
+    
+    // (CAUTION: THIS PART IS ACTUALLY ALLOCATED IN THE ROBLAS HANDLE)
+    rocblas_status perf_status = rocblasCall_trsm_mem<false,T,U>(handle,rocblas_side_left,m,n,batch_count,x_temp,x_temp_arr,invA,invA_arr);
```
tfalders, line 42: Same comments as non-batched,

rocsolver/library/src/lapack/roclapack_getrf.hpp
```diff
@@ -17,42 +17,50 @@
...
+    rocblas_int *ipiv;
+    if (pivot) ipiv = ipivA + id*strideP + shiftP;
 
     if (info[id] == 0 && iinfo[id] > 0)
         info[id] = iinfo[id] + j;
 
     int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
 
-    if (tid < n)
+    if (tid < n && pivot)
```
tfalders, line 31: Perhaps the initialization of ipiv should be moved here so that pivot only needs to be checked once.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -2246,13 +2246,483 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zbdsqr(rocblas_handle handle,
...
+    If numerical accuracy is compromised, use the legacy-LAPACK-like API GETF2 routines instead.  
+
+    @param[in]
+    handle    rocblas_handle.
+    @param[in]
+    m         rocblas_int. m >= 0.\n
+              The number of rows of the matrix A. 
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The number of colums of the matrix A.
```
tfalders, line 2275: columns
(check other functions as well)

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -2246,13 +2246,483 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zbdsqr(rocblas_handle handle,
...
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the m-by-n matrix A to be factored.
+              On exit, the factors L and U from the factorization.
+              The unit diagonal elements of L are not stored.
+    @param[in]
+    lda       rocblas_int. lda >= m.\n
+              Specifies the leading dimension of A. 
+    @param[out]
+    info      pointer to a rocblas_int on the GPU.\n
+              If info = 0, succesful exit.
```
tfalders, line 2286: successful
(check other functions as well)

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -2246,13 +2246,483 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zbdsqr(rocblas_handle handle,
...
+    handle    rocblas_handle.
+    @param[in]
+    m         rocblas_int. m >= 0.\n
+              The number of rows of all matrices A_i in the batch.
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The number of colums of all matrices A_i in the batch.
+    @param[inout]
+    A         pointer to type. Array on the GPU (the size depends on the value of strideA).\n
+              On entry, in contains the m-by-n matrices A_i to be factored.
```
tfalders, line 2427: Remove: in contains
(check other functions as well)

rocsolver/library/src/include/helpers.hpp
```diff
@@ -60,6 +60,21 @@ inline double machine_precision()
...
+    T *R = array;
+    return R;
+}
+
+template <typename T>
+T ** cast2nonConstPointer(T *const *array)
+{
+    T **R = (T**)array;
+    return R;
+}
```
cgmb, line 75: It's not immediately obvious to me why this is needed. I might have to build the code to see how the template parameters are getting inferred, but naively, I would expect the buffers this is being used on to always be non-const anyways.

rocsolver/library/src/lapack/roclapack_getrf.cpp
jzuniga-amd: good catch! 
Changed...

rocsolver/library/src/lapack/roclapack_getrf.cpp
jzuniga-amd: No, these arrays are created by the rocblas_handle in a space of memory allocated when the handle is created. This memory will be freed by the rocblas_handle destructor. What we need is to move all memory allocations to the handle but for that we need the rocblas team to export some extra auxiliary functions. It is in the TODO list.

rocsolver/library/src/lapack/roclapack_getrf_batched.cpp
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_getrf.hpp
jzuniga-amd: Good catch! Changed...

rocsolver/library/include/rocsolver-functions.h
jzuniga-amd: fixed

rocsolver/library/include/rocsolver-functions.h
jzuniga-amd: fixed

rocsolver/library/include/rocsolver-functions.h
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
```diff
@@ -67,25 +88,53 @@ extern "C" {
...
+}
+
+ROCSOLVER_EXPORT rocblas_status rocsolver_cgetrf_npvt_strided_batched(rocblas_handle handle, const rocblas_int m, const rocblas_int n,
+                 rocblas_float_complex *A, const rocblas_int lda, const rocblas_stride strideA, rocblas_int* info, const rocblas_int batch_count) 
+{
+    rocblas_int *ipiv;
+    return rocsolver_getrf_strided_batched_impl<rocblas_float_complex>(handle, m, n, A, lda, strideA, ipiv, 0, info, batch_count, 0);
+}
+
+ROCSOLVER_EXPORT rocblas_status rocsolver_zgetrf_npvt_strided_batched(rocblas_handle handle, const rocblas_int m, const rocblas_int n,
```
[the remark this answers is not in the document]
jzuniga-amd: We can definitely do this. Good catch! 
As it is in all rocsolver functions, however, I will open a ticket and leave it our of this PR.

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
```diff
@@ -19,18 +20,29 @@ rocblas_status rocsolver_getrf_strided_batched_impl(rocblas_handle handle, const
...
     hipMalloc(&scalars,size_1);
-    hipMalloc(&pivotGPU,size_2);
-    hipMalloc(&iinfo,size_3);
-    if (!scalars || (size_2 && !pivotGPU) || (size_3 && !iinfo))
+    hipMalloc(&pivot_val,size_2);
+    hipMalloc(&pivot_idx,size_3);
+    hipMalloc(&iinfo,size_4);
+    hipMalloc(&work,size_5);
+    if (!scalars || (size_2 && !pivot_val) || (size_3 && !pivot_idx) || (size_4 && !iinfo) || (size_5 && !work))
         return rocblas_status_memory_error;
```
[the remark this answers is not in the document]
jzuniga-amd: Memory leaks fixed. 
x_temp, x_temp_arr, invA_arr and invA do not need to be freed; this is done by rocblas_handle destructor (see my comment above)

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
```diff
@@ -19,18 +20,29 @@ rocblas_status rocsolver_getrf_strided_batched_impl(rocblas_handle handle, const
         return st;
 
     // memory managment
+    typedef typename std::conditional<!is_complex<T>, T, decltype(std::real(T{}))>::type S;
```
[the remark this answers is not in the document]
jzuniga-amd: Changed to the nicer C++ syntax. 
(This typedef may appear in some other functions... let's change those when we spot them)

rocsolver/clients/gtest/getf2_getrf_gtest.cpp
```diff
@@ -80,8 +81,102 @@ class GETRF : public ::TestWithParam<getf2_getrf_tuple> {
     virtual void TearDown() {}
 };
 
+class GETF2_NPVT : public ::TestWithParam<getf2_getrf_tuple> {
+protected:
+    GETF2_NPVT() {}
+    virtual ~GETF2_NPVT() {}
```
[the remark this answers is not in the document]
jzuniga-amd: We can definitely do this.
As it is in all rocsolver tests, however, I will open a ticket and leave it our of this PR.

rocsolver/clients/gtest/getf2_getrf_gtest.cpp
```diff
@@ -80,8 +81,102 @@ class GETRF : public ::TestWithParam<getf2_getrf_tuple> {
...
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+};
+
+class GETRF_NPVT : public ::TestWithParam<getf2_getrf_tuple> {
+protected:
+    GETRF_NPVT() {}
+    virtual ~GETRF_NPVT() {}
```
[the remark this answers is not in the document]
jzuniga-amd: idem

rocsolver/clients/include/testing_getf2_getrf_npvt.hpp
```diff
@@ -0,0 +1,377 @@
...
+    rocblas_local_handle handle;
+    rocblas_int m = argus.M;
+    rocblas_int n = argus.N;
+    rocblas_int lda = argus.lda;
+    rocblas_stride stA = argus.bsa;
+    rocblas_stride stP = argus.bsp;
+    rocblas_int bc = argus.batch_count;
+    rocblas_int hot_calls = argus.iters;
+
+    rocblas_stride stARes = argus.unit_check || argus.norm_check ? stA : 0;
```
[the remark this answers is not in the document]
jzuniga-amd: I agree. Changed...

tfalders:
CI appears to be failing due to WaveSize (which I use in the small getri kernel) being renamed to WAVESIZE.

rocsolver/library/src/include/helpers.hpp
jzuniga-amd: Exported rocblas_trsm_template function expects matrix B to be a double pointer (T**).  We call this function using the input matrix which is normally an array of constant pointers (this is just for the batched case), so the cast is required.

jzuniga-amd:
Yes, this is one of the merging conflicts that I am solving.

rocsolver/library/src/include/helpers.hpp
cgmb: Hmmm... As long as rocblas isn't writing to it, I suppose that's fine, but casting away const is... not ideal.

This is a reasonable temporary solution, but we should probably revisit it in the future. It sounds like the 'right' solution is changing the rocblas_trsm_template interface to be const-correct.

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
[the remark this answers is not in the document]
cgmb: Perhaps I'm misunderstanding how this works, but the case I'm thinking of is where a subset of allocations fail. e.g. the allocation for `scalars` succeeds, but the allocation for `pivot_val` fails. I think that would still  result in a leak.

That sort of error path should be quite unusual, so I don't think it's likely that it would be much of a problem in practice, but it's the sort of thing that's nicely handled by RAII.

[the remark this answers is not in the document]
jzuniga-amd: Yeah, quite unusual but it could happen. That is why we need to do all memory allocations with the handle (which follows RAII). Hopefully we can get this done for 3.9. As I mentioned, we just need rocblas to export some auxiliaries which now are only accessible from rocblas functions.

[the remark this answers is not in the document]
jzuniga-amd: There is no point to do it separately. Rocsolver and rocblas use the same handle and so all memory allocations should be manage by the handle (which re-uses memory for any library calls during the entire life if the handle --until the handle destructor is called) to ensure complete integration between both libraries as there are users that uses both in the same application.

rocsolver/library/src/include/helpers.hpp
jzuniga-amd: Const is on the pointer addresses not the data.  Rocblas does write to the pointed data but it does not change the address.   Normally arguments on the batched functions are T *const []; for some reason only for TRSM they use T * []. I don't now whether this is an omission or something else is happening. I will ask Torre. If this is an omission, we could ask them to change to T *const [] for consistency; otherwise, if they are changing the address of the pointed data (which would be weird... and we would be getting errors here) then we should think of something else.

rocsolver/library/src/lapack/roclapack_getrf_strided_batched.cpp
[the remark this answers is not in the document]
cgmb: Awesome. I was going to suggest a slab allocation as a potential optimization route. Great to hear it's already part-way there.

PR #128, opened 2020-08-05
getri fixes 2

tfalders:
Size 500 matrix with lda = 600 is giving test errors. Likely cause is incorrect behaviour in the internal trmm device function. Pending a fix, this PR removes the use of the internal trmm device function.

EDIT: Cause of the bug has been identified as incorrect workspace handling. I started using the workspace within the trmm function to fix a different bug, and forgot to update the workspace allocation and strides.

### DIFF
--- a/rocsolver/clients/gtest/getri_gtest.cpp
+++ b/rocsolver/clients/gtest/getri_gtest.cpp
@@ -28,7 +28,7 @@ const vector<vector<int>> matrix_size_range = {
 
 // for daily_lapack tests
 const vector<vector<int>> large_matrix_size_range = {
-    {192, 192}, {640, 640}, {1000, 1024}, {1200, 1230} 
+    {192, 192}, {500, 600}, {640, 640}, {1000, 1024}, {1200, 1230} 
 };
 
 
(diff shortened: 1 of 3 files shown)


cgmb:
I suppose this change stems from the test failures seen in: https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/243

tfalders:
> I suppose this change stems from the test failures seen in: [ROCmSoftwarePlatform/hipBLAS#243](https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/243)

That is correct. The bug was caught by test failures in hipBLAS, which tests a matrix size that was not tested in rocSOLVER.

jzuniga-amd:
Does this will affect the performance ? We should be getting something similar to the tests that I presented yesterday, otherwise we should revisit.

tfalders:
> Does this will affect the performance ? We should be getting something similar to the tests that I presented yesterday, otherwise we should revisit.

The change to the number of threads probably will. I'll change it back to what it was before, to line up with your tests.

cgmb:
There's a linker failure in one of those tests, complaining about rocblas_dtrsm. Any idea what's going on there?

tfalders:
> There's a linker failure in one of those tests, complaining about rocblas_dtrsm. Any idea what's going on there?

Whatever it is, it shouldn't be related to this PR. That's an odd error; I'm not sure what could be causing it.

PR #129, opened 2020-08-05
Add gfx908 to CI, and get rid of Ubuntu 16 instances

amdkila:


cgmb:
Cool. Does dropping Ubuntu 16.04 support mean we have a new minimum CMake version? IIRC, that was what was driving CMake 3.5 as the minimum requirement.

saadrahim:
if the library still compiles with the older cmake, then we don't need to update the minimum cmake version.

PR #132, opened 2020-08-06
Unit test updates

tfalders:
Addresses ticket SWDEV-247003

### DIFF
--- a/rocsolver/clients/benchmarks/client.cpp
+++ b/rocsolver/clients/benchmarks/client.cpp
@@ -31,6 +31,8 @@ namespace po = boost::program_options;
 int main(int argc, char *argv[]) 
 try
 {
+    rocblas_initialize();
+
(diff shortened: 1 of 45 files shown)


cgmb:
The changes look good, too, though I haven't examined the CI failures.

tfalders:
> The changes look good, too, though I haven't examined the CI failures.

The Debug, PreCheckin, and StaticLibrary failures all seem to be the standard failures that have been plaguing us for a while, but the Extended failures seem to be new. It's a test failure for gebrd which I've noticed on my local machine too. I don't think anyone's touched gebrd in months, so I'd suspect it's due to a change in rocBLAS. The test results for gebrd are also pretty finnicky, so it might be a failure of the error computation rather than a real bug - we should update the gebrd unit test to do implicit testing ASAP.

PR #133, opened 2020-08-07
Optional runtime checks and other robustness improvements

cgmb:
The main feature of this PR is the optional `-DARMOR_LEVEL=1` cmake flag, which enables a number of compiler safety features, such as standard library assertions, stack protections and other such runtime checks. It's not enabled by default.

There's also a variety of other changes for robustness or convenience. The individual commits are not closely related to each other, so each has its own explanatory commit message. I'll fast-forward merge once when it's approved, to preserve both the linear history without throwing away the commit messages.


cgmb:
I requested @TorreZuk to also take a look, just because there's a few build-related things, and changes that address (minor) issues in code shared with rocblas.

Future work:
- `ROCSOLVER_UNREACHABLE_X(msg)`
- `ROCSOLVER_ASSUME(expr)` and `ROCSOLVER_ASSUME_X(expr, msg)` for [`__builtin_assume`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-assume)

cgmb:
It might be better to set some of the build defines and compile options as `PRIVATE`. Then we could set `ARMOR_LEVEL=1` as the default in Debug builds. They shouldn't have any impact on the C API. I just wanted to get them enabled on all C and C++ executables in our project and `PUBLIC` would propagate them...

Maybe I should just define a function to set everything on a target. In my applications, I'd usually just set `CMAKE_CXX_FLAGS` for things I wanted enabled project-wide, but I'm not sure that works as well for libraries.

cgmb:
I think the container changes are the part of this that I'm least sure about. There are clear problems in that code (setlocale in d_vector; uninitialized members in host_vector), but it's also functionality that I don't fully understand. My first instinct is to delete buggy unused code, but I'm now thinking my gutting of host_vector was probably a mistake.

The typedefs I defined for the containers that inherit from `std::vector` are also unnecessary. I had been planning on making the inheritance from `std::vector` private, and exposing the functionality we actually depend on with `using` to bring things into the public interface. I was thinking that not all `std::vector` functions make sense for these objects, but maybe I was wrong about that. I think perhaps I should just leave that public and delete those typedefs.

cgmb:
I've dropped anything I felt iffy about from the change set, so I think it's ready to for another look.

The flags for the runtime checks are now private and do not affect the rocsolver binary interface or its CMake interface, so I have enabled them by default in Debug mode. It's mostly just turning on standard library assertions, though if you enable `-O1` or higher, the optimizer will insert some additional checks on standard library calls like `memcpy` due to [`_FORTIFY_SOURCE=1`](https://access.redhat.com/blogs/766093/posts/1976213). All checks can be disabled by configuring CMake with `-DARMOR_LEVEL=0`.

In the future, there may be additional checks we can add, such as integer overflow detection with `-fwrapv` and stack protection with `-fstack-protector-strong`, but those flags do not currently interact very well with GPU code.

I've trimmed the changes I made to the rocblascommon containers to the minimum required to ensure correctness, and to add that value_type typedef.

TorreZuk:
I will give a general comment which of course can be ignored if you don't want to follow it for SOLVER, but I think it is better if multiple PRs are made for independent changes.   I don't suggest you rebase -i to do this retroactively for this PR. But in the future for example use of the updated timer style from rocBLAS could be reviewed and merged independently and so should be in a PR of it's own.    Of course a for tiny changes it makes sense to bundle changes to reduce CI load but doing atomic style PR also allows squash and merge to tidy up a long commit history when you have work that should never be rolled back independently.

cgmb:
You're right, of course. Some of these changes were originally going to be part of a completely different PR that I got pulled off of to work on tracking down various other bugs. There were some fixes in there which I wanted to build upon, in case the bugs were related, so I just dragged the whole thing in.

I can pull them into a separate PR if that's what folks want. Just point to the commits to move. My git-fu is strong.

jzuniga-amd:
I do not have the experience on this matter to really appreciate the improvements that these additions will provide... but it all looks like good stuff to me. 

I only have one concern, though: this PR will introduce some new code and changes along rocblascommon. So far we have been trying to keep this code as close as possible to the original rocblas code (hence the name rocblascommon) by mainly commenting out the things that we don't use, and trying to update the things that we use.  Adding/changing too much things will make our copy of the rocblas common code to diverge sooner than later. So, I see 3 possible scenarios:

1. If the changes proposed here are not critical for the correct functioning of rocsolver library, maybe we can hold them back, wait for similar improvements to be done in rocblas code and then follow it.

2. We can propose these changes in a PR to rocBLAS repo, then we can port them back to our rocblascommon code. This would be the ideal path to follow IMO, but also the longest one. As mentioned before the so called "rocblas common code" is not something that rocblas team is looking to maintain as common for both libraries. On the other hand, I don't see why rocBLAS cannot benefit from these improvements too, they are not particular to rocSOLVER... 

3. We can maintain, from now on, our own library-infrastructure code and don't try to keep up with rocblas changes (certainly we now have the man-power to do it). In this case we may want to change the name of rocblascommon directory to something more appropriate. 

I am ok with either option. I just think this is something we can discuss before going forward.

cgmb:
There's no easy answers for a problem like that. I kind of figured on option 2, _but_ not waiting on rocBLAS to approve our changes. If we decide these changes are good, we can integrate immediately, just like how we handle any other code in our repository. I will then send a PR off to rocBLAS for anything common to both projects.

There may be changes introduced during the review process for rocBLAS, but there's nothing fundamentally different about those changes than any other changes made in rocBLAS. We can reconcile them on an as-needed basis, just as we do with all the other changes introduced by the rocBLAS team.

Or, at least, I'd suggest trying it that way for now and see how it works out.

cgmb:
> I do not have the experience on this matter to really appreciate the improvements that these additions will provide...

Just for an example, the errors in #138 were found through these kinds of checks. Though, that problem was found by extending these checks to `host_strided_batch_vector` and `device_strided_batch_vector` (which will be a change found in my next PR). Currently, in this PR, the debug bounds checks only apply to standard library containers (and, by extension, the rocblascommon containers that inherit from `std::vector`).

Still, I expect these sorts of checks will be helpful in catching bugs during development.

PR #135, opened 2020-08-11
Fix CI for static builds with noOptimizations

cgmb:
Overriding the build command drops the static flag. Instead, just
append the noOptimizations flag.

It's not possible to build a dynamic version of rocSOLVER with a static
version of rocBLAS. The rocBLAS symbols are marked as hidden, and
cannot be used in rocsolver.so. Attempting to use them results in a
link failure like this:

    /usr/bin/ld: ../staging/example-c-basic: hidden symbol `rocblas_dtrsm' in /opt/rocm-3.7.0-6/rocblas/lib/librocblas.a(rocblas_trsm.cpp.o) is referenced by DSO
    /usr/bin/ld: final link failed: Bad value
    clang-11: error: linker command failed with exit code 1 (use -v to see invocation)


cgmb:
I'm pretty sure this fixes it, but I'm not 100% sure how to test the Jenkins groovy configuration without actually kicking off a CI build, so the proof will be in the pudding.

cgmb:
Success! We still have some run-time failures, but this problem is fixed.

PR #138, opened 2020-08-19
fix larfg unit test

jzuniga-amd:


cgmb:
This change looks good, so I'm happy to approve it, but this is really just one instance of a class of bugs relating to edge cases in our tests for zero values and array sizes. Do we want to add all the fixes to this PR, or just merge this bit now?

jzuniga-amd:
Yes, I am adding all similar fixes to other tests as discussed on teams.

PR #139, opened 2020-08-20
Check if HIP_RUNTIME equals ROCclr or rocclr

eidenyoshida:


eidenyoshida:
> Would a case-insensitive string comparison be appropriate? Admittedly, the question is somewhat academic, as I'm not sure that's actually any cleaner in CMake.
> 
> It's fine as is. I just wonder if that would avoid having to revisit this change in the future.

A similar comment was made in rocFFT. I can implement whichever you prefer.
https://github.com/ROCmSoftwarePlatform/rocFFT-internal/pull/215

cgmb:
I would prefer the `TOLOWER` version.

cgmb:
Awesome. It still needs to go through CI checks before merging, but the change looks good.

cgmb:
While there are CI failures, they are unrelated. Squash and merge at your leisure.

PR #140, opened 2020-08-20
Add geqlf

tfalders:
Addresses ticket SWDEV-247071. Includes backwards direction support for larfb and larft.


rocsolver/library/src/lapack/roclapack_geqlf_batched.cpp
```diff
@@ -0,0 +1,105 @@
+/* ************************************************************************
+ * Copyright 2019-2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#define batched
```
cgmb, line 5: This appears to exist solely to control an `#ifdef` in `rocauxiliary_larfb.hpp`. I'll have to take a closer look later, but it seems a bit unusual.

A. It uses a variable-length array (VLA), which is an easy thing to do without noticing. VLAs allocate on the stack dynamically, and have a mixed reputation. For some uses, it's easy to accidentally stackoverflow if the array size is too large. However, if the array is small, it may be a performance win over allocating on the heap.

B. If `batched` is only defined for some translation units, I think that breaks the [One Definition Rule](https://en.cppreference.com/w/cpp/language/definition) (ODR) by giving the same function a different implementation in different files. That may not be a problem in practice, but it might become a problem if we start compiling with link-time optimizations.

This involves some components of the C++ language that I don't spend much time interacting with, so I could be wrong. Perhaps you could clarify why this is used and when/where we might not want it defined vs not defined.

rocsolver/library/src/auxiliary/rocauxiliary_larfb.hpp
```diff
@@ -156,131 +156,157 @@ rocblas_status rocsolver_larfb_template(rocblas_handle handle, const rocblas_sid
...
+        if (forward) {
+            uploV = rocblas_fill_upper;
+            offsetV1 = shiftV;
+            offsetV2 = shiftV + idx2D(0,k,ldv);
+        } else {
+            uploV = rocblas_fill_lower;
+            offsetV1 = shiftV + idx2D(0,(leftside ? m-k : n-k),ldv);
+            offsetV2 = shiftV;
+        }
     }
```
jzuniga-amd, line 233: Well done!! I like how the code is used for forward and backward cases by just changing the offsets...

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -3522,6 +3522,267 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgeqr2_strided_batched(rocblas_handle
...
+    a m-by-m orthogonal/unitary matrix represented as the product of Householder matrices
+
+        Q = H(k) * ... * H(2) * H(1), with k = min(m,n)
+
+    Each Householder matrix H(i), for i = 1,2,...,k, is given by
+
+        H(i) = I - ipiv[i-1] * v(i) * v(i)'
+    
+    where the first m-i elements of the Householder vector v(i) are zero, and v(i)[i] = 1. 
+
```
jzuniga-amd, line 3545: No, the LAST m-i elements are the zero elements not the FIRST!!!

tfalders: Oops! fixed

rocsolver/library/src/lapack/roclapack_geqlf_batched.cpp
cgmb: Just for reference, I noticed there was a similar discussion about VLAs in rocBLAS (SWDEV-223708 / PR 1117).

PR #142, opened 2020-08-26
fix memory management for functions calling TRSM

jzuniga-amd:
Ignore the (TODO) MEMORY SIZE QUERIES AND ALLOCATIONS TO BE DONE WITH ROCBLAS HANDLE. We still need to decide whether we want to keep pushing and see if we can coordinate with rocblas to use the rocblas_handle, or implement our own rocsolver RAII memory model. (For now, this manual hipMalloc+free model should work)

GETRS is still calling rocblas C trsm APIs. Change it to call the rocblas_trsm_template functions will be part of ticket SWDEV-247031, so I am not doing it here.
(For now, the rocblas fixes on PR #260 should take care of problems as their memory allocation should be correct)


cgmb:
I'm still seeing the same crash in `GETRS.__double_complex/37` with rocBLAS e3c91260 when building in debug mode (including after rebasing to discard all the changes I'd made in #133).

cgmb:
The code looks good and given that there's no regression, I'm happy to approve the change. I just surprised that I'm still seeing that issue after these changes.

PR #145, opened 2020-08-29
Set xnack and sram-ecc as needed per architecture

cgmb:
I don't understand the full reasoning behind these, but this matched
what rocBLAS will set.

edit: I figured out how to include the gfx900 flag, assuming we want that.

### DIFF
--- a/rocsolver/library/src/CMakeLists.txt
+++ b/rocsolver/library/src/CMakeLists.txt
@@ -122,8 +122,9 @@ if(CMAKE_CXX_COMPILER MATCHES ".*/hipcc$")
   # explicit xnack compilation flags
   target_compile_options( rocsolver PRIVATE
     -mno-xnack
-    -Xarch_gfx906 -mno-sram-ecc
-    -Xarch_gfx908 -msram-ecc
+    "SHELL:-Xarch_gfx900 -mno-sram-ecc"
+    "SHELL:-Xarch_gfx906 -mno-sram-ecc"
+    "SHELL:-Xarch_gfx908 -msram-ecc"
   )
 endif()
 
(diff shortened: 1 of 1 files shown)


cgmb:
@kzhuravl, I'm told you wish to be included on these PRs.

rocsolver/library/src/CMakeLists.txt
```diff
@@ -118,6 +118,26 @@ else()
   target_link_libraries( rocsolver PRIVATE hip::device hcc::hccshared )
 endif()
 
+if(CMAKE_CXX_COMPILER MATCHES ".*/hipcc$")
+  # explicit xnack compilation flags
+  set( gfx900_options "-Xarch_gfx900 -mno-sram-ecc" )
```
zaliu, line 123: Note that gfx900 does not support the target feature "sram-ecc", as such we should not reference sram-ecc for gfx900; cf. [documentation](https://llvm.org/docs/AMDGPUUsage.html#processors) for details.  Thanks.

cgmb: Ah. So that was intended. Will do. Thanks!

rocsolver/library/src/CMakeLists.txt
```diff
@@ -118,6 +118,24 @@ else()
   target_link_libraries( rocsolver PRIVATE hip::device hcc::hccshared )
 endif()
 
+# explicit xnack compilation flags
+if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
+  set( gfx906_options "-Xarch_gfx906 -mno-sram-ecc" )
```
zaliu, line 123: Note also due to a bug in the "old" backend, gfx906 is implemented so that sram-ecc defaults to ON.  At this point, we need to match the buggy implementation by setting `-Xarch_gfx906 -msram-ecc` for gfx906; cf. [Austin's comments](http://ontrack-internal.amd.com/browse/SWDEV-250318).  This (repeated `-msram-ecc`) will really test out if de-duplication is applied to the added flags in pre-3.12 cmake environment.

cgmb: Thanks for pointing that out!

PR #146, opened 2020-08-31
gebrd implicit test

tfalders:
Addresses item 1 of SWDEV-247005, and should address SWDEV-246839.

SWDEV-246839 is affecting QA, so the sooner we can get this change into staging, the better.

### DIFF
--- a/rocsolver/clients/include/testing_gebd2_gebrd.hpp
+++ b/rocsolver/clients/include/testing_gebd2_gebrd.hpp
@@ -165,7 +165,7 @@ void gebd2_gebrd_initData(const rocblas_handle handle,
 }
 
 
-template <bool STRIDED, bool GEBRD, typename S, typename T, typename Sd, typename Td, typename Ud, typename Sh, typename Th, typename Uh, bool COMPLEX = is_complex<T>>
+template <bool STRIDED, bool GEBRD, typename S, typename T, typename Sd, typename Td, typename Ud, typename Sh, typename Th, typename Uh>
 void gebd2_gebrd_getError(const rocblas_handle handle, 
                         const rocblas_int m, 
                         const rocblas_int n, 
@@ -189,6 +189,9 @@ void gebd2_gebrd_getError(const rocblas_handle handle,
                         Uh &hTaup, 
                         double *max_err)
 {
+    constexpr bool COMPLEX = is_complex<T>;
+    constexpr bool VERIFY_IMPLICIT_TEST = false;
+
     std::vector<T> hW(max(m,n));
 
     // input data initialization
@@ -197,8 +200,7 @@ void gebd2_gebrd_getError(const rocblas_handle handle,
 
     // execute computations
     // use verify_implicit_test to check correctness of the implicit test using CPU lapack
-    bool verify_implicit_test = false;
-    if (!verify_implicit_test)
+    if (!VERIFY_IMPLICIT_TEST)
     {
         // GPU lapack
         CHECK_ROCBLAS_ERROR(rocsolver_gebd2_gebrd(STRIDED,GEBRD,handle, m, n, dA.data(), lda, stA, dD.data(), stD, dE.data(), stE, dTauq.data(), stQ, dTaup.data(), stP, bc));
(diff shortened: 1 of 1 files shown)


cgmb:
Not a comment about the code, but maybe I can learn something while this work is fresh in your mind.

1. Is there an easy way to identify other functions (present or future) that may need similar treatment in our tests?
2. What is it about gebrd that makes the output less directly comparable against LAPACK?

I think some of this was mentioned in discussions a while back, but I don't remember the answers.

rocsolver/clients/include/testing_gebd2_gebrd.hpp
```diff
@@ -165,7 +165,7 @@ void gebd2_gebrd_initData(const rocblas_handle handle,
 }
 
 
-template <bool STRIDED, bool GEBRD, typename S, typename T, typename Sd, typename Td, typename Ud, typename Sh, typename Th, typename Uh>
+template <bool STRIDED, bool GEBRD, typename S, typename T, typename Sd, typename Td, typename Ud, typename Sh, typename Th, typename Uh, bool COMPLEX = is_complex<T>>
```
cgmb, line 168: Unless there's a need to override that default value and set `COMPLEX` to something other than the result of `is_complex<T>`, I would just make it a `constexpr` variable within the function.

Type-dependent values are usually simpler when defined via constexpr, with the exception being if you need to sometimes be able to specify them independently from the type.

tfalders:
> Not a comment about the code, but maybe I can learn something while this work is fresh in your mind.
> 
> 1. Is there an easy way to identify other functions (present or future) that may need similar treatment in our tests?
> 2. What is it about gebrd that makes the output less directly comparable against LAPACK?
> 
> I think some of this was mentioned in discussions a while back, but I don't remember the answers.

I think the only functions that we need to look out for are those that do not have a unique answer. More specifically, I would suggest being mindful of larfg - to my knowledge, it does not have a unique answer, and the sign of the answer it does give is based on the sign of a potentially small number (thus prone to instability). The factorization computed by gebrd, I believe, is not unique, and the function calls larfg while seemingly providing ample opportunity for larfg to change sign.

rocsolver/clients/include/testing_gebd2_gebrd.hpp
tfalders: Sure, I can take care of that. The template parameter list is long enough even without the COMPLEX bool.

tfalders: I think I will also turn verify_implicit_test into a constexpr, since it should not change during run time.

cgmb: Sure. There are gotchas for using it on functions and class members, but you can pretty much never go wrong adding constexpr to a local variable.

I would note that I wouldn't expect the compiled code to change in either of these cases, though. The if would be optimized away anyway due to [sparse conditional constant propagation (sccp)](https://en.m.wikipedia.org/wiki/Sparse_conditional_constant_propagation), which clang enables at -O1.

cgmb: I should clarify: I encourage you to use constexpr liberally for local variables. It's a great practice.

The optimizer can often do it's job without the annotation, but the most reliable way to ensure compilers (and other programmers reading your code) know that a variable is a compile-time constant is by marking it as such.

I just mentioned the optimizer behaviour because I think it's worth knowing. Not to dissuade you from your plan.

rocsolver/clients/include/testing_gebd2_gebrd.hpp
```diff
@@ -196,22 +196,104 @@ void gebd2_gebrd_getError(const rocblas_handle handle,
...
+                        taup + j, a + j + j*lda, lda, hW.data());
+                if (COMPLEX)
+                    cblas_lacgv(1, taup + j, 1);
+            }
+        }
     }
    
     // error is ||hA - hARes|| / ||hA||
     // (THIS DOES NOT ACCOUNT FOR NUMERICAL REPRODUCIBILITY ISSUES. 
     // IT MIGHT BE REVISITED IN THE FUTURE)
```
jzuniga-amd, line 295: You can remove this comment now as we are not comparing with LAPACK anymore.

tfalders:
I've addressed the comments and reformatted the code. I'll wait for CI before merging. If there are any other comments, please let me know before then.

PR #143, opened 2020-08-27
Format check

saadrahim:

PR #150, opened 2020-09-10
Move docs to root directory and minor improvements

cgmb:
While reviewing the docs for the recent SVD additons, I noticed a few unrelated issues and opportunities for improvement in the docs. I've bundled those changes into this PR.

There are a few fixups for minor mistakes, and I've configured doxygen and breathe so we can link to types. To show how linking is done, I've applied it to all the rocsolver additional types.


jzuniga-amd (commented):
LGTM!

Hey, do you want to use this PR to move the docs to the root directory and re-enable doc generation test on CI?

- docs must be moved out from rocsolver/docs. 
- also, I was thinking that maybe the alternative way to build the docs using the install.sh script and docker should put the outputs in docs/build (the same as the run_doc.sh script) and not build/docs as currently does. 

What do you guys think?

cgmb:
> I was thinking that maybe the alternative way to build the docs using the install.sh script and docker should put the outputs in docs/build (the same as the run_doc.sh script) and not build/docs as currently does

I agree that they should be the same, but IMO it's `run_docs.sh` that should change. There should be one place to find all build outputs.

If we need consistency across projects for compatibility with CI, maybe we could drop a symlink from `docs/build` -> `../build/docs` for compatibility and ask @saadrahim about migrating.

tfalders:
> > I was thinking that maybe the alternative way to build the docs using the install.sh script and docker should put the outputs in docs/build (the same as the run_doc.sh script) and not build/docs as currently does
> 
> I agree that they should be the same, but IMO it's `run_docs.sh` that should change. There should be one place to find all build outputs.
> 
> If we need consistency across projects for compatibility with CI, maybe we could drop a symlink from `docs/build` -> `../build/docs` for compatibility and ask @saadrahim about migrating.

I'm okay with whatever we decide to do with the doc build, including the current approach. I do like the idea of putting everything in the build folder, however.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -5836,9 +5836,9 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zpotrf_strided_batched(
    are in S; those that converged correspond to a subset of the singular values
    of A (not necessarily ordered).
     @param[in]
-    fast_alg    rocblas_workmode. \n
-                If set to rocblas_outofplace, the function will execute the fast
-   thin-SVD version of the algorithm when possible.
+    fast_alg    #rocblas_workmode. \n
+                If set to rocblas_outofplace, the function will execute the
+   fast thin-SVD version of the algorithm when possible.
```
tfalders, line 5841: Oh, I do not like the tabbing on this, but I guess it's an issue with the automatic formatting and not this PR. Hopefully we can fix it in the other formatting PR.

tfalders (approved):
LGTM!

jzuniga-amd:
> > I was thinking that maybe the alternative way to build the docs using the install.sh script and docker should put the outputs in docs/build (the same as the run_doc.sh script) and not build/docs as currently does
> 
> I agree that they should be the same, but IMO it's `run_docs.sh` that should change. There should be one place to find all build outputs.
> 
> If we need consistency across projects for compatibility with CI, maybe we could drop a symlink from `docs/build` -> `../build/docs` for compatibility and ask @saadrahim about migrating.

Yeah, why not. I like this idea too. I don't care much whether we have docs/build -> ../build/docs or build/docs -> ../docs/build, but we do need to have the generated docs in docs/build for the CI; I am not sure if all the other libraries will agree to change as they don't even offer any docs generation tool other than run_docs.sh

rocsolver/library/include/rocsolver-functions.h
cgmb: Yeah, I did not write it that way. I think #147 disables auto-reflowing comments. I'll have to check what we were doing before #143 was merged and fix formatting in our comments.

saadrahim:
Please do not use the install.sh script for any new features. The script needs to be deprecated. It encourages bad behaviour on the part of other teams. The script is not acceptable for our customers and not conducive to spack integration. It needs to be phased out.

jzuniga-amd (changes requested):
So, let's move docs to root and re-enable CI documentation test.
As for the install script (and the new --docs flag) discussion, unless everyone disagrees, I see no harm to keep it for now. Just be sure to have docs/build <-> build/docs link.

cgmb:
I moved `rocsolver/docs` to `docs`. There are now two symlinks for comptibility:
* `docs/docBin` -> `build/docs/doxygen`
* `docs/build` -> `build/docs/sphinx`

I think overall, this layout is vastly more understandable. I also cleaned up the `run_docs.sh` and `run_doxygen.sh` scripts so they can be called from any directory.

cgmb:
Apparently not compatible enough. The new static analysis handle symlinks. Hmm...

cgmb:
I'll contribute a fix to the static analysis tool. It's pretty straight-forward.

jzuniga-amd (approved):
LGTM. Good job...
Probably good to wait for your PR in rocJenkins to be merged first and see if errors here are fixed.

cgmb:
After a bit of thought, I swapped the direction of the symlinks to ensure they're never broken. They're no longer in the repository, but will be created when building the docs.

cgmb:
I pulled in the changes from the current develop branch to try to fix the (unrelated) CI failures.

tfalders (approved):
LGTM (again)!

cgmb:
Hmm... I don't see how those CI failures in the Centos 7 gfx908 pre-checkin tests could possibly be related, but they are very consistent, and do seem isolated to this PR.

cgmb:
We actually do have a failure that looks somewhat similar in the extended tests on the develop branch. I'm going to go with my instinct that the failure in this PR is unrelated and do the merge.

The CI has improved a lot, but we definitely still have more work to do on improving the signal-to-noise ratio.

PR #151, opened 2020-09-10
Prepare comments for formatting

cgmb:
This change adjusts all of our comments that could be moved around by clang-format so that they will remain in place. I figured it would be easier to review these changes seperately from the giant reformat.


tfalders (approved):
We might want to take another look at the formatting on all the different larf and larfg calls - and probably also at the template function calls in each of the cpp files - but other than that it looks good to me!

jzuniga-amd (approved):
It looks good to me...

**But please, do not merge now. I am preparing a fix for the current problem on CI (by enabling the remaining template functions, especially on GETRS and POTRF which are the problematic tests) and I want it to go on top of feature complete freeze as it will be probably required to patch 3.9.**

cgmb:
> We might want to take another look at the formatting on all the different larf and larfg calls - and probably also at the template function calls in each of the cpp files - but other than that it looks good to me!

Why do you say that? Is it just ugly, or is there something deeper you're referring to?

tfalders:
> > We might want to take another look at the formatting on all the different larf and larfg calls - and probably also at the template function calls in each of the cpp files - but other than that it looks good to me!
> 
> Why do you say that? Is it just ugly, or is there something deeper you're referring to?

Yeah, there's something about comments splitting up the lines of a function call that just rubs me the wrong way. It also spreads it out a lot more, while I prefer function calls to be compact.

The template calls in the cpp files also tend to have a unusual commenting style that I don't know would work well with this format. What I've often seen is A and shiftA on a different line from lda and strideA, while the comment is specifically about shiftA. If the comment about the shift is no longer beside the shift, but above A and shiftA (or, worse, all 4 arguments), it might complicate the readability.

jzuniga-amd:
Yeah, I agree with Troy. The comment about the zero shift/offset of the arrays should appear right besides the 0 in the call to the rocsolver_xx_template(). On the other hand, I suppose the formatting style moves everything around.

I would suggest that we do the same as with strides and batch_count, ie. define, for example

**// Working with the entire array A
rocblas_int shiftA = 0;** 

And then in the template call no comment is needed:

**return rocsolver_xx_template(... A, shiftA, ...)**

As for the comments when calling the larfx_template functions within the code, I would simply remove them.

cgmb:
@tfalders, could you make those changes to `rocauxiliary_larfb.cpp`, `roclapack_gebd2.hpp`, `roclapack_gelq2.hpp`, `roclapack_geql2.hpp` and `roclapack_geqr2.hpp`? I get the idea, but the details still confuse me a little, so it'll probably be a lot faster for you to show me. You have access to push the changes to my `prepare-comments-for-formatting` branch.

tfalders:
> @tfalders, could you make those changes to `rocauxiliary_larfb.cpp`, `roclapack_gebd2.hpp`, `roclapack_gelq2.hpp`, `roclapack_geql2.hpp` and `roclapack_geqr2.hpp`? I get the idea, but the details still confuse me a little, so it'll probably be a lot faster for you to show me. You have access to push the changes to my `prepare-comments-for-formatting` branch.

Sure thing. I'll go do that now.

tfalders:
Based on Juan's comment, I've committed some example changes. My understanding is that we should remove all comments from the inside of function calls. In the case of shift parameters in the cpp files, we create shift variables with a zero value (similar to how we have stride variables with a zero value) and extract the "will work on the entire matrix" comment and use it to describe the variables.

cgmb:
I resolved the conflicts with the larfb changes and will merge once the CI has run on the results.

PR #154, opened 2020-09-22
Fix docs format and warnings

cgmb:
The bulk of this change is reverting the formatting of the docs to before #143 was merged. Though, I did reapply a few of the changes it made that were improvements. It depends on #147.

I also grouped our functions together, so that doxygen can recognize that the documentation applies to all the functions in the group. There should no longer be *any* warnings when building the docs.

I spent a lot of time fighting against the pre-commit hook in making these changes. When doing partial commits or rebasing, it would introduce large numbers of unintended changes, often causing merge conflicts or simply incorrect commits, and it would silently reinstall itself whenever I built an older version of the software. Fixing that script is the final step before I can call the clang-format integration done.


cgmb:
I've disabled comment reflow, so this should now pass the format checks and can be merged before #147. I'd like to do so sooner rather than later, to avoid having to make too many commits like 51dc9ca779afbeac7dda9956ab855d3912cf82b3.

tfalders (approved):
The changelog is too big for me to fully look over in-depth, but in general the changes look good.

cgmb:
Yeah, creating this was `git checkout -p a70266ff3fb5f48a2357cfe57df28505cb34b1db -- rocsolver/library/include/rocsolver-functions.h` and two hours of hitting `y`, `n` or `s`. Plus a few crafty find-and-replace commands and a bit of manual cleanup for the SVD docs which didn't exist at that time. It's a very boring change.

The only real question is whether there were any corrections to the docs since a70266ff3fb5f48a2357cfe57df28505cb34b1db that I'd have accidentally reverted.

jzuniga-amd (approved):
Looks good...

Should we remove the extra .clang-format file in .jenkins in this PR?

cgmb:
Might as well, I suppose.

cgmb:
I'll just wait for the CI to run again and then merge.

PR #155, opened 2020-09-22
Disable git hook auto-install

cgmb:
Install hooks with `./install.sh --hooks`.

Making this process explicit may prevent surprises and frustration.

### DIFF
--- a/.githooks/install
+++ b/.githooks/install
@@ -1,8 +1,0 @@
-#!/usr/bin/env bash
-
-cd $(git rev-parse --git-dir)
-cd hooks
-
-echo "Installing hooks..."
-ln -s ../../.githooks/pre-commit pre-commit
-echo "Done!"
--- a/.githooks/install
+++ b/.githooks/install
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+cd $(git rev-parse --git-dir)
+cd hooks
+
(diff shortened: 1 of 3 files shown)


tfalders (changes requested):
LGTM, but I'm a little iffy on deleting the separate install script. I'd prefer if it were more like how we handle building the docs, with separate scripts that can be called through install.sh for convenience.

jzuniga-amd:
Same here, I think it is better to keep the hook install script. As per previous discussions, we need that ALL the functionality provided by the install.sh script can be achieved other ways. To build the library,  we have cmake + make.  To build the docs, we have run_doc.sh. To install hooks, we need something in case the install.sh is deprecated (any one can make the links manually, but I prefer to have the script).

cgmb:
I deleted it because it wasn't used or mentioned anywhere. Would we prefer to just remove the auto-install from install.sh and change the README to say run `.githooks/install`?

cgmb:
I think this addresses your concerns, and I like the simplicity. Thoughts?

CI failures are due to being unable to access `/dev/dri`. I assume there's a machine that needs rebooting.

tfalders (approved):
LGTM!

jzuniga-amd (approved):
LGTM

PR #158, opened 2020-09-24
Reduce repetition in Sphinx docs

cgmb:
The fix for doxygen's undocumented function warnings in commit 611b74408022ad9f0d9a9dd954873dfa56aa139a (#154) made the Sphinx docs very repetitive. This change ensures that the functions are considered documented, but suppresses everything but the function signature for all but one function in each group.


tfalders (approved):
LGTM!

jzuniga-amd (approved):
Looks good. 
Too much job just to silent the warnings, but if you guys don't mind writing the extra delimiters and the :outline: from now on, it is fine with me.

cgmb:
It is a little verbose, though we may be able to find a more concise way of doing it in the future. The thing I like about getting rid of the undocumented function warnings is that it makes the real problems easier to spot... or, at least, the subset of problems that doxygen warns about.

PR #159, opened 2020-09-29
New memory model

jzuniga-amd:


jzuniga-amd:
More functions added...
Only 2 more group of functions to modify.
(also fixed bugs in larfg complex case that were causing memory access faults)

jzuniga-amd:
I asked rocblas team to standardize constantness of batched pointers. This is done, and the fix required on our side is in this last commit. No more need to cast out const of array of pointers.

cgmb:
What's the bugfix in larfg?

cgmb:
It would be nice to have memory functions that take an array by pointer and size. It would reduce the number of parameters we're passing, which would ever so slightly reduce the calling overhead. More importantly, would make the calls a lot more similar between functions. If we wanted to add another allocation, it could be changed in one place, without having to update every function call and signature.

e.g.
```cpp
  enum labrd_workspaces {
    labrd_scalars, // constants in rocblas calls
    labrd_work_workArr, // arrays of pointers (for batched cases) and re-usable workspace
    labrd_norms, // extra requirements for calling LARFG
    labrd_count
  };

  // memory workspace sizes:
  size_t sizes[labrd_size_count];
  rocsolver_labrd_getMemorySize<T, false>(m, n, k, batch_count, sizes, labrd_size_count);

  if (rocblas_is_device_memory_size_query(handle))
    return rocblas_set_optimal_device_memory_sizev(handle, sizes, labrd_size_count);

  // memory workspace allocation
  rocblas_device_malloc mem(handle, sizes, labrd_size_count);

  if (!mem)
    return rocblas_status_memory_error;

  T* scalars = (T *)mem[labrd_scalars];
  T sca[] = {-1, 0, 1};
  RETURN_IF_HIP_ERROR(hipMemcpy(scalars, sca, size_scalars, hipMemcpyHostToDevice));

  // execution
  return rocsolver_labrd_template<S, T>(
      handle, m, n, k, A, shiftA, lda, strideA, D, strideD, E, strideE, tauq,
      strideQ, taup, strideP, X, shiftX, ldx, strideX, Y, shiftY, ldy, strideY,
      batch_count, scalars, mem[labrd_work_workArr], (T *)mem[labrd_size_norms]);
```

jzuniga-amd:
> What's the bugfix in larfg?

fix can be seen in rocauxiliary_larfg.hpp, in kernel set_taubeta for complex case. 

I am not sure if the binary operators (+, -, *, /) overloads for rocblas_complex types are handling correctly all the combinations (real-complex, complex-real, etc.)... original code was corrupting memory when a[0] was pure imaginary. Also, although mathematically correct, the way we were assigning the sing to beta was not the same as in LAPACK (this might be the cause why the signature in the diagonal of GEBD2/GEBRD was too sensitive when comparing with LAPACK results).

Anyways, at the end, for such a simple computation, I just re-wrote everything doing all the math in-place.

TorreZuk:
Do you have the original code for complex corruption you can pass on?  We should add unit test cases to rocBLAS or rocSOLVER to confirm and test this.

jzuniga-amd:
> It would be nice to have memory functions that take an array by pointer and size. It would reduce the number of parameters we're passing, which would ever so slightly reduce the calling overhead. More importantly, would make the calls a lot more similar between functions. If we wanted to add another allocation, it could be changed in one place, without having to update every function call and signature.

That's not a bad idea!!
Main limitation is the rocblas APIs (rocblas_device_malloc and rocblas_set_optimal_size). These are variadic templates that accept an arbitrary number of sizes. I am not sure if they will work with a pointer and count. It may require changes on rocblas side which is complicated.

Also, we need to change all the getMemorySize helpers to get pointer + count. And for the idea of "changed in on place" to make sense, we may also need to change all the template functions in rocsolver upfront... otherwise you need to call them with (... scalars, mem[labrd_work_workArr], (T *)mem[labrd_size_norms]), for example; so if the workspace parameters change, the call changes and the template signature must change as well. 
To make all the required changes upfront is a lot of work...

I do like how the use of the enum could avoid errors when getting the pointers from mem[]...
Anyways, for future PRs, we can think  on some solution in between.

rocsolver/library/src/lapack/roclapack_getf2.hpp
```diff
@@ -702,23 +701,45 @@ __global__ void getf2_check_singularity(
     pivot_val[id] = 1.0 / A[idx];
 }
 
-template <typename T, typename S>
-void rocsolver_getf2_getMemorySize(const rocblas_int m,
+template <bool ISBATCHED, typename T, typename S>
```
tfalders, line 704: Why ISBATCHED instead of BATCHED?

rocsolver/library/src/lapack/roclapack_getrs.cpp
```diff
@@ -21,42 +21,46 @@ rocsolver_getrs_impl(rocblas_handle handle, const rocblas_operation trans,
...
 
-  // memory managment
-  size_t size_1; // for TRSM x_temp
-  size_t size_2; // for TRSM x_temp_arr
-  size_t size_3; // for TRSM invA
-  size_t size_4; // for TRSM invA_arr
-  rocsolver_getrs_getMemorySize<false, T>(n, nrhs, batch_count, &size_1,
-                                          &size_2, &size_3, &size_4);
+  // memory workspace sizes:
+  // size of workspace (for callinf TRSM)
```
tfalders, line 35: Spelling: calling

rocsolver/library/src/lapack/roclapack_potrf.hpp
```diff
@@ -25,51 +25,82 @@ __global__ void chk_positive(rocblas_int *iinfo, rocblas_int *info, int j) {
...
+  rocblas_int blocksReset = (batch_count - 1) / BLOCKSIZE + 1;
+  dim3 gridReset(blocksReset, 1, 1);
+  dim3 threads(BLOCKSIZE, 1, 1);
+
+  // info=0 (starting with a positive definite matrix)
+  hipLaunchKernelGGL(reset_info, gridReset, threads, 0, stream, info,
+                     batch_count, 0);
+
+  // quick return
+  if (batch_count == 0)
```
tfalders, line 101: Should this be n == 0?

tfalders (commented):
Looking good so far. There's a lot to look through, so I may not have caught everything - I've only spotted one coding error. The rest are spelling. Most spelling errors occur in multiple files, but I've only left one comment for each - they should be easy to fix with a search and replace.

jzuniga-amd:
All functions added!
Turning this into a normal PR.
Addressing comments next...

rocsolver/library/src/auxiliary/rocauxiliary_labrd.cpp
```diff
@@ -21,6 +21,12 @@ rocsolver_labrd_impl(rocblas_handle handle, const rocblas_int m,
   if (st != rocblas_status_continue)
     return st;
 
+  // working with unshifted arrays
+  rocblas_int shiftA = 0;
+  rocblas_int shiftX = 0;
+  rocblas_int shiftY = 0;
+
+  // normal (non-bacthed non-strided) execution
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/auxiliary/rocauxiliary_bdsqr.cpp
```diff
@@ -22,35 +22,40 @@ rocsolver_bdsqr_impl(rocblas_handle handle, const rocblas_fill uplo,
   if (st != rocblas_status_continue)
     return st;
 
+  // working with unshifted arrays
+  rocblas_int shiftV = 0;
+  rocblas_int shiftU = 0;
+  rocblas_int shiftC = 0;
+
+  // normal (non-bacthed non-strided) execution
```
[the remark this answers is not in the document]
jzuniga-amd: My fingers refuse to write this work correctly! 
Then I think I copied it all over the place. 
Fixed...

rocsolver/library/src/auxiliary/rocauxiliary_orglq_unglq.hpp
```diff
@@ -19,31 +19,44 @@
...
-    *size_2 = max(max(s1, s2), s3);
+    // size of workspace is maximum of what is needed by larft and larfb.
+    // size of Abyx_tmptr is maximum of what is needed by orgl2/ungl2 and larfb.
+    rocsolver_larft_getMemorySize<T, BATCHED>(jb, batch_count, &unused, &s1,
+                                              &unused);
+    rocsolver_larfb_getMemorySize<T, BATCHED>(rocblas_side_left, m - jb, n, jb,
+                                              batch_count, &s2, &s3, &unused);
+
+    *size_work = max(s1, s2);
+    *size_Abyx_tmptr = *size_Abyx_tmptr >= s3 ? *size_Abyx_tmptr : s3;
```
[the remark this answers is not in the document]
jzuniga-amd: No reason really, I don't think it makes any difference... 
I think I was just writing over the old code. 
Changed to max

rocsolver/library/src/lapack/roclapack_gebrd_batched.cpp
```diff
@@ -21,69 +21,57 @@ rocblas_status rocsolver_gebrd_batched_impl(
   if (st != rocblas_status_continue)
     return st;
 
+  // working with unshifted arrays
+  rocblas_int shiftA = 0;
+  rocblas_int shiftX = 0;
+  rocblas_int shiftY = 0;
+
+  // bacthed execution
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_gebrd_batched.cpp
```diff
@@ -21,69 +21,57 @@ rocblas_status rocsolver_gebrd_batched_impl(
...
-  hipMalloc(&X, size_5);
-  hipMalloc(&Y, size_6);
-  if (!scalars || (size_2 && !work) || (size_3 && !workArr) ||
-      (size_4 && !diag) || (size_5 && !X) || (size_6 && !Y))
+  // memory workspace sizes:
+  // size for constants in rocblas calls
+  size_t size_scalars;
+  // size of arrays of pointers (for batched cases) and re-usable workspace
+  size_t size_work_workArr;
+  // extra requirements for callling GEDB2 and LABRD
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/auxiliary/rocauxiliary_labrd.hpp
```diff
@@ -62,11 +78,11 @@ rocblas_status rocsolver_labrd_template(
     const rocblas_int k, U A, const rocblas_int shiftA, const rocblas_int lda,
     const rocblas_stride strideA, S *D, const rocblas_stride strideD, S *E,
     const rocblas_stride strideE, T *tauq, const rocblas_stride strideQ,
-    T *taup, const rocblas_stride strideP, U X, const rocblas_int shiftX,
-    const rocblas_int ldx, const rocblas_stride strideX, U Y,
+    T *taup, const rocblas_stride strideP, T *X, const rocblas_int shiftX,
+    const rocblas_int ldx, const rocblas_stride strideX, T *Y,
```
cgmb, line 82: What is the distinction between `U` and `T*`?

tfalders (approved):
LGTM!

rocsolver/library/src/lapack/roclapack_getf2.hpp
jzuniga-amd: ISBATCHED = BATCHED || STRIDED
Just to keep the name consistent.

rocsolver/library/src/auxiliary/rocauxiliary_labrd.hpp
```diff
@@ -121,31 +141,31 @@ rocblas_status rocsolver_labrd_template(
             cast2constType<T>(scalars + 2), 0, A, shiftA + idx2D(j, j + 1, lda),
             lda, strideA, A, shiftA + idx2D(j, j, lda), 1, strideA,
             cast2constType<T>(scalars + 1), 0, Y, shiftY + idx2D(j + 1, j, ldy),
-            1, strideY, batch_count, workArr);
+            1, strideY, batch_count, (T **)work_workArr);
```
cgmb, line 144: It's unfortunate we need these casts, but I don't see an obvious way to avoid them.

rocsolver/library/src/auxiliary/rocauxiliary_larf.cpp
```diff
@@ -21,41 +21,47 @@ rocblas_status rocsolver_larf_impl(rocblas_handle handle,
   if (st != rocblas_status_continue)
     return st;
 
+  // working with unshifted arrays
+  rocblas_int shiftA = 0;
+  rocblas_int shiftx = 0;
+
+  // normal (non-bacthed non-strided) execution
```
cgmb, line 28: Maybe just run:
`find "$(git rev-parse --show-toplevel)" -name '*.cpp' -exec sed -i 's/bacthed/batched/g' {} \+`

rocsolver/library/src/lapack/roclapack_getrs.cpp
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_potrf.hpp
jzuniga-amd: Both conditions are quick return; n == 0 is a bit below in the code.
The thing is that when there is batch_count >0, even if the problems has zero dimension, we still need to return info set to zero.

rocsolver/library/src/lapack/roclapack_getrs_batched.cpp
```diff
@@ -21,40 +21,44 @@ rocblas_status rocsolver_getrs_batched_impl(
...
 
-  // memory managment
-  size_t size_1; // for TRSM x_temp
-  size_t size_2; // for TRSM x_temp_arr
-  size_t size_3; // for TRSM invA
-  size_t size_4; // for TRSM invA_arr
-  rocsolver_getrs_getMemorySize<true, T>(n, nrhs, batch_count, &size_1, &size_2,
-                                         &size_3, &size_4);
+  // memory workspace sizes:
+  // size of workspace (for callinf TRSM)
```
cgmb, line 33: `find "$(git rev-parse --show-toplevel)" -name '*.cpp' -exec sed -i 's/callinf/calling/g' {} \+`

rocsolver/library/src/auxiliary/rocauxiliary_orm2r_unm2r.cpp
```diff
@@ -22,45 +22,53 @@ rocsolver_orm2r_unm2r_impl(rocblas_handle handle, const rocblas_side side,
...
-  hipMalloc(&work, size_2);
-  hipMalloc(&workArr, size_3);
-  hipMalloc(&diag, size_4);
-  if (!scalars || (size_2 && !work) || (size_3 && !workArr) ||
-      (size_4 && !diag))
+  // memory workspace sizes:
+  // size for constants in rocblas calls
+  size_t size_scalars;
+  // size of arrays of pointers (for batched cases)
+  size_t size_workArr;
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_potrf.hpp
tfalders: Both conditions are currently for batch_count == 0, I believe the n == 0 condition is missing.

rocsolver/library/src/lapack/roclapack_gelq2_batched.cpp
```diff
@@ -21,43 +21,49 @@ rocsolver_gelq2_batched_impl(rocblas_handle handle, const rocblas_int m,
   if (st != rocblas_status_continue)
     return st;
 
+  // working with unshifted arrays
+  rocblas_int shiftA = 0;
+
+  // normal (non-batched non-strided) execution
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_geql2_batched.cpp
```diff
@@ -21,43 +21,49 @@ rocsolver_geql2_batched_impl(rocblas_handle handle, const rocblas_int m,
   if (st != rocblas_status_continue)
     return st;
 
+  // working with unshifted arrays
+  rocblas_int shiftA = 0;
+
+  // normal (non-batched non-strided) execution
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_potrf.hpp
```diff
@@ -25,51 +25,82 @@ __global__ void chk_positive(rocblas_int *iinfo, rocblas_int *info, int j) {
...
+    return rocblas_status_success;
+
   // everything must be executed with scalars on the host
```
cgmb, line 104: Just while we're looking at this file, why is that? What guides the decision between scalars on host vs device?

rocsolver/library/src/auxiliary/rocauxiliary_ormbr_unmbr.cpp
```diff
@@ -21,49 +21,52 @@ rocblas_status rocsolver_ormbr_unmbr_impl(
...
-  hipMalloc(&workArr, size_3);
-  hipMalloc(&trfact, size_4);
-  hipMalloc(&workTrmm, size_5);
-  if (!scalars || (size_2 && !work) || (size_3 && !workArr) ||
-      (size_4 && !trfact) || (size_5 && !workTrmm))
+  // memory workspace sizes:
+  // requirements for calling ORMQR/UNMQR or ORMLQ/UNMLQ
+  size_t size_scalars;
+  size_t size_workArr;
+  size_t size_AbyxORwork, size_diagORtmptr;
```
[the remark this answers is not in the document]
jzuniga-amd: Order rearranged...

The names are OK. I forgot to describe this case. 

diag_tmptr means that the array is used for both functions, and its size is max(size of diag, size of tmptr). Here we are saving space. 

diagORtmptr means that the array is used for diag, with is corresponding size, or for tmptr, with its corresponding size, depending on the algorithm (in this case depending on whether k > blocksize). Here I am just trying to avoid passing too many arguments: I could have passed both arrays, diag and tmptr; one of them will have size zero depending on the algorithm.

rocsolver/library/src/auxiliary/rocauxiliary_labrd.hpp
jzuniga-amd: There might be a couple of functions with inconsistent notation but, in general, we use T for the basic type in the precision (float, double, float_complex or double_complex), and we use U for the actual type of the passed argument. If the function is called from a normal or strided_batched stream, then U = T*, if the function is called in a batched stream, then U = T *const *

In this case in particular, A could be a simple array or an array of pointers, but we require X and Y to always be simple arrays.

rocsolver/library/src/auxiliary/rocauxiliary_larf.cpp
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_getrs_batched.cpp
jzuniga-amd: fixed

rocsolver/library/src/lapack/roclapack_potrf.hpp
jzuniga-amd: Rocblas has both modes, device mode and host mode. Rocsolver does not get scalars from the user API calls. When we internally need to pass scalars to rocblas we must decide where to put them. Having the scalars on the device requires an extra synchronizing hipmemcpy. So, we will choose host mode whenever is possible. It may require switching modes depending on whatever mode the user has set for the rocblas_handle, but the possible overhead of the switch is minimal compared to the overhead of the mem copy.

rocsolver/library/src/lapack/roclapack_potrf.hpp
jzuniga-amd: Oh... you are right!!!
Corrected.

cgmb (approved):

PR #160, opened 2020-10-05
memory model fix

jzuniga-amd:


cgmb (approved):

tfalders (approved):
LGTM! Though I'm confused as to why size_scalars needs to be passed to labrd and not gebd2.

PR #161, opened 2020-10-05
Info corrections

tfalders:
Implementing tests for SWDEV-247005 revealed minor bugs in how the info parameter is updated in potrf and bdsqr. The tests are not yet ready, but the bug fixes can be included in the next release.

### DIFF
--- a/rocsolver/library/src/auxiliary/rocauxiliary_bdsqr.hpp
+++ b/rocsolver/library/src/auxiliary/rocauxiliary_bdsqr.hpp
@@ -381,8 +381,10 @@ bdsqrKernel(const rocblas_int n, const rocblas_int nv, const rocblas_int nu,
       }
     }
   }
+  
+  info[bid] = 0;
 
-  // re-arange singular values/vectors if algorithm converged
+  // re-arrange singular values/vectors if algorithm converged
   if (k == 0) {
     // all positive
     for (rocblas_int ii = 0; ii < n; ++ii) {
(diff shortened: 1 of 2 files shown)


rocsolver/library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -381,8 +381,10 @@ bdsqrKernel(const rocblas_int n, const rocblas_int nv, const rocblas_int nu,
       }
     }
   }
+  
+  info[bid] = 0;
```
jzuniga-amd, line 385: good catch!!!

rocsolver/library/src/lapack/roclapack_potrf.hpp
```diff
@@ -16,7 +16,7 @@
 
 template <typename U>
 __global__ void chk_positive(rocblas_int *iinfo, rocblas_int *info, int j) {
-  int id = hipBlockIdx_x;
+  int id = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
```
jzuniga-amd, line 19: hmmm, I see the problem here!
info is size batch_count and it is divided into several threads in possible several blocks. ..

So, you also need to check that id < batch_count (as we do for reset_info)
chk_positive  should take batch_count as argument...

Or, otherwise call chk_positive with 1 thread per block and batch_count blocks.

jzuniga-amd (changes requested):

jzuniga-amd (approved):

cgmb (approved):
I would definitely prefer to have tests accompany bug fixes whenever possible, but I suppose it's fine to merge this now as long as we make sure we don't forget them.

My only question is what the root cause of this bug was. Is the hole in our testing strategy a one-off mistake that only affects these particular functions, or is there a class of tests that we're missing across the board?

tfalders:
It's a class of tests that we're missing, which is what the ticket is meant to address. I don't believe info is currently tested in any of the functions.

PR #163, opened 2020-10-07
Fix uninitialized values in getrf/getf2 nvpt and larft

cgmb:
This fixes some uses of uninitialized variables in certain cases in the non-pivoting getrf/getf2 functions, and in the larft quick return cases. There's also a couple minor tweaks that I made as was reading through the code.

If you disable optimizations, then rocsolver-test should now pass when running in debug mode. At least, for the checkin tests. Unfortunately, those take 37 minutes on my machine when we have no optimizations enabled. Given that the debug mode can identify problems not seen in other builds, perhaps we should be running the debug mode checkin tests in the CI with a frequency similar to that of the release mode extended tests?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,11 @@ endif( )
 set( ARMOR_LEVEL "${DEFAULT_ARMOR_LEVEL}" CACHE STRING "Enables increasingly expensive runtime correctness checks" )
 include( armor-config )
 
+# set the optimization level for debug mode
+if( CMAKE_BUILD_TYPE STREQUAL "Debug" )
+  add_compile_options( -O0 )
+endif( )
+
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build rocSOLVER as a shared library" ON )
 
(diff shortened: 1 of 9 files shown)


tfalders (approved):
LGTM!

jzuniga-amd (commented):
Looks good. Good catches Cory!!!

Running debug built locally did not give me any errors for _npvt or larft. But it looks like something we definitely need to fix anyway.

I do get errors in other functions, though, but I am not sure on to how to interpret them: when using default -O3, all the complex tests fail; when using -O0 the complex tests pass but I get segfault in GESVD. If I remove _GLIBCXX_ASSERTIONS nothing fails regardless of the optimization level.

So, for me, running tests in debug mode is still not very useful. We need to figure out what is going on in future PRs. For this one: if we are sure that debug mode will require -O0, shouldn't we add this to the CMake (and the install.sh) files?

cgmb:
The failures in debug mode with optimizations enabled concern me, but it's not clear whether the assertions are revealing a problem or causing a problem.

After gels is sorted out, I was going to send a PR that copied some of the changes the rocBLAS team has made to their build scripts. I figured I'd bundle -O0 into that, but maybe I'll just change it right now.

I'm running the tests again to see if I can reproduce that GESVD crash. Do you recall what test case segfaulted? We could disable the assertions, but I kind of want to see what's going on with that one.

jzuniga-amd (approved):
LGTM then!

(This is the first test failing for me with -O0
checkin_lapack/GESVD.__float/86...
We can work it out later. Let's merge this one, as well as the formatting stuff so we can close for feature complete)

cgmb:
Odd. That one works for me. I tested rocSOLVER version 3.10.0.186-74c86f3 (with rocBLAS 2.31.0.2833-e28c19f8) on 3809-STG1 with gfx906.

PR #164, opened 2020-10-07
Added gels batched

cgmb:
The core of implementation is in there, but there's a number of questions interspersed. Perhaps you can give me some pointers. I'm still working on the tests.

### DIFF
--- a/docs/source/userguide_api.rst
+++ b/docs/source/userguide_api.rst
@@ -687,6 +687,36 @@ rocsolver_<type>gelqf_strided_batched()
    :outline:
 .. doxygenfunction:: rocsolver_sgelqf_strided_batched
 
+rocsolver_<type>gels()
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. doxygenfunction:: rocsolver_zgels
+   :outline:
+.. doxygenfunction:: rocsolver_cgels
+   :outline:
+.. doxygenfunction:: rocsolver_dgels
+   :outline:
+.. doxygenfunction:: rocsolver_sgels
+
+rocsolver_<type>gels_batched()
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. doxygenfunction:: rocsolver_zgels_batched
+   :outline:
+.. doxygenfunction:: rocsolver_cgels_batched
+   :outline:
+.. doxygenfunction:: rocsolver_dgels_batched
+   :outline:
+.. doxygenfunction:: rocsolver_sgels_batched
+
+rocsolver_<type>gels_strided_batched()
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. doxygenfunction:: rocsolver_zgels_strided_batched
+   :outline:
+.. doxygenfunction:: rocsolver_cgels_strided_batched
+   :outline:
+.. doxygenfunction:: rocsolver_dgels_strided_batched
+   :outline:
+.. doxygenfunction:: rocsolver_sgels_strided_batched
+
 General Matrix Diagonalizations
 -------------------------------
 
--- a/rocsolver/clients/common/cblas_interface.cpp
+++ b/rocsolver/clients/common/cblas_interface.cpp
@@ -105,6 +105,51 @@ void zgetrs_(char* trans,
              int* ldb,
              int* info);
 
+void sgels_(char* trans,
+            int* m,
+            int* n,
+            int* nrhs,
+            float* A,
+            int* lda,
+            float* B,
+            int* ldb,
+            float* work,
+            int* lwork,
+            int* info);
+void dgels_(char* trans,
+            int* m,
+            int* n,
+            int* nrhs,
+            double* A,
+            int* lda,
+            double* B,
+            int* ldb,
+            double* work,
+            int* lwork,
+            int* info);
+void cgels_(char* trans,
+            int* m,
+            int* n,
+            int* nrhs,
+            rocblas_float_complex* A,
+            int* lda,
+            rocblas_float_complex* B,
+            int* ldb,
+            rocblas_float_complex* work,
+            int* lwork,
+            int* info);
+void zgels_(char* trans,
+            int* m,
+            int* n,
+            int* nrhs,
+            rocblas_double_complex* A,
+            int* lda,
+            rocblas_double_complex* B,
+            int* ldb,
+            rocblas_double_complex* work,
+            int* lwork,
+            int* info);
+
 void sgetri_(int* n, float* A, int* lda, int* ipiv, float* work, int* lwork, int* info);
 void dgetri_(int* n, double* A, int* lda, int* ipiv, double* work, int* lwork, int* info);
 void cgetri_(int* n,
@@ -4020,6 +4065,75 @@ void cblas_getrs<rocblas_double_complex>(rocblas_operation trans,
     zgetrs_(&transC, &n, &nrhs, A, &lda, ipiv, B, &ldb, &info);
 }
 
+// gels
+template <>
+void cblas_gels<float>(rocblas_operation transR,
+                       rocblas_int m,
+                       rocblas_int n,
+                       rocblas_int nrhs,
+                       float* A,
+                       rocblas_int lda,
+                       float* B,
+                       rocblas_int ldb,
+                       float* work,
+                       rocblas_int lwork)
+{
+    char trans = rocblas2char_operation(transR);
+    int info;
+    sgels_(&trans, &m, &n, &nrhs, A, &lda, B, &ldb, work, &lwork, &info);
+}
+
+template <>
+void cblas_gels<double>(rocblas_operation transR,
+                        rocblas_int m,
+                        rocblas_int n,
+                        rocblas_int nrhs,
+                        double* A,
+                        rocblas_int lda,
+                        double* B,
+                        rocblas_int ldb,
+                        double* work,
+                        rocblas_int lwork)
+{
+    char trans = rocblas2char_operation(transR);
+    int info;
+    dgels_(&trans, &m, &n, &nrhs, A, &lda, B, &ldb, work, &lwork, &info);
+}
+
+template <>
+void cblas_gels<rocblas_float_complex>(rocblas_operation transR,
+                                       rocblas_int m,
+                                       rocblas_int n,
+                                       rocblas_int nrhs,
+                                       rocblas_float_complex* A,
+                                       rocblas_int lda,
+                                       rocblas_float_complex* B,
+                                       rocblas_int ldb,
+                                       rocblas_float_complex* work,
+                                       rocblas_int lwork)
+{
+    char trans = rocblas2char_operation(transR);
+    int info;
+    cgels_(&trans, &m, &n, &nrhs, A, &lda, B, &ldb, work, &lwork, &info);
+}
+
+template <>
+void cblas_gels<rocblas_double_complex>(rocblas_operation transR,
+                                        rocblas_int m,
+                                        rocblas_int n,
+                                        rocblas_int nrhs,
+                                        rocblas_double_complex* A,
+                                        rocblas_int lda,
+                                        rocblas_double_complex* B,
+                                        rocblas_int ldb,
+                                        rocblas_double_complex* work,
+                                        rocblas_int lwork)
+{
+    char trans = rocblas2char_operation(transR);
+    int info;
+    zgels_(&trans, &m, &n, &nrhs, A, &lda, B, &ldb, work, &lwork, &info);
+}
+
 // getri
 template <>
 void cblas_getri<float>(rocblas_int n,
@@ -4027,10 +4141,10 @@ void cblas_getri<float>(rocblas_int n,
                         rocblas_int lda,
                         rocblas_int* ipiv,
                         float* work,
-                        rocblas_int* lwork,
+                        rocblas_int lwork,
                         rocblas_int* info)
 {
-    sgetri_(&n, A, &lda, ipiv, work, lwork, info);
+    sgetri_(&n, A, &lda, ipiv, work, &lwork, info);
 }
 
 template <>
@@ -4039,10 +4153,10 @@ void cblas_getri<double>(rocblas_int n,
                          rocblas_int lda,
                          rocblas_int* ipiv,
                          double* work,
-                         rocblas_int* lwork,
+                         rocblas_int lwork,
                          rocblas_int* info)
 {
-    dgetri_(&n, A, &lda, ipiv, work, lwork, info);
+    dgetri_(&n, A, &lda, ipiv, work, &lwork, info);
 }
 
 template <>
@@ -4051,10 +4165,10 @@ void cblas_getri<rocblas_float_complex>(rocblas_int n,
                                         rocblas_int lda,
                                         rocblas_int* ipiv,
                                         rocblas_float_complex* work,
-                                        rocblas_int* lwork,
+                                        rocblas_int lwork,
                                         rocblas_int* info)
 {
-    cgetri_(&n, A, &lda, ipiv, work, lwork, info);
+    cgetri_(&n, A, &lda, ipiv, work, &lwork, info);
 }
 
 template <>
@@ -4063,10 +4177,10 @@ void cblas_getri<rocblas_double_complex>(rocblas_int n,
                                          rocblas_int lda,
                                          rocblas_int* ipiv,
                                          rocblas_double_complex* work,
-                                         rocblas_int* lwork,
+                                         rocblas_int lwork,
                                          rocblas_int* info)
 {
-    zgetri_(&n, A, &lda, ipiv, work, lwork, info);
+    zgetri_(&n, A, &lda, ipiv, work, &lwork, info);
 }
 
 // geqrf
--- a/rocsolver/clients/gtest/gels_gtest.cpp
+++ b/rocsolver/clients/gtest/gels_gtest.cpp
@@ -34,12 +34,12 @@ const vector<gels_params_A> matrix_sizeA_range = {
     {1, -1, 1, 1},
     {10, 10, 10, 1},
     {10, 10, 1, 10},
-    {40, 20, 40, 20},
     // not yet implemented
     {10, 1, 10, 10},
     // normal (valid) samples
     {20, 20, 20, 20},
     {30, 20, 40, 30},
+    {40, 20, 40, 40},
 };
 const vector<gels_params_B> matrix_sizeB_range = {
     // quick return
--- a/rocsolver/clients/gtest/gels_gtest.cpp
+++ b/rocsolver/clients/gtest/gels_gtest.cpp
@@ -32,14 +32,13 @@ const vector<gels_params_A> matrix_sizeA_range = {
     // invalid
     {-1, 1, 1, 1},
     {1, -1, 1, 1},
-    {1, 1, -1, 1},
-    {1, 1, 1, -1},
     {10, 10, 10, 1},
     {10, 10, 1, 10},
-    {10, 1, 10, 10},
-    /// normal (valid) samples
-    {20, 20, 20, 20},
     {40, 20, 40, 20},
+    // not yet implemented
+    {10, 1, 10, 10},
+    // normal (valid) samples
+    {20, 20, 20, 20},
     {30, 20, 40, 30},
 };
 const vector<gels_params_B> matrix_sizeB_range = {
@@ -47,6 +46,7 @@ const vector<gels_params_B> matrix_sizeB_range = {
     {0, 'N'},
     // invalid
     {-1, 'N'},
+    // not yet implemented
     {1, 'T'},
     {1, 'C'},
     // normal (valid) samples
@@ -57,10 +57,10 @@ const vector<gels_params_B> matrix_sizeB_range = {
 
 // for daily_lapack tests
 const vector<gels_params_A> large_matrix_sizeA_range = {
-    {75, 25, 75, 25},
+    {75, 25, 75, 75},
     {150, 150, 150, 150},
-    {500, 50, 600, 60},
-    {1000, 500, 1000, 500},
+    {500, 50, 600, 600},
+    {1000, 500, 1000, 1000},
 };
 const vector<gels_params_B> large_matrix_sizeB_range = {
     {100, 'N'},
@@ -88,7 +88,7 @@ Arguments gels_setup_arguments(gels_tuple tup)
 
     // only testing standard use case for strides
     // strides are ignored in normal and batched tests
-    arg.bsa = arg.lda * arg.M;
+    arg.bsa = arg.lda * arg.N;
     arg.bsc = arg.ldc * arg.K;
 
     return arg;
--- a/rocsolver/clients/gtest/gels_gtest.cpp
+++ b/rocsolver/clients/gtest/gels_gtest.cpp
@@ -16,7 +16,7 @@ typedef std::tuple<int, rocsolver_op_char> gels_params_B;
 
 typedef std::tuple<gels_params_A, gels_params_B> gels_tuple;
 
-// each A_range tuple is a {M, N, lda, ldc};
+// each A_range tuple is a {M, N, lda, ldb};
 
 // each B_range tuple is a {nrhs, trans};
 
@@ -79,7 +79,7 @@ Arguments gels_setup_arguments(gels_tuple tup)
     arg.M = std::get<0>(matrix_sizeA);
     arg.N = std::get<1>(matrix_sizeA);
     arg.lda = std::get<2>(matrix_sizeA);
-    arg.ldc = std::get<3>(matrix_sizeA);
+    arg.ldb = std::get<3>(matrix_sizeA);
 
     arg.K = std::get<0>(matrix_sizeB);
     arg.transA_option = std::get<1>(matrix_sizeB);
@@ -89,7 +89,7 @@ Arguments gels_setup_arguments(gels_tuple tup)
     // only testing standard use case for strides
     // strides are ignored in normal and batched tests
     arg.bsa = arg.lda * arg.N;
-    arg.bsc = arg.ldc * arg.K;
+    arg.bsb = arg.ldb * arg.K;
 
     return arg;
 }
--- a/rocsolver/clients/gtest/gels_gtest.cpp
+++ b/rocsolver/clients/gtest/gels_gtest.cpp
@@ -1,0 +1,257 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "testing_gels.hpp"
+
+using ::testing::Combine;
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<int, int, int, int> gels_params_A;
+typedef std::tuple<int, rocsolver_op_char> gels_params_B;
+
+typedef std::tuple<gels_params_A, gels_params_B> gels_tuple;
+
+// clang-format off
+// each A_range tuple is
+namespace A_range { enum { M, N, lda, ldc }; }
+
+// each B_range tuple is
+namespace B_range { enum { nrhs, trans }; }
+// clang-format on
+
+// case when N = nrhs = 0 will also execute the bad arguments test
+// (null handle, null pointers and invalid values)
+
+// TODO: Add tests for singularity check
+
+// for checkin_lapack tests
+const vector<gels_params_A> matrix_sizeA_range = {
+    // quick return
+    {0, 0, 0, 0},
+    // invalid
+    {-1, 1, 1, 1},
+    {1, -1, 1, 1},
+    {1, 1, -1, 1},
+    {1, 1, 1, -1},
+    {10, 10, 10, 1},
+    {10, 10, 1, 10},
+    {10, 1, 10, 10},
+    /// normal (valid) samples
+    {20, 20, 20, 20},
+    {40, 20, 40, 20},
+    {30, 20, 40, 30},
+};
+const vector<gels_params_B> matrix_sizeB_range = {
+    // quick return
+    {0, 'N'},
+    // invalid
+    {-1, 'N'},
+    {1, 'T'},
+    {1, 'C'},
+    // normal (valid) samples
+    {10, 'N'},
+    {20, 'N'},
+    {30, 'N'},
+};
+
+// for daily_lapack tests
+const vector<gels_params_A> large_matrix_sizeA_range = {
+    {75, 25, 75, 25},
+    {150, 150, 150, 150},
+    {500, 50, 600, 60},
+    {1000, 500, 1000, 500},
+};
+const vector<gels_params_B> large_matrix_sizeB_range = {
+    {100, 'N'},
+    {200, 'N'},
+    {500, 'N'},
+    {1000, 'N'},
+};
+
+Arguments gels_setup_arguments(gels_tuple tup)
+{
+    gels_params_A matrix_sizeA = std::get<0>(tup);
+    gels_params_B matrix_sizeB = std::get<1>(tup);
+
+    Arguments arg;
+
+    arg.M = std::get<A_range::M>(matrix_sizeA);
+    arg.N = std::get<A_range::N>(matrix_sizeA);
+    arg.lda = std::get<A_range::lda>(matrix_sizeA);
+    arg.ldc = std::get<A_range::ldc>(matrix_sizeA);
+
+    arg.K = std::get<B_range::nrhs>(matrix_sizeB);
+    arg.transA_option = std::get<B_range::trans>(matrix_sizeB);
+
+    arg.timing = 0;
+
+    // only testing standard use case for strides
+    // strides are ignored in normal and batched tests
+    arg.bsa = arg.lda * arg.M;
+    arg.bsc = arg.ldc * arg.K;
+
+    return arg;
+}
+
+class GELS : public ::TestWithParam<gels_tuple>
+{
+protected:
+    GELS() {}
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+};
+
+// non-batch tests
+
+TEST_P(GELS, __float)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, false, float>();
+
+    arg.batch_count = 1;
+    testing_gels<false, false, float>(arg);
+}
+
+TEST_P(GELS, __double)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, false, double>();
+
+    arg.batch_count = 1;
+    testing_gels<false, false, double>(arg);
+}
+
+TEST_P(GELS, __float_complex)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, false, rocblas_float_complex>();
+
+    arg.batch_count = 1;
+    testing_gels<false, false, rocblas_float_complex>(arg);
+}
+
+TEST_P(GELS, __double_complex)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, false, rocblas_double_complex>();
+
+    arg.batch_count = 1;
+    testing_gels<false, false, rocblas_double_complex>(arg);
+}
+
+// batched tests
+
+TEST_P(GELS, batched__float)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<true, true, float>();
+
+    arg.batch_count = 3;
+    testing_gels<true, true, float>(arg);
+}
+
+TEST_P(GELS, batched__double)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<true, true, double>();
+
+    arg.batch_count = 3;
+    testing_gels<true, true, double>(arg);
+}
+
+TEST_P(GELS, batched__float_complex)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<true, true, rocblas_float_complex>();
+
+    arg.batch_count = 3;
+    testing_gels<true, true, rocblas_float_complex>(arg);
+}
+
+TEST_P(GELS, batched__double_complex)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<true, true, rocblas_double_complex>();
+
+    arg.batch_count = 3;
+    testing_gels<true, true, rocblas_double_complex>(arg);
+}
+
+// strided_batched tests
+
+TEST_P(GELS, strided_batched__float)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, true, float>();
+
+    arg.batch_count = 3;
+    testing_gels<false, true, float>(arg);
+}
+
+TEST_P(GELS, strided_batched__double)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, true, double>();
+
+    arg.batch_count = 3;
+    testing_gels<false, true, double>(arg);
+}
+
+TEST_P(GELS, strided_batched__float_complex)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, true, rocblas_float_complex>();
+
+    arg.batch_count = 3;
+    testing_gels<false, true, rocblas_float_complex>(arg);
+}
+
+TEST_P(GELS, strided_batched__double_complex)
+{
+    Arguments arg = gels_setup_arguments(GetParam());
+
+    if(arg.M == 0 && arg.N == 0)
+        testing_gels_bad_arg<false, true, rocblas_double_complex>();
+
+    arg.batch_count = 3;
+    testing_gels<false, true, rocblas_double_complex>(arg);
+}
+
+// daily_lapack tests normal execution with medium to large sizes
+INSTANTIATE_TEST_SUITE_P(daily_lapack,
+                         GELS,
+                         Combine(ValuesIn(large_matrix_sizeA_range),
+                                 ValuesIn(large_matrix_sizeB_range)));
+
+// checkin_lapack tests normal execution with small sizes, invalid sizes,
+// quick returns, and corner cases
+INSTANTIATE_TEST_SUITE_P(checkin_lapack,
+                         GELS,
+                         Combine(ValuesIn(matrix_sizeA_range), ValuesIn(matrix_sizeB_range)));
--- a/rocsolver/clients/gtest/gels_gtest.cpp
+++ b/rocsolver/clients/gtest/gels_gtest.cpp
@@ -16,13 +16,9 @@ typedef std::tuple<int, rocsolver_op_char> gels_params_B;
 
 typedef std::tuple<gels_params_A, gels_params_B> gels_tuple;
 
-// clang-format off
-// each A_range tuple is
-namespace A_range { enum { M, N, lda, ldc }; }
+// each A_range tuple is a {M, N, lda, ldc};
 
-// each B_range tuple is
-namespace B_range { enum { nrhs, trans }; }
-// clang-format on
+// each B_range tuple is a {nrhs, trans};
 
 // case when N = nrhs = 0 will also execute the bad arguments test
 // (null handle, null pointers and invalid values)
@@ -80,13 +76,13 @@ Arguments gels_setup_arguments(gels_tuple tup)
 
     Arguments arg;
 
-    arg.M = std::get<A_range::M>(matrix_sizeA);
-    arg.N = std::get<A_range::N>(matrix_sizeA);
-    arg.lda = std::get<A_range::lda>(matrix_sizeA);
-    arg.ldc = std::get<A_range::ldc>(matrix_sizeA);
+    arg.M = std::get<0>(matrix_sizeA);
+    arg.N = std::get<1>(matrix_sizeA);
+    arg.lda = std::get<2>(matrix_sizeA);
+    arg.ldc = std::get<3>(matrix_sizeA);
 
-    arg.K = std::get<B_range::nrhs>(matrix_sizeB);
-    arg.transA_option = std::get<B_range::trans>(matrix_sizeB);
+    arg.K = std::get<0>(matrix_sizeB);
+    arg.transA_option = std::get<1>(matrix_sizeB);
 
     arg.timing = 0;
 
--- a/rocsolver/clients/include/rocsolver.hpp
+++ b/rocsolver/clients/include/rocsolver.hpp
@@ -8,6 +8,15 @@
 #include "clientcommon.hpp"
 #include "rocsolver.h"
 
+// Most functions within this file exist to provide a consistent interface for our templated tests.
+// Function overloading is used to select between the float, double, rocblas_float_complex
+// and rocblas_double_complex variants, and to distinguish the batched case (T**) from the normal
+// and strided_batched cases (T*).
+//
+// The normal and strided_batched cases are distinguished from each other by passing a boolean
+// parameter, STRIDED. Variants such as the the blocked and unblocked versions of algorithms,
+// may be provided in similar ways.
+
 // The following functions are not included in the public API and must be
 // declared
 
@@ -2723,6 +2732,170 @@ inline rocblas_status rocsolver_gelq2_gelqf(bool STRIDED,
 {
     return GELQF ? rocsolver_zgelqf_batched(handle, m, n, A, lda, ipiv, stP, bc)
                  : rocsolver_zgelq2_batched(handle, m, n, A, lda, ipiv, stP, bc);
+}
+/********************************************************/
+
+/******************** GELS ********************/
+// normal and strided_batched
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     float* A,
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     float* C,
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    if(STRIDED)
+        return rocsolver_sgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+                                               info, bc);
+    else
+        return rocsolver_sgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+}
+
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     double* A,
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     double* C,
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    if(STRIDED)
+        return rocsolver_dgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+                                               info, bc);
+    else
+        return rocsolver_dgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+}
+
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     rocblas_float_complex* A,
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     rocblas_float_complex* C,
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    if(STRIDED)
+        return rocsolver_cgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+                                               info, bc);
+    else
+        return rocsolver_cgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+}
+
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     rocblas_double_complex* A,
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     rocblas_double_complex* C,
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    if(STRIDED)
+        return rocsolver_zgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+                                               info, bc);
+    else
+        return rocsolver_zgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+}
+
+// batched
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     float* const A[],
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     float* const C[],
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    return rocsolver_sgels_batched(handle, trans, m, n, nrhs, A, lda, C, ldc, info, bc);
+}
+
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     double* const A[],
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     double* const C[],
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    return rocsolver_dgels_batched(handle, trans, m, n, nrhs, A, lda, C, ldc, info, bc);
+}
+
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     rocblas_float_complex* const A[],
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     rocblas_float_complex* const C[],
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    return rocsolver_cgels_batched(handle, trans, m, n, nrhs, A, lda, C, ldc, info, bc);
+}
+
+inline rocblas_status rocsolver_gels(bool STRIDED,
+                                     rocblas_handle handle,
+                                     rocblas_operation trans,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     rocblas_int nrhs,
+                                     rocblas_double_complex* const A[],
+                                     rocblas_int lda,
+                                     rocblas_stride stA,
+                                     rocblas_double_complex* const C[],
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    return rocsolver_zgels_batched(handle, trans, m, n, nrhs, A, lda, C, ldc, info, bc);
 }
 /********************************************************/
 
--- a/rocsolver/clients/include/rocsolver.hpp
+++ b/rocsolver/clients/include/rocsolver.hpp
@@ -14,8 +14,8 @@
 // and strided_batched cases (T*).
 //
 // The normal and strided_batched cases are distinguished from each other by passing a boolean
-// parameter, STRIDED. Variants such as the the blocked and unblocked versions of algorithms,
-// may be provided in similar ways.
+// parameter, STRIDED. Variants such as the blocked and unblocked versions of algorithms, may be
+// provided in similar ways.
 
 // The following functions are not included in the public API and must be
 // declared
--- a/rocsolver/clients/include/rocsolver.hpp
+++ b/rocsolver/clients/include/rocsolver.hpp
@@ -2746,17 +2746,17 @@ inline rocblas_status rocsolver_gels(bool STRIDED,
                                      float* A,
                                      rocblas_int lda,
                                      rocblas_stride stA,
-                                     float* C,
-                                     rocblas_int ldc,
-                                     rocblas_stride stC,
+                                     float* B,
+                                     rocblas_int ldb,
+                                     rocblas_stride stB,
                                      rocblas_int* info,
                                      rocblas_int bc)
 {
     if(STRIDED)
-        return rocsolver_sgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+        return rocsolver_sgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, B, ldb, stB,
                                                info, bc);
     else
-        return rocsolver_sgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+        return rocsolver_sgels(handle, trans, m, n, nrhs, A, lda, B, ldb, info);
 }
 
 inline rocblas_status rocsolver_gels(bool STRIDED,
@@ -2768,17 +2768,17 @@ inline rocblas_status rocsolver_gels(bool STRIDED,
                                      double* A,
                                      rocblas_int lda,
                                      rocblas_stride stA,
-                                     double* C,
-                                     rocblas_int ldc,
-                                     rocblas_stride stC,
+                                     double* B,
+                                     rocblas_int ldb,
+                                     rocblas_stride stB,
                                      rocblas_int* info,
                                      rocblas_int bc)
 {
     if(STRIDED)
-        return rocsolver_dgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+        return rocsolver_dgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, B, ldb, stB,
                                                info, bc);
     else
-        return rocsolver_dgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+        return rocsolver_dgels(handle, trans, m, n, nrhs, A, lda, B, ldb, info);
 }
 
 inline rocblas_status rocsolver_gels(bool STRIDED,
@@ -2790,17 +2790,17 @@ inline rocblas_status rocsolver_gels(bool STRIDED,
                                      rocblas_float_complex* A,
                                      rocblas_int lda,
                                      rocblas_stride stA,
-                                     rocblas_float_complex* C,
-                                     rocblas_int ldc,
-                                     rocblas_stride stC,
+                                     rocblas_float_complex* B,
+                                     rocblas_int ldb,
+                                     rocblas_stride stB,
                                      rocblas_int* info,
                                      rocblas_int bc)
 {
     if(STRIDED)
-        return rocsolver_cgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, C, ldc, stC,
+        return rocsolver_cgels_strided_batched(handle, trans, m, n, nrhs, A, lda, stA, B, ldb, stB,
                                                info, bc);
     else
-        return rocsolver_cgels(handle, trans, m, n, nrhs, A, lda, C, ldc, info);
+        return rocsolver_cgels(handle, trans, m, n, nrhs, A, lda, B, ldb, info);
 }
 
 inline rocblas_status rocsolver_gels(bool STRIDED,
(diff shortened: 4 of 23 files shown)


rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+#include "rocsolver.h"
+#include "roclapack_geqrf.hpp"
+#include "../auxiliary/rocauxiliary_ormqr_unmqr.hpp"
+
+template <typename T>
+rocblas_status
+rocsolver_gels_argCheck(rocblas_operation trans, const rocblas_int m,
+                        const rocblas_int n, const rocblas_int nrhs, T A,
+                        const rocblas_int lda, T C, const rocblas_int ldc,
+                        const rocblas_int batch_count) {
```
tfalders, line 23: Typename T is usually used to refer to the base type of the array, e.g. float, double, etc...
Typename U is used to refer to the actual array type, e.g. float*, float *const *, etc...
Please add typename U and change A and C to type U.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+template <bool BATCHED, typename T>
+rocblas_status rocsolver_gels_template(
+    rocblas_handle handle, rocblas_operation trans, const rocblas_int m,
+    const rocblas_int n, const rocblas_int nrhs, T *const *A,
+    const rocblas_int lda, T *const *C, const rocblas_int ldc,
+    rocblas_int *info, rocblas_int* solution_info,
+    const rocblas_int batch_count, T* scalars, void* void_work, void* void_workArr, void* void_diag_trfact, void* void_trfact_workTrmm) {
+  static_assert(BATCHED, "only the batched version is implemented");
+  // quick return
+  if (n == 0 || nrhs == 0 || batch_count == 0) {
```
tfalders, line 86: Or m == 0

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+  // everything must be executed with scalars on the host
+  rocblas_pointer_mode old_mode;
+  rocblas_get_pointer_mode(handle, &old_mode);
+  rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+
+  // TODO: fixup this allocation
+  const rocblas_int pivot_count_per_batch = std::min(m, n);
+  T *dIpiv;
+  hipMalloc(&dIpiv, sizeof(T) * pivot_count_per_batch * batch_count);
```
tfalders, line 101: ipiv should be allocated in the impl function and passed as a workspace. You'll need to add it to getMemorySize as well.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+  constexpr rocblas_int shiftA = 0;
+  constexpr rocblas_int shiftC = 0;
+  constexpr rocblas_stride strideA = 0;
+  constexpr rocblas_stride strideC = 0;
+  constexpr rocblas_stride strideP = 0;
+  constexpr bool STRIDED = false;
```
tfalders, line 108: These should be passed as parameters to the template function. The impl function can set them to zero if they are not needed.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+  {
+    // note: m > n
+    // compute QR factorization of A
+    T *work = (T *)void_work;
+    T *workArr = (T *)void_workArr;
+    T *diag = (T *)void_diag_trfact;
+    T **trfact = (T **)void_trfact_workTrmm;
```
tfalders, line 115: We usually perform the cast in the function argument list, particularly when the cast is only used once as it is here.

tfalders: Which will also let you get rid of the extraneous curly braces.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+    rocsolver_geqrf_template<BATCHED, STRIDED>(
+        handle, m, n, A, shiftA, lda, strideA, dIpiv, strideP, batch_count,
+        scalars, work, workArr, diag, trfact);
+    // what if the result is not success?
```
tfalders, line 119: If the result is not a success, then you haven't checked the arguments or allocated the workspaces properly. If the algorithm is unsuccessful, that will show up in any info parameters. The zero-on-the-diagonal check in trtrs (which you will need to add) is used to check whether gels will execute successfully, and can be used to set info.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+  }
+  {
+      T *work = (T *)void_work;
+      T *workArr = (T *)void_workArr;
+      T *trfact = (T *)void_diag_trfact;
+      T **workTrmm = (T **)void_trfact_workTrmm;
+      rocsolver_ormqr_unmqr_template<BATCHED,STRIDED>(
+          handle, rocblas_side_left, rocblas_operation_transpose, m, nrhs, n, A,
```
tfalders, line 127: Why not rocblas_operation_conjugate_transpose?

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+      // again, what about failure?
+  }
+  {
+      // this is implementing strtrs inline
+      // TODO: singularity check
+      void *x_temp = void_work;
+      void *x_temp_arr = void_workArr;
+      void *invA =  void_diag_trfact;
+      void *invA_arr = void_trfact_workTrmm;
+      constexpr bool optimal_memory = true;
```
tfalders, line 139: optimal_memory should be set in the impl function and passed to the template function.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,112 @@
+/* ************************************************************************
+ * Copyright (c) 2019-2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "roclapack_gels.hpp"
+#include "roclapack_geqrf.hpp"
+#include "rocblas.hpp"
+#include "../auxiliary/rocauxiliary_ormqr_unmqr.hpp"
```
tfalders, line 8: Are geqrf and ormqr needed here? Can these includes be removed?

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,112 @@
...
+//          = 'B' or 'b',   SLAMCH := base
+//          = 'P' or 'p',   SLAMCH('P') := eps*base
+
+template <typename T>
+rocblas_status rocsolver_gels_batched_impl(
+    rocblas_handle handle, rocblas_operation trans, const rocblas_int m,
+    const rocblas_int n, const rocblas_int nrhs, T *const *A,
+    const rocblas_int lda, T *const *C, const rocblas_int ldc,
+    rocblas_int *info, rocblas_int* solution_info,
+    const rocblas_int batch_count) {
```
tfalders, line 20: Add typename U and change A and C to type U.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  // logging is missing ???
+
+  // argument checking
+  rocblas_status st =
+      rocsolver_gels_argCheck(trans, m, n, nrhs, A, lda, C, ldc, batch_count);
+  if (st != rocblas_status_continue)
+    return st;
+
+  constexpr bool BATCHED = true;
```
tfalders, line 32: The rest of the library does not do this. Please remove and use true where you would use BATCHED.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,112 @@
...
+  if (rocblas_is_device_memory_size_query(handle))
+    return rocblas_set_optimal_device_memory_size(handle, size_scalars,
+                                                  size_2, size_3,
+                                                  size_4, size_5);
+  // memory workspace allocation
+  rocblas_device_malloc mem(handle, size_scalars, size_2, size_3, size_4, size_5);
+  if (!mem)
+    return rocblas_status_memory_error;
+
+  T *scalars = (T *)mem[0];
```
tfalders, line 47: Declarations of the workspace arrays is done before the declaration of mem, and they are all void pointers. The arrays other than scalars should be declared and assigned as well.

tfalders (changes requested):

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+  return rocblas_status_continue;
+}
+
+template <bool BATCHED, typename T>
+void rocsolver_gels_getMemorySize(const rocblas_int m, const rocblas_int n, const rocblas_int nrhs,
+                                   const rocblas_int batch_count,
+                                   size_t *size_scalars, size_t *size_2,
+                                   size_t *size_3, size_t *size_4, size_t *size_5) {
+  // if quick return no workspace needed
+  if (m == 0 || n == 0 || batch_count == 0) {
```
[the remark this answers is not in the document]
tfalders: One more thing: info needs to be set even if you do a quick return. Take a look at potf2 for an example.

rocsolver/library/src/lapack/roclapack_gels.hpp
tfalders: One more thing: info needs to be set even if you do a quick return. Take a look at potf2 for an example.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+  *size_5 = std::max({geqrf_trfact, ormqr_workTrmm, trsm_invA_arr});
+}
+
+template <bool BATCHED, typename T>
+rocblas_status rocsolver_gels_template(
+    rocblas_handle handle, rocblas_operation trans, const rocblas_int m,
+    const rocblas_int n, const rocblas_int nrhs, T *const *A,
+    const rocblas_int lda, T *const *C, const rocblas_int ldc,
+    rocblas_int *info, rocblas_int* solution_info,
+    const rocblas_int batch_count, T* scalars, void* void_work, void* void_workArr, void* void_diag_trfact, void* void_trfact_workTrmm) {
```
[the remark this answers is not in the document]
tfalders: STRIDED should probably be added as a template parameter as well.

rocsolver/library/src/lapack/roclapack_gels.hpp
tfalders: Oh, I'm guessing you've not been looking at the complex version of the function. Usually, when I implement LAPACK functions in rocSOLVER, I base it off the complex version of the function. This saves time later so that we don't have to go in after the fact to add complex support, and often the only thing I need to do to make it work for the real version as well is to make any conjugations conditional on type T being complex.

tfalders:
I imagine that we'll be wanting to add the non-batched and strided batched versions of gels to this PR as well, but that's something to discuss with Juan. It should be fairly easy to do once typename U has been added to the template parameters and the shift and stride parameters are passed as arguments to the template function.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
cgmb: They're required for `rocsolver_geqrf_getMemorySize` and `rocsolver_ormqr_unmqr_getMemorySize`.

tfalders: Yeah, but are they needed in the cpp file? Seems like that would be needed only in the hpp.

cgmb: Ah. Yeah. You're right! That's duplicated and can be removed.

I originally implemented batched gels in the cpp file. I later moved it to the header as I began changing my implementation to make it consistent with the rest of the library... that incomplete transition from making it work to making it consistent was the source of a lot of the things you've noted in your PR comments, but allowed me to learn one thing at a time rather than dealing with everything at once.

On the bright side, this PR has finally made the structure of the rocSOLVER library code clear to me, so I can just build things consistently from the start going forward...

docs/source/userguide_api.rst
```diff
@@ -705,6 +705,13 @@ rocsolver_<type>gelqf_strided_batched()
    :outline:
 .. doxygenfunction:: rocsolver_sgelqf_strided_batched
 
+rocsolver_<type>gels_batched()
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. doxygenfunction:: rocsolver_zgels_batched
+.. doxygenfunction:: rocsolver_cgels_batched
+.. doxygenfunction:: rocsolver_dgels_batched
+.. doxygenfunction:: rocsolver_sgels_batched
```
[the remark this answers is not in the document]
cgmb: Indeed. Started before that change; finished after. Fixed.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -6077,6 +6077,95 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(
     const rocblas_int batch_count);
 //! @}
 
+///@{
+/*! \brief GELS_BATCHED solves batches of overdetermined linear systems defined
+    by the array of m-by-n matrices A, and the array of n-by-nrhs matricies C.
```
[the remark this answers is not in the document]
cgmb: Fixed... And turned on spellcheck to catch a few other things.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -6077,6 +6077,95 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(
...
+    info      pointer to rocblas_int. Array of batch_count integers on the GPU.\n
+              If info_i = 0, successful exit for solution of A_i.
+              If info_i = j > 0, the solution for A_i could not be computed because
+              the diagonal element at A_i(j,j) is 0.
+    @param[out]
+    solution_info  pointer to rocblas_int (optional).
+              If info_i = 0, successful exit for solution of A_i.
+              If info_i = j > 0, the solution for A_i could not be computed because
+              the diagonal element at A_i(j,j) is 0.
+              TODO: why is this seperated into two for just this function?
```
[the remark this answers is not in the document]
cgmb: That makes sense.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+    return rocblas_status_invalid_size;
+
+  // 3. invalid pointers
+  if ((n && !A) || (n && !C))
+    return rocblas_status_invalid_pointer;
+
+  return rocblas_status_continue;
+}
+
+template <bool BATCHED, typename T>
```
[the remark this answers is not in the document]
cgmb: Sure!

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: Sounds good. My big question is how I should decide whether it should be true or false. I just picked true since optimal sounds good.

tfalders: So far, we always set it to true, because we always try to allocate the optimal amount of memory. I don't know if or when we'll be revisiting that in the future.

cgmb:
With the way tests are done, it's difficult to get test coverage of important cases (like parameter validity checks) without having all three implemented, so I will be including the non-batched and strided batched implementations into this PR as well. Originally, I'd thought it would be simpler to just do the one and then iterate on it as necessary, but I was mistaken.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+void rocsolver_gels_getMemorySize(const rocblas_int m, const rocblas_int n, const rocblas_int nrhs,
+                                   const rocblas_int batch_count,
+                                   size_t *size_scalars, size_t *size_2,
+                                   size_t *size_3, size_t *size_4, size_t *size_5) {
```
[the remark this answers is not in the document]
cgmb: Done, though this really makes me wish we had a data structure to bundle them all together. That's a lot of names!

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
cgmb: Done.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7169,6 +7169,319 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
                                                                  const rocblas_int batch_count);
 //! @}
 
+///@{
+/*! \brief GELS solves batches of overdetermined linear systems defined
```
tfalders, line 7173: overdetermined or underdetermined. If we don't have underdetermined yet, then return rocblas_status_not_implemented from the hpp template and make sure the gtest.cpp file won't create an underdetermined scenario. I imagine that we're going to implement it eventually, so to save on hassle make sure you code everything else under the assumption that we have underdetermined working. Take a look at larfb and larft prior to PR #140 for an example.

Also, this is not the batched version. Remove references to batches or multiple matrices.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7169,6 +7169,319 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    The problem solved by this function is of the form
+
+        min X | C - A * X |
+
+    where each result X is the least-squares solution to minimization problem i.
+
+    @param[in]
+    handle    rocblas_handle.
+    @param[in]
+    trans     rocblas_operation. Must be rocblas_operation_none.
```
tfalders, line 7186: Add a description of what trans does. Again, write under the assumption that we will match the functionality of vanilla LAPACK.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7169,6 +7169,319 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    @param[in]
+    m         rocblas_int. m >= 0.\n
+              The number of rows of all matrices A in the batch.
+    @param[in]
+    n         rocblas_int. m >= n >= 0.\n
+              The number of columns of all matrices A in the batch;
+              also the number of rows of all matrices C in the batch.
+    @param[in]
+    nrhs      rocblas_int. nrhs >= 0.\n
+              The number of columns of all matrices C and X in the batch;
```
tfalders, line 7196: I don't think we need to mention X here.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7169,6 +7169,319 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+              i.e., the matrices on the right hand side.
+    @param[inout]
+    A         array of pointers to type. Each pointer points to an array on the
+              GPU of dimension lda*n.\n
+              On entry, the m-by-n matrices A.
+              On exit, the QR factorization of A from GEQRF.
```
tfalders, line 7202: Instead of 'from', I'd prefer 'as returned by'.

rocsolver/library/src/lapack/roclapack_gels.cpp
```diff
@@ -0,0 +1,135 @@
...
+    if(st != rocblas_status_continue)
+        return st;
+
+    const rocblas_int batch_count = 1;
+    const rocblas_int shiftA = 0;
+    const rocblas_stride strideA = 0;
+    const rocblas_int shiftC = 0;
+    const rocblas_stride strideC = 0;
+    const rocblas_stride strideP = 0;
+    const bool optim_mem = true;
```
tfalders, line 35: Try to match how these are organized and commented in other functions. ATM, it looks quite messy.

rocsolver/library/src/lapack/roclapack_gels.cpp
```diff
@@ -0,0 +1,135 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "roclapack_gels.hpp"
+
+template <typename T>
```
tfalders, line 7: Add typename U to the template parameters and replace T* with U.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,142 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "roclapack_gels.hpp"
+
+template <typename T>
```
tfalders, line 7: Add typename U and replace T* const* with U.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,142 @@
...
+    if(st != rocblas_status_continue)
+        return st;
+
+    // working with unshifted arrays
+    const rocblas_int shiftA = 0;
+    const rocblas_stride strideA = 0;
+    const rocblas_int shiftC = 0;
+    const rocblas_stride strideC = 0;
+    const rocblas_stride strideP = 0;
+    const bool optim_mem = true;
```
tfalders, line 36: Same comment as for non-batched version. strideP should be non-zero since ipiv is not a batched array.

rocsolver/library/src/lapack/roclapack_gels_strided_batched.cpp
```diff
@@ -0,0 +1,152 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "roclapack_gels.hpp"
+
+template <typename T>
```
tfalders, line 7: Same comment as non-batch version

rocsolver/library/src/lapack/roclapack_gels_strided_batched.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    // argument checking
+    rocblas_status st = rocsolver_gels_argCheck(trans, m, n, nrhs, A, lda, C, ldc, batch_count);
+    if(st != rocblas_status_continue)
+        return st;
+
+    // working with unshifted arrays
+    const rocblas_int shiftA = 0;
+    const rocblas_int shiftC = 0;
+    const rocblas_stride strideP = 0;
```
tfalders, line 35: strideP should be non-zero since ipiv is not a batched array. It should probably also be separated from the shift parameters.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+                                  const rocblas_int batch_count,
+                                  size_t* size_scalars,
+                                  size_t* size_work_x_temp,
+                                  size_t* size_workArr_temp_arr,
+                                  size_t* size_diag_trfac_invA,
+                                  size_t* size_trfact_workTrmm_invA_arr,
+                                  size_t* size_ipiv)
+{
+    // if quick return no workspace needed
+    if(m == 0 || n == 0 || batch_count == 0)
```
tfalders, line 31: Or nrhs == 0

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+                                       T A,
+                                       const rocblas_int lda,
+                                       T C,
+                                       const rocblas_int ldc,
+                                       const rocblas_int batch_count = 1)
+{
+    // order is important for unit tests:
+    // 1. invalid/non-supported values
+    if(trans != rocblas_operation_none)
+        return rocblas_status_invalid_value;
```
tfalders, line 80: Might need to add cases for transpose in complex case and conj transpose in real case. Again, we should consult Juan.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    // 2. invalid size
+    if(m < 0 || n < 0 || nrhs < 0 || m < n || lda < 1 || lda < m || ldc < 1 || ldc < m
+       || ldc < n /*|| batch_count < 0*/)
```
tfalders, line 84: lda < 1 and ldc < 1 are unnecessary. Also, why is batch_count commented out?

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+        return rocblas_status_invalid_size;
+
+    // 3. invalid pointers
+    if((n && !A) || (nrhs && !C))
```
tfalders, line 88: m * n && !A
C will need different cases depending on trans, but for no trans it is m * nrhs && !C.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+                                       const rocblas_int batch_count,
+                                       T* scalars,
+                                       void* work_x_temp,
+                                       void* workArr_temp_arr,
+                                       void* diag_trfac_invA,
+                                       void* trfact_workTrmm_invA_arr,
+                                       bool optim_mem)
+{
+    // quick return
+    if(n == 0 || nrhs == 0 || batch_count == 0)
```
tfalders, line 120: Or m == 0.
You will need to set info to zero before returning if batch_count > 0. See potrf for an example.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+    rocblas_get_pointer_mode(handle, &old_mode);
+    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+
+    // note: m > n
+    // compute QR factorization of A
+    rocsolver_geqrf_template<BATCHED, STRIDED>(
+        handle, m, n, A, shiftA, lda, strideA, ipiv, strideP, batch_count, scalars, (T*)work_x_temp,
+        (T*)workArr_temp_arr, (T*)diag_trfac_invA, (T**)trfact_workTrmm_invA_arr);
+    rocsolver_ormqr_unmqr_template<BATCHED, STRIDED>(
+        handle, rocblas_side_left, rocblas_operation_transpose, m, nrhs, n, A, shiftA, lda, strideA,
```
tfalders, line 139: Make it conjugate transpose.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,157 @@
...
+    // do the equivalent of strtrs
+
+    // TODO: singularity check
+    const T one = 1; // constant 1 in host
+    // solve U*X = B, overwriting B with X
+    rocblasCall_trsm<BATCHED, T>(handle, rocblas_side_left, rocblas_fill_upper,
+                                 rocblas_operation_none, rocblas_diagonal_non_unit, n, nrhs, &one,
+                                 A, shiftA, lda, strideA, C, shiftC, ldc, strideC, batch_count,
+                                 optim_mem, work_x_temp, workArr_temp_arr, diag_trfac_invA,
+                                 trfact_workTrmm_invA_arr /*, workArr ? */);
```
tfalders, line 151: As long as A and B are the same type, I think excluding workArr should be fine.

tfalders (commented):
It's a vast improvement, but there's more work to be done. Hope these comments are helpful.

cgmb:
Sorry for pinging you. I was just pushing up some intermediate changes and marking things off as I went along. This is currently just a mess of half-finished work. Some bits, such as the commented-out `batch_count` check, are like that because I was verifying that the tests will fail when the code is wrong.

With that said, there are a number of very useful comments in there (such as the empty matrix exceptions to the parameter validity checks, which I'd been wondering about), and I will definitely incorporate your feedback.

When it's done, I'll mark the PR as "Ready for Review" and it will no longer be a draft. I'm working through some issues with my memory allocations...

rocsolver/library/src/lapack/roclapack_gels_strided_batched.cpp
cgmb: This may be the cause of my confusion. Thank you.

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: Mmm... Is the lda < 1 / ldc < 1 thing a FORTRAN thing?

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: I was wondering about that. I thought this was for all empty matrices, but it wasn't obvious from looking at other functions that was the case. Thanks!

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: Why does the type need to match?

rocsolver/clients/include/rocsolver.hpp
```diff
@@ -2604,6 +2604,170 @@ inline rocblas_status rocsolver_gelq2_gelqf(bool STRIDED,
...
+                                     rocblas_stride stA,
+                                     float* const C[],
+                                     rocblas_int ldc,
+                                     rocblas_stride stC,
+                                     rocblas_int* info,
+                                     rocblas_int bc)
+{
+    assert(!STRIDED);
+    (void)stA;
+    (void)stC;
```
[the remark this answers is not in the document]
cgmb: It's a reminder to myself that those variables are unused.

rocsolver/library/src/lapack/roclapack_gels.hpp
tfalders: The rocBLAS template will expect these to be the same type - either batched or strided. When one is strided and the other is batched, we have to match the type, which we do by putting pointers to the strided array in workArr and using workArr instead of the strided array.

rocsolver/library/src/lapack/roclapack_gels.hpp
tfalders: Possibly. It could also be a small area where we decided to do things differently than LAPACK.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,519 @@
...
+{
+    // handle
+    EXPECT_ROCBLAS_STATUS(
+        rocsolver_gels(STRIDED, nullptr, trans, m, n, nrhs, dA, lda, stA, dC, ldc, stC, info, bc),
+        rocblas_status_invalid_handle);
+
+    // values
+    EXPECT_ROCBLAS_STATUS(rocsolver_gels(STRIDED, handle, rocblas_operation(-1), m, n, nrhs, dA,
+                                         lda, stA, dC, ldc, stC, info, bc),
+                          rocblas_status_invalid_value)
```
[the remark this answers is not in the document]
cgmb: It would be nice to have the full LAPACK interface, but I've just been tasked with providing parity with CUDA to start with. As I understand it, they don't support those options, so I'm pretty sure it's fine to just stick with `rocblas_operation_none`.

There should definitely be a check that we return a status reflecting that condition, and it's on my todo list before calling this done. The main question I had on that is whether `rocblas_status_invalid_value` is the appropriate status, or if it should be something else.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,519 @@
...
+        << "Must report error when operation is invalid";
+
+    // sizes
```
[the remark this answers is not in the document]
cgmb: Yeah, I wrote these to ensure I covered all the cases, then discovered they were normally tested elsewhere. I can see why it's done that way (we're using a cross-product of inputs and we need test code filter out the invalid ones anyway), and will be deleting these tests.

rocsolver/library/include/rocsolver-functions.h
cgmb: Note that CUDA only supports the overdetermined case.

tfalders: My understanding is that our plans are to achieve parity with LAPACK. We aren't limited by what is and is not supported by CUDA.

cgmb: Awesome. This initial work on GELS was just scoped out to help port applications from CUDA to HIP, but I will definitely make sure it's ready for expansion to the full LAPACK interface.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
cgmb: I don't understand the reasoning for this. I thought we use U to handle cases where the underlying type could be either `T*` or `T**` (or some const variation thereof). Why template it on U if the type never varies?

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
cgmb: OpenGL has ruined me. I keep expecting `stride == 0` to mean "tightly packed".

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,519 @@
...
+ * ************************************************************************ */
+
+#include "cblas_interface.h"
+#include "clientcommon.hpp"
+#include "norm.hpp"
+#include "rocsolver.hpp"
+#include "rocsolver_arguments.hpp"
+#include "rocsolver_test.hpp"
+
+template <bool STRIDED, typename U>
```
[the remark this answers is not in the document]
cgmb: `U` was actually originally named `T` in the code I started with. `getrf`, I think? I changed it to U when I realized it was not the normal type `T` that we use elsewhere.

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: I tried to add a test for that case, and realized it's not possible for that first condition to occur, as we only support `m >= n`. If `m == 0`, then it must also be that `n == 0`.

rocsolver/clients/include/testing_gels.hpp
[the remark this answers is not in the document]
tfalders: Yeah, I realized while reviewing Juan's PR that this function is handled a little differently. We might want to think about looking at all these bad arg functions and maybe putting U on all of them instead of T.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
tfalders: I think there's a good argument to be made that U should not be used in the cpp files, but at the moment it's the library convention and I want to ensure consistency.

rocsolver/library/src/lapack/roclapack_gels.hpp
tfalders: This is why I suggest writing most code as if we have the full gels functionality, and only coding differently in the template function and gtest file. There are some assumptions you're baking in that will be easy to miss or forget about when it comes time to implement the remaining functionality.

cgmb: My strategy has traditionally been the opposite. If you only add code when there exists a test that will fail when it's missing, you can ensure that the tests cover all the cases.

With that said, I came to the same conclusion as you here: I should check m*n for clarity and to help ensure things aren't missed in changes later. In the long-term, though, I hope we may be able to make test improvements that make missing cases more obvious.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
cgmb: This is something that really threw me off when I was trying to understand how the library was structured. It felt like a real achievement when I figured out that U was ultimately just a single type in this case. That was the breakthrough that made the overall structure of the library clear to me, when I could contrast them with the functions in the header that actually did require a templated U.

I'm fine with changing this back to U, but I would like to add a task to our backlog to globally replace the unconstrained U template with a typedef that explicitly depends on T and the strided/strided_batched flags. Using a real type definition rather than an (undocumented?) naming convention would make the function's expected inputs more obvious.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,154 @@
...
+    *size_4 = 0;
+    *size_5 = 0;
+    return;
+  }
+
+  size_t geqrf_scalars, geqrf_work, geqrf_workArr, geqrf_diag, geqrf_trfact;
+  rocsolver_geqrf_getMemorySize<T,BATCHED>(m, n, batch_count,
+    &geqrf_scalars, &geqrf_work, &geqrf_workArr, &geqrf_diag, &geqrf_trfact);
+
+  size_t ormqr_scalars, ormqr_work, ormqr_workArr, ormqr_trfact, ormqr_workTrmm;
```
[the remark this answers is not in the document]
cgmb: `geqrf_work` and `ormqr_work` are the same size? I didn't realize that! I only did that for the scalars. I'll be sure to address that before I mark this as resolved.

Rather than pass 'unused', I marked the expression `geqrf_scalars == ormqr_scalars` as an assumption, which in Release mode can be used by the optimizer as a guarantee, and which will be checked for truth in Debug mode (where `ROCSOLVER_VERIFY_ASSUMPTIONS` is defined by default).

When I first added that assumption, I was surprised that it was immediately failing! I hadn't yet implemented the quick return, and so not all my inputs were properly validated yet. I was hitting the quick return on geqrf, but not on ormqr, resulting in differing sizes. That's almost certainly something that I would have caught anyways as I continued to implement the tests, but it was nice to have solid proof that the assertions worked.

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: My tests are passing without this change, so something about the test coverage is lacking. This is an easy fix to make, but I don't know how to test it. Do you know?

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,249 @@
...
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<vector<int>, vector<int>> gels_tuple;
+
+// each A_range vector is a {N, lda, ldb};
+
+// each B_range vector is a {nrhs, trans};
```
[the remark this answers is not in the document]
cgmb: I hadn't written this section yet. It has been entirely reworked.

I replaced the comments with constants, gave the parameters proper types, and defined an Arguments struct for the GELS' parameters. I know that's not entirely consistent with the existing tests, but it shares the same overall structure, even if not the same types.

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: I... think I understand what you're saying. I'm sure it will click next time I work with this function. Thanks!

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+Arguments gels_setup_arguments(gels_tuple tup)
+{
+    vector<int> matrix_sizeA = std::get<0>(tup);
+    vector<int> matrix_sizeB = std::get<1>(tup);
+
+    Arguments arg;
+
+    arg.M = matrix_sizeA[0];
+    arg.N = matrix_sizeB[0];
```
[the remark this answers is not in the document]
cgmb: I created a GELS_Arguments class so I could use nrhs for nrhs (among other things).

rocsolver/library/include/rocsolver-functions.h
cgmb: LAPACK includes that tidbit in their docs. I also included it because I appreciated having the reminder.

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
tfalders: Not work, workArr. Typically, functions have the same workArr size, with some exceptions.

rocsolver/library/src/lapack/roclapack_gels.hpp
tfalders: I'll have to take a look, but if you have tests for the complex cases, then perhaps the imaginary parts of the matrix entries aren't being initialized to be non-zero?

rocsolver/clients/gtest/gels_gtest.cpp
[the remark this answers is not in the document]
tfalders: No! We have a single argument class so that command line args to rocsolver-bench can all be assigned to that class. If you start creating new arguments classes, you'll have to map command line args to each class in client.cpp.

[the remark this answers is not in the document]
cgmb: Mmm... that's unfortunate. Well, at least the benchmark client seems to clarify which options I should use in the help text. Looking at `Arguments` itself, there's nothing that indicates that `nrhs` should be passed as `K`! You can understand why I was so confused trying to choose the mapping for these tests.

I'll swap it back to the old Arguments class, and pass the parameters as specified documented in the benchmark client. I think the relationships might have been more obvious had the benchmark arguments been left as dynamically typed, and it was up to the individual tests to extract what they needed from it... but that's a discussion for another time.

rocsolver/clients/common/cblas_interface.cpp
```diff
@@ -3815,6 +3860,75 @@ void cblas_getrs<rocblas_double_complex>(rocblas_operation trans,
...
+void cblas_gels<float>(rocblas_operation transR,
+                       rocblas_int m,
+                       rocblas_int n,
+                       rocblas_int nrhs,
+                       float* A,
+                       rocblas_int lda,
+                       float* B,
+                       rocblas_int ldb,
+                       float* work,
+                       rocblas_int* lwork)
```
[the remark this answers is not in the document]
cgmb: I copied how `cblas_getri` handled it, though I see `cblas_geqrf` does it differently.

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
cgmb: I'm finding that they differ for T=float, m=20, n=20m nrhs=10, batch_count=1. geqrf_workArr is returning 80, while ormqr_workArr is returning 4.

rocsolver/clients/common/cblas_interface.cpp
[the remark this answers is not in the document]
tfalders: Huh. This might be an error on getri too. Oops!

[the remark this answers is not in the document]
cgmb: I'll go ahead and correct both of them in this PR.

rocsolver/clients/gtest/gels_gtest.cpp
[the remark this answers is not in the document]
cgmb: I've added a comment to Arguments the rocsolver-bench program documents the parameters.

rocsolver/clients/include/testing_gels.hpp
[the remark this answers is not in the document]
cgmb: I left this as-is. We can revisit this in a sweep of all the tests.

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,249 @@
...
+    {30, 2},
+*/
+};
+
+// for daily_lapack tests
+const vector<vector<int>> large_matrix_sizeA_range = {
+    // {70, 70, 100}, {192, 192, 192}, {600, 700, 645}, {1000, 1000, 1000}, {1000, 2000, 2000}
+};
+const vector<vector<int>> large_matrix_sizeB_range = {
+    //    {100, 0}, {150, 0}, {200, 1}, {524, 2}, {1000, 2},
```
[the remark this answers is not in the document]
cgmb: conjugate transpose is not supported yet. Is there any significance to the values of 150 and 524?

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: My only question here is how to test this.

cgmb:
There's some discussion to be had on how best to document cases where we don't support the full LAPACK functionality yet, and I would like to expand the test coverage in our next iteration, but I'm generally happy with this PR now. We'll get the docs sorted out, and I'll swap in whatever is decided on later today. Thank you very much for the guidance, @tfalders.

@jzuniga-amd, the one thing I think is really important to do before merge is to ensure we have test cases that fail when `rocsolver_ormqr_unmqr_template` is called with `rocblas_operation_transpose` rather than `rocblas_operation_conjugate_transpose`.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+    size_t ormqr_scalars, ormqr_work, ormqr_workArr, ormqr_trfact, ormqr_workTrmm;
+    rocsolver_ormqr_unmqr_getMemorySize<T, BATCHED>(rocblas_side_left, m, nrhs, n, batch_count,
+                                                    &ormqr_scalars, &ormqr_work, &ormqr_workArr,
+                                                    &ormqr_trfact, &ormqr_workTrmm);
+
+    size_t trsm_x_temp, trsm_x_temp_arr, trsm_invA, trsm_invA_arr;
+    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, n, nrhs, batch_count, &trsm_x_temp,
+                                     &trsm_x_temp_arr, &trsm_invA, &trsm_invA_arr);
+
+    ROCSOLVER_ASSUME_X(geqrf_scalars == ormqr_scalars, "GEQRF and ORMQR use the same scalars");
```
tfalders, line 55: Is this line necessary? ormqr_scalars can instead be named unused.

cgmb: I mean, technically no. This line documents why ormqr_scalars is omitted from the calculation below, and in the debug build, it will check that the belief that that optimization is founded upon is valid.

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,257 @@
...
+typedef std::tuple<int, rocsolver_op_char> gels_params_B;
+
+typedef std::tuple<gels_params_A, gels_params_B> gels_tuple;
+
+// clang-format off
+// each A_range tuple is
+namespace A_range { enum { M, N, lda, ldc }; }
+
+// each B_range tuple is
+namespace B_range { enum { nrhs, trans }; }
```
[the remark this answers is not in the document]
cgmb: This expresses in code what is expressed in comments in other classes. It's even formatted to look almost identical! But, ok. I'll replace it. I don't want to hold up this feature over a trivial change.

rocsolver/library/src/lapack/roclapack_gels.cpp
```diff
@@ -0,0 +1,141 @@
...
+        return st;
+
+    // working with unshifted arrays
+    const rocblas_int shiftA = 0;
+    const rocblas_int shiftC = 0;
+
+    // normal (non-batched non-strided) execution
+    const rocblas_stride strideA = 0;
+    const rocblas_stride strideC = 0;
+    const rocblas_stride strideP = 0;
```
jzuniga-amd, line 36: strideP is an internal value that only make sense in the context of the template function, it should be calculated inside the template_function body, not passed as a parameter.  

If gels_template() is called from another place in rocsolver (not this implementation), we may not know a suitable value for strideP. Passing an arbitrary value may not be compatible with what the gels algorithm will put inside the array ipiv, which is a workspace array, not part of the API.

See also related comments inside roclapack_gels.hpp below.

rocsolver/library/src/lapack/roclapack_gels_batched.cpp
```diff
@@ -0,0 +1,147 @@
...
+        return st;
+
+    // working with unshifted arrays
+    const rocblas_int shiftA = 0;
+    const rocblas_int shiftC = 0;
+
+    // batched execution
+    const rocblas_stride strideA = 0;
+    const rocblas_stride strideC = 0;
+    const rocblas_stride strideP = std::min(m, n);
```
jzuniga-amd, line 37: See related comments above

rocsolver/library/src/lapack/roclapack_gels_strided_batched.cpp
```diff
@@ -0,0 +1,157 @@
...
+    rocblas_status st = rocsolver_gels_argCheck(trans, m, n, nrhs, A, lda, C, ldc, info, batch_count);
+    if(st != rocblas_status_continue)
+        return st;
+
+    // working with unshifted arrays
+    const rocblas_int shiftA = 0;
+    const rocblas_int shiftC = 0;
+
+    // contiguous pivot array
+    const rocblas_stride strideP = std::min(m, n);
```
jzuniga-amd, line 37: idem

jzuniga-amd (changes requested):
I will review the documentation in rocsolver-functions.h once it is updated.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7287,6 +7287,334 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    \p rocblas_operation_transpose or \p rocblas_operation_conjugate_transpose.
+
+    @param[in]
+    handle    rocblas_handle.
+    @param[in]
+    trans     rocblas_operation.\n
+              Specifies the form of the system of equations.
+    @param[in]
+    m         rocblas_int. m >= 0.\n
+              The number of rows of all matrices A in the batch.
```
jzuniga-amd, line 7324: the number of rows of matrix A...
this is the non batched version

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7287,6 +7287,334 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The number of columns of matrix A;
+              also the number of rows of C.
```
jzuniga-amd, line 7328: No, this is not correct!
The number of rows of C depends on whether the system is transposed or not. 
n is just the number of columns of A

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7287,6 +7287,334 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    @param[in]
+    nrhs      rocblas_int. nrhs >= 0.\n
+              The number of columns of matrices C and X;
+              i.e., the columns on the right hand side.
+    @param[inout]
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the m-by-n matrix A.
+              On exit, the QR factorization of A as returned by GEQRF.
```
jzuniga-amd, line 7336: The orthogonal factorization of A as returned by GEQRF or GELQF

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7287,6 +7287,334 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    @param[in]
+    lda       rocblas_int. lda >= m.\n
+              Specifies the leading dimension of matrix A.
+    @param[inout]
+    C         pointer to type. Array on the GPU of dimension ldc*nrhs.\n
+              On entry, the m-by-nrhs matrix C.
```
jzuniga-amd, line 7342: Dimensions of C are: m-by-nrhs if non-transposed, or n-by-nrhs if transposed !

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7287,6 +7287,334 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+              On exit, the first n rows of each matrix contain the solution vectors.
```
jzuniga-amd, line 7343: This is not correct in general. 
We can just say something like:
"On exit, when info = 0, C is overwritten by the solution vectors stored as columns. "

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: In this case, it's just an assertion. When reviewing various PRs, it was oven unclear to me why certain variables in the memory calculation were unused. I seem to even recall asking Troy about one in your code and he didn't know either. If I were to read the code without this, I would wonder: "How do you know there's enough room for ormqr to do its work if you didn't actually use all the values returned from its getMemorySize?"

This answers that question with: "I know because GEQRF and ORMQR use the same scalars. (It's the same buffer and thus must be the same size.)"

In debug mode, this condition is checked for validity. It doesn't have a good error message yet, but I plan to continue to improve these checks iteratively.

In release mode, this condition is given to the optimizer as an invariant that can be assumed true without checking, and used to improve code generation. That probably has no relevance to this case, but it will be nice to have this macro in the codebase already for cases in which it does.

This approach caught an error when it was suggested that I should ignore the size of ormqr_workArr. Since it is not necessarily possible to catch out-of-bounds reads or writes via tests, I think trying to catch mistakes in memory allocation early is worthwhile.

There's [a similar macro in Qt with more extensive documentation](https://doc.qt.io/qt-5/qtglobal.html#Q_ASSUME), but in either case, in release mode it's just a wrapper around clang's [__builtin_assume](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-assume).

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -7287,6 +7287,334 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetri_strided_batched(rocblas_handle
...
+    @param[in]
+    ldc       rocblas_int. ldc >= max(m,n).\n
+              Specifies the leading dimension of matrix C.
+    @param[out]
+    info      pointer to rocblas_int on the GPU.\n
+              If info = 0, successful exit.
+              If info = j > 0, the solution for A could not be computed because
+              the diagonal element at A(j,j) is 0.
```
jzuniga-amd, line 7351: Yes,  the element (j,j) of array A is zero, but this array does not contain matrix A anymore, it has its orthogonal factorization (in other words, the original matrix A could have zeros along the diagonal, but that does not mean it is singular). 

The correct documentation here should read something like:

If info = j > 0, the solution could not be computed because input matrix A is singular; the j-th diagonal element of its triangular factor is zero.

rocsolver/library/include/rocsolver-functions.h
cgmb: Whoops. I really thought I'd got all those.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+        (T*)workArr_temp_arr, (T*)diag_trfac_invA, (T**)trfact_workTrmm_invA_arr);
+    rocsolver_ormqr_unmqr_template<BATCHED, STRIDED>(
+        handle, rocblas_side_left, rocblas_operation_conjugate_transpose, m, nrhs, n, A, shiftA,
+        lda, strideA, ipiv, strideP, C, shiftC, ldc, strideC, batch_count, scalars, (T*)work_x_temp,
+        (T*)workArr_temp_arr, (T*)diag_trfac_invA, (T**)trfact_workTrmm_invA_arr);
+
+    // do the equivalent of strtrs
+    const rocblas_int check_threads = min(((n - 1) / 64 + 1) * 64, BLOCKSIZE);
+    hipLaunchKernelGGL(check_singularity<T>, dim3(batch_count, 1, 1), dim3(1, check_threads, 1), 0,
+                       stream, n, A, shiftA, lda, strideA, info);
```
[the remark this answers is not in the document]
cgmb: That is very informative. I'd assumed trsm handled that itself, but what you're saying makes much more sense. Thanks for the help with this. I'll leave that strtrs comment everything below it in this file untouched to avoid conflicting with your changes.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+    // handle
+    EXPECT_ROCBLAS_STATUS(
+        rocsolver_gels(STRIDED, nullptr, trans, m, n, nrhs, dA, lda, stA, dC, ldc, stC, info, bc),
+        rocblas_status_invalid_handle);
+
+    // values
+    EXPECT_ROCBLAS_STATUS(rocsolver_gels(STRIDED, handle, rocblas_operation(-1), m, n, nrhs, dA,
+                                         lda, stA, dC, ldc, stC, info, bc),
+                          rocblas_status_invalid_value)
+        << "Must report error when operation is invalid";
```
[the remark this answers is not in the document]
cgmb: These are printed when the test fails. I wrote them to describe in plain English what requirement the software is not meeting when it fails the test.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+    EXPECT_ROCBLAS_STATUS(rocsolver_gels(STRIDED, handle, rocblas_operation_transpose, m, n, nrhs,
+                                         dA, lda, stA, dC, ldc, stC, info, bc),
+                          rocblas_status_invalid_value)
+        << "Must report error when operation is transpose (unsupported)";
+    EXPECT_ROCBLAS_STATUS(rocsolver_gels(STRIDED, handle, rocblas_operation_conjugate_transpose, m,
+                                         n, nrhs, dA, lda, stA, dC, ldc, stC, info, bc),
+                          rocblas_status_invalid_value)
+        << "Must report error when opration is conjugate transpose (unsupported)";
+
```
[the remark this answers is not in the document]
cgmb: Rodger.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+        << "Should normally report error when C is null";
+    EXPECT_ROCBLAS_STATUS(
+        rocsolver_gels(STRIDED, handle, trans, m, n, nrhs, dA, lda, stA, dC, ldc, stC, nullptr, bc),
+        rocblas_status_invalid_pointer)
+        << "Should normally report error when info is null";
+
+    // quick return with invalid pointers
+    EXPECT_ROCBLAS_STATUS(rocsolver_gels(STRIDED, handle, trans, m, 0, nrhs, (U) nullptr, lda, stA,
+                                         dC, ldc, stC, info, bc),
+                          rocblas_status_success)
```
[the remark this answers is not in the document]
cgmb: Added.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+    {
+        rocblas_init<T>(hA, true);
+        rocblas_init<T>(hC, true);
+
+        // scale A to avoid singularities
+        for(rocblas_int b = 0; b < bc; ++b)
+        {
+            for(rocblas_int i = 0; i < m; i++)
+            {
+                for(rocblas_int j = 0; j < m; j++)
```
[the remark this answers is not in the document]
cgmb: Hmmm.... I copy/pasted this from getrs. I suppose that must operate on square matrices. At some point we should probably pull this into a function where those sorts of parameter differences are more obvious.

It makes no difference to correctness, but this (and getrs) iterate over i and j in the wrong order. That would also be easier to fix globally if it was defined in one place.

In any case. I fixed the correctness issue.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+
+    // error is ||hC - hCRes|| / ||hC||
+    // (THIS DOES NOT ACCOUNT FOR NUMERICAL REPRODUCIBILITY ISSUES.
+    // IT MIGHT BE REVISITED IN THE FUTURE)
+    // using vector-induced infinity norm
+    double err;
+    *max_err = 0;
+    for(rocblas_int b = 0; b < bc; ++b)
+    {
+        err = norm_error('I', m, nrhs, ldc, hC[b], hCRes[b]);
```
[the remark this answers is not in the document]
cgmb: Ok.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+                              rocblas_status_invalid_value);
+
+        if(argus.timing)
+            ROCSOLVER_BENCH_INFORM(2);
+
+        return;
+    }
+
+    // determine sizes
+    size_t size_A = size_t(lda) * m;
```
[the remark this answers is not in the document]
cgmb: Fixed.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+    size_t size_C = size_t(ldc) * nrhs;
+    double max_error = 0, gpu_time_used = 0, cpu_time_used = 0;
+
+    size_t size_CRes = (argus.unit_check || argus.norm_check) ? size_C : 0;
+
+    // check invalid sizes
+    bool invalid_size
+        = (m < 0 || n < 0 || nrhs < 0 || m < n || lda < m || ldc < m || ldc < n || bc < 0);
```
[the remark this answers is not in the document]
cgmb: Moved m < n to non-supported values.

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -0,0 +1,498 @@
...
+        // collect performance data
+        if(argus.timing)
+            gels_getPerfData<STRIDED, T>(handle, trans, m, n, nrhs, dA, lda, stA, dC, ldc, stC,
+                                         dInfo, bc, hA, hC, hInfo, &gpu_time_used, &cpu_time_used,
+                                         hot_calls, argus.perf);
+    }
+    // validate results for rocsolver-test
+    // using m * machine_precision as tolerance
+    if(argus.unit_check)
+        rocsolver_test_check<T>(max_error, m);
```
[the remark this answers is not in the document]
cgmb: I haven't had a chance to look closely at what the `rocsolver_test_check` does, so please do check that I understood you correctly.

rocsolver/library/include/rocsolver-functions.h
cgmb: Removed.

rocsolver/library/include/rocsolver-functions.h
cgmb: Fixed.

rocsolver/library/include/rocsolver-functions.h
cgmb: Fixed.

rocsolver/library/include/rocsolver-functions.h
cgmb: Changed.

rocsolver/library/include/rocsolver-functions.h
cgmb: That makes a lot more sense.

rocsolver/clients/include/testing_gels.hpp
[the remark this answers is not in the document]
cgmb: Or rather, added with rocblas_not_implemented as the return value, as this case is not currently possible.

rocsolver/clients/include/testing_gels.hpp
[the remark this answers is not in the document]
cgmb: Hmmm... I'm not understanding something here, because I'm segfaulting on the C host -> device transfer after I replaced the bounds of j with n. I assume I got i and j backwards, but if A is an lda*n array, then shouldn't each n be lda apart? Ugh. Well, I'll revert that change and debug it.

I wish it was easier to see what was actually allocated.

rocsolver/library/include/rocsolver-functions.h
cgmb: Though, is this not an issue in the brief? I've removed "an n-by-nrhs matrix C" at the top. Maybe take a look at the replacement to ensure I'm not accidentally implying something I shouldn't. I don't speak the same language as mathematicians.

rocsolver/library/src/lapack/roclapack_gels.hpp
cgmb: Done.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+                                       rocblas_int* info,
+                                       const rocblas_int batch_count = 1)
+{
+    // order is important for unit tests:
+    // 1. invalid/non-supported values
+    if(trans != rocblas_operation_none)
+        return rocblas_status_invalid_value;
+
+    // 2. invalid size
+    if(m < 0 || n < 0 || nrhs < 0 || m < n || lda < m || ldc < m || ldc < n || batch_count < 0)
```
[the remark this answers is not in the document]
cgmb: Done.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+        return rocblas_status_success;
+
+    // everything must be executed with scalars on the host
+    rocblas_pointer_mode old_mode;
+    rocblas_get_pointer_mode(handle, &old_mode);
+    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+
+    // TODO: apply scaling to improve accuracy over a larger range of values
+
+    // note: m >= n
```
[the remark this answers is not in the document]
cgmb: That's already covered by the arg check.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+                                       U A,
+                                       const rocblas_int shiftA,
+                                       const rocblas_int lda,
+                                       const rocblas_stride strideA,
+                                       U C,
+                                       const rocblas_int shiftC,
+                                       const rocblas_int ldc,
+                                       const rocblas_stride strideC,
+                                       T* ipiv,
+                                       const rocblas_stride strideP,
```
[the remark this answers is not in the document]
cgmb: Done.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+                                       rocblas_int* info,
+                                       const rocblas_int batch_count,
+                                       T* scalars,
+                                       void* work_x_temp,
+                                       void* workArr_temp_arr,
+                                       void* diag_trfac_invA,
+                                       void* trfact_workTrmm_invA_arr,
```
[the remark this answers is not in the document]
cgmb: Done.

cgmb:
To do:
- [ ] Update GELS_BATCHED docs
- [ ] Update GELS_STRIDED_BATCHED docs
- [ ] Zero C in quick return
- [ ] Rename C to B
- [ ] Skip TRSM on failed singularity check

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+    rocblas_int blocksReset = (batch_count - 1) / BLOCKSIZE + 1;
+    dim3 gridReset(blocksReset, 1, 1);
+    dim3 threads(BLOCKSIZE, 1, 1);
+
+    // info=0 (starting with a nonsingular matrix)
+    hipLaunchKernelGGL(reset_info, gridReset, threads, 0, stream, info, batch_count, 0);
+
+    // quick return if A or C are empty
+    if(m == 0 || n == 0 || nrhs == 0)
+        return rocblas_status_success;
```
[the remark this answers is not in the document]
cgmb: Ah. There can be no test case. It's not possible. For `rocblas_operation_none` and `m >= n`, C contains n rows of solutions.

- If `nrhs == 0`, then C is empty.
- If `n == 0` then C is empty.
- If `m == 0` then since `m >= n`, we know `n == 0`.

I'll add a TODO there to remind us of this (though that sort of error should be caught by tests anyway).

rocsolver/clients/include/testing_gels.hpp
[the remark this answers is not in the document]
cgmb: Was calling initData with n and m reversed. Fixed.

rocsolver/library/src/lapack/roclapack_gels.hpp
```diff
@@ -0,0 +1,172 @@
...
+        handle, rocblas_side_left, rocblas_operation_conjugate_transpose, m, nrhs, n, A, shiftA,
+        lda, strideA, ipiv, strideP, C, shiftC, ldc, strideC, batch_count, scalars, (T*)work_x_temp,
+        (T*)workArr_temp_arr, (T*)diag_trfac_invA, (T**)trfact_workTrmm_invA_arr);
+
+    // do the equivalent of strtrs
+    const rocblas_int check_threads = min(((n - 1) / 64 + 1) * 64, BLOCKSIZE);
+    hipLaunchKernelGGL(check_singularity<T>, dim3(batch_count, 1, 1), dim3(1, check_threads, 1), 0,
+                       stream, n, A, shiftA, lda, strideA, info);
+    const T one = 1; // constant 1 in host memory
+    // solve U*X = B, overwriting B with X
```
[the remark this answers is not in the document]
cgmb: Fixed this, because the switch from B to C would make it even more confusing and we probably need a merge for these lines anyway.

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
cgmb: Turned this comment into an assertion with a comment to explain why it holds true.

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,257 @@
...
+using namespace std;
+
+typedef std::tuple<int, int, int, int> gels_params_A;
+typedef std::tuple<int, rocsolver_op_char> gels_params_B;
+
+typedef std::tuple<gels_params_A, gels_params_B> gels_tuple;
+
+// clang-format off
+// each A_range tuple is
+namespace A_range { enum { M, N, lda, ldc }; }
```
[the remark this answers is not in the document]
cgmb: Fixed.

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,257 @@
...
+    {-1, 1, 1, 1},
+    {1, -1, 1, 1},
+    {1, 1, -1, 1},
+    {1, 1, 1, -1},
+    {10, 10, 10, 1},
+    {10, 10, 1, 10},
+    {10, 1, 10, 10},
+    /// normal (valid) samples
+    {20, 20, 20, 20},
+    {40, 20, 40, 20},
```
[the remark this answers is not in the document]
jzuniga-amd: This has nothing to do with the test structure, the test executes whatever you tell it to execute. 

The problem is that if you did not understand that ldb < max(m,n) was invalid, and you were relying on this test to validate your computations, then the test was useless because the code returned invalid_size before executing any operation. 

In other words, what I was asking is to change the test by increasing the value of ldb, not to label it as invalid... you already had a bunch of those.

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
jzuniga-amd: No, for rocblas_operation_none and m >= n, the size of C is m-by-nrhs!
Its first n rows are computed as the solution of Rx=C (with TRSM), the last m-n rows are a constant residual computed as Q'*C (with ORMQR_UNMQR). As they are constant, they do not intervene in the optimization, but they do exist. 

So, if n=0 (quick return), but m*nrhs != 0, C must be set to zero.
In general, if m*nrhs>0 || n*nrhs >0, C exists and must be set to zero before quick return.

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
jzuniga-amd: Ok, that is fine too..

The idea was to have everything ready so that when you complete the functionality, you only substitute _rocblas_not_implemented by your code. As it is right now, when the time come, you will have to change arg check, and add the case-selection logic anyways. Up to you.

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
jzuniga-amd: Unfortunately, this will take more thinking than what I was expected... we cannot simply re-use the mechanism we implemented in GETRI. Let's leave it for now and fix it as a defect in a future PR.

jzuniga-amd (approved):
Looks good!

rocsolver/library/src/lapack/roclapack_gels.hpp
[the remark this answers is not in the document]
cgmb: Oh. It does not contain solution vectors, but it does need to contain residual sums. Got it.

PR #166, opened 2020-10-09
Don't build msgpack examples

tfalders:
Ported install fixes from https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/365 into rocSOLVER.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -167,7 +167,7 @@ install_msgpack_from_source( )
       cd "${build_dir}/deps"
       git clone -b cpp-3.0.1 https://github.com/msgpack/msgpack-c.git
       cd msgpack-c
-      CXX=${cxx} CC=${cc} ${cmake_executable} -DMSGPACK_BUILD_TESTS=OFF .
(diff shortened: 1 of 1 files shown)


cgmb (approved):
LGTM

TorreZuk (approved):
LGTM.  Not sure how you can also get into 3.10 branch.

cgmb:
Just to copy the reasoning over here, the msgpack examples use the option `-std=c++98`, which won't work with the bleeding-edge ROCm compilers. We use C++14 for everything. I'm not sure what other versions are supported.

There's no reason to build the examples anyways, so we might as well save time and effort by skipping them.

cgmb:
We should probably let Juan handle getting it into staging when he gets back. If it absolutely can't wait, then either Troy or I can do it with Colin's approval.

tfalders:
Porting changes from https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/367 as well.

jzuniga-amd (approved):
LGTM

PR #169, opened 2020-10-22
fix SLES builds for amd-targetid

jzuniga-amd:


TorreZuk (approved):
You removed -lgcc_s -rtlib=compiler-rt but these are needed, we shifted them to the library flags in rocblas from the clients, that is why they are gone.  See:
target_link_libraries( rocblas PRIVATE hip::device --rtlib=compiler-rt --unwindlib=libgcc )
The unwindlib is gcc_s.
Otherwise okay, but was this reported in amd-targetid branch?   I just merged to master and will come back to develop branch, can't say what branches are req. for you.

TorreZuk (commented):
LGTM, I am leaving the if (hipcc) for now, just deleting the else, but guess nvcc may never be used at this level.   Just as well to delete the if as can add it back if ever required.

tfalders (approved):
LGTM!

cgmb (approved):
:+1:

cgmb:
@TorreZuk. Do we actually ever test building rocBLAS or rocSOLVER with nvcc? If not, then I'd definitely agree with you. I don't have a lot of faith in untested code paths. Might as well write our code as simply as possible for the platforms we support.

PR #172, opened 2020-10-27
Tridiagonalization of symmetric/hermitian matrices - 1/3

jzuniga-amd:
Dividing the Tridiagonalization functionality in 3 parts to facilitate reviewing process:
1/3: Unblocked algorithm (SYTD2/HETD2),
2/3: Block auxiliary (LATRD), and
3/3: Blocked algorithm (SYTRD/HETRD)

### DIFF
--- a/rocsolver/clients/include/testing_sytxx_hetxx.hpp
+++ b/rocsolver/clients/include/testing_sytxx_hetxx.hpp
@@ -238,7 +238,7 @@ void sytxx_hetxx_getError(const rocblas_handle handle,
         {
             for(rocblas_int i = 0; i < n - 2; ++i)
                 a[i + (n - 1) * lda] = 0;
-            a[n - 2 + (n - 1) * lda] = a[n - 1 + (n - 2) * lda];
+            a[(n - 2) + (n - 1) * lda] = a[(n - 1) + (n - 2) * lda];
 
             // for each column
             for(rocblas_int j = n - 2; j >= 0; --j)
@@ -247,7 +247,7 @@ void sytxx_hetxx_getError(const rocblas_handle handle,
                 for(rocblas_int i = 0; i < j - 1; ++i)
                     a[i + j * lda] = 0;
                 if(j > 0)
-                    a[j - 1 + j * lda] = a[j + (j - 1) * lda];
+                    a[(j - 1) + j * lda] = a[j + (j - 1) * lda];
                 for(rocblas_int i = j + 2; i < n; ++i)
                 {
                     v[i - j - 1] = a[i + j * lda];
@@ -257,7 +257,7 @@ void sytxx_hetxx_getError(const rocblas_handle handle,
 
                 // apply householder reflector
                 cblas_larf(rocblas_side_left, n - 1 - j, n - j, v.data(), 1, t + j,
-                           a + j + 1 + j * lda, lda, hW.data());
+                           a + (j + 1) + j * lda, lda, hW.data());
                 if(COMPLEX)
                     cblas_lacgv(1, t + j, 1);
                 cblas_larf(rocblas_side_right, n - j, n - 1 - j, v.data(), 1, t + j,
@@ -282,7 +282,7 @@ void sytxx_hetxx_getError(const rocblas_handle handle,
                 }
                 v[j - 1] = 1;
                 if(j < n - 1)
-                    a[j + 1 + j * lda] = a[j + (j + 1) * lda];
+                    a[(j + 1) + j * lda] = a[j + (j + 1) * lda];
                 for(rocblas_int i = j + 2; i < n; ++i)
                     a[i + j * lda] = 0;
 
--- a/rocsolver/clients/include/testing_sytxx_hetxx.hpp
+++ b/rocsolver/clients/include/testing_sytxx_hetxx.hpp
@@ -30,39 +30,40 @@ void sytxx_hetxx_checkBadArgs(const rocblas_handle handle,
                           rocblas_status_invalid_handle);
 
     // values
-    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, rocblas_fill_full, n, dA, lda, stA, dD,
-                                                stD, dE, stE, dTau, stP, bc),
+    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, rocblas_fill_full, n, dA,
+                                                lda, stA, dD, stD, dE, stE, dTau, stP, bc),
                           rocblas_status_invalid_value);
 
     // sizes (only check batch_count if applicable)
     if(STRIDED)
-        EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA, dD,
-                                                    stD, dE, stE, dTau, stP, -1),
+        EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA,
+                                                    dD, stD, dE, stE, dTau, stP, -1),
                               rocblas_status_invalid_size);
 
     // pointers
-    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, (T) nullptr, lda, stA,
-                                                dD, stD, dE, stE, dTau, stP, bc),
+    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, (T) nullptr, lda,
+                                                stA, dD, stD, dE, stE, dTau, stP, bc),
                           rocblas_status_invalid_pointer);
     EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA,
                                                 (S) nullptr, stD, dE, stE, dTau, stP, bc),
                           rocblas_status_invalid_pointer);
-    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA, dD, stD,
-                                                (S) nullptr, stE, dTau, stP, bc),
+    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA, dD,
+                                                stD, (S) nullptr, stE, dTau, stP, bc),
                           rocblas_status_invalid_pointer);
-    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA, dD, stD,
-                                                dE, stE, (U) nullptr, stP, bc),
+    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, n, dA, lda, stA, dD,
+                                                stD, dE, stE, (U) nullptr, stP, bc),
                           rocblas_status_invalid_pointer);
 
     // quick return with invalid pointers
-    EXPECT_ROCBLAS_STATUS(rocsolver_sytxx_hetxx(STRIDED, SYTRD, handle, uplo, 0, (T) nullptr, lda, stA,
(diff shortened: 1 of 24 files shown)


rocsolver/library/src/lapack/roclapack_sytd2_hetd2.cpp
```diff
@@ -0,0 +1,132 @@
+/* ************************************************************************
+ * Copyright (c) 2019-2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "roclapack_sytd2_hetd2.hpp"
+
+template <typename S, typename T, typename U>
```
cgmb, line 7: Is `S` *always* just `decltype(std::real(T{}))`, or can it sometimes be different?

rocsolver/library/src/lapack/roclapack_sytd2_hetd2.hpp
```diff
@@ -0,0 +1,380 @@
...
+        S* e = E + b * strideE;
+
+        // diagonal
+        d[i] = a[i + i * lda];
+
+        // off-diagonal
+        if(i < n - 1)
+        {
+            if(lower)
+                a[i + 1 + i * lda] = T(e[i]);
```
tfalders, line 133: See comment for the testing hpp file about putting the i coordinate in brackets. Also applies to line 169.

tfalders (approved):
Looks good to me! I have some comments regarding typographical errors, but nothing major.

rocsolver/clients/include/testing_sytxx_hetxx.hpp
```diff
@@ -0,0 +1,585 @@
...
+    for(rocblas_int b = 0; b < bc; ++b)
+    {
+        T* a = hARes[b];
+        T* t = hTau[b];
+
+        if(uplo == rocblas_fill_lower)
+        {
+            for(rocblas_int i = 0; i < n - 2; ++i)
+                a[i + (n - 1) * lda] = 0;
+            a[n - 2 + (n - 1) * lda] = a[n - 1 + (n - 2) * lda];
```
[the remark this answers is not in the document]
jzuniga-amd: done

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+    \brief SYTD2 computes the tridiagonal form of a real symmetric matrix A.
+
+    \details
+    (This is the unblocked version of the algorithm).
+
+    The tridiagonal form is given by:
+
+        T = Q' * A * Q
+
+    where T is symmetric tribidiagonal and Q is an orthogonal matrix represented as the product
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+    \brief HETD2 computes the tridiagonal form of a complex hermitian matrix A.
+
+    \details
+    (This is the unblocked version of the algorithm).
+
+    The tridiagonal form is given by:
+
+        T = Q' * A * Q
+
+    where T is symmetric tribidiagonal and Q is an unitary matrix represented as the product
```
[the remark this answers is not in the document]
jzuniga-amd: changed

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+    n         rocblas_int. n >= 0.\n
+              The number of rows and columns of the matrix A.
+    @param[inout]
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the matrix to be factored.
+              On exit, if upper, then the elements on the diagonal and superdiagonal
+              contain the tridiagonal form T; the elements above the superdiagonal contain
+              the i-1 non-zero elements of vectors v(i) stored as columns.
+              If lower, then the elements on the diagonal and subdiagonal
+              contain the tridiagonal form T; the elements below the sudiagonal contain
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+              the n-i-1 non-zero elements of vectors v(i) stored as columns.
+    @param[in]
+    lda       rocblas_int. lda >= n.\n
+              specifies the leading dimension of A.
+    @param[out]
+    D         pointer to real type. Array on the GPU of dimension n.\n
+              The diagonal elements of T.
+    @param[out]
+    E         pointer to realtype. Array on the GPU of dimension n-1.\n
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+
+    where tau[i] is the corresponding Householder scalar. When uplo indicates lower, the first i
+    elements of the Householder vector v(i) are zero, and v(i)[i+1] = 1. If uplo indicates upper,
+    the last n-i elements of the Householder vector v(i) are zero, and v(i)[i] = 1.
+
+    @param[in]
+    handle    rocblas_handle.
+    @param[in]
+    uplo      rocblas_fill.\n
+              Specifies whether the upper or lower part of the symmetric matrix A is stored.
```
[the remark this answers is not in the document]
jzuniga-amd: changed

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+              If lower, then the elements on the diagonal and subdiagonal
+              contain the tridiagonal form T_j; the elements below the sudiagonal contain
+              the n-i-1 non-zero elements of vectors v_j(i) stored as columns.
+    @param[in]
+    lda       rocblas_int. lda >= n.\n
+              specifies the leading dimension of A_j.
+    @param[in]
+    strideA   rocblas_stride.\n
+              Stride from the start of one matrix A_j and the next one A_(j+1).
+              There is no restriction for the value of strideA. Normal use case is strideD >= lda*n.
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+              contain the tridiagonal form T; the elements below the sudiagonal contain
+              the n-i-1 non-zero elements of vectors v(i) stored as columns.
+    @param[in]
+    lda       rocblas_int. lda >= n.\n
+              specifies the leading dimension of A.
+    @param[out]
+    D         pointer to real type. Array on the GPU of dimension n.\n
+              The diagonal elements of T.
+    @param[out]
+    E         pointer to realtype. Array on the GPU of dimension n-1.\n
```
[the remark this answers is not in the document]
jzuniga-amd: fixed

rocsolver/library/src/include/rocblas.hpp
```diff
@@ -60,13 +60,42 @@ rocblas_status rocblasCall_dot(rocblas_handle handle,
...
+                               rocblas_int incx,
+                               rocblas_stride stridex,
+                               T* const y[],
+                               rocblas_int offsety,
+                               rocblas_int incy,
+                               rocblas_stride stridey,
+                               rocblas_int batch_count,
+                               T* results,
+                               T* workspace,
+                               T** work = nullptr)
```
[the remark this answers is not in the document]
jzuniga-amd: Yeah, I guess no. Removed.

rocsolver/library/src/lapack/roclapack_sytd2_hetd2.hpp
jzuniga-amd: done

rocsolver/library/src/lapack/roclapack_sytd2_hetd2.cpp
jzuniga-amd: Yes, normally. That's the convention we have been trying to follow.

cgmb: Since it doesn't change, could we replace `S` with the templated typedef `real_t<T>` [defined by utility.h in rocblascommon](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/rocm-3.9.0/rocblascommon/library/src/include/utility.h#L369)? It seems short enough to use directly in place of `S`, and would avoid having to define / pass around `S`.

I just applied that suggestion to gebrd to show what it would look like: https://github.com/cgmb/rocSOLVER/commit/9c5f7adfa1a37097df2c3e68c59566c34a493c2f

Making that change in this PR would be inconsistent with the overall library, but if it's a good idea, perhaps we could leave this as-is and instead make a task to apply it globally?

jzuniga-amd: Yeah, it could be done this way...
I certainly would not apply real_t<T> to only this PR. It should be a different PR that apply the change globally, if we want to do it.

IMO this would be too much work for very little gain. I don't really see a clear advantage of using real_t<T> vs S. If it is just to avoid passing S when calling the template, this could be achieved with the current code: if I call `return rocsolver_sytd2_hetd2_impl(..)` the compiler should be able to deduce S (and T) for us. However, I prefer `return rocsolver_sytd2_hetd2_impl<float, rocblas_float_complex>(..)` (and S and T) as this indicates me that the function at hand, in the complex case, mixes complex and real types; not all complex functions do this.

cgmb: > If it is just to avoid passing S when calling the template, this could be achieved with the current code: if I call return rocsolver_sytd2_hetd2_impl(..) the compiler should be able to deduce S (and T) for us.

That deduces S and T independently, but defining the relationship between S and T explicitly in the code does two things:
1. It makes it clear to the compiler what the relation between S and T is, and thus helps it (sometimes) generate a better error message when you pass the wrong parameters.
2. It makes it clear to the programmer what the relationship between S and T is.

We could instead just document our conventions for these sorts of things, but my first instinct is usually to embed them within the code. Eliminating S would also end the questions I wonder when looking at function calls: "does S come first, or does T?" and, "do I need to provide S for this function?"

There's an argument to be made for keeping T, even if it can be deduced from the arguments (as that also disambiguates T even when the parameter list contains errors), but the S appears to be entirely superfluous in terms of functionality.

> However, I prefer return rocsolver_sytd2_hetd2_impl<float, rocblas_float_complex>(..) (and S and T) as this indicates me that the function at hand, in the complex case, mixes complex and real types; not all complex functions do this.

Doesn't the function signature do that? I kinda thought that this bit was boring boilerplate marshalling between C and C++, with the actually interesting stuff being either the C API on the one side, or the actual kernels / implementation on the other. I viewed the process of making this code similar as being an improvement, but it might just be that I don't understand the algorithm and thus don't appreciate the hint provided by specifying S at the call site.

cgmb: (We could also eliminate U in much the same way. Defining the expected type of U based on the other template parameters would affect the appearance of the code less, but improve the compiler error messages more, as we always use type deduction for that one.)

jzuniga-amd: Sounds good Cory, if you think is useful, we can create a ticket to request these modifications in a future PR for all the files that need them.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -8211,6 +8211,1156 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
                                                                  rocblas_int* info,
                                                                  const rocblas_int batch_count);
```
cgmb, line 8213: I think you may have reused a `//! @}` from the functions above. The one that should be here is missing.

jzuniga-amd: Good catch!
Fixed

cgmb (approved):

PR #173, opened 2020-10-28
Add -a|--architecture flag to install.sh

cgmb:
`./install.sh -a "$(mygpu)"` is handy for development, as it reduces the
number of GPU kernels that must be compiled, and thus the overall
build times. This matches the rocBLAS flag.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,6 +59,10 @@ Options:
 
   -n | --no-optimizations     Pass this flag to disable optimizations for small sizes.
 
+  -a | --architecture         Set GPU architecture target, e.g. "gfx803;gfx900;gfx906;gfx908".
+                              If you don't know the architecture of the GPU in your local machine, it can be
+                              queried by running "mygpu".
+
   --docs                      (experimental) Pass this flag to build the documentation from source.
                               Official documentation is available online at https://rocsolver.readthedocs.io/
                               Building locally with this flag will require docker on your machine. If you are
@@ -316,6 +320,7 @@ build_type=Release
 build_relocatable=false
 build_docs=false
 optimal=true
+architecture=all
 
 rocm_path=/opt/rocm
 if ! [ -z ${ROCM_PATH+x} ]; then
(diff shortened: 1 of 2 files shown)


cgmb:
Moving the setup for `AMDGPU_TARGETS` to before `find_package(hip REQUIRED)` allows to be better `hip::device` to include some better flags. Related: https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/0ded5cfb3113178ad9239dabe596b08292f2b062. Perhaps I should add a comment.

This patch worked great when I wrote it, but it probably will need some rework due to the xnack changes. I opened this PR partly just to run it on the CI with the new containers.

cgmb:
Hmm... The CI appears to somewhat misconfigured at the moment, so that wasn't very useful.

cgmb:
The CI seems to be working again, and the build is succeeding, so this is ready for review. See also [Paul's comments](https://github.com/ROCmSoftwarePlatform/sample-app/pull/2#discussion_r494425189).

TorreZuk (approved):
This looks fine to me, of note as we don't define  -DAMDGPU_TARGETS in rocBLAS install I found I had to use FORCE on the cache variable otherwise if you are rebuilding the cache var doesn't always get updated if it is already exists.  One feature you can also consider in a separate PR is for install.sh to consider taking the environment variable if set to be passed in as if it were the -a flag.  This style seems to be in use elsewhere.  If we implement it can avoid special flags for limited or new architecture rebuilds across multiple libs.  I may get to this in rocBLAS soon to see if there are any issues.  Tensile is why we are a bit of an outlier in rocBLAS.

jzuniga-amd (approved):
LGTM

install.sh
```diff
@@ -316,6 +320,7 @@ build_type=Release
 build_relocatable=false
 build_docs=false
 optimal=true
+architecture=
```
tfalders, line 323: I'm not familiar with what happens when the equals sign doesn't have a right-hand side. Is there a reason it's written this way?

tfalders (approved):
LGTM, but I'm confused about line 323 in install.sh.

install.sh
cgmb: This sets the variable to be an empty string. Odd as it looks, there's a certain kind of consistency to it. Even in the line above `optimal=true` sets the value of optimal to be `"true"` (a string, not a boolean) while `architecture=` sets the value of architecture to be `""`

You could get away without this line, as the access of an undefined variable would return an empty string anyway, but it's easier to check for bugs in the script when all uses of undefined variables are errors.

PR #175, opened 2020-10-31
Enable docs build on CI and minor fixes

cgmb:
This is a rollup of the various docs changes I've made over the past month or so.

Summary:
- Minor fixups of typos and broken links
- Moved deprecated items to their own section at the bottom of the page
- Enabled the docs build on the CI
- `./install.sh --docs` now prints clickable links to the built docs

PR #176, opened 2020-10-31
Enable checkin tests in debug mode on CI

cgmb:
This change adds a new CI job that runs the checkin tests in Debug
mode.

On my local machine, running all checkin tests in debug mode takes
37 minutes, but 35 of those minutes are spent on the SVD tests. I've
excluded them from this new CI job (for now) to keep the job fast.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/CHANGELOG.md merge=union
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -39,14 +39,15 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false)
 def runTestCommand (platform, project, gfilter)
 {
     String sudo = auxiliary.sudo(platform.jenkinsLabel)
+    String buildType = project.buildName.contains('Debug') ? 'debug' : 'release'
     def command = """#!/usr/bin/env bash
                 set -x
-                cd ${project.paths.project_build_prefix}/build/release/clients/staging
+                cd ${project.paths.project_build_prefix}/build/${buildType}/clients/staging
                 ${sudo} GTEST_LISTENER=NO_PASS_LINE_IN_LOG ./rocsolver-test --gtest_output=xml --gtest_color=yes --gtest_filter=${gfilter}
                 """
 
     platform.runCommand(this, command)
-    junit "${project.paths.project_build_prefix}/build/release/clients/staging/*.xml"
+    junit "${project.paths.project_build_prefix}/build/${buildType}/clients/staging/*.xml"
 }
 
 def runPackageCommand(platform, project)
--- a/.jenkins/debug.groovy
+++ b/.jenkins/debug.groovy
@@ -32,6 +32,15 @@ def runCI =
         commonGroovy.runCompileCommand(platform, project, jobName)
     }
 
+    def testCommand =
+    {
+        platform, project->
+
+        // Skip the SVD tests as they're too slow in debug mode
+        def gfilter = '*checkin_lapack*-*SVD*'
+        commonGroovy.runTestCommand(platform, project, gfilter)
+    }
+
     def packageCommand =
     {
         platform, project->
@@ -39,7 +48,7 @@ def runCI =
         commonGroovy.runPackageCommand(platform, project)
     }
 
-    buildProject(prj, formatCheck, nodes.dockerArray, compileCommand, null, packageCommand)
+    buildProject(prj, formatCheck, nodes.dockerArray, compileCommand, testCommand, packageCommand)
 
 }
 
--- a/.jenkins/debug.groovy
+++ b/.jenkins/debug.groovy
@@ -36,8 +36,7 @@ def runCI =
     {
         platform, project->
 
-        // Skip the SVD tests as they're too slow in debug mode
-        def gfilter = '*checkin_lapack*-*SVD*'
+        def gfilter = '*checkin_lapack*'
         commonGroovy.runTestCommand(platform, project, gfilter)
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Removed
 
 ### Fixed
-- Fixed default launch bounds causing run-time failures in Debug mode
+- Fixed runtime errors in debug mode caused by incorrect kernel launch bounds
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](https://rocsolver.readthedocs.io/en/latest/).
 
+## [(Unreleased) rocSOLVER for ROCm 4.1.0]
+### Added
+
+### Optimizations
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+
+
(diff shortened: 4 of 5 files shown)


cgmb:
BTW, the failure in the static analysis is due to a trivial formatting error that stuck onto develop. It's fixed in #175.

jzuniga-amd (commented):
Who is running unit tests in debug mode? CI, QA, only developers?
What is purpose of running the tests if not all functionality is tested?
If someone is relying on these tests, I think that we need to generate a subset of them that take less time, rather than remove them completely.

cgmb:
Just to be clear, this change doesn't remove any tests. This changes the CI from running no tests at all on the Debug build, to running all the checkin tests except for SVD. All the CI jobs that existed prior to this change should remain exactly the same.

All the issues you've raised are valid concerns. It would be nice to include SVD in the Debug test runs. I suspect that it uses matrices that are larger than the other tests, and this just gets unreasonably slow to execute when optimizations are disabled. We could perhaps try moving some of the larger SVD matrix sizes from checkin to daily?

jzuniga-amd:
That is my point, if nobody is currently testing in debug mode, what is the problem with our tests taking 40 minutes? Is somebody complaining about it?, is there a restriction in CI?
If is just for the developers, any developer can disable/enable any test locally as needed, I see no point in officially introducing a test suite that is incomplete.

Now, the problem with SVD tests is that all the calculations needed for the implicit testing are beefy and happen on the host; without compiler optimization (-O0) I bet all these loops are taking the hit. There are no large matrices (n is always less than 40). The other problem is that the SVD algorithm can follow many paths and, in consequence, there are naturally many tests to execute to cover all the paths. Generating a useful subset of the tests to use on debug mode is not as simple as moving some of them to daily_lapack. Besides, if we are creating a new test category (debug_lapack, or something), it should be done for all the functions, not only SVD...

So, I would suggest to leave the entire checkin_lapack tests for now, if @saadrahim and the math-CI team agree. In the future we can create the debug_lapack category if required. BTW, why do we need to execute tests in debug mode? is this something math-CI will start doing for all libraries?

cgmb:
To expand on the motivation and reasoning behind this PR, I think this change is good because:

- It required almost no development effort.
- It requires almost no CI server resources.
- I expect it to require little ongoing maintenance.
- It would have caught bugs before they were merged (e.g. the bugs fixed in #163).

> BTW, why do we need to execute tests in debug mode?

We do not need to, but I suggested this because I perceived it as a change that had moderate benefits and minimal costs.

> is this something math-CI will start doing for all libraries?

I don't think there are any plans to do so, though I think we would if we had unlimited hardware to run them.

> What is purpose of running the tests if not all functionality is tested?

My philosophy is that all test suites are incomplete. Adding tests is thus a matter of cost vs. benefit.

If these added tests help catch bugs in GETRF and other functions, that's useful even if they do nothing to help with catching bugs in GESVD. The tests would certainly be more useful if they did both, but to cover SVD, we would either need to expend either additional development effort or additional CI time.

Either of those things would be reasonable, but I think expanding the debug mode tests to include SVD is potentially an item we could put on the backlog and prioritize separately. Though, that's obviously not necessary if Saad thinks we have the spare resources on the CI to just expand it now.

saadrahim:
@cgmb says that additional testing in debug mode could have caught #163. That's good enough for me to say we should add tests to cover it. 

From the CI perspective, we have been building in debug mode only simply because I was uncertain of the value of running tests. I will leave it to the individual libraries to decide on what you need. I support any increased test activity that will reduce developer work load.

jzuniga-amd:
@saadrahim the question is whether we can add the entire checkin_lapack test suite to the debug jobs, or if we should generate a subset that runs faster. Currently checkin_lapack tests would take ~40min in debug mode. Is there a restriction or maximum acceptable running time for these extra checks?

saadrahim:
Please go ahead and add 40 minute tests.

cgmb:
Sounds good to me! As Saad feels the CI can handle the additional load, I have have no objections to enabling them all.

TorreZuk:
Just like the rocblas tensile tip PR I'll ask how often is this going to run?    Will we add the same to rocBLAS and any others who request it?    I would think for rocBLAS a rare run would be good (weekly or on demand).   One could argue the developer should run debug builds anyway when coding a new feature or run with valgrind as it points out errors that may not even crash in debug.   So will this build turn on the AddressSanitizer flags in lieu of valgrind for the tests or that is too slow?

cgmb:
> I would think for rocBLAS a rare run would be good (weekly or on demand).

I'd actually originally proposed to run this more infrequently in rocSOLVER (like the extended tests), but I'm not sure how to do that. Weekly seems a bit infrequent, though.

> So will this build turn on the AddressSanitizer flags in lieu of valgrind for the tests or that is too slow?

Not sure what the plan is, but that was what I had originally been trying to do when I added extra debug mode checks in #133. I had to cut the AddressSanitizer flags, though, since they were broken at the time.

jzuniga-amd (approved):
All checkin_lapack tests in looks good to me.

cgmb:
This failure looks like that crash you mentioned. Never seen it on my hardware, so it's nice to have an example on CI.
```
:0:rocdevice.cpp            :2316: 2253119848406 us: Device::callbackQueue aborting with status: 0x100f

/var/jenkins_home/workspace/_hipclang_debug_rocSOLVER_PR-176@tmp/durable-347c158f/script.sh: line 4: 19093 Aborted                 (core dumped) GTEST_LISTENER=NO_PASS_LINE_IN_LOG ./rocsolver-test --gtest_output=xml --gtest_color=yes --gtest_filter=*checkin_lapack*

script returned exit code 134 
```

Looks like `HSA_STATUS_ERROR_INVALID_ISA = 0x100F`.

https://github.com/ROCm-Developer-Tools/ROCclr/blob/rocm-3.9.0/device/rocm/rocdevice.cpp#L2179
https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/rocm-3.9.0/src/inc/hsa.h#L195

cgmb:
It's just a guess, but I'm thinking that without optimizations, GESVD is overflowing the vector register file (VGPR), as that seems like [it matches the error code](https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/rocm-3.9.0/src/core/runtime/amd_aql_queue.cpp#L847) and it would make sense that problem would only show up with certain hardware. I'll open a ticket to investigate the issue further.

I'm not sure if this is a rocsolver bug or an llvm-amdgpu bug, but IMO, this debug test run is already paying dividends. (It will be mildly embarrassing if this crash ultimately turns out to have been my fault for disabling optimizations and enabling assertions, causing us to exceed platform limits, but sometimes learning requires that you take chances, make mistakes and get messy.)

cgmb:
It seems that we're launching kernels compiled for a maximum of 256 thread per block with 1024 threads (32x32). I've fixed this by specifying `--gpu-max-threads-per-block=1024` to increase the compiler default from 256. From my understanding, that will be the default soon anyway. My guess is that this doesn't affect release mode, because when optimizations are enabled, the kernels are simply not exceeding the memory limits for 1024 thread per block even when compiled for 256 threads per block.

From a performance perspective, we should probably be specifying this parameter on a per-kernel basis using `__launch_bounds__(MAX_THREADS_PER_BLOCK)`. I'm not sure if we want to tackle that now or later. I need to add a CHANGELOG entry for this, but aside from that, I think it's ready to merge.

jzuniga-amd (approved):
LGTM!
We can add tightest launch bounds to all kernels that don't have them later.

tfalders (approved):
LGTM! Thanks for catching that launch bounds bug, btw.

PR #177, opened 2020-11-03
Tridiagonalization of symmetric/hermitian matrices - 2/3

jzuniga-amd:
Dividing the Tridiagonalization functionality in 3 parts to facilitate reviewing process:
1/3: Unblocked algorithm (SYTD2/HETD2),
2/3: Block auxiliary (LATRD), and
3/3: Blocked algorithm (SYTRD/HETRD)


tfalders (approved):
LGTM! I skimmed most of the template function body since the tests should reveal any errors there.

rocsolver/clients/common/cblas_interface.cpp
```diff
@@ -275,6 +275,30 @@ void zlarfb_(char* side,
              rocblas_double_complex* W,
              int* ldw);
 
+void slatrd_(char* uplo, int* n, int* k, float* A, int* lda, float* E, float* tau, float* W, int* ldw);
+
```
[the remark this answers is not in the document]
jzuniga-amd: You are right. I removed the spaces and reorganized a bit this file.

rocsolver/clients/include/testing_latrd.hpp
```diff
@@ -0,0 +1,372 @@
...
+                       const rocblas_int hot_calls,
+                       const bool perf)
+{
+    if(!perf)
+    {
+        latrd_initData<true, false, T>(handle, n, dA, lda, hA);
+
+        // cpu-lapack performance
+        *cpu_time_used = get_time_us();
+        memset(hW[0], 0, ldw * k * sizeof(T));
```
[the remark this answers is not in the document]
jzuniga-amd: Neither legacy LAPACK labrd nor latrd formally initialize W, X or Y.  I guess in old FORTRAN systems, arrays were initialized to zero; this is not the case now. If you look at the outputs from LAPACK labrd or latrd, there is garbage in those entries that are supposed to be zero. 

In rocSOLVER we have two options: 
a. do not initialize the arrays, as in the original rocsolver_labrd (in this case we should clearly mention in the documentation that passing a pointer for the arrays W, X or Y is not enough, and that it is the user responsibility to initialize the memory), or 
b. initialize the arrays, as I do in rocsolver_latrd and the modified rocsolver_labrd (in this case the user should only pass a valid pointer, and the functions will return a clean array with the results).

With option b (the one I am proposing here), I would like to add to the cpu_time, the time that the user would require to initialize the arrays W, X, Y. This will make the comparisons more fair. 

What do you think?

rocsolver/clients/include/testing_latrd.hpp
```diff
@@ -0,0 +1,372 @@
...
+    CHECK_ROCBLAS_ERROR(rocsolver_latrd(handle, uplo, n, k, dA.data(), lda, dE.data(), dTau.data(),
+                                        dW.data(), ldw));
+    CHECK_HIP_ERROR(hARes.transfer_from(dA));
+    CHECK_HIP_ERROR(hWRes.transfer_from(dW));
+
+    // CPU lapack
+    memset(hW[0], 0, ldw * k * sizeof(T));
+    cblas_latrd(uplo, n, k, hA[0], lda, hE[0], hTau[0], hW[0], ldw);
+
+    // error is max(||hA - hARes|| / ||hA||, ||hW - hWRes|| / ||hW||)
```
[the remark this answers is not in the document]
jzuniga-amd: Not necessarily. The values of E are the results of the application of the householder reflectors (the beta values in larfg). So if A (which contains the Householder vectors) and W are correct, E should be correct. If we try to compare the computed E with the resulted E from LAPACK we may have different signatures... so we will require an implicit testing which is more complicated.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -682,6 +682,124 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlabrd(rocblas_handle handle,
...
+    @param[in]
+    uplo      rocblas_fill.\n
+              Specifies whether the upper or lower part of the matrix A is stored.
+              If uplo indicates lower (or upper), then the upper (or lower)
+              part of A is not used.
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The number of rows and columns of the matrix A.
+    @param[in]
+    k         rocblas_int. min(m,n) >= k >= 0.\n
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -682,6 +682,124 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlabrd(rocblas_handle handle,
...
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the n-by-n matrix to be reduced.
+              On exit, if uplo is lower, the first k columns have been reduced to tridiagonal form
+              (given in the diagonal elements of A and the array E), the elements below the diagonal
+              contain the vectors v(i) stored as columns.
+              If uplo is upper, the last k columns have been reduced to tridiagonal form
+              (given in the diagonal elements of A and the array E), the elements above the diagonal
+              contain the vectors v(i) stored as columns.
+    @param[in]
+    lda       rocblas_int. lda >= m.\n
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

rocsolver/clients/include/testing_latrd.hpp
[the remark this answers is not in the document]
tfalders: That makes sense. I wrote this comment before I realized that labrd and latrd initialize the arrays and are included in the GPU timing.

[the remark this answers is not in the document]
cgmb: There are sometimes optimizations on new allocations of zeroed pages in CPU land. e.g. https://lwn.net/Articles/117881/
I don't know if that sort of optimization will ever be relevant to a GPU API, but it would not be unreasonable to ask the user to zero it themselves, especially if that that is what is required by the LAPACK interface.

In the presence of doubt over the best way to design the API, I probably would mildly favour leaving the zeroing up to the user, as that is a decision which gives them the most flexibility, and allows us to reverse our decision without breaking backwards compatibility. But, if we already have similar API calls that are doing the initialization, it would definitely make more sense to stay consistent with them.

I suppose even if we do choose Option B and later discover that there are benefits to Option A, we can always provide `rocsolver_labrd_noinit` (or whatever) to provide such an alternative interface in a backwards-compatible manner. (Or we could break compatibility and bump the major version, but I'm loathe to do that unless it's absolutely necessary.)

In short, I have no strong opinions one way or another. I would have defaulted to Option A, but if you like Option B better, that's fine too. In either case, there is a reasonable way for us to correct this decision in the future, which IMO is the most important thing.

rocsolver/clients/gtest/latrd_gtest.cpp
```diff
@@ -0,0 +1,133 @@
...
+Arguments latrd_setup_arguments(latrd_tuple tup)
+{
+    vector<int> matrix_size = std::get<0>(tup);
+    vector<int> op_size = std::get<1>(tup);
+
+    Arguments arg;
+
+    arg.N = matrix_size[0];
+    arg.lda = matrix_size[1];
+    arg.ldb = matrix_size[2]; //ldw
```
cgmb, line 66: I don't know how to improve this right now, so it's fine as-is, but at some point it might be worth looking at how we're structuring our tests. I find the parameter mapping to be a particularly confusing aspect.

rocsolver/library/src/auxiliary/rocauxiliary_latrd.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+    if(!mem)
+        return rocblas_status_memory_error;
+
+    scalars = mem[0];
+    work = mem[1];
+    norms = mem[2];
+    workArr = mem[3];
+    T sca[] = {-1, 0, 1};
+    RETURN_IF_HIP_ERROR(hipMemcpy((T*)scalars, sca, size_scalars, hipMemcpyHostToDevice));
```
cgmb, line 66: I know we do this everywhere, but it seems a bit weird to me that it's done on the default stream. Is this something that's going to be replaced with a different strategy in the future?

cgmb (approved):

rocsolver/clients/gtest/latrd_gtest.cpp
cgmb: Why do we use a single `Arguments` class for all tests? This mapping between `ldb` and `ldw` seems like it would be unecessary if we defined an `LATRD_Arguments` class.

rocsolver/clients/include/testing_latrd.hpp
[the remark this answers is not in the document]
jzuniga-amd: Yeah, I am not sure either which option would be better. I kind of like option B because, by implementing the initialization in a kernel, we could eventually do a better (faster) job than what most users can do by using hipmemset or hipmemcpy (not all users would be able to write their own kernels to do the initializations). Also, I feel awkward asking the users to initialize output arrays. 

Anyways, I will leave it as it for now (option B)... is not much of a problem to review it any time if needed. I also contacted some people working closer to the applications and final users and LAPACK experts. If I hear from them right away, I could revert any decision in the upcoming Tridiagonalization - 3/3 PR, if needed.

rocsolver/clients/gtest/latrd_gtest.cpp
jzuniga-amd: Definitely, we should at some point review the argument class as this also affects severely the benchmark client. There is already a ticket opened while ago (SWDEV-247028), we just need to find some time to work on it

rocsolver/library/src/auxiliary/rocauxiliary_latrd.cpp
jzuniga-amd: why is this weird? what would be advantage/difference of using a different stream?

cgmb: If I understand this correctly, doing the memcpy on the default stream introduces a global synchronization point. The memcpy won't start until all jobs on all other streams have finished, and it will thus also block the rest of the algorithm from running. In workloads that are heavily parallelized via streams, I could imagine that might have a significant performance impact.

dcyca: All GPU operations should be executed in a user-specified stream.

rocBLAS provides an API function to associate a stream with a handle, and all subsequent operations using that handle are executed in the stream.  This approach is common in most GPU libraries, although a few new GPUs have adopted a stateless approach where the stream (and any required workspace memory) must be passed into each function.

TorreZuk: Dan they use the same handle as rocBLAS I think.  Should be simple changes to just switch to hipMemcpyAsync and get the stream from the handle same as rocBLAS.  Uploads nothing to worry about, downloads you might have to add a synchronization somewhere if you use the async form and need the results, otherwise just leave it to the user to synchronize if they are waiting on host results.   You probably wouldn't see much difference until you start multi-stream/thread/gpu benching.

jzuniga-amd: Yes, I think that is the main point here. rocSOLVER is not multi-stream application, but the rocblas_handle has now a working API for the user to chose the single stream to work with (which defaults to stream 0). 

So, although we do need the synchronization on the host (there is no point to continue with the computations until the transfer of the scalars  is done), we don't necessarily want to wait on other operations on the GPU (in case the user has other processes running in different streams, or even multiple rocSOLVER instances with different handles). 

So, if you agree I will open a ticket to make all the scalars transfers tied to the stream in the rocblas_handle (whatever that is) --followed by the required hipStreamsychronize. 

@TorreZuk, I think this should be reviewed in rocBLAS as well. There are functions which have internal transfers. I know that TRTRI and TRSM have internal mem copies (as the results of the inverse of the diagonal blocks have to be moved around), and these are all hipMemcpy. 
Why do you think the sychronization is not required on the uploads (host to device)? Is hipmemcpyasync actually synchronous if memory not pined?

TorreZuk: Sure we can review it, most of the internal memory copies in rocBLAS should now have been replaced with hipMemcpyAsync.  hipMemcpyWithStream I haven't had a chance to evaluate.   hipmemcpyasync should be queue processed in the stream to preserve.   hipMemcpyAsync with hipMemcpyDeviceToDevice should be equivalent to a kernel that moves memory in terms of queue order ( I hope ).   Relating to a different PR but hipMemsetAsync I expect should be equivalent to a kernel that zeros device memory, so must be done before the next kernel on that stream.

jzuniga-amd: Nevermind! 
Yes, in this case we only need to change hipMemcpy by hipMemcpyAsync with the handle stream as it all happens in the same stream...

jzuniga-amd: Will open ticket to make change globally. Leaving this PR as is, for now.

rocsolver/clients/include/testing_latrd.hpp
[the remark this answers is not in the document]
jzuniga-amd: Ok, I heard back from Jakub, and the mystery is solved. 
While it is true that the resulted W, or X and Y in LABRD, could have some uninitialized elements (with garbage), those elements are the entries that the algorithm never touches, and this is because they are not needed...

I am not sure why lapack LATRD returns an n-by-k array W, but the only rows really needed to update the unreduced part of A, are the first k rows if uplo = upper, or the last k rows if uplo = lower, and these always have computed values. 

So, we don't really need to internally initialize W, and the users don't have to do it either if they don't want. And we don't need to test the entire array W (or X and Y in LABRD) as those uninitialized elements could be very different in the CPU and GPU cases. 

So, I am fixing the tests and removing the initializations.

cgmb:
Sounds like we have a great plan. :+1:

PR #178, opened 2020-11-05
Added sterf and steqr

tfalders:
Addresses SWDEV-247078.

Note that both sterf and steqr have been implemented without any parallelism (similar to bdsqr). steqr is quite slow at the moment. This can probably be addressed by adding some parallelism to lasr in the near future.

### DIFF
--- a/rocblascommon/clients/include/utility.hpp
+++ b/rocblascommon/clients/include/utility.hpp
@@ -148,7 +148,7 @@ inline void rocblas_print_matrix(T* CPU_result, T* GPU_result, size_t m, size_t 
         {
             rocblas_cout << "matrix  col " << i << ", row " << j
                          << ", CPU result=" << CPU_result[j + i * lda]
-                         << ", GPU result=" << GPU_result[j + i * lda] << "\n";
+                         << ", GPU result=" << GPU_result[j + i * lda] << std::endl;
         }
 }
 
--- a/rocsolver/clients/gtest/steqr_gtest.cpp
+++ b/rocsolver/clients/gtest/steqr_gtest.cpp
@@ -39,7 +39,7 @@ const vector<vector<int>> matrix_size_range = {
     {35, 40}};
 
 // for daily_lapack tests
-const vector<vector<int>> large_matrix_size_range = {{192, 192}, {640, 640}, {1000, 1024}};
+const vector<vector<int>> large_matrix_size_range = {{192, 192}};
 
 Arguments steqr_setup_arguments(steqr_tuple tup)
 {
--- a/rocsolver/clients/gtest/steqr_gtest.cpp
+++ b/rocsolver/clients/gtest/steqr_gtest.cpp
@@ -1,0 +1,115 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "testing_steqr.hpp"
+
+using ::testing::Combine;
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<vector<int>, vector<int>> steqr_tuple;
+
+// each size_range vector is a {N, ldc}
+
+// each op_range vector is a {e}
+// if e = 0, then evect = 'N'
+// if e = 1, then evect = 'I'
+// if e = 2, then evect = 'V'
+
+// case when N == 0 and evect == N will also execute the bad arguments test
+// (null handle, null pointers and invalid values)
+
+const vector<vector<int>> op_range = {{0}, {1}, {2}};
+
+// for checkin_lapack tests
+const vector<vector<int>> matrix_size_range = {
+    // quick return
+    {0, 1},
+    // invalid
+    {-1, 1},
+    // invalid for case evect != N
+    {2, 1},
+    // normal (valid) samples
+    {12, 12},
+    {20, 30},
+    {35, 40}};
+
+// for daily_lapack tests
+const vector<vector<int>> large_matrix_size_range = {{192, 192}, {640, 640}, {1000, 1024}};
+
+Arguments steqr_setup_arguments(steqr_tuple tup)
+{
+    vector<int> size = std::get<0>(tup);
+    vector<int> op = std::get<1>(tup);
+
+    Arguments arg;
+
+    arg.N = size[0];
+    arg.ldc = size[1];
+
+    arg.evect = (op[0] == 0 ? 'N' : (op[0] == 1 ? 'I' : 'V'));
+
+    arg.timing = 0;
+
+    return arg;
+}
+
+class STEQR : public ::TestWithParam<steqr_tuple>
+{
+protected:
+    STEQR() {}
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+};
+
+TEST_P(STEQR, __float)
+{
+    Arguments arg = steqr_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N')
+        testing_steqr_bad_arg<float, float>();
+
+    testing_steqr<float, float>(arg);
+}
+
+TEST_P(STEQR, __double)
+{
+    Arguments arg = steqr_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N')
+        testing_steqr_bad_arg<double, double>();
+
+    testing_steqr<double, double>(arg);
+}
+
+TEST_P(STEQR, __float_complex)
+{
+    Arguments arg = steqr_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N')
+        testing_steqr_bad_arg<float, rocblas_float_complex>();
+
+    testing_steqr<float, rocblas_float_complex>(arg);
+}
+
+TEST_P(STEQR, __double_complex)
+{
+    Arguments arg = steqr_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N')
+        testing_steqr_bad_arg<double, rocblas_double_complex>();
+
+    testing_steqr<double, rocblas_double_complex>(arg);
+}
+
+INSTANTIATE_TEST_SUITE_P(daily_lapack,
+                         STEQR,
+                         Combine(ValuesIn(large_matrix_size_range), ValuesIn(op_range)));
+
+INSTANTIATE_TEST_SUITE_P(checkin_lapack,
+                         STEQR,
+                         Combine(ValuesIn(matrix_size_range), ValuesIn(op_range)));
--- a/rocsolver/clients/gtest/steqr_gtest.cpp
+++ b/rocsolver/clients/gtest/steqr_gtest.cpp
@@ -39,7 +39,7 @@ const vector<vector<int>> matrix_size_range = {
     {35, 40}};
 
 // for daily_lapack tests
-const vector<vector<int>> large_matrix_size_range = {{192, 192}};
+const vector<vector<int>> large_matrix_size_range = {{192, 192}, {256, 270}, {300, 300}};
 
 Arguments steqr_setup_arguments(steqr_tuple tup)
 {
--- a/rocsolver/clients/include/testing_steqr.hpp
+++ b/rocsolver/clients/include/testing_steqr.hpp
@@ -1,0 +1,356 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "cblas_interface.h"
+#include "clientcommon.hpp"
+#include "norm.hpp"
+#include "rocsolver.hpp"
+#include "rocsolver_arguments.hpp"
+#include "rocsolver_test.hpp"
+
+template <typename S, typename T, typename U>
+void steqr_checkBadArgs(const rocblas_handle handle,
+                        const rocblas_evect compc,
+                        const rocblas_int n,
+                        S dD,
+                        S dE,
+                        T dC,
+                        const rocblas_int ldc,
+                        U dInfo)
+{
+    // handle
+    EXPECT_ROCBLAS_STATUS(rocsolver_steqr(nullptr, compc, n, dD, dE, dC, ldc, dInfo),
+                          rocblas_status_invalid_handle);
+
+    // values
+    EXPECT_ROCBLAS_STATUS(rocsolver_steqr(handle, rocblas_evect(-1), n, dD, dE, dC, ldc, dInfo),
+                          rocblas_status_invalid_value);
+
+    // pointers
+    EXPECT_ROCBLAS_STATUS(rocsolver_steqr(handle, compc, n, (S) nullptr, dE, dC, ldc, dInfo),
+                          rocblas_status_invalid_pointer);
+    EXPECT_ROCBLAS_STATUS(rocsolver_steqr(handle, compc, n, dD, (S) nullptr, dC, ldc, dInfo),
+                          rocblas_status_invalid_pointer);
+    EXPECT_ROCBLAS_STATUS(rocsolver_steqr(handle, compc, n, dD, dE, (T) nullptr, ldc, dInfo),
+                          rocblas_status_invalid_pointer);
+    EXPECT_ROCBLAS_STATUS(rocsolver_steqr(handle, compc, n, dD, dE, dC, ldc, (U) nullptr),
+                          rocblas_status_invalid_pointer);
+
+    // quick return with invalid pointers
+    EXPECT_ROCBLAS_STATUS(
+        rocsolver_steqr(handle, compc, 0, (S) nullptr, (S) nullptr, (T) nullptr, ldc, dInfo),
+        rocblas_status_success);
+}
+
+template <typename S, typename T>
+void testing_steqr_bad_arg()
+{
+    // safe arguments
+    rocblas_local_handle handle;
+    rocblas_evect compc = rocblas_evect_original;
+    rocblas_int n = 1;
+    rocblas_int ldc = 1;
+
+    // memory allocations
+    device_strided_batch_vector<S> dD(1, 1, 1, 1);
+    device_strided_batch_vector<S> dE(1, 1, 1, 1);
+    device_strided_batch_vector<T> dC(1, 1, 1, 1);
+    device_strided_batch_vector<rocblas_int> dInfo(1, 1, 1, 1);
+    CHECK_HIP_ERROR(dD.memcheck());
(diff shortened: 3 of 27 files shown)


rocblascommon/clients/include/utility.hpp
```diff
@@ -148,7 +148,7 @@ inline void rocblas_print_matrix(T* CPU_result, T* GPU_result, size_t m, size_t
         {
             rocblas_cout << "matrix  col " << i << ", row " << j
                          << ", CPU result=" << CPU_result[j + i * lda]
-                         << ", GPU result=" << GPU_result[j + i * lda] << "\n";
+                         << ", GPU result=" << GPU_result[j + i * lda] << std::endl;
```
cgmb, line 151: This adds a buffer flush to the printing of each entry. Is that the intent? It seems like a reasonable change. Just wondering if there's a particular problem you encountered.

rocsolver/clients/include/testing_steqr.hpp
```diff
@@ -0,0 +1,407 @@
...
+    {
+        rocblas_int lda = n;
+        T alpha;
+        T beta = 0;
+        for(int j = 0; j < n; j++)
+        {
+            alpha = T(1) / hD[0][j];
+            cblas_symv_hemv(rocblas_fill_upper, n, alpha, hA[0], lda, hCRes[0] + j * ldc, 1, beta,
+                            hC[0] + j * ldc, 1);
+        }
```
cgmb, line 217: Interesting. So this is an iterative solver? Does n iterations guarantee a solution? Does less than n iterations 
 have a meaningful result? Just personal curiosity.

rocsolver/library/include/rocsolver-extra-types.h
```diff
@@ -36,14 +36,25 @@ typedef enum rocblas_svect_
...
 /*! \brief Used to enable the use of fast algorithms (with out-of-place
  *computations) in some of the routines
  ********************************************************************************/
 typedef enum rocblas_workmode_
 {
-    rocblas_outofplace = 201, /**< Out-of-place computations are allowed; this
+    rocblas_outofplace = 211, /**< Out-of-place computations are allowed; this
                                requires enough free memory. */
-    rocblas_inplace = 202, /**< When not enough memory, this forces in-place computations  */
+    rocblas_inplace = 212, /**< When not enough memory, this forces in-place computations  */
```
cgmb, line 57: This breaks binary compatibility, and is going to require bumping the rocSOLVER version to 4.0. Given that it's the reason we're bumping the major version, it will probably need a mention in the changelog.

That's actually rather convenient this time, since it means we'll get to match the ROCm version again, but we should make sure that we've picked values that are not going to require changing again.

rocsolver/library/include/rocsolver-functions.h
```diff
@@ -3178,13 +3178,148 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zbdsqr(rocblas_handle handle,
...
+    diagonal elements D and the array of symmetric off-diagonal elements E
+    as returned by, e.g., SYTRD or HETRD. If the tridiagonal matrix is the
+    reduced form of a full symmetric/Hermitian matrix as returned by, e.g.,
+    SYTRD or HETRD, then the eigenvectors of the original matrix can also
+    be found.
+
+    @param[in]
+    handle    rocblas_handle.
+    @param[in]
+    compC     #rocblas_evect.\n
```
cgmb, line 3247: I like these links. I wish it worked for the original rocblas types, too. I'd thought that was going to be a straight-forward extension, as cross-linking the docs is supported in doxygen, but as far as I can tell, there's no existing way to bring that information into sphinx. :(

Though, I hadn't realized until recently that the rocsolver-defined types in the function signature were clickable!

rocsolver/library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -0,0 +1,420 @@
...
+
+#include "rocauxiliary_sterf.hpp"
+#include "rocblas.hpp"
+#include "rocblas_device_functions.hpp"
+#include "rocsolver.h"
+
+/****************************************************************************
+(TODO:THIS IS BASIC IMPLEMENTATION. THE ONLY PARALLELISM INTRODUCED HERE IS
+  FOR THE BATCHED VERSIONS (A DIFFERENT THREAD WORKS ON EACH INSTANCE OF THE
+  BATCH).
```
cgmb, line 21: Error: Mismatched parentheses!

A trivial typo. I just dressed up as a LISP compiler for Halloween and I still haven't gotten back to normal yet.

rocsolver/library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -0,0 +1,420 @@
...
+
+    S* D = DD + (bid * strideD);
+    S* E = EE + (bid * strideE);
+    T* C = load_ptr_batch<T>(CC, bid, shiftC, strideC);
+    S* work = WW + (bid * strideW);
+
+    rocblas_int m, l, lsv, lend, lendsv;
+    rocblas_int l1 = 0;
+    rocblas_int iters = 0;
+    S b, f, g, anorm, p, r, c, s;
```
cgmb, line 76: Are all of these used for communication across iterations of the loop, or could some of them be declared within it?

rocsolver/library/src/auxiliary/rocauxiliary_sterf.hpp
```diff
@@ -0,0 +1,362 @@
...
+
+    // quick return
+    if(n <= 1)
+        return rocblas_status_success;
+
+    T eps = get_epsilon<T>();
+    T ssfmin = get_safemin<T>();
+    T ssfmax = T(1.0) / ssfmin;
+    ssfmin = sqrt(ssfmin) / (eps * eps);
+    ssfmax = sqrt(ssfmax) / T(3.0);
```
cgmb, line 354: Great job matching the types. You've done it consistently throughout. 👍

cgmb:
`rocblas_evect` needs to be added to the types in `userguide_api.rst`

rocblascommon/clients/include/utility.hpp
tfalders: That was the intent, yes. I often print matrices to help me debug problems, and before this change they would be printed after all the test cases printed. It was not only hard to see where one pair of matrices ends and the other begins, but it was also difficult to associate it with the test case that was printing it.

rocsolver/clients/include/testing_steqr.hpp
tfalders: No, this is multiplying A with each of its n eigenvectors and dividing by the associated eigenvalue. I had to make it an implicit test because eigenvectors are not unique under scaling.

rocsolver/library/include/rocsolver-extra-types.h
tfalders: Yeah, this was an iffy change for me. I can change the values back and give the evect enum new values. I wanted to have a nice flow from svect to evect, but if it's not worth breaking compatibility over.

rocsolver/library/src/auxiliary/rocauxiliary_steqr.hpp
tfalders: The ints are used to communicate across all loops. The floats are used across a couple inner loops.

I might do as you suggest for sterf, since I have several floats pulling double duty there and it's a bit confusing.

rocsolver/library/include/rocsolver-extra-types.h
cgmb: I hadn't realized this until now, but apparently [the rocblas enum values are supposed to be consistent with other implementations](https://rocblas.readthedocs.io/en/master/functions.html#enums). Is the same true of the rocsolver extra types?

rocsolver/clients/include/testing_steqr.hpp
cgmb: That's such a good description of what's being done and why. IMO, that would make a great comment to put directly in the code.

rocsolver/library/src/include/rocblas_device_functions.hpp
```diff
@@ -267,4 +268,375 @@ __device__ void gemm_kernel(const rocblas_int m,
...
+    {
+        c = 0;
+        s = 1;
+        r = g;
+    }
+    else
+    {
+        r = sqrt(f * f + g * g);
+        c = f / r;
+        s = g / r;
```
jzuniga-amd, line 294: These computations could, in theory, overflow. Not super important for now, but for consistency, as you are scaling the main algorithm, you should be applying the overflow/underflow scale guards here as well. 
Alternatively, I would suggest that we use the lartg implemented in BDSQR. It is the same thing except that, as mentioned above, it gives you the rotation, not the transpose. That implementation guards against overflow without scaling. 

You should be able to use bdsqr_lartg in STEQR as long as you take the transpose before applying the rotation; I think is easier to modify there, because of the symmetry, rather than in BDSQR.  

(If you agree, I can try and make the modifications along with the correction of the offset in bdsqr_lasr. Ideally we should be using the same Givens-rotations auxiliaries everywhere.)

rocsolver/library/src/include/rocblas_device_functions.hpp
```diff
@@ -267,4 +268,375 @@ __device__ void gemm_kernel(const rocblas_int m,
...
+                    A[i + j * lda] = cs * temp - sn * A[i + (j - 1) * lda];
+                    A[i + (j - 1) * lda] = cs * A[i + (j - 1) * lda] + sn * temp;
+                }
+            }
+        }
+    }
+}
+
+/** LAE2 computes the eigenvalues of a 2x2 symmetric matrix
+    [ a b ]
```
jzuniga-amd, line 388: This is not BLAS/rocBLAS functionality! 
We should move it to common_device.hpp. 

(Maybe later we can think of better names for these two files: common_device.hpp and rocblas_device_functions.hpp)

rocsolver/library/src/include/rocblas_device_functions.hpp
```diff
@@ -267,4 +268,375 @@ __device__ void gemm_kernel(const rocblas_int m,
...
+    }
+    else
+    {
+        rt1 = T(0.5) * rt;
+        rt2 = T(-0.5) * rt;
+    }
+}
+
+/** LAEV2 computes the eigenvalues and eigenvectors of a 2x2 symmetric matrix
+    [ a b ]
```
jzuniga-amd, line 440: This is not BLAS/rocBLAS functionality!
We should move it to common_device.hpp.

rocsolver/library/src/include/rocblas_device_functions.hpp
```diff
@@ -267,4 +268,375 @@ __device__ void gemm_kernel(const rocblas_int m,
...
+
+    if(sgn1 == sgn2)
+    {
+        temp1 = cs1;
+        cs1 = -sn1;
+        sn1 = temp1;
+    }
+}
+
+/** LASRT_INCREASING sorts an array D in increasing order.
```
jzuniga-amd, line 540: This is not BLAS/rocBLAS functionality!
We should move it to common_device.hpp.

jzuniga-amd (changes requested):

rocsolver/library/src/include/rocblas_device_functions.hpp
cgmb: Perhaps we could add a brief comment to the top `rocblas_device_functions.hpp` that describes what sort of functions that should go within it, and maybe also why we have rocblas functionality implemented in a rocsolver header. Even just a sentence or two would help a lot, I think.

tfalders: I figured it made sense to keep all device functions that implement a full LAPACK or BLAS function in one place, and to have common_device be responsible for other non-LAPACK helper functions. Perhaps we can change the name of rocblas_device_functions to lapack_device_functions? It's up to you - I can certainly move things into common_device.

jzuniga-amd:
@tfalders I submitted changes to your branch as per comments above. Please merge (if you agree with them) so that they can be reflected here. I will continue my review afterwards.

rocsolver/library/src/include/rocblas_device_functions.hpp
```diff
@@ -267,4 +268,379 @@ __device__ void gemm_kernel(const rocblas_int m,
...
+        s = 1;
+        r = -g;
+    }
+    else
+    {
+        T t;
+        if(std::abs(g) > std::abs(f))
+        {
+            t = -f / g;
+            s = 1 / T(std::sqrt(1 + t * t));
```
cgmb, line 296: I don't think this cast should be necessary. `std::sqrt` takes `T` and returns `T`. If it can be removed, you're better off without it, as having an explicit cast there could potentially hide a mismatched type.

cgmb (commented):
I left one more minor comment, but it's not terribly important. The only thing reason I haven't marked this approved is because that question about our enum values troubles me. It's definitely not something specific to your PR, but since you're adding a new enum, we should figure it out ASAP.

rocsolver/library/include/rocsolver-extra-types.h
jzuniga-amd: there are few C/C++ Lapack implementations, and most use chars. If we want, we can follow this [reference](https://bitbucket.org/icl/magma/src/hipMAGMA/include/magma_types.h). Not a bad idea (except for the major version bumping?) 
We can leave it for now and discuss it later; Alternatively we can also add functions that translate otherlibrary_enum -> rocsolver_enum and vice-versa...

cgmb: This is something we should bring up with @doctorcolinsmith, as breaking binary compatibility would be felt most strongly by the FORTRAN bindings. I think we can accept this PR with `rocblas_evect` as-is, and update everything at once if we choose to do so.

cgmb (approved):
The `std::sqrt` change is a mere nice-to-have. LGTM.

rocsolver/library/src/include/rocblas_device_functions.hpp
jzuniga-amd: Right, I think the cast could be removed...

rocsolver/library/src/include/rocblas_device_functions.hpp
tfalders: I'll rename the files in a separate PR. Should be possible to automate it, but it will clutter up this PR with extra stuff.

rocsolver/clients/gtest/steqr_gtest.cpp
```diff
@@ -0,0 +1,115 @@
...
+    {-1, 1},
+    // invalid for case evect != N
+    {2, 1},
+    // normal (valid) samples
+    {12, 12},
+    {20, 30},
+    {35, 40}};
+
+// for daily_lapack tests
+const vector<vector<int>> large_matrix_size_range = {{192, 192}};
```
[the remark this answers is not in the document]
tfalders: Sure. They'll still be rather small matrices though, since steqr is quite slow at the moment.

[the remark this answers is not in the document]
tfalders: Although the slowness could also be due to not having splits in the matrix when compC != N. I'll try out your suggestions for initialization and see how that impacts the runtime.

rocsolver/library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -0,0 +1,428 @@
...
+        }
+
+        lsv = l = l1;
+        lendsv = lend = m;
+        l1 = m + 1;
+        if(lend == l)
+            continue;
+
+        // Scale submatrix
+        steqr_inf_norm(l, lend, D, E, anorm);
```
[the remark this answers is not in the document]
tfalders: The inf norm is used in vanilla LAPACK for steqr. I wouldn't mind changing it to the max norm if you don't think it will affect the algorithm.

rocsolver/clients/gtest/steqr_gtest.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+// each op_range vector is a {e}
+// if e = 0, then evect = 'N'
+// if e = 1, then evect = 'I'
+// if e = 2, then evect = 'V'
+
+// case when N == 0 and evect == N will also execute the bad arguments test
+// (null handle, null pointers and invalid values)
+
+const vector<vector<int>> op_range = {{0}, {1}, {2}};
```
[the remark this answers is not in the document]
tfalders: I might keep it the way it is just for consistency with the test cases for other functions.

rocsolver/library/src/auxiliary/rocauxiliary_steqr.hpp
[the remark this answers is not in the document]
jzuniga-amd: I have not checked the LAPACK code. Let me take a look... I don't see why the inf norm will be needed (any option we take should not really affect the algorithm, I think)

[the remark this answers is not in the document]
jzuniga-amd: The only norm that I see in LAPACK steqr is line 269:
anorm = dlanst( 'M', lend-l+1, d( l ), e( l ) )
it is a 'max' norm!

As I said, you can leave the inf norm, but there are some extra FLOPS there that are not really needed. Also, there is the advantage of re-using the code in sterf as mentioned in the other comments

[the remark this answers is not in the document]
tfalders: Looks like the real versions of steqr use the max norm, but the complex versions use the inf norm. I don't know if there's a reason for that.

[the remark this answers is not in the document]
jzuniga-amd: I don't see why!
D and E are always real, even in the complex case. 
I would prefer to use find_max everywhere. But can always review later. 
(If you can test find_max that would be good)

rocsolver/library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -0,0 +1,407 @@
...
+#include "rocsolver.h"
+
+/****************************************************************************
+(TODO:THIS IS BASIC IMPLEMENTATION. THE ONLY PARALLELISM INTRODUCED HERE IS
+  FOR THE BATCHED VERSIONS (A DIFFERENT THREAD WORKS ON EACH INSTANCE OF THE
+  BATCH).)
+***************************************************************************/
+
+/** STERF_KERNEL implements the main loop of the sterf algorithm
+    to compute the eigenvalues of a symmetric tridiagonal matrix given by D
```
jzuniga-amd, line 25: STEQR

jzuniga-amd (approved):
LGTM

PR #186, opened 2020-11-17
Add debug asserts to client containers

cgmb:
I've added some debug assertions to the client containers. The focus here was on changing as little as possible about how these functions work, so the checks are rather limited. (e.g. In the batched cases, they ensure that you're not exceeding the batch count, but they can't tell if you access one of the matrices in the batch with an out-of-bounds index.)

I had entirely reworked the client containers to be able to check more, but let's just start with this. It's a much smaller change than that was going to be, and I think it will be easier to upstream into rocBLAS.

### DIFF
--- a/rocblascommon/clients/include/device_batch_vector.hpp
+++ b/rocblascommon/clients/include/device_batch_vector.hpp
@@ -207,9 +207,14 @@ public:
         //
         for(rocblas_int batch_index = 0; batch_index < this->m_batch_count; ++batch_index)
         {
-            if(hipSuccess
-               != (hip_err = hipMemcpy((*this)[batch_index], that[batch_index],
-                                       sizeof(T) * this->nmemb(), hipMemcpyHostToDevice)))
+            T* dst = (*this)[batch_index];
+            const T* src = that[batch_index];
+            const size_t bytes = sizeof(T) * this->nmemb();
+            assert(dst || bytes == 0);
+            assert(src || bytes == 0);
+            assert(this->nmemb() == that.nmemb());
+            hip_err = hipMemcpy(dst, src, bytes, hipMemcpyHostToDevice);
+            if(hipSuccess != hip_err)
             {
                 return hip_err;
             }
--- a/rocblascommon/clients/include/device_batch_vector.hpp
+++ b/rocblascommon/clients/include/device_batch_vector.hpp
@@ -154,7 +154,7 @@ public:
     //!
     T* operator[](rocblas_int batch_index)
     {
-        assert(this->m_data || batch_index == 0);
+        assert(this->m_data);
         assert(batch_index < this->m_batch_count);
         return this->m_data[batch_index];
     }
@@ -166,7 +166,7 @@ public:
     //!
     const T* operator[](rocblas_int batch_index) const
     {
-        assert(this->m_data || batch_index == 0);
+        assert(this->m_data);
         assert(batch_index < this->m_batch_count);
         return this->m_data[batch_index];
     }
--- a/rocblascommon/clients/include/device_batch_vector.hpp
+++ b/rocblascommon/clients/include/device_batch_vector.hpp
@@ -3,6 +3,8 @@
 //
 #pragma once
 
+#include <assert.h>
+
 #include "d_vector.hpp"
 
 //
@@ -152,6 +154,7 @@ public:
     //!
     T* operator[](rocblas_int batch_index)
     {
+        assert(this->m_data || batch_index == 0);
         return this->m_data[batch_index];
     }
 
@@ -162,6 +165,7 @@ public:
     //!
     const T* operator[](rocblas_int batch_index) const
     {
+        assert(this->m_data || batch_index == 0);
         return this->m_data[batch_index];
     }
 
--- a/rocblascommon/clients/include/device_batch_vector.hpp
+++ b/rocblascommon/clients/include/device_batch_vector.hpp
@@ -155,6 +155,7 @@ public:
     T* operator[](rocblas_int batch_index)
     {
         assert(this->m_data || batch_index == 0);
+        assert(batch_index < this->m_batch_count);
         return this->m_data[batch_index];
     }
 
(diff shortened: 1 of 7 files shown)


rocblascommon/clients/include/device_batch_vector.hpp
```diff
@@ -152,6 +154,8 @@ class device_batch_vector : private d_vector<T, PAD, U>
     //!
     T* operator[](rocblas_int batch_index)
     {
+        assert(this->m_data || batch_index == 0);
```
tfalders, line 157: Should that be batch_count instead of batch_index? I don't see how m_data being null is solved by looking at the first batch entry.

rocblascommon/clients/include/device_strided_batch_vector.hpp
```diff
@@ -153,9 +155,12 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
     //!
     T* operator[](rocblas_int batch_index)
     {
-        return (this->m_stride >= 0)
-            ? this->m_data + batch_index * this->m_stride
-            : this->m_data + (batch_index + 1 - this->m_batch_count) * this->m_stride;
+        auto offset = (this->m_stride >= 0)
+            ? batch_index * this->m_stride
+            : (batch_index + 1 - this->m_batch_count) * this->m_stride;
+        assert(this->m_data || offset == 0);
```
tfalders, line 161: Similar comment as for device_batch_vector

rocblascommon/clients/include/host_batch_vector.hpp
```diff
@@ -104,13 +105,23 @@ class host_batch_vector
...
+    }
+
     //!
     //! @brief Random access to the vectors.
     //! @param batch_index the batch index.
     //! @return The mutable pointer.
     //!
     T* operator[](rocblas_int batch_index)
     {
+        assert(this->m_data || batch_index == 0);
```
tfalders, line 123: Similar comment as for device_batch_vector

rocblascommon/clients/include/host_strided_batch_vector.hpp
```diff
@@ -148,16 +150,27 @@ class host_strided_batch_vector
...
     //!
     T* operator[](rocblas_int batch_index)
     {
-        return (this->m_stride >= 0)
-            ? this->m_data + this->m_stride * batch_index
-            : this->m_data + (batch_index + 1 - this->m_batch_count) * this->m_stride;
+        auto offset = (this->m_stride >= 0)
+            ? this->m_stride * batch_index
+            : (batch_index + 1 - this->m_batch_count) * this->m_stride;
+        assert(this->m_data || offset == 0);
```
tfalders, line 171: Similar comment as for device_batch_vector

tfalders (approved):
LGTM, though I do have a question regarding one of the checks (repeated across several files).

rocblascommon/clients/include/device_strided_batch_vector.hpp
cgmb: I think it's really confusing, but the client containers sometimes depend on returning `((T*)nullptr) + 0`. I put on my language lawyer hat, and as far as I can tell, it is in fact valid to do pointer arithmetic on a null pointer, as long as you're adding or subtracting `0`. The result is simply a null pointer. I'm trying not to change how anything works this time around, so my assertions were written to accept that as being ok.

You're right about `device_batch_vector`, though. That one actually dereferences the null pointer. `((T*)nullptr)[0]` is not valid, so I should remove that `|| batched_index == 0`.

rocblascommon/clients/include/device_batch_vector.hpp
cgmb: `m_data[anything]` would be a problem if `m_data` is null (for any reason). I've removed `|| batch_index == 0` entirely from the batch vectors.

I also realized I wasn't checking that the index was non-negative. I added that to all the operator[] checks.

cgmb: My local tests continue to pass. Yay!

jzuniga-amd (approved):
Nice!! LGTM

rocblascommon/clients/include/host_batch_vector.hpp
```diff
@@ -104,13 +105,24 @@ class host_batch_vector
         return 0;
     }
 
+    //!
+    //! @brief Returns the number of members in each vector.
+    //!
```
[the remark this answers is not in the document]
cgmb: `nmemb` seems to always correspond to the number of elements in the array that is allocated (or in the batched case, in each array). It was previously mentioned down below in `try_initialize_memory`.

I had the same question when I first saw it, so I'm not a huge fan of the name. Given my limited understanding of what the name is supposed to imply, I've just tried to document the how the abbreviation expands.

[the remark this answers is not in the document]
jzuniga-amd: I don't think nmemb is an abbreviation of what it returns (although I guess this should be the best practice?)... 

Is not important but the description "members of a vector" sounds strange to me. 
`m_n` is the number of "elements" in the vector, and `m_n\*m_inc` is the size of the array that holds the vector (because its elements are separated by `inc`)

rocblascommon/clients/include/host_batch_vector.hpp
```diff
@@ -155,7 +170,12 @@ class host_batch_vector
             size_t num_bytes = this->n() * std::abs(this->inc()) * sizeof(T);
             for(rocblas_int batch_index = 0; batch_index < this->m_batch_count; ++batch_index)
             {
-                memcpy((*this)[batch_index], that[batch_index], num_bytes);
+                T* dst = (*this)[batch_index];
+                const T* src = that[batch_index];
+                assert(dst);
+                assert(src);
+                assert(num_bytes == that.n() * std::abs(that.inc()) * sizeof(T));
```
jzuniga-amd, line 177: I guess you could use the `nmemb()` getter here ?, otherwise why we bother to define it.

cgmb:
Just FYI, I left this open as I figured I would submit the matching changes into rocBLAS and incorporate their feedback.

PR #187, opened 2020-11-25
fix install script

jzuniga-amd:
The directory where rocSOLVER package installs should be, in this order,
1. the directory specified by the user with --install_dir, or
2. the directory specified in the system by the environment variable ROCM_PATH, or
3. the default /opt/rocm

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -311,6 +311,11 @@ supported_distro
 # #################################################
 # global variables
 # #################################################
+rocm_path=/opt/rocm
+if ! [ -z ${ROCM_PATH+x} ]; then
+  rocm_path=${ROCM_PATH}
+fi
+
 install_package=false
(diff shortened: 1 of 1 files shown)


tfalders (approved):
LGTM!

cgmb:
What's the expected behaviour for folks using CMake directly? Is it supposed to default to `ROCM_PATH` when `-DCPACK_PACKAGING_INSTALL_PREFIX` is not specified? If so, it might make more sense to avoid passing `-DCPACK_PACKAGING_INSTALL_PREFIX` at all that case, to avoid duplicating logic for defaults between bash and cmake.

If this fix is urgent, please feel free to go ahead and merge regardless.

cgmb (approved):

PR #188, opened 2020-11-27
Fixed erroneous zaxpy method signature

tfalders:
The method signature for zaxpy in norm.hpp had alpha as a double, whereas the LAPACK signature has alpha as a double complex. This is likely the cause of complex test failures when armor is enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,10 +163,10 @@ endif( )
 set( ARMOR_LEVEL "${DEFAULT_ARMOR_LEVEL}" CACHE STRING "Enables increasingly expensive runtime correctness checks" )
 include( armor-config )
 
-# set the optimization level for debug mode
-if( CMAKE_BUILD_TYPE STREQUAL "Debug" )
-  add_compile_options( -O0 )
-endif( )
+# # set the optimization level for debug mode
+# if( CMAKE_BUILD_TYPE STREQUAL "Debug" )
+#   add_compile_options( -O0 )
+# endif( )
 
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build rocSOLVER as a shared library" ON )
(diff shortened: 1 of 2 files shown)


cgmb (approved):
After the you pointed out that enabling standard library assertions broke the debug build, I spent a day trying to figure out what the problem was. I made two builds that differed only by the presence of assertions or not, and isolated a single function that would fail.

I logged the arrays at the entrance of every function, and I found that they were identical up until after xaxpy, but I didn't know how to debug the Fortran, and other work came in, so that's where I stopped.

This was definitely the cause of some particularly annoying problems. Glad to see it identified and resolved.

jzuniga-amd (commented):
Cool. Nice catch Troy!!
Does this mean that we can reenable -O3 in debug mode? This will speed up CI tests.

tfalders:
> Cool. Nice catch Troy!!
> Does this mean that we can reenable -O3 in debug mode? This will speed up CI tests.

I don't know what errors we were getting with -O3 in debug mode, but if they were complex test failures, then I imagine so.

jzuniga-amd:
Yes, all the complex tests were failing in debug mode with -O3 and the assertions enabled. -O0 was set for them to pass, but the tests take too long to complete. It would be nice to have them run faster as we have included them in CI. 
I removed the -O0 in a local build of your branch and all the tests pass now. 
So, I would leave the -O3, unless having compiler optimizations make useless the assertions and all the job that @cgmb has done for debug mode. ?

tfalders:
Assuming that @cgmb is okay with it, would you like me to remove the -O0 flag as part of this PR? Do I need to replace it with -O3, or will it build with -O3 automatically?

cgmb:
All the stuff I've done on debug checks should work fine with optimizations enabled.

Though, there's reasons beyond this bug that made `-O0` nice for debug mode. The faster compile times and the improved rocgdb experience make it my default for any work that's not performance-related.

I suppose I could just add the flag myself with an environment variable, so maybe it doesn't matter all that much. I'm fine with whatever you decide.

jzuniga-amd (approved):
LGTM!
Please remove the -O0 from CMake

PR #189, opened 2020-11-30
Updated changelog

tfalders:
Minor update to changelog as I forgot to do in most recent PR.


jzuniga-amd (commented):
"fixed unstable norm computation" are big words for numerical analysts, and this is not quite what the fix was about. I would rather say: 
"fixed bug caused by mixing precisions when computing the norm of errors in complex unit tests." or "fixed incorrect function signature in complex unit test: or something like this...

jzuniga-amd (approved):
Looks good, thanks.

cgmb (approved):

PR #190, opened 2020-12-01
File reorganization

tfalders:
Reorganizing files as per https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/178#discussion_r520259462

Summary of changes:
- Renamed common_device.hpp to common_device_helpers.hpp
- Renamed helpers.hpp to common_host_helpers.hpp
- Renamed rocblas_device_helpers.hpp to lapack_device_helpers.hpp
- Removed include for common_device_helpers.hpp from all files except rocblas.hpp in light of the fact that it is used everywhere but only included by some function headers, and to reflect the include pattern of common_host_helpers.hpp

### DIFF
--- a/rocsolver/library/src/include/common_host_helpers.hpp
+++ b/rocsolver/library/src/include/common_host_helpers.hpp
@@ -2,8 +2,8 @@
  * Copyright (c) 2019-2020 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
-#ifndef HELPERS_H
-#define HELPERS_H
+#ifndef COMMON_HOST_HELPERS_H
+#define COMMON_HOST_HELPERS_H
 
 #include <cassert>
 #include <cstdlib>
--- a/rocsolver/library/src/include/common_host_helpers.hpp
+++ b/rocsolver/library/src/include/common_host_helpers.hpp
@@ -12,6 +12,14 @@
 #include <iostream>
 #include <limits>
 
+/*
+ * ===========================================================================
+ *    common location for functions that are used across several rocSOLVER
+ *    routines, excepting device functions and kernels (see
+ *    common_device_helpers.hpp and lapack_device_functions.hpp).
+ * ===========================================================================
+ */
+
 template <typename T, std::enable_if_t<!is_complex<T>, int> = 0>
 constexpr double get_epsilon()
 {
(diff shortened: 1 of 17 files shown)


tfalders:
As part of this PR, we might want to look at reconsidering the purpose of libcommon.hpp. Right now it only has two includes and is included by only one file (common_device_helpers.hpp). I would propose either deleting libcommon.hpp or using it as a single include header file with all common includes (e.g. rocsolver.h, rocblas.hpp, common_device_helpers.hpp, etc...).

rocsolver/library/src/include/lapack_device_functions.hpp
```diff
@@ -1,13 +1,22 @@
 /* ************************************************************************
  * Copyright (c) 2019-2020 Advanced Micro Devices, Inc.
  * ************************************************************************ */
+
 #pragma once
-#ifndef _ROCBLAS_DEVICE_FUNCTIONS_HPP_
-#define _ROCBLAS_DEVICE_FUNCTIONS_HPP_
+#ifndef _LAPACK_DEVICE_FUNCTIONS_HPP_
+#define _LAPACK_DEVICE_FUNCTIONS_HPP_
```
cgmb, line 7: It's not really worth the effort of changing, but technically, identifiers that begin with an underscore followed by an uppercase letter are reserved for the standard library. In practice, it's unlikely to matter.

cgmb (approved):
This looks good to me. I have no real opinion on `libcommon.hpp` at the moment, so I'm happy with whatever you'd like to do.

rocsolver/library/src/include/common_host_helpers.hpp
```diff
@@ -2,8 +2,8 @@
  * Copyright (c) 2019-2020 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
-#ifndef HELPERS_H
-#define HELPERS_H
+#ifndef COMMON_HOST_HELPERS_H
+#define COMMON_HOST_HELPERS_H
```
jzuniga-amd, line 6: Please also add a brief description of what is in this file...
Same as you did with common_device_helpers and lapack_device_functions

jzuniga-amd (approved):
Looks good to me!!

The purpose of libcommon was to include all the headers comming from "rocblas common" in one place. At the end we only needed utility.h, but probably we will need more in the future? Maybe when we implement logging? 

I don't know, we can leave it for now or remove it and add it later if we want. Either way is fine by me.

cgmb:
:+1:

PR #191, opened 2020-12-01
Initialize scalars on user stream

cgmb:
This change moves the copying of scalars from the default stream to the user-specified stream from the `rocblas_handle`. I'm still a little unclear with the semantics, but I think it will do a synchronous memcpy into a pinned memory, then enqueue a job to copy from pinned memory into device memory.

I also consolidated some code in helpers.hpp, since I was there anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Fixed
 - Fixed runtime errors in debug mode caused by incorrect kernel launch bounds
 - Fixed complex unit test bug caused by incorrect zaxpy function signature
+- Moved a small memory transfer from the default stream to the stream set in `rocblas_handle`
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Fixed
 - Fixed runtime errors in debug mode caused by incorrect kernel launch bounds
 - Fixed complex unit test bug caused by incorrect zaxpy function signature
-- Moved a small memory transfer from the default stream to the stream set in `rocblas_handle`
+- Eliminated a small memory transfer that was being done on the default stream
 
 
 
--- a/rocsolver/library/src/auxiliary/rocauxiliary_labrd.cpp
+++ b/rocsolver/library/src/auxiliary/rocauxiliary_labrd.cpp
@@ -69,7 +69,8 @@ rocblas_status rocsolver_labrd_impl(rocblas_handle handle,
     scalars = mem[0];
     work_workArr = mem[1];
     norms = mem[2];
-    RETURN_IF_HIP_ERROR(init_scalars(handle, (T*)scalars, size_scalars));
+    if(size_scalars > 0)
+        init_scalars(handle, (T*)scalars);
 
     // execution
     return rocsolver_labrd_template<S, T>(handle, m, n, k, A, shiftA, lda, strideA, D, strideD, E,
--- a/rocsolver/library/src/auxiliary/rocauxiliary_labrd.cpp
+++ b/rocsolver/library/src/auxiliary/rocauxiliary_labrd.cpp
@@ -69,8 +69,7 @@ rocblas_status rocsolver_labrd_impl(rocblas_handle handle,
     scalars = mem[0];
     work_workArr = mem[1];
     norms = mem[2];
-    T sca[] = {-1, 0, 1};
-    RETURN_IF_HIP_ERROR(hipMemcpy((T*)scalars, sca, size_scalars, hipMemcpyHostToDevice));
+    RETURN_IF_HIP_ERROR(init_scalars(handle, (T*)scalars, size_scalars));
 
     // execution
     return rocsolver_labrd_template<S, T>(handle, m, n, k, A, shiftA, lda, strideA, D, strideD, E,
--- a/rocsolver/library/src/auxiliary/rocauxiliary_larf.cpp
+++ b/rocsolver/library/src/auxiliary/rocauxiliary_larf.cpp
@@ -57,8 +57,7 @@ rocblas_status rocsolver_larf_impl(rocblas_handle handle,
     scalars = mem[0];
     Abyx = mem[1];
     workArr = mem[2];
-    T sca[] = {-1, 0, 1};
-    RETURN_IF_HIP_ERROR(hipMemcpy((T*)scalars, sca, size_scalars, hipMemcpyHostToDevice));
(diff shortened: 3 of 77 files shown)


tfalders (approved):
LGTM, though I'm a little skeptical of the changes to the cast2const functions. If everything still works though, I suppose it's okay.

cgmb:
What makes you say that? I think the return type already ensures the result is correct, so the assignment is unnecessary. That said, perhaps it was added for a reason?

Perhaps more importantly, is there a good way to verify that this is actually being done on the stream specified? Ultimately, all versions of memcpy are implemented in [ihipMemcpy](https://github.com/ROCm-Developer-Tools/HIP/blob/rocm-3.10.0/rocclr/hip_memory.cpp#L135), and there appear to be fallbacks that drop back to the default stream.

cgmb:
After a little more investigation, I think the staging buffers are just one of multiple possible implementations. I'll move the source data into global constant memory rather than relying on the Async to actually be host-synchronous.

I actually liked [this other implementation](https://github.com/cgmb/rocSOLVER/commit/c7bb2c3d0cc8d48752026626ca77824ac1fdd8b5) better, as it seems simpler and more consistent to just initialize these in a kernel, but I was a little surprised that the performance was slightly worse in my tests.

jzuniga-amd (commented):
I don't mind for the host to wait for the transfer of scalars to complete. There is, in general, very little things that the host can do before calling the first kernel (which will explicitly synchronize with the transfer anyways). So, the real objective here is not necessarily to make the transfer asynchronous, but to wait, if necessary, on the stream specified by the user, which could be different from the default stream. 

Now, if hipMemcpyAsync without pinned memory will work with the default stream and not with the provided stream, then it is not really useful... I am not sure whether this is really the behavior, it would be nice if we can confirm it.   

That being said, I also like better your alternative solution, using a kernel to generate the scalars... good idea!. This way we are sure we are using the stream provided by the user, and in addition, we get the host asynchrony. 

The only thing, though, is that init_scalars() should be called from the implementations of the C APIs, not in the template functions! In other words, the scalars should be generated (or transferred) only by the main C API called by the user. If you do it in the template functions, you will be running the generation kernel many times (by the main function, and by all the functions that this one calls internally).

cgmb:
I built a debug copy of the HIP runtime and traced this through. We end up taking [this path](https://github.com/ROCm-Developer-Tools/HIP/blob/rocm-3.10.0/rocclr/hip_memory.cpp#L161-L172). I was a bit concerned about that null stream bit, but we don't end up going down that path. I still don't really understand what `queueDevice != dstMemory->getContext().devices()[0]` is checking, but it returns false on my machine.

cgmb:
I'd forgotten that the kernel would potentially get called multiple times when put in the template function. That might actually explain the small, persistent difference in performance between the two approaches. It's not a big deal, but I'll try the kernel again after fixing that. If that fixes the minor performance difference, I'll swap out the memcpy for the kernel, and we can get this merged.

Either solution works. I prefer the kernel for a few reasons. A big one is that I want to understand the rocSOLVER performance characteristics and error-handling really well, and there's less to study when we use fewer types of operations. If I can eliminate the memcpy with no measurable performance cost, that's the solution I'll go with (assuming Troy is ok with that too).

cgmb:
Moving the initialization back out of the template functions did indeed seem to fix the (small, but persistent) overhead I was seeing.

I think this is ready, if you folks are happy with it.

jzuniga-amd (approved):
Great! Looks good...

tfalders (commented):
Looks good, though I would personally prefer if the contents of the init_scalars.hpp file were divided amongst common_host_helpers.hpp and common_device_helpers.hpp.

jzuniga-amd:
Right, not a bad idea! 
Put init_scalars() in common_host_helpers and iota_n in common_device_helpers and we don't need init_scalars.hpp.

cgmb:
I didn't do that because it will require including the device header from the host header. If the separation is purely organizational, then there's no problem with that. I wasn't sure that was ok, though.

TorreZuk (commented):
Can't say I understand how the scalars will vary from the default and why you don't inline intialization in the template kernels.  But will just say you might want to consider the template init_scalars as a prototype only and define it in a .cpp and compile once with all explicit instantiations for all the required data types.  Otherwise it is compiled over and over.

cgmb:
@TorreZuk, I did that originally, but I wasn't sure how to organize it in our current setup.

My next PR was actually going to be marking all our common device functions as extern and explicitly instantiating them in a cpp file specifically for that reason.

cgmb:
I'm going to merge as-is to get the functionality in there. There's further reorganization on my todo list anyway, so there will be other convenient opportunities to move this to wherever you'd prefer.

cgmb:
@TorreZuk, I tested out a few different ways of trying to optimize the build time. I checked which templates existed in rocsolver.so by dumping the kernels with `/opt/rocm/bin/extractkernel -i build/debug/rocsolver-install/rocsolver/lib/librocsolver.so.0.1` and greped through the symbols to find which templates had been instantiated (and how many times) with `fgrep -rI init_scalars . | cut -d'<' -f 2 | cut -d'>' -f 1 | xargs c++filt | sort` and compared against `fgrep -rI init_scalars . | cut -d'<' -f 2 | cut -d'>' -f 1 | xargs c++filt | sort | uniq` to find how many instantiations I could save by doing things explicitly.

A plain build of the develop branch took me ~29m35s.

I tried using `extern template __global__ void iota_n<float>(float* first, uint32_t count, float value)` in the header and  to declare that an explicit instantiation for `T = float` and put `template __global__ void iota_n<float>(float* first, uint32_t count, float value)` in an implementation file. I did the same for all relevent `T` and most of the global functions in `common_device_helpers.hpp`. This had no meaningful effect on compile times. (~29m30s)

I tried only declaring the template function prototype in the header and putting both the definition and explicit instantiations in an implementation file. This had no meaningful effect on compile times. (~29m40s)

I tried putting everything back and just putting non-templated prototypes for the 4 versions of init_scalars in `init_scalars.hpp` and put the host template function and the device template function in `init_scalars.cpp`. This had no meaningful effect on compile times. (~29m30s)

The compiler is definitely spending most of its time on optimizing GPU kernels, but as far as I can tell, these small device kernels don't have much impact, despite being having dozens or even hundreds of instantiations across different translation units.

TorreZuk:
Well it is a pattern that may require more functions.   I didn't follow your first paragraph, you were looking for instantiations in the .so but that is after the linker removes duplicates, what you should have looked for is duplicates in the temporary object files.  Also the template prototype the compiler knows will be defined elsewhere, the kernel could be in the .cpp with the template definition.   Can discuss in a chat another time, but your analysis shows this function didn't seem to cause any measurable overhead so this  can be deferred until another more expensive function is to be shared.

PR #192, opened 2020-12-07
Skip pointer arg check if in query mode

tfalders:
In order to calculate appropriate memory sizes for hipSOLVER bufferSize functions, we need to be able to call rocSOLVER in query mode with some null or invalid pointers. This PR accommodates that by skipping the invalid pointer checks in argCheck functions when we are in query mode.


jzuniga-amd (approved):
LGTM

cgmb (approved):
👍

PR #193, opened 2020-12-08
fix bug in getf2 get mem size

jzuniga-amd:
(This bug could be related with problems we are observing in NAVI21 with hipBLAS)


cgmb (approved):
Looks good. I took a look at all the `getMemorySize` functions and I don't see this problem anywhere else. Though, I only looked at that first quick-return section (not the second, like exists here for small-size optimizations).

What sort of problems did this cause?

tfalders (approved):
LGTM!

PR #194, opened 2020-12-13
Switch to pragma once for internal headers

cgmb:
- Switch to `#pragma once` for internal headers
  - We already were using this in some of the rocblascommon files. I figured we might as well just use them everywhere internally. It avoids the need to update the guard when renaming files, and ensures we cannot have symbol collisions.
- Fix missing include guard in init_scalars.hpp
- ~Use consistent naming for externally-visible include guards~
  - ~All guards now consistently start with `ROCSOLVER` and reserved symbols are avoided.~ (I will investigate what other teams are doing before changing the externally-visible headers.)
  - Changed C sample from C11 to C99 to more carefully test our header compatibility.


rocsolver/clients/include/norm.hpp
```diff
@@ -3,8 +3,7 @@
  * ************************************************************************ */
 
 #pragma once
-#ifndef _NORM_HPP
-#define _NORM_HPP
+#pragma once
```
tfalders, line 6: Is that a second #pragma once?

tfalders (approved):
I do like consistency. LGTM!

It might be worth leaving rocsolver.h as it was originally, though, in case there's anyone using rocSOLVER with an #ifdef _ROCSOLVER_H_ somewhere in their code. That said, if we're going to change it, there's no time like the present.

rocsolver/clients/include/norm.hpp
cgmb: Whoops. Good catch. There's no point in using two `#pragma once` statements. My script just didn't recognize this case.

The pattern of using `#pragma once` followed by an include guard was typically used by Windows developers, as in the past, that would speed up compilation on MSVC. However, [there's no point in using both together these days](https://docs.microsoft.com/en-us/cpp/preprocessor/once?view=msvc-160). I will just remove the guard and leave the existing pragma.

cgmb:
The defines are undocumented, so I'm not too concerned about breaking code that relies upon them. On the other hand, we might want to coordinate a standard convention across the ROCm libraries, to ensure we only ever change it once. I'll undo my changes to the macros in the user-visible headers, and we can deal with the matter at a later date.

rocsolver/clients/include/norm.hpp
jzuniga-amd: you still have 2 pragma once!!

jzuniga-amd (approved):
LGTM

cgmb:
It looks like there are some issues with the extended tests. Given that pre-checkin passed on an earlier container, I'm going to go chalk that up to CI changes.

PR #195, opened 2020-12-15
Clients updates (reallocation and stream)

jzuniga-amd:
Now that we fixed the automatic reallocation in rocBLAS, and all our GPU operations are tied to the stream provided by the user, some updates are needed in the unit tests and bench:

1. Synchronization for timing is done on the provided stream; CPU operations don't need synchronization. 
2. Removed manual reallocation of workspace unless we set REALLOC=0. (rocBLAS owns the automatic reallocation switch, but in case it is disabled in the future, we can update our tests by simply setting REALLOC to zero). --We can also use the logic in the if statement to enable memory queries from rocsolver-bench in the future--.

### DIFF
--- a/rocsolver/clients/include/rocsolver_test.hpp
+++ b/rocsolver/clients/include/rocsolver_test.hpp
@@ -11,7 +11,7 @@
 
 // If REALLOC is 0, automatic reallocation is disable and we will manually
 // reallocate workspace if the memory query resulted in more than DEFAULT_MEM
-#define REALLOC 1         
+#define REALLOC 1
 #define DEFAULT_MEM 33554432
 
 #define ROCSOLVER_BENCH_INFORM(case)                                       \
--- a/rocsolver/clients/include/rocsolver_test.hpp
+++ b/rocsolver/clients/include/rocsolver_test.hpp
@@ -10,7 +10,7 @@
 #include <limits>
 
 // If USE_ROCBLAS_REALLOC_ON_DEMAND is false, automatic reallocation is disable and we will manually
-// reallocate workspace 
+// reallocate workspace
 #define USE_ROCBLAS_REALLOC_ON_DEMAND true
 
 #define ROCSOLVER_BENCH_INFORM(case)                                       \
--- a/rocsolver/clients/include/rocsolver_test.hpp
+++ b/rocsolver/clients/include/rocsolver_test.hpp
@@ -10,9 +10,8 @@
 #include <limits>
 
 // If USE_ROCBLAS_REALLOC_ON_DEMAND is false, automatic reallocation is disable and we will manually
-// reallocate workspace if the memory query resulted in more than DEFAULT_MEM
+// reallocate workspace 
 #define USE_ROCBLAS_REALLOC_ON_DEMAND true
-#define DEFAULT_MEM 33554432
 
 #define ROCSOLVER_BENCH_INFORM(case)                                       \
     do                                                                     \
--- a/rocsolver/clients/include/rocsolver_test.hpp
+++ b/rocsolver/clients/include/rocsolver_test.hpp
@@ -9,9 +9,9 @@
 #include <cstdarg>
 #include <limits>
 
-// If REALLOC is 0, automatic reallocation is disable and we will manually
+// If USE_ROCBLAS_REALLOC_ON_DEMAND is false, automatic reallocation is disable and we will manually
 // reallocate workspace if the memory query resulted in more than DEFAULT_MEM
-#define REALLOC 1
+#define USE_ROCBLAS_REALLOC_ON_DEMAND true
 #define DEFAULT_MEM 33554432
 
 #define ROCSOLVER_BENCH_INFORM(case)                                       \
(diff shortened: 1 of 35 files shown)


rocsolver/clients/include/rocsolver_test.hpp
```diff
@@ -9,6 +9,11 @@
 #include <cstdarg>
 #include <limits>
 
+// If REALLOC is 0, automatic reallocation is disable and we will manually
+// reallocate workspace if the memory query resulted in more than DEFAULT_MEM
+#define REALLOC 1
+#define DEFAULT_MEM 33554432
```
cgmb, line 15: - What's special about 33554432?

- Maybe use `true` / `false` for `REALLOC` to make it obvious that there's only two possible values.

- I think the name `REALLOC` is a little confusing, since we end up reallocating either way. It's really just choosing who does the reallocation. If it was called `USE_ROCBLAS_AUTO_REALLOC` or something of that sort, you could probably delete the comment.

rocsolver/clients/include/testing_gebd2_gebrd.hpp
```diff
@@ -357,14 +357,14 @@ void gebd2_gebrd_getPerfData(const rocblas_handle handle,
...
         // cpu-lapack performance (only if not in perf mode)
-        *cpu_time_used = get_time_us();
+        *cpu_time_used = get_time_us_no_sync();
         for(rocblas_int b = 0; b < bc; ++b)
         {
             GEBRD ? cblas_gebrd<S, T>(m, n, hA[b], lda, hD[b], hE[b], hTauq[b], hTaup[b], hW.data(),
                                       max(m, n))
                   : cblas_gebd2<S, T>(m, n, hA[b], lda, hD[b], hE[b], hTauq[b], hTaup[b], hW.data());
         }
-        *cpu_time_used = get_time_us() - *cpu_time_used;
```
cgmb, line 367: This seems good. Does the removal of the sync from the CPU code actually affect anything in practice? I assumed it was unnecessary, but harmless.

cgmb (commented):
That was an ugly todo. I like this change.

rocsolver/clients/include/rocsolver_test.hpp
jzuniga-amd: 32MB is the default size you get at rocblas_handle creation. If your function won't require more than that, there is no need to reallocate workspace.
I will change to USE_ROCBLAS_REALLOC_ON_DEMAND = true,false

rocsolver/clients/include/testing_gebd2_gebrd.hpp
jzuniga-amd: Yes, it is unnecessary and most of the time harmless: I would suppose that the users running the rocsolver-bench won't have any other process using the GPU at the same time but, just in case, better to make it explicit that we don't need any device synchronization to time CPU LAPACK.

rocsolver/clients/include/rocsolver_test.hpp
cgmb: Whoops. I didn't notice my calculator was using Megabytes rather than Mebibytes. That all sounds good to me.

cgmb (approved):

rocsolver/clients/include/testing_bdsqr.hpp
```diff
@@ -481,6 +484,21 @@ void testing_bdsqr(Arguments argus)
...
+    {
+        CHECK_ROCBLAS_ERROR(rocblas_start_device_memory_size_query(handle));
+        CHECK_ALLOC_QUERY(rocsolver_bdsqr(handle, uplo, n, nv, nu, nc, (S*)nullptr, (S*)nullptr,
+                                          (T*)nullptr, ldv, (T*)nullptr, ldu, (T*)nullptr, ldc,
+                                          (rocblas_int*)nullptr));
+
+        size_t size;
+        CHECK_ROCBLAS_ERROR(rocblas_stop_device_memory_size_query(handle, &size));
+
+        if(size > DEFAULT_MEM)
```
tfalders, line 498: Should we perhaps compare size to the result of rocblas_get_device_memory_size instead?

tfalders (approved):
LGTM, though I do have one suggestion.

rocsolver/clients/include/testing_bdsqr.hpp
jzuniga-amd: Right! That would be the correct way to do it. I forgot that users can actually modify the default by setting an environment variable. Will change it.

jzuniga-amd: A simpler option is to always do the reallocation with size. I don't think it really matters.

tfalders: That could also work. If we run the tests with automatic memory reallocation, this block shouldn't trigger and it won't impact the test run time.

jzuniga-amd: going for the easiest way...

cgmb:
👍

PR #199, opened 2021-01-02
Build system restructuring - 1/5

cgmb:
This first step is some CMake cleanup. There's a lot of changes here,
but they should still result in more or less the same build commands
being generated. This particular commit is not terribly exciting stuff;
I'm just doing some tidying up:

- Cleanup CMake comments
- Reorganize helper functions
- Simplify function calls
- Update package description

The end goal is a cleaner, simpler build configuration. Step 2 is going
to be applying an auto-formatter. Step 3 is going to be porting some
changes rocBLAS has already made. Step 4 is going to be look at
how we handle our dependencies. And Step 5 is going to be some
more misc cleanup.

I split the changes into steps just for clarity, as it would be difficult to
follow what was done if they were all squashed together.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,30 +1,34 @@
 # ########################################################################
-# Copyright (c) 2019-2020 Advanced Micro Devices, Inc.
+# Copyright (c) 2019-2021 Advanced Micro Devices, Inc.
 # ########################################################################
 
 cmake_minimum_required( VERSION 3.8 )
 
-# We use C++14 features, this will add compile option: -std=c++14
 set( CMAKE_CXX_STANDARD 14 )
-# Without this line, it will add -std=gnu++14 instead, which has some issues.
 set( CMAKE_CXX_EXTENSIONS OFF )
+set( CMAKE_CXX_STANDARD_REQUIRED ON )
 
 # This has to be initialized before the project() command appears
-# Set the default of CMAKE_BUILD_TYPE to be release, unless user specifies with -D.  MSVC_IDE does not use CMAKE_BUILD_TYPE
+# Set the default of CMAKE_BUILD_TYPE to Release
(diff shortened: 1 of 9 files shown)


TorreZuk (commented):
LGTM but you changed ROCM_USE_FLOAT16 from a global to only be used in the library.  It changes the type if following rocblas and so the tests likely would want to use that same type (if you start using half types).   I would apply it to the clients as well or you could be creating a subtle bug in the future.   Not sure why the globals should generate warning/errors for these cases where the design of global or project settings are actually better than setting it for each target manually.   But that is a different discussion.

cgmb:
> you changed ROCM_USE_FLOAT16 from a global to only be used in the library

I think you may be mistaken. I moved the setting of `ROCM_USE_FLOAT16` on the library from `CMakeLists.txt` to `library/src/CMakeLists.txt`, but it's a private target property on the library either way, so the location of the call shouldn't make a difference. It was the only library-specific setting in the main file, so I just put it with the other library-specific settings. The clients are setting that define separately in their own files, and I left that aspect unchanged.

> Not sure why the globals should generate warning/errors for these cases where the design of global or project settings are actually better than setting it for each target manually.

I agree. It is pretty gnarly to be duplicating common settings on all of our targets. For now, I tend towards that ugliness because I know it's safe.

TorreZuk:
Very good, I see now it was library target based define even at the top level which is maybe why I didn't read it correctly.   Shifting it down to the target level file makes sense.   I think globals could go into toolchain files and not generate warnings as that is what the message says it expects, but this is good.

TorreZuk (approved):

jzuniga-amd (approved):
LGTM

tfalders (approved):
I don't know CMake very well, but LGTM!

PR #168, opened 2020-10-18
Unit test improvement

jzuniga-amd:

PR #203, opened 2021-01-15
Fix B when GELS info is nonzero

cgmb:
This change is to save and restore B when info is non-zero, to ensure that we only change B iff info is zero. I've looked into some better solutions, but they will probably require some coordination with rocBLAS, and I figure it's better to get a simple fix in now and improve it later.

I don't think GELS has actually been in a release yet, so I suppose this doesn't need a changelog entry.


rocsolver/clients/include/testing_gels.hpp
```diff
@@ -260,22 +293,23 @@ void gels_getPerfData(const rocblas_handle handle,
...
         gels_initData<true, false, T>(handle, trans, m, n, nrhs, dA, lda, stA, dB, ldb, stB, dInfo,
-                                      bc, hA, hB, hInfo);
+                                      bc, hA, rankd, hB, hInfo);
         // cpu-lapack performance (only if not in perf mode)
         *cpu_time_used = get_time_us_no_sync();
         for(rocblas_int b = 0; b < bc; ++b)
         {
-            cblas_gels<T>(trans, m, n, nrhs, hA[b], lda, hB[b], ldb, hW.data(), sizeW);
+            rocblas_int info;
+            cblas_gels<T>(trans, m, n, nrhs, hA[b], lda, hB[b], ldb, hW.data(), sizeW, &info);
```
tfalders, line 302: I'd suggest hInfo[b] instead of info.

rocsolver/clients/include/rocsolver_arguments.hpp
```diff
@@ -59,6 +59,8 @@ class Arguments
     rocblas_int bsp = 128;
     rocblas_int bs5 = 128;
 
+    bool rankd = false; // rank deficient
```
tfalders, line 62: This seems to serve the same purpose as the "singular" field and should be combined with it. It may worth renaming it to something more general, such as "degenerate".

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -11,35 +11,36 @@ using ::testing::Values;
...
     // normal (valid) samples
-    {20, 20, 20, 20},
-    {30, 20, 40, 30},
-    {40, 20, 40, 40},
+    {20, 20, 20, 20, 0},
+    {20, 20, 20, 20, 1},
+    {30, 20, 40, 30, 0},
+    {30, 20, 40, 30, 1},
+    {40, 20, 40, 40, 0},
+    {40, 20, 40, 40, 1},
```
tfalders, line 43: There is a fairly new convention for testing degenerate cases that would be a good fit for what you're doing. See getri_gtest.cpp for an example.

tfalders (commented):
The testing behaviour in this PR reminds me a lot of the singular tests that were added in PR #168. For consistency, please study one of them (e.g. getri) and adopt a similar style.

cgmb:
Yeah, I saw getri after finishing and tried to adapt what I'd written to match, though, it seems my adjustments were incomplete. I'll update to match.

cgmb:
I think it's consistent now.

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -38,7 +36,10 @@ const vector<gels_params_A> matrix_sizeA_range = {
     {10, 1, 10, 10},
     // normal (valid) samples
     {20, 20, 20, 20},
+    {20, 20, 20, 20},
     {30, 20, 40, 30},
+    {30, 20, 40, 30},
+    {40, 20, 40, 40},
     {40, 20, 40, 40},
```
tfalders, line 43: The duplicated test cases can be removed. Also, you need to set arg.singular in gels_setup_arguments. You can add an extra value to the test case arrays like you had before to indicate if singular tests are on/off, but you won't need to duplicate any of the cases. Something like this will work:
    {20, 20, 20, 20, 1},
    {30, 20, 40, 30, 0},
    {40, 20, 40, 40, 1},

rocsolver/clients/gtest/gels_gtest.cpp
```diff
@@ -57,10 +58,8 @@ const vector<gels_params_B> matrix_sizeB_range = {
 
 // for daily_lapack tests
 const vector<gels_params_A> large_matrix_sizeA_range = {
-    {75, 25, 75, 75},
-    {150, 150, 150, 150},
-    {500, 50, 600, 600},
-    {1000, 500, 1000, 1000},
+    {75, 25, 75, 75},    {150, 150, 150, 150},    {150, 150, 150, 150},
+    {500, 50, 600, 600}, {1000, 500, 1000, 1000},
```
tfalders, line 62: Add the extra value to enable singular tests here too.

tfalders (changes requested):
Almost there. Some small changes to gels_gtest.cpp, and I think it should be good to go.

tfalders (approved):
LGTM!

jzuniga-amd (commented):
Looks good!
Some minor changes. Please also see PR I opened in your branch with other changes that we did not catch in the original GELS PR.

jzuniga-amd:
Also, you may need a changelog entry as this fix wont be in 4.1 :(

rocsolver/clients/include/testing_gels.hpp
```diff
@@ -168,6 +173,27 @@ void gels_initData(const rocblas_handle handle,
...
+                do
+                {
+                    if(n < m) // zero random row
+                    {
+                        rocblas_int i = sample_index(rocblas_rng);
+                        for(rocblas_int j = 0; j < n; j++)
+                            hA[b][i + j * lda] = 0;
+                    }
+                    else // zero random col
+                    {
```
jzuniga-amd, line 190: to ensure rank deficiency, if m >= n you should introduce a **zero column**; if m < n a **zero row**.

cgmb: Upon encountering some failing tests, I swapped the sign figuring I'd somehow got it backwards. (I was actually going to come back and check that before marking your previous comments as resolved.) In any case, you're absolutely right. I've corrected the error.

It might be nice to isolate and test these matrix generators at some point. The outputs they generate are not very visible, but they're important (and unlike a lot of helper code for the tests, the correctness of those outputs doesn't get implicitly tested by the tests they're used in).

jzuniga-amd (approved):
LGTM

PR #204, opened 2021-01-15
Multi-level logging

tfalders:
Adds multi-level logging functionality to all rocSOLVER functions.


rocsolver/library/src/include/rocsolver_logger.hpp
```diff
@@ -0,0 +1,231 @@
...
+#include "common_host_helpers.hpp"
+#include "common_ostream_helpers.hpp"
+#include "rocsolver.h"
+#include <unordered_map>
+
+/***************************************************************************
+ * rocSOLVER logging macros
+ ***************************************************************************/
+
+#define ROCSOLVER_ENTER_TOP(name, ...)                                              \
```
cgmb, line 16: This could be made less invasive if these enter functions created an object on the stack to represent the scope. Log entering the scope in the object's constructor and log leaving the scope in its destructor. The object would store the name of the scope, so that you only have to specify it once at the top of the function. Upon destruction, it can use the stored name.

Doing it that way would avoid having to think about logging at all when leaving the scope.

cgmb (commented):
It's nice to see these todos being removed!

This change affects almost every facet of the library, with implications that extend from the API, to the ABI, to how we write each function of the implementation, so there's a lot to review. I'm not going to be able to get through all of it today, but I do have a couple of suggestions to start.

rocsolver/library/src/include/rocsolver_logger.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+        call_stack[handle].pop_back();
+    }
+
+    friend rocblas_status rocsolver_logging_initialize(const rocblas_layer_mode layer_mode,
+                                                       const rocblas_int max_levels);
+    friend rocblas_status rocsolver_logging_cleanup(void);
+};
+
+extern rocsolver_logger* logger;
```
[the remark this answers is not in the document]
tfalders: Yeah, I agree that the logger name is a little too general and could be renamed or moved into a different scope. At one point I was attempting to make this variable a static member of the rocsolver_logger class, but I ran into a linker issue and left it as it was. I may try again.

rocsolver/library/src/include/rocsolver_logger.hpp
```diff
@@ -0,0 +1,231 @@
...
+    // logging function to be called upon entering a top-level (i.e. impl) function
+    template <typename T, typename... Ts>
+    void log_enter_top_level(rocblas_handle handle,
+                             const char* func_prefix,
+                             const char* func_name,
+                             Ts... args)
+    {
+        int level = call_stack[handle].size();
+        if(level != 0)
+            ROCSOLVER_UNREACHABLE();
```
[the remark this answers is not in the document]
tfalders: Sounds good. I can remove the 'if' on that if I change it, correct?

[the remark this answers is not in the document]
cgmb: Yup.

rocsolver/library/src/include/rocsolver_logger.hpp
tfalders: I like the idea, but will I have to get rid of the "if (logger != nullptr && logger->is_logging_enabled)" conditionals in order to keep the object out of the scope of an if statement? Part of my philosophy with this PR is to keep things lightweight if the logging isn't enabled, so I hope that any such object can be kept within the scope of the if statement.

cgmb: The `if` would need to be within the constructor of the object, but there should be negligible overhead as long as:
1. The definition of the constructor and destructor are available inline.
2. The object has only [trivial types](https://en.cppreference.com/w/cpp/types/is_trivial) as member variables.

cgmb: That said, I'll put together a small proof-of-concept to check.

rocsolver/library/src/include/rocsolver_logger.hpp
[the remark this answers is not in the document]
cgmb: That actually raises a more important point: initialization. I think a global variable like `logger` could potentially be uninitialized if a user called a rocsolver function before entering `main` (since the order of initialization of global variables in different translation units is unspecified). A rocsolver function could be called before main if, for example, a user created a global variable and defined its constructor to call into rocsolver.

The typical solution to that problem is the singleton pattern. Static variables declared within functions will be initialized the first time program execution reaches them.  You can find [a few examples of this pattern in `rocblas_ostream`](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/6b820a8eb67396acb6002612748196b8205a6e17/rocblascommon/library/src/include/rocblas_ostream.hpp#L146-L158). So maybe the naming point is moot.

That said, I'm still looking wistfully at tying the logging state to a handle. I don't like dealing with global mutable state. It can be rather tricky to get right.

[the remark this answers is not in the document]
cgmb: Ah, yeah. That is constant initialization, so that aspect of initialization isn't anything to worry about.

rocsolver/library/src/rocsolver_logger.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return 'c';
+}
+template <>
+char rocsolver_logger::get_precision<rocblas_double_complex>()
+{
+    return 'z';
+}
+
+double rocsolver_logger::get_time_us()
+{
```
jzuniga-amd, line 99: Similarly, these are in rocblascommon/clients/common/utility...
Is there a way to re-use this code? or is there an advantage of having a library and a clients version of these helpers?

jzuniga-amd (commented):
Code looks good to me. Some not very important comments...
I will be testing the functionality this morning.

rocsolver/library/src/rocsolver_logger.cpp
tfalders: I think the main advantage is it makes the organizational structure of the code easier to understand. If a user wanted, for some reason, to get rid of all the "client" directories, as it stands now I think it should still be possible to build the library without them. I don't think we can build the clients if all "library" directories were deleted, but I'm a little more okay with including library code in clients than the other way around.

It might be worth it for us to create a "common" directory that isn't split into "client" and "library" subdirectories (unlike "rocblascommon"), where we can place all the code that is common to both - mainly the output and timing code. I believe Cory also had some ideas for reorganizing the library, like moving "rocblascommon" into the "library" and "client" directories. I can try to put something together in a separate PR if this is something we want to pursue.

cgmb: Thumbs up for a common directory that is used in both client and library. That sounds like a good way to prevent duplication, and it would still be consistent with my desire to put the client-only parts of rocblascommon under clients, and the library-only parts of rocblascommon under library.

jzuniga-amd: Sounds like a good plan. So, should we leave this "add logging" PR as is and reorganize in a separate PR?... (I think that would be better)

tfalders: I agree. This PR is big enough as it is without reorganizing the directory structure.

rocsolver/library/src/include/rocsolver_logger.hpp
tfalders: I think I've managed to develop a version of this idea that I'm happy with. I'll push it along with the other changes once it's ready.

rocsolver/library/src/rocsolver_logger.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+extern "C" {
+
+rocblas_status rocsolver_logging_initialize(const rocblas_layer_mode layer_mode,
+                                            const rocblas_int max_levels)
+{
+    rocsolver_logger::_mutex.lock();
+
+    if(rocsolver_logger::_instance != nullptr)
+        return rocblas_status_internal_error;
```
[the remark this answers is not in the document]
tfalders: Good catch! I will fix that up.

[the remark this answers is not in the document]
tfalders: Oh, I see that's been handled in your PR. Thanks!

rocsolver/library/src/include/rocsolver_logger.hpp
cgmb: That's a little fancier than what I was thinking, but it looks good to me. The only thing I'd suggest is explicitly marking the copy constructor and assignment operator as deleted on your scope object.

rocsolver/library/src/rocsolver_logger.cpp
```diff
@@ -0,0 +1,264 @@
...
+std::mutex rocsolver_logger::_mutex;
+
+/***************************************************************************
+ * Open logging streams
+ ***************************************************************************/
+
+auto rocsolver_logger::open_log_stream(const char* environment_variable_name)
+{
+    const char* logfile;
+    if((logfile = getenv(environment_variable_name)) != nullptr
```
cgmb, line 22: It *probably* makes no difference, but I would use `std::getenv`, as it has slightly stronger threading guarantees than the C version.

rocsolver/library/src/rocsolver_logger.cpp
```diff
@@ -0,0 +1,264 @@
...
+        rocsolver_logger::_mutex.unlock(); 
+        return rocblas_status_internal_error;
+    }
+
+    auto logger = rocsolver_logger::_instance = new rocsolver_logger();
+
+    // set layer_mode from environment variable ROCSOLVER_LAYER or to default
+    const char* str_layer_mode = getenv("ROCSOLVER_LAYER");
+    if(str_layer_mode)
+        logger->layer_mode = static_cast<rocblas_layer_mode>(strtol(str_layer_mode, 0, 0));
```
cgmb, line 104: `strtol` can fail and/or return values larger than can be stored in `rocblas_layer_mode`. There should probably be a check for that. Not sure if that should result in `rocblas_status_internal_error` or just defaulting to `rocblas_layer_mode_none`.

The same is true of `max_levels` if `INT_MAX < LONG_MAX`.

tfalders: I believe I copied that straight from rocBLAS. I'm not sure how to fix it, but such a fix should probably also go into rocBLAS.

rocsolver/library/src/include/rocsolver_logger.hpp
```diff
@@ -65,12 +54,14 @@
 struct rocsolver_log_entry
```
doctorcolinsmith, line 54: As a sanity check, confirm that there is an implicit move constructor generated for this struct.  I'm not sure if that happens when there is an explicit default constructor defined.  Could help avoid some string copies.

cgmb: [It does](https://en.cppreference.com/w/cpp/language/move_constructor#Implicitly-declared_move_constructor). It's just if you define a custom copy constructor / assignment operator / destructor that the implicit move constructor will be suppressed.

docs/source/userguide_logging.rst
```diff
@@ -0,0 +1,98 @@
...
+Multiple host threads
+================================================
+
+The logging facilities for rocSOLVER assume that each ``rocblas_handle`` is associated with at
+most one host thread. When using rocSOLVER's multi-level logging setup, it is recommended to
+create a separate ``rocblas_handle`` for each host thread.
+
+Calling any of these logging functions, especially ``rocsolver_log_begin`` and ``rocsolver_log_end``,
+while another rocSOLVER routine is running on another thread may lead to undefined behaviour
+and is not recommended.
```
cgmb, line 94: This is accurate, but the statement should be stronger. The words 'may' and 'especially' are out-of-place here. We need to communicate the interface requirements, which we define (and are statements made with certainty).

Based on your writing here, I think we want to say something like:
> The `rocsolver_log_*` functions are not thread-safe. It is undefined behavior to call a log function while any rocsolver function is executing on another host thread.
>
> Once enabled, logging data collection is thread-safe. However, note that trace logging will likely result in garbled trace trees if rocSOLVER routines are called from multiple host threads.

Though, I'm a little unclear on the threading guarantee for trace logging. Is it a guarantee that the worst that can happen is that the output is garbled? If it could cause a crash, we probably don't need to go into as much detail about what the result is. In that case, we'd just state that it's not thread-safe to perform rocsolver operations on multiple threads with trace logging enabled.

docs/source/userguide_logging.rst
```diff
@@ -0,0 +1,98 @@
...
+set, then the corresponding logging output is streamed to standard error.
+
+The results of profile logging, if enabled, can be printed using ``rocsolver_log_write_profile``
+or ``rocsolver_log_flush_profile``. Once logging facilities are no longer required (e.g. at
+program termination), the user must call ``rocsolver_log_end`` to free the data structures used
+for logging.
+
+Note that when profile logging is enabled, memory usage will increase. If the program exits
+without calling ``rocsolver_log_write_profile`` or ``rocsolver_log_end``, then profile
+logging will not be outputted before the program exits.
```
cgmb, line 82: You might just want to say that profile logging will not automatically be printed on program exit (like is done in rocblas?). You already covered how to manually print in the previous paragraph.

cgmb (approved):
The ABI looks fine. Our benchmark and test clients will be tied to the particular version of rocSOLVER that they were built against, but normal library users will not be.

I had a comment about the documentation, but that's it. This will be a very handy feature. Nice work!

docs/source/userguide_logging.rst
tfalders: I believe the worst that can happen for trace logging is garbled output, though I have not tested it. Calling rocSOLVER routines should only affect the call_stack and profile members of rocsolver_logger, and I'm pretty sure all R/W operations on those are mutex-locked.

cgmb: Sounds good to me.

cgmb:
I have no further comments. LGTM. :+1:

jzuniga-amd (commented):
Very nice!!!
One last simple modification...

rocsolver/library/src/rocsolver_logger.cpp
```diff
@@ -0,0 +1,367 @@
...
+/***************************************************************************
+ * Profile log printing
+ ***************************************************************************/
+
+void rocsolver_logger::write_profile(rocsolver_profile_map::iterator start,
+                                     rocsolver_profile_map::iterator end)
+{
+    for(auto it = start; it != end; ++it)
+    {
+        rocsolver_profile_entry& entry = it->second;
```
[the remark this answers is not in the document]
tfalders: I thought about doing that, but wasn't sure if we wanted to trim the profile output as well. I'll get on that right now.

jzuniga-amd (approved):
LGTM.

PR #206, opened 2021-01-20
Remove support for hcc - build system - 2/5

cgmb:
Most of the changes here are following the lead from rocBLAS, which has
already removed their support for hcc.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -5,6 +5,8 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false)
 {
     project.paths.construct_build_prefix()
 
+    String compiler = 'hipcc'
+    String hipClang = ''
     String sles = platform.jenkinsLabel.contains('sles') ? '/usr/bin/sudo --preserve-env' : ''
     String debug = project.buildName.contains('Debug') ? '-g' : ''
     String centos = platform.jenkinsLabel.contains('centos') ? 'source scl_source enable devtoolset-7' : ''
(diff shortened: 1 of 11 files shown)


TorreZuk (approved):
Didn't spot anything but get your docs as well as hcc may be mentioned as was in rocBLAS.

.jenkins/common.groovy
```diff
@@ -5,8 +5,8 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false)
 {
     project.paths.construct_build_prefix()
 
-    String compiler = jobName.contains('hipclang') ? 'hipcc' : 'hcc'
-    String hipClang = jobName.contains('hipclang') ? '--hip-clang' : ''
+    String compiler = 'hipcc'
+    String hipClang = ''
```
tfalders, line 9: Should this be '--hip-clang'?

tfalders (approved):
LGTM!

.jenkins/common.groovy
cgmb: That's a good eye you've got there, but the inconsistency in this pair of line changes is because the original code didn't actually work anymore.

That hipClang property specifies what options to pass to build for the CI to choose between hip-clang and hcc, but since the default is now hip-clang, both branches of that ternary operator will result in `./install.sh` doing the same thing. I'd remove the property entirely, but the CI expects it to exist.

cgmb:
> Didn't spot anything but get your docs as well as hcc may be mentioned as was in rocBLAS.

Thanks for the tip! When I was preparing this PR, I did a grep for HCC and hcc, so those cases should be covered. I did update a few spots in the build instructions.

jzuniga-amd (approved):

PR #208, opened 2021-01-26
Add cleanup option to install script

eidenyoshida:
Targeting large intermediate build directories to reduce stage 2 container size. Removing _CPack_Packages will save 3.9 gigabytes

### DIFF
--- a/.jenkins/precheckin.groovy
+++ b/.jenkins/precheckin.groovy
@@ -15,7 +15,7 @@ def runCI =
 
     def prj = new rocProject('rocSOLVER', 'PreCheckin')
     // customize for project
-    prj.paths.build_command = './install.sh -c'
+    prj.paths.build_command = './install.sh -c --cleanup'
 
     // Define test architectures, optional rocm version argument is available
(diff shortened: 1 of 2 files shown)


cgmb:
On my local machine, that appears to save 4GB. You could trim another ~2GB with a `make clean` (or if you would like to keep the generated binaries, a bit less with `find -name '*.o' -delete`).

cgmb (approved):

cgmb:
LGTM. I only tested a debug build, but that went from 3.3G -> 1.1G.

jzuniga-amd (approved):
LGTM

tfalders (approved):
LGTM!

cgmb:
@eidenyoshida, squash and merge whenever you're ready.

cgmb:
I don't know how we missed this, but this change has significantly increased the build time for rocSOLVER on the CI. It would be fine if we did the build and the package creation in one go, but the CI does it as two steps, with the cleanup of the temporary files in the middle.

PR #209, opened 2021-01-28
Improve dependency isolation - build system - 3/5

cgmb:
These changes to the dependency list will require testing on all supported platforms, so I'm going to be using the CI to run some tests as I go along. I'll switch this out of draft mode when it's ready for review.

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -35,7 +35,7 @@ find_package( Threads REQUIRED )
 
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
-find_package( cblas REQUIRED )
+find_package( cblas REQUIRED CONFIG )
 
 set( rocsolver_benchmark_common
       ../common/lapack_host_reference.cpp
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -35,10 +35,7 @@ find_package( Threads REQUIRED )
 
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
-find_package( cblas CONFIG REQUIRED )
-if( NOT cblas_FOUND )
-  message( FATAL_ERROR "cblas is a required dependency and is not found;  try adding cblas path to CMAKE_PREFIX_PATH" )
-endif( )
+find_package( cblas REQUIRED )
 
 set( rocsolver_benchmark_common
       ../common/lapack_host_reference.cpp
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -35,10 +35,7 @@ endif( )
 
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
-find_package( cblas CONFIG REQUIRED )
-if( NOT cblas_FOUND )
-  message( FATAL_ERROR "cblas is a required dependency and is not found;  try adding cblas path to CMAKE_PREFIX_PATH" )
-endif( )
+find_package( cblas REQUIRED )
 
 find_package( GTest REQUIRED )
 
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -35,7 +35,7 @@ endif( )
 
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
-find_package( cblas REQUIRED )
+find_package( cblas REQUIRED CONFIG )
 
 find_package( GTest REQUIRED )
 
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -42,7 +42,6 @@ set( fortran_samples
 foreach( exe ${c_samples} ${cpp_samples} ${fortran_samples} )
   target_link_libraries( ${exe} PRIVATE roc::rocsolver roc::rocblas )
 
-  set_target_properties( ${exe} PROPERTIES CXX_EXTENSIONS NO )
   set_target_properties( ${exe} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
 
   target_include_directories( ${exe}
@@ -76,9 +75,11 @@ foreach( exe ${c_samples} ${cpp_samples} ${fortran_samples} )
 endforeach( )
 
 foreach( exe ${cpp_samples} )
-  if( NOT CUDA_FOUND )
-    target_link_libraries( ${exe} PRIVATE hip::device )
-  endif()
+  set_target_properties( ${exe} PROPERTIES
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED ON
+    CXX_EXTENSIONS OFF
+  )
 endforeach( )
 
 foreach( exe ${c_samples} )
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -57,6 +57,10 @@ foreach( exe ${c_samples} ${cpp_samples} ${fortran_samples} )
     # GCC or hip-clang needs specific flags to turn on f16c intrinsics
     target_compile_options( ${exe} PRIVATE -mf16c )
   endif( )
+
+  if( NOT CUDA_FOUND )
+    target_link_libraries( ${exe} PRIVATE hip::host )
+  endif()
 endforeach( )
 
 foreach( exe ${cpp_samples} )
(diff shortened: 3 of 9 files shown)


cgmb:
@tfalders, @jzuniga-amd. Ok. I finally know what to do here. The code doesn't reflect it yet, but I wanted to communicate the design. My proposal is basically this: 

## What
Change the install path for `./install.sh -d` from `/usr/local` to `<builddir>/deps/local` and make that the preferred location for finding dependencies (if a dependency isn't satisfied by something in `<builddir>/deps/local`, then `/usr/local` will be checked afterwards).

## Why
We've had problems with users of multiple ROCm libraries running `./install.sh -d` in parallel. Since they all overwrite each other in the global `/usr/local`, we sometimes will see build failures caused by building half the library with one copy and half the library with a different copy. By avoiding writing to `/usr/local`, we avoid causing such breakages in other libraries, and by preferring our own copies in `<builddir>/deps/local`, we can avoid being broken by other libraries that continue to use the old behaviour.

## User Impact
The main difference in visible behaviour is this: before when you ran `./install.sh -d`, you only needed to do so once per machine. Now, however, you will need to do so whenever you have a different build directory. That is a change in behaviour, but it doesn't break either the mathci or rocmci build scripts. As for external users, I think most people will only have one copy of the source code on their machines and will probably not notice the difference. Though I will mention it in the changelog.

## Extra Benefits:
- When we installed into `/usr/local`, we had to coordinate versions with all other ROCm libraries, and would get breakages on the CI if we failed to do so. This will avoid the need for that.
- There was no way to cleanly uninstall the installed libraries that had been built from source. With them installed into the build directory, they now are just removed when you delete the build dir. (This is also a bonus for upgrading versions. A way to cleanly remove the old version helps guarantee we won't have problems when updating versions.)

tfalders:
Will there still be the option to install to `usr/local` as before?

cgmb:
I wasn't planning on it, though I could add one. Maybe `--global-deps`?

tfalders:
Sounds good to me!

cgmb:
This is ready for review. The changes I've made are probably easiest to understand if you go through the commits. Most of them have descriptions of why I made the change I did. Not all of them are truly necessary. I included a few quality-of-life improvements to the build scripts that could be split off if we want to keep them separate.

I'm currently running through a set of manual tests on various platforms to exercise more cases than the math-ci will check.

clients/benchmarks/CMakeLists.txt
```diff
@@ -35,10 +35,7 @@ find_package( Threads REQUIRED )
 
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
-find_package( cblas CONFIG REQUIRED )
-if( NOT cblas_FOUND )
-  message( FATAL_ERROR "cblas is a required dependency and is not found;  try adding cblas path to CMAKE_PREFIX_PATH" )
-endif( )
+find_package( cblas REQUIRED )
```
TorreZuk, line 38: From what I read the CONFIG style is the more modern approach, based on your lack of other parameters this will first try the older MODULE style then I expect fail and fallback to CONFIG mode.  Why not just keep CONFIG I think these have cblas-config.cmake (same for lapack)

cgmb:
I'm requesting a few other reviewers, as I believe this is a fairly common problem among our library install.sh scripts.

TorreZuk (commented):
While the cleanup is good, I will likely suggest we move to python for the cmake configuration layer and leave only maybe dependency installation where required in the bash script.  Keep this in mind as this could happen as soon as next release cycle.   Not sure what can be moved fully into the cmake vs. dep install.sh.   As for isolation we should also try to unify the dependencies wherever possible as presenting a library dependency set that is as simple and consistent as possible should be one of our goals.

clients/gtest/CMakeLists.txt
```diff
@@ -35,10 +35,7 @@ endif( )
 
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
-find_package( cblas CONFIG REQUIRED )
-if( NOT cblas_FOUND )
-  message( FATAL_ERROR "cblas is a required dependency and is not found;  try adding cblas path to CMAKE_PREFIX_PATH" )
```
saadrahim, line 40: What's the reason behind removing the warning?

clients/samples/CMakeLists.txt
```diff
@@ -76,9 +75,11 @@ foreach( exe ${c_samples} ${cpp_samples} ${fortran_samples} )
 endforeach( )
 
 foreach( exe ${cpp_samples} )
-  if( NOT CUDA_FOUND )
-    target_link_libraries( ${exe} PRIVATE hip::device )
```
saadrahim, line 80: was this an unused path? CUDA support will be needed for most libraries soon.

saadrahim:
math-ci does not install dependencies with install.sh. You are responsible for testing it prior to it reaching rocm-ci. math-ci installs the same dependencies for all the libraries into a common docker container.

Personally, I am against spending time on dependencies as it should be user managed. For shared computers, we should be moving to modules to get dependencies.

clients/gtest/CMakeLists.txt
eidenyoshida: The REQUIRED option in the find_package below fails the build if blas is not found, so the message is not necessary

cgmb: The `REQUIRED` argument to `find_package` halts processing if the package can't be found, so this message should never be reached.

clients/samples/CMakeLists.txt
cgmb: If I understand correctly, `hip::device` is only needed by code that directly uses kernels. The rocsolver library should be ensuring `hip::host` is linked via its interface requirements, and so our samples shouldn't need special handling.

That's good to know about CUDA. There's lots of special-casing for CUDA in our CMake and I'd be very surprised if it actually works given that we've not been building for it. I really hate to see broken code lying around, so I was going to suggest we should either fix it or remove it.

clients/benchmarks/CMakeLists.txt
cgmb: My reasoning was basically:
- Searching a few extra paths for the module is basically instantaneous.
- Using the default CMake behaviour is less code.
- Users and developers are usually more familiar with how the default works.

That said, it has been marked as CONFIG since forever. Maybe I should be considering that as our default... Hmmm....

I'll switch this back to CONFIG.

clients/samples/CMakeLists.txt
cgmb: I just realized that the reasoning I gave you about `hip::host` implies that a bunch of the code on the lines above this comment are unnecessary. So, I removed it in 37829194c741bfd306fb29db15c30d7e7b3c1c64.

cgmb:
I've tested on CentOS 7, CentOS 8, SLES 15 and Ubuntu 18.04. 

What I'm getting from @TorreZuk and @saadrahim is that there are going to be major changes in the near future, so there's not much enthusiasm for spending a lot of time on this. Some of what was done here is definitely going to be replaced.  My take: then let's just go ahead and merge. It's not going to hurt those efforts. It may even help (particularly trimming the dependency list). And it fixes things right now.

> As for isolation we should also try to unify the dependencies wherever possible as presenting a library dependency set that is as simple and consistent as possible should be one of our goals.

Agreed. We will always need to ensure that there exists a version of each dependency that satisfies the build requirements for all libraries as that's a standard requirement for most distros. The CI will still always be testing that requirement.

cgmb:
There is one rather important issue that I'm skipping out on dealing with: the dependency on rocBLAS. The dependency on rocblas is not being publicly declared in either our exported CMake config or in our DEB/RPM packages, which seems weird to me. However, it was like that before my PR and will be like that after my PR.

cgmb:
Would it make this easier to approve if I just change the default back to `/usr/local`? I don't feel like I've sold you on changing how dependencies are handled, but we might as well take the code cleanup even if we don't take the design changes.

tfalders (approved):
LGTM, though I'm not an expert on the build system by any means. As long as cleaning up the build directory won't cause the install to break and we have the option to change the install location, I'm fine with installing the dependencies to the build directory by default.

cgmb:
After some discussion on this subject and further reflection, I feel the main concept of this PR (dependency isolation) has been rejected in favour of dependency coordination.

Rather than gut this PR of its original purpose, I'm just going to close it. There are pieces worth salvaging, but they can be moved into smaller, more focused PRs. For now, the fix for users running `./install.sh -d` in parallel can simply be a note in the `--help` text that says not to do that.

Coordination is indeed a better solution than isolation. It's just more difficult. Fundamentally, the problem is that we lack a place where that coordination occurs. So, the path forward is to find or create a place for the coordination.

Thank you everyone for the feedback.

> I recall that after we had conducted thousands of experiments on a certain project without solving the problem, one of my associates, after we had conducted the crowning experiment and it had proved a failure, expressed discouragement and disgust over our having failed ‘to find out anything.’ I cheerily assured him that we had learned something. For we had learned for a certainty that the thing couldn’t be done that way, and that we would have to try some other way. We sometimes learn a lot from our failures if we have put into the effort the best thought and work we are capable of.

~ Thomas Edison, in *American Magazine* Vol. 91 (January 1921)

PR #210, opened 2021-02-03
File reorganization 2

tfalders:
Reorganizing files as per https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/204#discussion_r564686314

Summary of changes:

- Removed top-level rocsolver directory
    - rocsolver/clients -> clients
    - rocsolver/library -> library
- Removed top-level rocblascommon directory
    - rocblascommon/clients/common & rocblascommon/clients/include -> clients/rocblascommon
    - rocblascommon/cmake -> cmake
    - rocblascommon/deps -> deps
    - rocblascommon/library/src -> library/src/rocblascommon
- Created new top-level common directory for code used by both the library and clients
    - New location for rocblas_ostream.hpp and cpp (now renamed to rocsolver_ostream.hpp and cpp)
    - New location for rocsolver_datatype2string.hpp (merged with rocblas_datatype2string)
        - New location for get_precision function (now renamed to rocblas2char_precision)
    - New location for common_ostream_helpers.hpp (now renamed to common_host_helpers.hpp)
        - Renamed old common_host_helpers.hpp and common_device_helpers.hpp to lib_host_helpers.hpp and lib_device_helpers.hpp for clarity
        - New location for get_time_us functions
        - New location for print_device_matrix functions
        - New location for rocblas_print_matrix functions (now renamed to print_host_matrix)

If these changes are adopted, note that new files created in the old directory structure will need to be moved into the directory structure, as git apparently moves files and not folders. I'd recommend visiting the branch in my fork to get a better view of the new directory structure.


cgmb:
I like it. I notice you also fixed the location of the sample programs.

Please do change back the copyright headers for `clients/include/testing_*.hpp`. Git doesn't actually track file moves, either. It determines whether a file was moved by looking at the lines added and removed in different files. The heuristics can generally deal with small modifications like that, but we might as well avoid unnecessary changes.

tfalders:
> Please do change back the copyright headers for `clients/include/testing_*.hpp`. Git doesn't actually track file moves, either. It determines whether a file was moved by looking at the lines added and removed in different files. The heuristics can generally deal with small modifications like that, but we might as well avoid unnecessary changes during the move.

It looks like all the testing_*.hpp files were edited in https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/198, but the copyright wasn't updated. I'll fix the copyright on the other files, but it might be correct to keep these ones.

cgmb (approved):
I'm in favour. This organization makes a lot of sense, and it should make it a little easier to share code between the library and the clients. :+1:

jzuniga-amd (commented):
It looks good! 

I like the `common` directory sharing the common code for the library and the clients. I think having two separate directories: `library/rocblascommon` and `clients/rocblascommon` is clearer. The purpose of the original rocblascommon was to have, in a single place, all the imported code to facilitate its maintenance and synchronization with the rocblas copy, and with the idea that, at some point, this common code could be maintained independently (not owned by rocblas), and added as a subcomponent (submodule, subtree, etc) by both libraries. In practice these ideas were never pursued and at this point our copy and rocblas copy have diverged, (and we have moved code to some other files, etc.) so that it has no more sense to keep this top-level rocblascommon. I like the splitting!... (if some kind of common code will be shared between libraries with this windows rocmath project, we may need to re-organize things again, though.)

Just a couple of comments:

1. I would extract the `cblas_interface.cpp` file from `clients/common` and put it directly in `clients/` (eliminating `clients/common` and only leaving `clients/rocblascommon`).

2. Also, I think is time to rename this file (and its header `cblas_interface.hpp`) to something else. It is more than a cblas interface... I would name it `reference_lapack_interface` or something like this.

3. Our ostream is completely independent from rocblas (and built entirely in rocsolver); should we rename the `rocblas_cout/cerr` macros as `rocsolver_cout/cerr` now that we are touching all files anyways?

tfalders:
> 1. I would extract the `cblas_interface.cpp` file from `clients/common` and put it directly in `clients/` (eliminating `clients/common` and only leaving `clients/rocblascommon`).

Personally, I'd prefer to keep this directory, as it is present in other libraries and keeps the cpp files nicely partitioned into folders.

> 2. Also, I think is time to rename this file (and its header `cblas_interface.hpp`) to something else. It is more than a cblas interface... I would name it `reference_lapack_interface` or something like this.

Sure! How about `lapack_host_reference`, to keep with some of our existing naming schemes?

> 3. Our ostream is completely independent from rocblas (and built entirely in rocsolver); should we rename the `rocblas_cout/cerr` macros as `rocsolver_cout/cerr` now that we are touching all files anyways?

I don't see why not. I'll add it in.

jzuniga-amd:
1. Sounds good. Not big deal. We may probably want to to the same with `library/src/buildinfo.cpp` and `library/src/rocsolver_logger.cpp`, put them into a `library/src/common` directory (or something like this)? Up to you. I am fine with any format...

2. Yeah, lapack_host_reference sounds good.

jzuniga-amd (approved):

tfalders:
> 1. Sounds good. Not big deal. We may probably want to to the same with `library/src/buildinfo.cpp` and `library/src/rocsolver_logger.cpp`, put them into a `library/src/common` directory (or something like this)? Up to you. I am fine with any format...

Sounds reasonable; I'll add it in!

I'll merge the PR after CI passes, unless you'd like me to wait. It may create conflicts in https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/212.

cgmb:
Could I make one last suggestion? These files should all be `*.hpp`:
```
./clients/include/lapack_host_reference.h
./library/src/rocblascommon/definitions.h
./library/src/rocblascommon/utility.h
```
The two rocblascommon files have already been renamed in rocblas, and we might as well fix lapack_host_reference while we're at it.

tfalders:
I seem to have accidentally committed a test file from the branch I'm currently working on: clients/include/testing_sygv_hegv.hpp. Sorry about that. It's not included anywhere right now, so I don't think it will cause issues, but I thought I'd give everyone a heads up.

PR #211, opened 2021-02-03
Added sygs2/hegs2 and  sygst/hegst

tfalders:
Addresses ticket SWDEV-247080


rocsolver/library/src/include/rocblas.hpp
```diff
@@ -790,8 +790,8 @@ rocblas_status rocblasCall_syr2_her2(rocblas_handle handle,
                                      rocblas_int incy,
                                      rocblas_stride stridey,
                                      V A,
-                                     rocblas_int lda,
                                      rocblas_int offsetA,
+                                     rocblas_int lda,
```
doctorcolinsmith, line 794: This argument reordering is potentially a breaking change in the API, which we should only do after a deprecation and major version update to ROCm.

Was this template in 4.1?  If so, the argument ordering needs to be preserved.

doctorcolinsmith (changes requested):
Recheck this for breaking API changes.

rocsolver/library/src/include/rocblas.hpp
tfalders: The functions in rocblas.hpp are only used internally. They are not exposed in the public API.

cgmb: To expand on the design and how we ensure API stability: All public API functions are declared in [library/include](https://github.com/ROCmSoftwarePlatform/rocSOLVER/tree/develop/rocsolver/library/include) and are always C functions.

We do have a private API that includes C++ functions, but it is used only by the rocsolver-test and rocsolver-benchmark clients. We make no promises about its stability (which is why it's private). It does not affect user code that sticks to the documented rocsolver functions.

I think that covers everything. As I understand it, libraries like hipBLAS just use the rocsolver public API, so there's no other interface to worry about. Did I get that right, @tfalders?

cgmb: (Note: This function is not a part of the private API, either. It's entirely internal.)

tfalders: By private API I assume you mean rocsolver.hpp? I'm not sure private API is how I'd describe it - it's really just a wrapper around rocSOLVER functions for the sake of convenience and is not intended to ever be used in user code.

There's a small handful of functions called by hipBLAS that are not in the public API (i.e. not in library/include), but none of them are affected by this PR. So as far as external libraries are concerned, we should only have to worry about the public API and the handful of cpp files that define these functions.

cgmb: > By private API I assume you mean rocsolver.hpp?

No... or at least, not in general. The functions declared in the `extern "C"` block at the top of that file are part of the private API, but I mean all undocumented functions that are marked with `ROCSOLVER_EXPORT`. That includes things like `rocsolver_ostream& operator<<(rocsolver_ostream& os, double x)`.

tfalders: Oh, my mistake; I got confused. I see what you mean by private API now. To my knowledge, the private API is used by `rocsolver-test` and `rocsolver-bench` (as you pointed out), the rocSOLVER library (through `rocsolver_ostream`), and the hipBLAS library (through the private functions I mentioned). We don't have any private API functions for hipSOLVER yet, but it's possible that we will have to add some later on too.

doctorcolinsmith: For clarification, any item should be considered public if it appears in the ABI as public symbol or is declared in a header file available in the packaged distribution.  It doesn't matter if the intent is to use it for the test clients or not.

cgmb: Good to know. That should be fairly straightforward for rocSOLVER to achieve, but we should talk more about it some time.

cgmb (approved):
:+1:

library/include/rocsolver-functions.h
```diff
@@ -10063,6 +10063,1038 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhetrd_strided_batched(rocblas_handle
                                                                  const rocblas_int batch_count);
 //! @}
 
+/*! @{
+    \brief SYGS2 reduces a symmetric-definite generalized eigenproblem to standard form.
```
jzuniga-amd, line 10067: Same here: it is important not to omit the word "real" in "real symmetric-definite" as there could also be a complex symmetric case which, again, is another problem... 
(also applies to all the similar functions below)

library/include/rocsolver-functions.h
```diff
@@ -10063,6 +10063,1038 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhetrd_strided_batched(rocblas_handle
...
+              Specifies whether the upper or lower part of the matrix A is stored, and
+              whether the factorization applied to B was upper or lower triangular.
+              If uplo indicates lower (or upper), then the upper (or lower) parts of A and
+              B are not used.
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The matrix dimensions.
+    @param[inout]
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the matrix A. On exit, the transformed matrix.
```
jzuniga-amd, line 10112: I would like to be more specific here and say: "On exit, the transformed matrix A associated with the equivalent standard eigenproblem" (same for all the other functions below)

library/include/rocsolver-functions.h
```diff
@@ -10063,6 +10063,1038 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhetrd_strided_batched(rocblas_handle
...
+    @param[in]
+    lda       rocblas_int. lda >= n.\n
+              Specifies the leading dimension of A.
+    @param[out]
+    B         pointer to type. Array on the GPU of dimension ldb*n.\n
+              The triangular factor of the matrix B, as returned by POTRF.
+    @param[in]
+    ldb       rocblas_int. lda >= n.\n
```
jzuniga-amd, line 10120: ldb >= n
(same for all the functions below)

library/include/rocsolver-functions.h
```diff
@@ -10063,6 +10063,1038 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhetrd_strided_batched(rocblas_handle
...
+    @param[out]
+    B         pointer to type. Array on the GPU (the size depends on the value of strideB).\n
+              The triangular factors of the matrices B_i, as returned by POTRF.
+    @param[in]
+    ldb       rocblas_int. lda >= n.\n
+              Specifies the leading dimension of B_i.
+    @param[in]
+    strideB   rocblas_stride.\n
+              Stride from the start of one matrix B_i and the next one B_(i+1).
+              There is no restriction for the value of strideB. Normal use case is strideB >= lda*n.
```
jzuniga-amd, line 10454: strideB >= ldb*n
(same for other similar functions below)

jzuniga-amd (changes requested):
Some initial comments and requested changes. I still need to go through the actual algorithms implementations and tests...

library/src/include/rocblas.hpp
```diff
@@ -1271,6 +1295,148 @@ rocblas_status rocblasCall_symv_hemv(rocblas_handle handle,
...
+                  "bc:", batch_count);
+
+    return rocblas_symm_template<false>(
+        handle, side, uplo, m, n, cast2constType<T>(alpha), cast2constType<T>(A), offsetA, lda,
+        strideA, cast2constType<T>(B), offsetB, ldb, strideB, cast2constType<T>(beta), C, offsetC,
+        ldc, strideC, batch_count);
+}
+
+// hemm
+template <typename T, typename U, typename V, std::enable_if_t<is_complex<T>, int> = 0>
```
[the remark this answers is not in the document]
tfalders: Most rocBLAS functions that have symmetric and hermitian variants have separate templates for sy and he - this one, for whatever reason, doesn't. I still separated them to be consistent with the other sy/he functions and just in case the templates are later separated in rocBLAS, but technically we could combine them.

library/src/lapack/roclapack_sygst_hegst.hpp
```diff
@@ -0,0 +1,313 @@
...
+    // everything must be executed with scalars on the host
+    rocblas_pointer_mode old_mode;
+    rocblas_get_pointer_mode(handle, &old_mode);
+    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+
+    S s_one = 1;
+    T t_one = 1;
+    T t_half = 0.5;
+    T t_minone = -1;
+    T t_minhalf = -0.5;
```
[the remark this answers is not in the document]
tfalders: I'm getting errors from cast2constType when I mark these as const. Also, I'm not sure if the rocBLAS templates accept pointers to constants.

library/src/lapack/roclapack_sygst_hegst.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+    // quick return
+    if(n == 0 || batch_count == 0)
+        return rocblas_status_success;
+
+    hipStream_t stream;
+    rocblas_get_stream(handle, &stream);
+
+    rocblas_int bs = xxGST_xxGS2_BLOCKSIZE;
+    rocblas_int nb = xxGST_xxGS2_SWITCHSIZE;
```
jzuniga-amd, line 108: bs is never used... 

Strictly speaking, there is no switch-size here as, in theory, the complexity of blocked and unblocked algorithm must be the same. So, the only parameter to tune will be the block-size.... 

We may have mixed names in other functions, and all this will change once we start optimizations and tuning. So, not big deal, but if you can eliminate switchsize and make nb=blocksize=64, that will be fine, it will save us some work later.

jzuniga-amd (changes requested):
Looks good to me. Just a few minor additions...

jzuniga-amd (approved):
LGTM!!
There is conflict in changelog file. Maybe because I removed the version 3.12.0. Almost certainly it will be the version we release at feature complete day but we don't know yet... we still have some days to break backward compatibility :)-

tfalders:
Awesome! Conflict should be addressed now. I'll merge after CI passes.

PR #215, opened 2021-02-19
Symmetric Eigensolvers

jzuniga-amd:
This PR also fixes some bugs in STERF and STEQR
Also, as rocSOLVER now includes functions in all LAPACK matrix computations streams, I am also reorganizing the functions (in the documentation, and CMakeLists files) accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,14 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Optimizations
 
 ### Changed
+- Sorting method in STERF as original quick-sort was failing for large sizes.
 
 ### Removed
 - Removed hcc compiler support
 
 ### Fixed
 - Fixed GELS overwriting B even when info != 0
+- Error when calling STEQR with n=1 from bacthed routines
 
 
 ## [rocSOLVER 3.11.0 for ROCm 4.1.0]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 - Reductions of generalized symmetric- and hermitian-definite eigenproblems:
     - SYGS2, SYGST (with batched and strided\_batched versions)
     - HEGS2, HEGST (with batched and strided\_batched versions)
+- Symmetric and Hermitian matrix Eigensolvers
+    - SYEV (with batched and strided\_batched versions)
+    - HEEV (with batched and strided\_batched versions)
 
 ### Optimizations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 
 ### Fixed
 - Fixed GELS overwriting B even when info != 0
-- Error when calling STEQR with n=1 from bacthed routines
+- Error when calling STEQR with n=1 from batched routines
 
 
 ## [rocSOLVER 3.11.0 for ROCm 4.1.0]
--- a/clients/common/lapack_host_reference.cpp
+++ b/clients/common/lapack_host_reference.cpp
@@ -5073,6 +5073,8 @@ void cblas_sygst_hegst<rocblas_double_complex>(rocblas_eform itype,
     zhegst_(&itypeI, &uploC, &n, A, &lda, B, &ldb, &info);
 }
 
+// sygst & hegst
+
 template <>
 void cblas_syev_heev<float, float>(rocblas_evect evect,
                                    rocblas_fill uplo,
--- a/clients/common/lapack_host_reference.cpp
+++ b/clients/common/lapack_host_reference.cpp
@@ -1447,6 +1447,29 @@ void zhegst_(int* itype,
              int* ldb,
              int* info);
 
+void ssyev_(char* jobz, char* uplo, int* n, float* A, int* lda, float* D, float* E, int* sizeW, int* info);
+void dsyev_(char* jobz, char* uplo, int* n, double* A, int* lda, double* D, double* E, int* sizeW, int* info);
+void cheev_(char* jobz,
+            char* uplo,
+            int* n,
+            rocblas_float_complex* A,
+            int* lda,
+            float* D,
+            rocblas_float_complex* work,
+            int* sizeW,
+            float* E,
+            int* info);
+void zheev_(char* jobz,
+            char* uplo,
+            int* n,
+            rocblas_double_complex* A,
+            int* lda,
+            double* D,
+            rocblas_double_complex* work,
+            int* sizeW,
+            double* E,
+            int* info);
+
 #ifdef __cplusplus
 }
 #endif
@@ -5048,4 +5071,70 @@ void cblas_sygst_hegst<rocblas_double_complex>(rocblas_eform itype,
     int itypeI = rocblas2char_eform(itype) - '0';
     char uploC = rocblas2char_fill(uplo);
     zhegst_(&itypeI, &uploC, &n, A, &lda, B, &ldb, &info);
+}
+
+template <>
+void cblas_syev_heev<float, float>(rocblas_evect evect,
+                                   rocblas_fill uplo,
+                                   rocblas_int n,
+                                   float* A,
+                                   rocblas_int lda,
+                                   float* D,
+                                   float* E,
+                                   rocblas_int sizeW,
+                                   rocblas_int* info)
+{
+    char jobz = rocblas2char_evect(evect);
+    char uploC = rocblas2char_fill(uplo);
+    ssyev_(&jobz, &uploC, &n, A, &lda, D, E, &sizeW, info);
+}
+
+template <>
+void cblas_syev_heev<double, double>(rocblas_evect evect,
+                                     rocblas_fill uplo,
+                                     rocblas_int n,
+                                     double* A,
+                                     rocblas_int lda,
+                                     double* D,
+                                     double* E,
+                                     rocblas_int sizeW,
+                                     rocblas_int* info)
+{
+    char jobz = rocblas2char_evect(evect);
+    char uploC = rocblas2char_fill(uplo);
+    dsyev_(&jobz, &uploC, &n, A, &lda, D, E, &sizeW, info);
+}
+
+template <>
+void cblas_syev_heev<rocblas_float_complex, float>(rocblas_evect evect,
+                                                   rocblas_fill uplo,
+                                                   rocblas_int n,
+                                                   rocblas_float_complex* A,
+                                                   rocblas_int lda,
+                                                   float* D,
+                                                   float* E,
+                                                   rocblas_int sizeW,
+                                                   rocblas_int* info)
+{
+    char jobz = rocblas2char_evect(evect);
+    char uploC = rocblas2char_fill(uplo);
+    rocblas_float_complex work[sizeW];
+    cheev_(&jobz, &uploC, &n, A, &lda, D, work, &sizeW, E, info);
+}
+
+template <>
+void cblas_syev_heev<rocblas_double_complex, double>(rocblas_evect evect,
+                                                     rocblas_fill uplo,
+                                                     rocblas_int n,
+                                                     rocblas_double_complex* A,
+                                                     rocblas_int lda,
+                                                     double* D,
+                                                     double* E,
+                                                     rocblas_int sizeW,
+                                                     rocblas_int* info)
+{
+    char jobz = rocblas2char_evect(evect);
+    char uploC = rocblas2char_fill(uplo);
+    rocblas_double_complex work[sizeW];
+    zheev_(&jobz, &uploC, &n, A, &lda, D, work, &sizeW, E, info);
 }
--- a/clients/gtest/syev_heev_gtest.cpp
+++ b/clients/gtest/syev_heev_gtest.cpp
@@ -1,0 +1,229 @@
+/* ************************************************************************
+ * Copyright (c) 2021 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "testing_syev_heev.hpp"
+
+using ::testing::Combine;
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<vector<int>, vector<char>> syev_heev_tuple;
+
+// each size_range vector is a {n, lda}
+
+// each op_range vector is a {evect, uplo}
+
+// case when n == 0, evect == N, and uplo = L will also execute the bad arguments test
+// (null handle, null pointers and invalid values)
+
+const vector<vector<char>> op_range = {{'N', 'L'}, {'N', 'U'}, {'V', 'L'}, {'V', 'U'}};
+
+// for checkin_lapack tests
+const vector<vector<int>> size_range = {
+    // quick return
+    {0, 1},
+    // invalid
+    {-1, 1},
+    {10, 5},
+    // normal (valid) samples
+    {1, 1},
+    {12, 12},
+    {20, 30},
+    {35, 35},
+    {50, 60}};
+
+// for daily_lapack tests
+const vector<vector<int>> large_size_range = {{192, 192}, {256, 270}, {300, 300}, {512, 550}};
+
+Arguments syev_heev_setup_arguments(syev_heev_tuple tup)
+{
+    vector<int> size = std::get<0>(tup);
+    vector<char> op = std::get<1>(tup);
+
+    Arguments arg;
+
+    arg.N = size[0];
+    arg.lda = size[1];
+
+    arg.evect = op[0];
+    arg.uplo_option = op[1];
+
+    // only testing standard use case for strides
+    // strides are ignored in normal and batched tests
+    arg.bsa = arg.lda * arg.N; // strideA
+    arg.bsb = arg.N; // strideD
+    arg.bsc = arg.N; // strideE
+
+    arg.timing = 0;
+
+    return arg;
+}
+
+class SYEV : public ::TestWithParam<syev_heev_tuple>
+{
+protected:
+    SYEV() {}
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+};
+
+class HEEV : public ::TestWithParam<syev_heev_tuple>
+{
+protected:
+    HEEV() {}
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+};
+
+// non-batch tests
+
+TEST_P(SYEV, __float)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<false, false, float>();
+
+    arg.batch_count = 1;
+    testing_syev_heev<false, false, float>(arg);
+}
+
+TEST_P(SYEV, __double)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<false, false, double>();
+
+    arg.batch_count = 1;
+    testing_syev_heev<false, false, double>(arg);
+}
+
+TEST_P(HEEV, __float_complex)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<false, false, rocblas_float_complex>();
+
+    arg.batch_count = 1;
+    testing_syev_heev<false, false, rocblas_float_complex>(arg);
+}
+
+TEST_P(HEEV, __double_complex)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<false, false, rocblas_double_complex>();
+
+    arg.batch_count = 1;
+    testing_syev_heev<false, false, rocblas_double_complex>(arg);
+}
+
+// batched tests
+
+TEST_P(SYEV, batched__float)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<true, true, float>();
+
+    arg.batch_count = 3;
+    testing_syev_heev<true, true, float>(arg);
+}
+
+TEST_P(SYEV, batched__double)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<true, true, double>();
+
+    arg.batch_count = 3;
+    testing_syev_heev<true, true, double>(arg);
+}
+
+TEST_P(HEEV, batched__float_complex)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
+
+    if(arg.N == 0 && arg.evect == 'N' && arg.uplo_option == 'L')
+        testing_syev_heev_bad_arg<true, true, rocblas_float_complex>();
+
+    arg.batch_count = 3;
+    testing_syev_heev<true, true, rocblas_float_complex>(arg);
+}
+
+TEST_P(HEEV, batched__double_complex)
+{
+    Arguments arg = syev_heev_setup_arguments(GetParam());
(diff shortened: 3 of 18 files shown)


clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+                            U dinfo,
+                            const rocblas_int bc)
+{
+    // handle
+    EXPECT_ROCBLAS_STATUS(rocsolver_syev_heev(STRIDED, nullptr, evect, uplo, n, dA, lda, stA, dD,
+                                              stD, dE, stE, dinfo, bc),
+                          rocblas_status_invalid_handle);
+
+    // values
+    EXPECT_ROCBLAS_STATUS(rocsolver_syev_heev(STRIDED, handle, rocblas_evect(-1), uplo, n, dA, lda,
```
tfalders, line 35: Do we also need a case for rocblas_evect_tridiagonal?

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    if(GPU)
+    {
+        // now copy to the GPU
+        CHECK_HIP_ERROR(dA.transfer_from(hA));
+    }
+}
+
+template <bool STRIDED, typename T, typename Wd, typename Td, typename Id, typename Wh, typename Th, typename Ih>
```
tfalders, line 172: (2) The type names are used differently from normal, which is confusing. Please use Td/Th in place of Wd/Wh and Sd/Sh in place of Td/Th.

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+                }
+            }
+
+            // make copy of original data to test vectors if required
+            if(test && evect == rocblas_evect_original)
+            {
+                for(rocblas_int i = 0; i < n; i++)
+                {
+                    for(rocblas_int j = 0; j < n; j++)
+                        A[b * lda * n + i + j * lda] = hA[b][i + j * lda];
```
tfalders, line 159: Using Th as the type for A would make this less confusing.

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+    rocblas_int lwork = 5 * n;
+    std::vector<T> A(lda * n * bc);
+
+    // input data initialization
+    syev_heev_initData<true, true, T>(handle, evect, n, dA, lda, bc, hA, A);
+
+    // execute computations
+    // CPU lapack
+    for(rocblas_int b = 0; b < bc; ++b)
+        cblas_syev_heev<T>(evect, uplo, n, hA[b], lda, hD[b], hE[b], lwork, hinfo[b]);
```
tfalders, line 204: Any reason why the CPU computation is before the GPU computation? Usually we put it after.

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+                // error is ||hA - hARes|| / ||hA||
+                // using frobenius norm
+                err = norm_error('F', n, n, lda, hA[b], hAres[b]);
+                *max_err = err > *max_err ? err : *max_err;
+            }
+        }
+    }
+}
+
+template <bool STRIDED, typename T, typename Wd, typename Td, typename Id, typename Wh, typename Th, typename Ih>
```
tfalders, line 265: See (2)

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+    size_t size_Dres = 0;
+    size_t size_Ares = 0;
+    size_t size_A = size_t(lda) * n;
+    size_t size_D = n;
+    size_t size_E = size_D;
+    if(argus.unit_check || argus.norm_check)
+    {
+        size_Ares = size_A;
+        size_Dres = size_D;
+    }
```
tfalders, line 380: This is me being nitpicky, but our usual style for declaring result array sizes uses fewer lines of code.

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+                                                  (rocblas_int*)nullptr, bc));
+
+        size_t size;
+        CHECK_ROCBLAS_ERROR(rocblas_stop_device_memory_size_query(handle, &size));
+        CHECK_ROCBLAS_ERROR(rocblas_set_device_memory_size(handle, size));
+    }
+
+    // memory allocations (all cases)
+    // host
+    host_strided_batch_vector<S> hE(5 * n, 1, 5 * n, bc);
```
tfalders, line 425: I'm guessing that vanilla LAPACK requires more space than size_E provides?

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
...
+    {
+        if(!argus.perf)
+        {
+            rocsolver_cout << "\n============================================\n";
+            rocsolver_cout << "Arguments:\n";
+            rocsolver_cout << "============================================\n";
+            if(BATCHED)
+            {
+                rocsolver_bench_output("evect", "uplo", "n", "lda", "strideD", "strideE", "batch_c");
+                rocsolver_bench_output(evect, uplo, n, lda, stD, stE, bc);
```
tfalders, line 534: Use evectC and uploC. Once boost::format is removed from rocsolver_bench_output, these will get formatted through a stringstream, which I don't think will accept our enums. Same for other output below.

clients/include/testing_syev_heev.hpp
```diff
@@ -0,0 +1,570 @@
+/* ************************************************************************
+ * Copyright (c) 2020-2021 Advanced Micro Devices, Inc.
```
tfalders, line 2: (3) Copyright should just be 2021, correct?

library/src/lapack/roclapack_syev_heev.cpp
```diff
@@ -0,0 +1,144 @@
+/* ************************************************************************
+ * Copyright (c) 2019-2021 Advanced Micro Devices, Inc.
```
tfalders, line 2: See (3)

library/src/lapack/roclapack_syev_heev.hpp
```diff
@@ -0,0 +1,231 @@
+/************************************************************************
+ * Derived from the BSD3-licensed
+ * LAPACK routine (version 3.7.0) --
+ *     Univ. of Tennessee, Univ. of California Berkeley,
+ *     Univ. of Colorado Denver and NAG Ltd..
+ *     December 2016
+ * Copyright (c) 2020-2021 Advanced Micro Devices, Inc.
```
tfalders, line 7: See (3)

library/src/lapack/roclapack_syev_heev_strided_batched.cpp
```diff
@@ -0,0 +1,162 @@
+/* ************************************************************************
+ * Copyright (c) 2019-2021 Advanced Micro Devices, Inc.
```
tfalders, line 2: See (3)

library/src/lapack/roclapack_syev_heev_batched.cpp
```diff
@@ -0,0 +1,160 @@
+/* ************************************************************************
+ * Copyright (c) 2019-2021 Advanced Micro Devices, Inc.
```
tfalders, line 2: See (3)

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 - Reductions of generalized symmetric- and hermitian-definite eigenproblems:
     - SYGS2, SYGST (with batched and strided\_batched versions)
     - HEGS2, HEGST (with batched and strided\_batched versions)
+- Symmetric and Hermitian matrix Eigensolvers
+    - SYEV (with batched and strided\_batched versions)
+    - HEEV (with batched and strided\_batched versions)
 
 ### Optimizations
```
tfalders, line 16: Should the sterf and steqr bug fixes be mentioned in the changelog as well?

library/src/lapack/roclapack_syev_heev.hpp
```diff
@@ -0,0 +1,231 @@
...
+#pragma once
+
+#include "../auxiliary/rocauxiliary_orgtr_ungtr.hpp"
+#include "../auxiliary/rocauxiliary_steqr.hpp"
+#include "../auxiliary/rocauxiliary_sterf.hpp"
+#include "rocblas.hpp"
+#include "roclapack_sytrd_hetrd.hpp"
+#include "rocsolver.h"
+
+/** Set results for the sacalr case (n=1) **/
```
tfalders, line 19: Spelling error on scalar

library/src/lapack/roclapack_syev_heev.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // 3. invalid pointers
+    if((n && !A) || (n && !E) || (n && !D) || (batch_count && !info))
+        return rocblas_status_invalid_pointer;
+
+    return rocblas_status_continue;
+}
+
+/** Helper to calculate workspace sizes **/
+template <bool BATCHED, typename T, typename S>
```
tfalders, line 99: See (1)

library/src/lapack/roclapack_syev_heev.hpp
```diff
@@ -0,0 +1,231 @@
...
+                                            S* D,
+                                            S* E,
+                                            rocblas_int* info,
+                                            const rocblas_int batch_count = 1)
+{
+    // order is important for unit tests:
+
+    // 1. invalid/non-supported values
+    if((evect != rocblas_evect_original && evect != rocblas_evect_none
+        && evect != rocblas_evect_tridiagonal)
```
tfalders, line 79: I don't believe rocblas_evect_tridiagonal should be a supported value here, especially since many of the if statements only cover rocblas_evect_original.

tfalders (commented):
Looks good, though I have some nitpicky comments to ensure library consistency.

clients/common/lapack_host_reference.cpp
```diff
@@ -5049,3 +5072,69 @@ void cblas_sygst_hegst<rocblas_double_complex>(rocblas_eform itype,
     char uploC = rocblas2char_fill(uplo);
     zhegst_(&itypeI, &uploC, &n, A, &lda, B, &ldb, &info);
 }
+
```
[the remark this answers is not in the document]
jzuniga-amd: added

library/src/lapack/roclapack_syev_heev.cpp
```diff
@@ -0,0 +1,144 @@
...
+ * ************************************************************************ */
+
+#include "roclapack_syev_heev.hpp"
+
+template <typename T, typename S, typename W>
```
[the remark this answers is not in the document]
jzuniga-amd: Mmm, I have not realized that all functions with mixed types have this template...
If there is going to be a convention, I would have preferred it to be <T,S,U> because: 
1. T is always the base precision (s,d,c or z) so is the one that you should always pass when calling a function as it cannot be always inferred by the compiler especially if the matrix is an array of pointers. 
2. The parameters of type S (float when T = rocblas_float_complex, or double when T = rocblas_double_complex) are normally just simple arrays so you don't need to pass S in the template. 

So, with the 
`<typename S, typename T, typename U> function(U arrayA, S* arrayD)` 
signature, the call should be  `function<S,T>(A,D)` even if S could be inferred from D. 
On the other hand, with the 
`<typename T, typename S, typename U> function(U arrayA, S* arrayD)` 
signature, the call can be simply `function<T>(A,D)`...

What do you think? 
(I have no problem changing this tho)

library/include/rocsolver-functions.h
```diff
@@ -11109,6 +11109,487 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhegst_strided_batched(rocblas_handle
...
+    @param[in]
+    lda         rocblas_int. lda >= n.\n
+                Specifies the leading dimension of matrices A_j.
+    @param[out]
+    D           pointer to type. Array on the GPU (the side depends on the value of strideD).\n
+                The eigenvalues of A_j in increasing order.
+    @param[in]
+    strideD     rocblas_stride.\n
+                Stride from the start of one vector D_j to the next one D_(j+1).
+                There is no restriction for the value of strideD. Normal use case is strideD = n.
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

library/include/rocsolver-functions.h
```diff
@@ -11109,6 +11109,487 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhegst_strided_batched(rocblas_handle
...
+                On exit, if info > 0, it contains the unconverged off-diagonal elements of T_j
+                (or properly speaking, a tridiagonal matrix equivalent to T_j). The diagonal elements
+                of this matrix are in D_j; those that converged correspond to a subset of the
+                eigenvalues of A_j (not necessarily ordered).
+    @param[in]
+    strideE     rocblas_stride.\n
+                Stride from the start of one vector E_j to the next one E_(j+1).
+                There is no restriction for the value of strideE. Normal use case is strideE = n.
+    @param[out]
+    info        pointer to rocblas_int. Array of batch_count integres on the GPU.\n
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

clients/include/testing_syev_heev.hpp
jzuniga-amd: No, SYEV/HEEV's problem starts with the original matrix A. It makes no sense to compute the eigenvectors of the intermediate tridiagonal form. What I am doing is:
if(evect == rocblas_evect_original) then "compute eigenvectors". 
else "do not compute eigenvectors"

clients/include/testing_syev_heev.hpp
jzuniga-amd: changed

clients/include/testing_syev_heev.hpp
jzuniga-amd: And how would you know if Th is a `host_batch_vector` or a `host_strided_batch_vector` ?
How would you initialize A? like `Th A(lda*n, 1, bc)` or like `Th A(lda*n, 1, lda*n, bc)` ?

I can simply say `host_strided_batch_vector<T> A(lda*n, lda*n, bc)`... but I think is simpler the way it is as a simple standard vector. It is only a temporary array that only lives in _getError()

clients/include/testing_syev_heev.hpp
jzuniga-amd: No, no reason... changed.

clients/include/testing_syev_heev.hpp
jzuniga-amd: changed.

clients/include/testing_syev_heev.hpp
jzuniga-amd: Yes, it does...

clients/include/testing_syev_heev.hpp
jzuniga-amd: Right!!!
Corrected.

clients/include/testing_syev_heev.hpp
jzuniga-amd: Corrected.

CHANGELOG.md
jzuniga-amd: Yeah, why not... 
Added.

library/src/lapack/roclapack_syev_heev.hpp
jzuniga-amd: corrected

library/src/lapack/roclapack_syev_heev.hpp
jzuniga-amd: I don't see the problem. If a user calls the API with a value of evect different from the enum values, we should give an error. 
If the value of evect is none or tridiagonal, we don't compute eigenvector.
If the value of evect is original, then we compute the eigenvectors.

clients/gtest/syev_heev_gtest.cpp
```diff
@@ -0,0 +1,229 @@
...
+typedef std::tuple<vector<int>, vector<char>> syev_heev_tuple;
+
+// each size_range vector is a {n, lda}
+
+// each op_range vector is a {evect, uplo}
+
+// case when n == 0, evect == N, and uplo = L will also execute the bad arguments test
+// (null handle, null pointers and invalid values)
+
+const vector<vector<char>> op_range = {{'N', 'L'}, {'N', 'U'}, {'V', 'L'}, {'V', 'U'}};
```
cgmb, line 23: You can get nicer output from test failures by using `rocsolver_op_char` in place of `char`. The default GTest printer for char assumes the character may be invisible when printed, so it also prints the numeric value. I specialized `rocsolver_op_char` to just print the character itself, as we know it will be a letter.

There's implicit conversions to and from `rocsolver_op_char` and `char`, so you should be able to use it as a drop-in replacement.

library/src/auxiliary/rocauxiliary_sterf.hpp
```diff
@@ -255,7 +255,31 @@ __global__ void sterf_kernel(const rocblas_int n,
             info[bid]++;
 
     // Sort eigenvalues
-    lasrt_increasing(n, D, stack + bid * (2 * 32));
+    /** (TODO: the quick-sort method implemented in lasrt_increasing fails for some cases.
+        Substituting it here with a simple sorting algorithm. If more performance is required in
+        the future, lasrt_increasing should be debugged or another quick-sort method
+        could be implemented) **/
```
cgmb, line 261: Is that selection sort? I'm surprised by the original quick sort, though. I thought radix sort was the popular one on GPUs. I suppose there's enough separate sorts to be done that you get sufficient parallelism with a standard sort?

library/src/lapack/roclapack_syev_heev.cpp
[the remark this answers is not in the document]
cgmb: I agree with Juan for the reasons he presented, but I'd be happy with just taking the path of least-resistance at the moment and giving this more thought at a later date.

library/src/lapack/roclapack_syev_heev.hpp
```diff
@@ -0,0 +1,231 @@
...
+                            const rocblas_stride strideD,
+                            rocblas_int bc)
+{
+    int b = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+
+    if(b < bc)
+    {
+        T* A = load_ptr_batch<T>(AA, b, 0, strideA);
+        S* D = DD + b * strideD;
+        D[0] = A[0].real();
```
cgmb, line 55: Is `D[0] = std::real(A[0]);` equivalent? If so, could the two `scalar_case` implementations be combined into one?

library/src/lapack/roclapack_syev_heev.cpp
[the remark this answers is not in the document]
tfalders: That makes sense. We can leave this as it is with the understanding that the rest of the library will be updated to the new convention at a later date.

clients/include/testing_syev_heev.hpp
tfalders: My initial thought was to allocate it along with the other arrays (possibly empty like we do for the result arrays), but making it always a strided_batch_vector would also work. It wouldn't be of type Th in that case, but the indexing would still be cleaned up a bit. It's not a critical request though, so I won't press if you prefer it the way it is.

library/src/lapack/roclapack_syev_heev.hpp
```diff
@@ -0,0 +1,231 @@
...
+        rocsolver_steqr_getMemorySize<S, T>(evect, n, batch_count, &w3);
+    }
+    else
+    {
+        // extra requirements for computing only the eigenvalues (sterf)
+        rocsolver_sterf_getMemorySize<T>(n, batch_count, &w2);
+    }
+
+    // get max values
+    *size_work_stack = std::max(w1, std::max(w2, w3));
```
cgmb, line 148: `std::max` has an overload for initializer lists, so you can extend it to an arbitrary number of elements just by adding `{}` brackets. e.g.

    *size_work_stack = std::max({w1, w2, w3});

No need to change anything if you like it as-is. I just thought I'd mention it in case you were unaware.

library/include/rocsolver-functions.h
```diff
@@ -11109,6 +11109,487 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhegst_strided_batched(rocblas_handle
...
+    \details
+    The eigenvalues are returned in ascending order. The eigenvectors are computed depending
+    on the value of evect. The computed eigenvectors are orthonormal.
+
+    @param[in]
+    handle      rocblas_handle.
+    @param[in]
+    evect       #rocblas_evect.\n
+                Specifies whether the eigenvectors are to be computed.
+                If evect =  rocblas_evect_original the eigenvectors are computed.
```
cgmb, line 11125: `If evect =  rocblas_evect_original the eigenvectors are computed.` should have a comma after the if clause, and there's an extra space after the `=`.

My suggestion is, `If evect = rocblas_evect_original, the eigenvectors are computed.`

cgmb: I was going to also suggest `If evect is rocblas_evect_original`, because the docs are rendered as

    Specifies whether the eigenvectors are to be computed. If evect =
    rocblas_evect_original the eigenvectors are computed.

and English looks better wrapped over multiple lines than math expressions do... However, a search through our docs shows that we have always been using `=` for this sort of thing.

IDK. I would probably have always chosen `is` over `=`, except when contrasting equality with other relational operators. Another option would be to replace U+0020 (space) with U+00A0 (non-breaking space) in these mini-equations. Or just force a newline before the if. Or just leave it as-is, because it's not a big deal.

library/src/lapack/roclapack_syev_heev.hpp
tfalders: Part of my concern is that a user could pass the tridiagonal argument and expect it to do something different from what it actually does, especially since the documentation doesn't explicitly define this behaviour. Throwing an error makes more sense to me from a user perspective and would be consistent with LAPACK.

tfalders: This would be like if sytrd treated rocblas_fill_full as rocblas_fill_upper instead of throwing an error. From a user perspective it's unclear. Why upper/no eigenvectors instead of lower/all eigenvectors?

library/src/lapack/roclapack_syev_heev.cpp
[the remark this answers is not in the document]
jzuniga-amd: Ok. I Will leave it as is and open a ticket to update the other few functions with <S,T,U> signature.

library/src/lapack/roclapack_syev_heev.hpp
jzuniga-amd: oh, I see what you meant...
changed.

CHANGELOG.md
```diff
@@ -9,16 +9,21 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
...
 
 ### Changed
+- Sorting method in STERF as original quick-sort was failing for large sizes.
 
 ### Removed
 - Removed hcc compiler support
 
 ### Fixed
 - Fixed GELS overwriting B even when info != 0
+- Error when calling STEQR with n=1 from bacthed routines
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

clients/gtest/syev_heev_gtest.cpp
jzuniga-amd: cool...
Changed.

library/src/auxiliary/rocauxiliary_sterf.hpp
jzuniga-amd: I don't think Troy implemented quick-sort because it was the better method to use with GPUs. It is just the method used in LAPACK and it is faster than the simple comparison-based method. As I said, if in the future we see that the sorting part is the bottleneck here, we can implement radix, fix quick-sort or whatever is best. Right now the sorting is not even parallelized as there is only 1 thread working on each matrix in the batch.

library/src/lapack/roclapack_syev_heev.hpp
jzuniga-amd: In theory yes. But this never works for me. I think rocBLAS injected its overloads for rocblas_complex types into the std namespace (see rocblas_complex_types.h).
I have never been able to make it work so I make the two template implementations. If you know how to solve it let me know. Leaving this as it is for now...

library/src/lapack/roclapack_syev_heev.hpp
jzuniga-amd: oh cool, changed...

library/include/rocsolver-functions.h
jzuniga-amd: changed

tfalders (approved):
Awesome, LGTM!

library/src/lapack/roclapack_syev_heev.hpp
cgmb: Ah. I think I'm finally starting to understand what's going on there.

cgmb (approved):
LGTM!

PR #217, opened 2021-02-20
Trim dependency list

cgmb:
Many of our dependencies have been copied from rocBLAS, but they are not needed for building rocSOLVER. You can install a rocBLAS binary package and build rocSOLVER without these tools. If you are building rocBLAS from source, you will need them, but that is handled by the rocBLAS install script.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -179,7 +179,7 @@ install_packages( )
   fi
 
   # dependencies needed to build the rocsolver library
-  local library_dependencies_ubuntu=( "make" "cmake-curses-gui")
+  local library_dependencies_ubuntu=( "make" "cmake")
   local library_dependencies_centos_7=( "epel-release" "make" "cmake3" "rpm-build")
   local library_dependencies_centos_8=( "epel-release" "make" "cmake3" "rpm-build")
   local library_dependencies_fedora=( "make" "cmake" "rpm-build")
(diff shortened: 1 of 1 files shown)


cgmb:
This is a low priority, as it has no effect on the actual rocSOLVER release. While the change is done and ready for review, I think there's more important things to look at next week. I'll reopen this on March 1st.

tfalders (approved):
LGTM! Though, is rocm-dev really not needed by rocSOLVER?

cgmb:
A ROCm toolchain is required, but there's a few reasons not to install rocm-dev in our install.sh script:
1. rocm-dev changes over time. If you check out rocSOLVER 3.8, and run `./install.sh -d`, it will install ROCm 4.0, because that's the current official release. But there's no guarantee that you can compile rocSOLVER 3.8 using the ROCm 4.0 toolchain.
2. If you're compiling rocsolver from source, there's a decent chance it's because you want to build it using a different ROCm stack (like one that you've also compiled from source).
3. The documentation in install.sh states: "this does not install rocBLAS nor ROCm software stack" and I presume the latter is referring to rocm-dev. Removing it is consistent with the existing documentation.

cgmb:
I should note that OpenMP is still referenced in our CMake files. I added it to this PR while I was porting [Torre's rocBLAS clients changes](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/505), but I suppose that means this change should be ordered after that one.

jzuniga-amd (approved):
LGTM

PR #218, opened 2021-02-22
Add rocblas dependency to CMake and OS packages - build system - 4/5

cgmb:
After installing rocm-dkms and setting PATH as specified in [the ROCm install docs](https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#ubuntu), it should be possible to build [this sample app](https://gist.github.com/cgmb/4ab8e70355cf61021b2c8d04f39a4b03) using the minimal build script and `CMakeLists.txt`.

The rocblas dependency should be listed in the deb or rpm package, and the CMake usage requirements we generate (`rocsolver-config.cmake`) also should list the rocblas dependency.


tfalders (approved):
LGTM! (Again, with the caveat that I'm not an expert on the build system)

library/CMakeLists.txt
```diff
@@ -58,17 +58,10 @@ set( INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_INCLUDEDIR} )
...
-  set( RPM_REQUIREMENTS "hip-rocclr >= 3.5.0" )
-  set( CPACK_DEBIAN_PACKAGE_DEPENDS "hip-rocclr (>= 3.5.0)" )
-elseif( HIP_RUNTIME STREQUAL "cuda")
-  set( RPM_REQUIREMENTS "hip-nvcc >= 3.5.0" )
-  set( CPACK_DEBIAN_PACKAGE_DEPENDS "hip-nvcc (>= 3.5.0)" )
-else()
-  set( RPM_REQUIREMENTS "rocm-dev >= 2.5.27" )
-  set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocm-dev (>= 2.5.27)" )
-endif()
+if( NOT USE_CUDA )
```
saadrahim, line 61: What about the CUDA path? Was it never supported?

clients/gtest/CMakeLists.txt
```diff
@@ -104,13 +104,12 @@ add_executable( rocsolver-test ${roclapack_test_source} ${rocsolver_test_source}
...
 add_armor_flags( rocsolver-test "${ARMOR_LEVEL}" )
 
-target_compile_definitions( rocsolver-test PRIVATE BUILD_WITH_TENSILE=0 GOOGLE_TEST )
+target_compile_definitions( rocsolver-test PRIVATE GOOGLE_TEST )
 
 # Internal header includes
 target_include_directories( rocsolver-test
   PRIVATE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../library/include>
```
TorreZuk, line 113: If you build without install don't you need this to find shared helpers for ostream etc.?

clients/samples/CMakeLists.txt
```diff
@@ -40,45 +40,15 @@ set( fortran_samples
...
-  endif( )
-
   if( CMAKE_COMPILER_IS_GNUCXX OR CXX_VERSION_STRING MATCHES "clang" )
     # GCC or hip-clang needs specific flags to turn on f16c intrinsics
     target_compile_options( ${exe} PRIVATE -mf16c )
   endif( )
-
-  if( NOT CUDA_FOUND )
-    target_link_libraries( ${exe} PRIVATE hip::host )
-  endif()
```
TorreZuk, line 75: So none of your samples have hip code to be compiled I guess, they just call solver "C" functions, okay.

library/CMakeLists.txt
```diff
@@ -58,17 +58,10 @@ set( INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_INCLUDEDIR} )
...
+  set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocblas (>= 2.37.0)" )
+  set( CPACK_RPM_PACKAGE_REQUIRES "rocblas >= 2.37.0" )
+endif( )
```
TorreZuk, line 64: The next code which github won't let me select has (OS_ID_sles) LLVM, check rocblas as I think this was removed... not sure if you need it.

library/src/CMakeLists.txt
```diff
@@ -133,7 +133,14 @@ add_library( rocsolver
 
 add_library( roc::rocsolver ALIAS rocsolver )
 
-target_link_libraries( rocsolver PRIVATE hip::device --rtlib=compiler-rt --unwindlib=libgcc )
+target_link_libraries( rocsolver
+  PUBLIC
+    roc::rocblas
```
TorreZuk, line 138: How did you manage to link before?   So if you remove this does it link?

library/CMakeLists.txt
```diff
@@ -58,17 +58,10 @@ set( INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_INCLUDEDIR} )
...
-elseif( HIP_RUNTIME STREQUAL "cuda")
-  set( RPM_REQUIREMENTS "hip-nvcc >= 3.5.0" )
-  set( CPACK_DEBIAN_PACKAGE_DEPENDS "hip-nvcc (>= 3.5.0)" )
-else()
-  set( RPM_REQUIREMENTS "rocm-dev >= 2.5.27" )
-  set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocm-dev (>= 2.5.27)" )
-endif()
+if( NOT USE_CUDA )
+  set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocblas (>= 2.37.0)" )
+  set( CPACK_RPM_PACKAGE_REQUIRES "rocblas >= 2.37.0" )
```
TorreZuk, line 63: Due to internal interface usage this could be rocblas (== 2.37.0), not sure of the == syntax

TorreZuk (approved):
You can discuss as you might also consider find_package (rocblas x.y.z EXACT ....)  until the internal interface solidifies.

jzuniga-amd (commented):
Looks good to me. 

My only concern is with the addition of rocblas package requirement. If we are going to add this, then, as Torre mentioned, the versioning should be an exact match. The >= won't work... if the C++ template interface changes in rocblas 2.38, for example, then this version of rocsolver won't build even if the rocblas package requirement (>=2.37) is fulfilled. 

Also, will the condition pass even if the user has a manually installed rocBLAS and pass its location with --rocblas_dir ?

library/CMakeLists.txt
cgmb: You're right. Will do.

library/src/CMakeLists.txt
cgmb: We didn't link before. The references to rocblas in rocsolver don't need to be resolved at library build time. We just required all users to manually link rocblas after rocsolver (including our own clients).

Our CMake used to set up the rocBLAS include directories after `find_package`, but [I removed that](https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/199/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL196) in #199 thinking it was redundant. In my defense, if we were doing things the right way, it would be.

At the moment, `./install.sh --rocblas_dir=<dir>` is broken. You can only properly build against an installed copy of rocBLAS. Publicly linking `roc::rocblas` will fix that by adding the rocblas include paths to the rocsolver usage requirements.

clients/samples/CMakeLists.txt
cgmb: Yup.

clients/gtest/CMakeLists.txt
cgmb: We've moved our shared code into a `common` directory at the rocSOLVER project root, so the only code in `library/include` is now library code. It shouldn't be necessary for the clients to reach into the library directories.

(Though, the directory gets added to the rocsolver-test include directories anyways, as it's a part of the rocsolver library public build interface... it's more of a philosophical separation than an actual practical difference.)

library/CMakeLists.txt
cgmb: To my knowledge, all of the CUDA paths in rocSOLVER are residual from when the library structure was copied from rocBLAS.

That particular `if` was copied from hipBLAS. I suppose they do things that way because they depend on unpackaged CUDA libraries when USE_CUDA is true. That wouldn't apply to the hypothetical rocSOLVER that could be built against CUDA. I suppose we would depend on... rocblas-alt?

Or maybe I should just leave the `if` out entirely, and just not worry about how we package for CUDA until we actually build and test with it?

cgmb:
> Also, will the condition pass even if the user has a manually installed rocBLAS and pass its location with --rocblas_dir ?

No. You will not be able to install a rocSOLVER package unless you also install the corresponding rocBLAS package. You will be able to use rocSOLVER from a build tree with rocBLAS from a build tree, though.

Though, this is not a unique problem. I already faced that issue when I was building HIP from source. Our existing packages depend on rocclr, and if I built that from source but didn't build a package, I found I'd have to create a dummy rocclr package to satisfy the dependency check and allow rocsolver to install.

library/CMakeLists.txt
saadrahim: I agree with leaving the if out, a little confusing as we don't support CUDA with rocSOLVER yet.

jzuniga-amd (approved):
LGTM. After discussion on --rocblas_dir option, it is all fine for me.

library/CMakeLists.txt
cgmb: [It appears to still be there in rocBLAS.](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/bb2519979da0b761c610953d14c88b087572ba67/library/CMakeLists.txt#L143-L146) Is that a pending change?

library/CMakeLists.txt
cgmb: We can't use an exact comparison, because our versioning scheme uses the debian_revision / tweak number for the git hash. I'm debating between `2.37.1 > X > 2.37.0` and `2.38 > X > 2.37`. Which would you prefer?

TorreZuk: Whatever works I would think is fine

cgmb: Ok. I prefer the latter. It has a nice clean hierarchy: major version bump can break external interface, minor version bump can break internal interface, patch version bump cannot break any interface.

cgmb:
I had been doing some manual testing yesterday and was going to merge this first thing in morning, but seeing GESVD failing in the extended tests spooked me. I can't imagine how this change would cause those problems, but I want to see the tests pass before I merge.

cgmb:
After some investigation, the failures are unrelated, so I'm going to go ahead and merge.

PR #220, opened 2021-02-23
fix get_blksize function for GETRF

jzuniga-amd:


library/src/lapack/roclapack_getrf.hpp
```diff
@@ -14,6 +14,77 @@
 #include "roclapack_getf2.hpp"
 #include "rocsolver.h"
 
+template <bool ISBATCHED, bool PIVOT>
+inline rocblas_int get_blksize(rocblas_int dim);
```
cgmb, line 18: `inline` is unnecessary on the template declaration. Template functions are implicitly inline.

The template function specializations below require inline, since once you fill in all the blanks on a template function, it's no longer actually a template... it's just a function.

cgmb (approved):
My first instinct is all fixes should include a test that verifies the fix... but this problem doesn't change the correctness of the results. 

Nice job catching that. We could probably trim the line count by expressing it differently, but I'm not too concerned. If the tests pass, LGTM.

library/src/lapack/roclapack_getrf.hpp
```diff
@@ -14,6 +14,77 @@
...
+
+template <>
+inline rocblas_int get_blksize<false, false>(rocblas_int dim)
+{
+    rocblas_int max;
+    rocblas_int i;
+    std::vector<rocblas_int> size{GETRF_NPVT_BLKSIZES_NORMAL};
+    std::vector<rocblas_int> intervals{GETRF_NPVT_INTERVALS_NORMAL};
```
cgmb, line 26: If you do decide to make any changes, I'd recommend skipping the vector and just using a plain array.
```
rocblas_int size[] = { GETRF_NPVT_BLKSIZES_NORMAL };
rocblas_int intervals[] = { GETRF_NPVT_INTERVALS_NORMAL };
```

cgmb (commented):
I'm actually just going to rescind the approval, since this is still marked as draft. I guess what I'm saying is that this looks good. If you make it even better, then great!

library/src/lapack/roclapack_getrf.hpp
```diff
@@ -199,6 +218,7 @@ rocblas_status rocsolver_getrf_template(rocblas_handle handle,
 
     rocblas_int blk = get_blksize<ISBATCHED, PIVOT>(dim);
 
+    rocsolver_cout << blk << "...\n";
```
tfalders, line 221: This should probably be removed.

tfalders (approved):
LGTM!

library/src/lapack/roclapack_getrf.hpp
jzuniga-amd: oops forgot to remove this...
removed.

cgmb:
LGTM. If the test pass, it's good.

cgmb (approved):

PR #221, opened 2021-02-25
Added sygv/hegv

tfalders:
Addresses SWDEV-247081. Also fixes some missed abstractions in the test fixtures for sygst and syev.

Note that an early version of testing_sygv_hegv.hpp was accidentally committed in a previous commit, so the full file won't show up in the diff.


jzuniga-amd (approved):
LGTM.

cgmb (approved):
I have nothing to add. LGTM. 👍

PR #222, opened 2021-02-25
Add missing gemm logging

tfalders:
Thin SVD commit added some gemm calls without logging. This adds logging to those calls.


jzuniga-amd (approved):
Good catch.
LGTM

cgmb (approved):
LGTM! :+1:

PR #223, opened 2021-02-26
Minimize visible symbols in shared library

cgmb:
This PR removes some of the public symbols in `rocsolver.so` introduced in #204. There has not been a release since that change, so we can remove these without breaking binary compatibility with any released versions.

I eliminate the dependency of the clients on the library for these symbols by duplicating the sources. I am not an expert on the linking process, so I am putting them in different namespaces when they're compiled for each target. That will ensure they're completely separate from each other. I don't think the namespace is actually necessary, but I want to play it safe. We can remove it at a later date when we're more certain that it's unnecessary.

The change also builds upon #210 to clean up our include paths a little. I've made it so that anything in an `include/` directory can be included without specifying a path. Headers that are not in an `include/` directory can be included using the path starting from `library/src/` or `clients/`, depending on whether the source file is a part of a client or a part of the library. That eliminates all cases where an include path is specified starting with `..`.

I also made a small tweak to the logging docs, since the API section felt a bit sparse. The user guide has plenty of information, so I just added a link to make it easier to find.

Finally, I adjusted our clang-format settings to minimize the disruption caused by the addition of the namespaces in the common code, and I figured I might as well throw in a couple additions to `.githooks/pre-commit --format` that I used while developing this change.


cgmb:
Nearly forgot, but this also contains a minor fix to #218. Apparently, `<` means less than or equal in Debian package specifications. Its use is discouraged in favour of `<<` for less than or `<=` for less than or equal.

tfalders (approved):
LGTM, though I do have a question. It looks like you're creating libraries `client-common` and `rocsolver-common` - will that result in separate .so files or is it simply an organizational tool?

cgmb:
> It looks like you're creating libraries client-common and rocsolver-common - will that result in separate .so files or is it simply an organizational tool?

It's entirely an organizational tool.

The rocsolver-common and clients-common libraries I defined are [CMake interface libraries](https://cmake.org/cmake/help/v3.9/command/add_library.html#interface-libraries). Despite the name, they're not real libraries at all. They're just a collection of properties that are added to the libraries or executables that 'link' to them.

jzuniga-amd (approved):
LGTM. Nice clean-up!

PR #224, opened 2021-02-26
add known-bug tests to GESVD

jzuniga-amd:


tfalders (approved):
LGTM, though knownbug_opt_range is different from large_opt_range. Won't this remove some tests?

cgmb (approved):
As long as it builds and runs fine, seems ok to me.

jzuniga-amd:
> LGTM, though knownbug_opt_range is different from large_opt_range. Won't this remove some tests?

True. I can add them back. The removed tests don't fail that's why I did not include them in knownbugs.

PR #226, opened 2021-03-02
Fix rocBLAS version

tfalders:
rocBLAS version numbers are odd on develop and staging, even on master, and bump up by 2 each release cycle. This PR changes our bump script to bump the rocBLAS version by 2 and allows rocSOLVER to build against both the develop and master versions.

EDIT: I've made additional changes to the bump script that make it so that the script should be applied after the feature complete merge, rather than before. This will mean that the version number during the development cycle reflects the rocSOLVER version that will be released at the end of the current cycle.


library/CMakeLists.txt
```diff
@@ -58,8 +58,8 @@ set( INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_INCLUDEDIR} )
 add_subdirectory( src )
 
 # Package specific CPACK vars
-set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocblas (>= 2.38), rocblas (<< 2.39)" )
```
cgmb, line 61: So, there's a reason why I specified the range in such an awkward way. If you build the a rocBLAS package, you'll get something like this:

```sh
cgmb@localhost:/tmp# dpkg -I /src/rocBLAS/build/release/rocblas_2.38.0-7c1771df_amd64.deb 
 new debian package, version 2.0.
 size 7235538 bytes: control archive=1281 bytes.
     320 bytes,    10 lines      control              
    2212 bytes,    25 lines      md5sums              
      96 bytes,     2 lines      postinst             
      76 bytes,     2 lines      prerm                
 Architecture: amd64
 Depends: hip-rocclr (>= 3.5.0)
 Description: rocBLAS is AMD's library for BLAS on ROCm. It is implemented in HIP and optimized for AMD GPUs.
 Maintainer: rocBLAS Maintainer <rocblas-maintainer@amd.com>
 Package: rocblas
 Priority: optional
 Section: devel
 Version: 2.38.0-7c1771df
 Installed-Size: 197476
```

If I understand correctly, you're wanting this to be compatible with the next rocBLAS version on master for the next version (which will increment by 2). That version doesn't exist, but we can imagine what will happen if we use the most recent version and decrement our expectations by 2:

```sh
cgmb@localhost:/tmp# dpkg --compare-versions 2.38.0-7c1771df le 2.38 && echo True || echo False 
False
```

So basically, I guess what I'm saying is, this should be `<< 2.41` rather than `<= 2.40`. I don't think the RPM packages suffer from this problem, but we might as well make them match the deb packages anyway.

tfalders: Fixed

tfalders:
Since rocBLAS is apparently moving to using a single version number across develop and master, I'll update this PR to reflect that.

bump_rocsolver_version.sh
```diff
@@ -1,22 +1,22 @@
 #!/bin/sh
 
 # This script needs to be edited to bump rocsolver version for new release.
-# - run this script in develop before merging develop/staging into master
+# - run this script in develop after merging develop/staging into master
 
-OLD_ROCSOLVER_VERSION="3.11.0"  #Released in February/March 2021
-NEW_ROCSOLVER_VERSION="3.12.0"  #To release in April/May 2021
+OLD_ROCSOLVER_VERSION="3.13.0"  #To release in April/May 2021
+NEW_ROCSOLVER_VERSION="3.14.0"
```
jzuniga-amd, line 7: Using the dates never worked... So, I suggest we simply indicate the ROCm number: 
OLD_ROCSOLVER_VERSION="3.13.0"  #distributed with ROCm 4.3
NEW_ROCSOLVER_VERSION="3.14.0" #distributed with ROCm 4.4

or something like this...

bump_rocsolver_version.sh
```diff
@@ -1,22 +1,22 @@
 #!/bin/sh
 
 # This script needs to be edited to bump rocsolver version for new release.
-# - run this script in develop before merging develop/staging into master
+# - run this script in develop after merging develop/staging into master
```
jzuniga-amd, line 4: To be more clear, we should say:
"run this script in develop after merging develop/staging into master at feature-complete date"

jzuniga-amd (approved):
LGTM

cgmb (approved):
That was a bit of a pain, but I'm glad we found simple solution. It sounds like this is less work for rocBLAS, too. 👍

PR #227, opened 2021-03-08
Cleanup clients build system - 5/5

cgmb:
This is basically updating the clients CMake to a more modern style. It follows more or less the same pattern as Torre did in [his change](https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/d197906e6ce579032660ef67977674dbc9348375). However, we don't actually use OpenMP, so we can cut more and we don't need to bump the minimum version. (At some point this iteration we may want to bump to match rocBLAS, but this is already a lot in one PR.)

I thought I saw some issues with librt, but I'd been having trouble reproducing them. I'll leave this in draft until I'm sure they're not a problem.

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -10,12 +10,6 @@ add_executable( rocsolver-bench client.cpp )
 
 add_armor_flags( rocsolver-bench "${ARMOR_LEVEL}" )
 
-# Internal header includes
-target_include_directories( rocsolver-bench
-  PRIVATE
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
-)
-
 # External header includes
 target_include_directories( rocsolver-bench
   SYSTEM PRIVATE
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -2,9 +2,6 @@
 # Copyright (c) 2016-2021 Advanced Micro Devices, Inc.
 # ########################################################################
 
-set( THREADS_PREFER_PTHREAD_FLAG ON )
-find_package( Threads REQUIRED )
-
 # Linking lapack library requires fortran flags
 enable_language( Fortran )
 find_package( cblas REQUIRED CONFIG )
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -2,25 +2,12 @@
 # Copyright (c) 2016-2021 Advanced Micro Devices, Inc.
 # ########################################################################
 
-if(EXISTS /etc/redhat-release)
-    if(CXX_VERSION_STRING MATCHES "clang")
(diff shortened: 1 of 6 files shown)


cgmb:
The CI seems happy. I still wonder if we're handling `librt` and `libm` correctly, but this doesn't really change anything in that respect.

clients/benchmarks/CMakeLists.txt
```diff
@@ -2,88 +2,27 @@
...
+    $<BUILD_INTERFACE:${CBLAS_INCLUDE_DIRS}>
+    )
 
-    target_link_libraries( rocsolver-bench PRIVATE cblas lapack roc::rocsolver )
-endif()
+target_link_libraries( rocsolver-bench PRIVATE
+  cblas
+  lapack
+  Threads::Threads
+  roc::rocsolver
```
TorreZuk, line 24: I think this should be reordered consumer to provider, in static linking you find symbols in the later libs referred to by the earlier lib.  solver lapack cblas threads.

clients/gtest/CMakeLists.txt
```diff
@@ -95,71 +79,19 @@ add_armor_flags( rocsolver-test "${ARMOR_LEVEL}" )
...
-#        $<BUILD_INTERFACE:${BLIS_INCLUDE_DIR}>
-        )
-
-    target_link_libraries( rocsolver-test PRIVATE ${GTEST_LIBRARIES} cblas lapack roc::rocsolver )
-endif()
+target_link_libraries( rocsolver-test PRIVATE
+  cblas
+  lapack
+  GTest::GTest
+  roc::rocsolver
```
TorreZuk, line 93: Same reordering here.

TorreZuk (approved):
Good cleanup, minor comment made.

clients/benchmarks/CMakeLists.txt
cgmb: That's actually why I put rocsolver last. None of these libraries should be consumed by `roc::rocsolver`.

The `roc::rocsolver` interface should include everything the library needs. If `roc::rocsolver` depended on `Threads::Threads`, but didn't correctly list it in its interface requirements, then putting `roc::rocsolver` first would hide that mistake. Putting `roc::rocsolver` last makes it more likely that missing library interface requirements will cause a link failure on our CI, so we can catch mistakes earlier.

jzuniga-amd (approved):
LGTM. 
The implicit tests are getting more complicated, and we are coming to a point that the time required to execute the full tests-suite is long enough to maybe justify some parallelization on the CPU side. So, either we keep openMP organized and ready for future use, or we add it back later... just something to keep in mind.

clients/benchmarks/CMakeLists.txt
TorreZuk: Okay that's good, just looked backwards as the external libs I usually see listed last, but none of these are local as there is no client-common lib or fortran wrapper.

cgmb:
Thanks for informing me! That's is good to keep in mind.

My preference is to ruthlessly cut code that is not actually used. It's hard to feel confident in the correctness of a piece of code if nothing actually depends on it working correctly. IMO, it's total fine to remove OpenMP now and add it back following Torre's example when we parallelize the tests (regardless of whether that's tomorrow, next week, next month or next year).

tfalders (approved):

PR #228, opened 2021-03-08
Fix gesvd tests

jzuniga-amd:
Fix GESVD tests marked as "known bugs" by re-organizing the implicit test to avoid combining  gemm_strided_batched- with gemm_bached- based computations, which seems to be producing numerical artifacts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 
 ### Known Issues
 
+### Deprecated
+
+### Removed
+
+### Fixed
+- Thin-SVD failing tests. 
+
+### Known Issues
+
+### Security
 
 ## [rocSOLVER 3.12.0 for ROCm 4.2.0]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Removed
 
 ### Fixed
-- Thin-SVD known failing tests. 
+- Fixed known Thin-SVD failing tests.
 
 ### Known Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Removed
 
 ### Fixed
-- Fixed known Thin-SVD failing tests.
+- Fixed known issues with Thin-SVD. The problem was identified in the test specification, not in the thin-SVD
+  implementation or the rocBLAS gemm\_batched routines.
 
 ### Known Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,19 +12,10 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Removed
 
 ### Fixed
+- Thin-SVD known failing tests. 
 
 ### Known Issues
 
-### Deprecated
-
-### Removed
-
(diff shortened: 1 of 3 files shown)


clients/include/testing_gesvd.hpp
```diff
@@ -276,11 +276,13 @@ void gesvd_getError(const rocblas_handle handle,
...
     gesvd_initData<true, true, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A);
 
-    // execute computations
-    // complementary execution (to compute all singular vectors if needed)
-    CHECK_ROCBLAS_ERROR(rocsolver_gesvd(
-        STRIDED, handle, left_svectT, right_svectT, mT, nT, dA.data(), lda, stA, dS.data(), stS,
-        dUT.data(), lduT, stUT, dVT.data(), ldvT, stVT, dE.data(), stE, fa, dinfo.data(), bc));
+    // execute computations:
+    // complementary execution to compute all singular vectors if needed (always in-place to ensure
+    // we don't combine results computed by gemm_bacthed with results computed by gemm_strided_batched)
```
tfalders, line 281: Spelling error on batched

tfalders (approved):
LGTM! I'm curious, though, if this is purely a test issue or if a user could experience the same artifacts in their code by using the out-of-place computation.

jzuniga-amd:
> LGTM! I'm curious, though, if this is purely a test issue or if a user could experience the same artifacts in their code by using the out-of-place computation.

Short answer: yes, this is purely a test issue.

Long answer: the problem has nothing to do with the algorithm (in-place, out-of-place, thin-svd, normal svd, etc.) The problem arises when you extract the, let's say left vectors, from a bidiagonalization computed with gemm_batched and the right vectors from a bidiagonalization computed with gemm_strided_batched. Both bidiagonalizations are correct, and both sets of vectors as well, they are just not "compatible" as part of a single SVD. This makes sense, of course... one does not compute left and right vectors in two separate calls, but again, for testing purposes, we have to. 

If the tested case is `gesvd_batched(left_vect = none, right_vect = all, ...)`, for example, I still need the left vectors for the implicit test, but I cannot run `gesvd_batched(left_vect = all, right_vect = all)` cause this would be testing a different path. I need to make two separate calls, the required call  `gesvd_batched(left_vect = none, right_vect = all, ...)`, and a complementary call `gesvd_batched(left_vect = all, right_vect = none, ...)`. So, the solution in this PR is to ensure that the complementary call always executes the bidiagonalization in the same way as the required call does (ie. both using gemm_batched, or both using gemm_strided_batched).

clients/include/testing_gesvd.hpp
jzuniga-amd: corrected.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Removed
 
 ### Fixed
+- Fixed known Thin-SVD failing tests.
```
cgmb, line 15: We should be clear that the problem was entirely within the test specification, not the implementation.

@doctorcolinsmith, should we be changing the Known Issues entry of the 4.2 CHANGELOG now that we have more information about the issue?

jzuniga-amd: IMO, I don't think we need to change the 4.2 changelog entry. The problem at that moment, for the 4.2 version, was not identified...  (besides, changing would imply a hotfix to the master branch).

cgmb: I'm fine with that. Maybe I shouldn't bug Colin over this stuff, but I like to point out process weirdness in hopes that one day it can be made less weird. :p

cgmb (approved):

PR #229, opened 2021-03-08
Client improvements 2

tfalders:
Addresses SWDEV-275379.

Summary of proposed changes:
- Improved emulation of boost's variables_map (I plan to port these changes over to rocBLAS and hipBLAS if accepted)
- Convert Arguments class into a map of variables with easy-to-use get and set methods
    - New argument variables only need to be added to client.cpp
    - Getter method can return a default value if an actual value has not been provided
    - Slight drawback in the form of more verbose code and increased compilation time
    - Some argument variables (namely, test options) are still kept as public fields for ease-of-use


clients/include/testing_gesvd.hpp
```diff
@@ -460,21 +460,21 @@ void testing_gesvd(Arguments argus)
...
+    rocblas_int m = argus.get<rocblas_int>("m");
+    rocblas_int n = argus.get<rocblas_int>("n");
+    rocblas_int lda = argus.get<rocblas_int>("lda", m);
+    rocblas_int ldu = argus.get<rocblas_int>("ldu", m);
+    rocblas_int ldv = argus.get<rocblas_int>("ldv", (rightvC == 'A' ? n : min(m, n)));
+    rocblas_stride stA = argus.get<rocblas_stride>("strideA", lda * n);
+    rocblas_stride stS = argus.get<rocblas_stride>("strideS", min(m, n));
+    rocblas_stride stU = argus.get<rocblas_stride>("strideU", ldu * m);
+    rocblas_stride stV = argus.get<rocblas_stride>("strideV", ldv * n);
+    rocblas_stride stE = argus.get<rocblas_stride>("strideE", min(m, n) - 1);
```
cgmb, line 474: A couple questions, so I can understand a bit better what the user experience would be like:
- What happens if the given strideE is not a valid argument (e.g. not an integer)?
- What happens if I make a typo and specify `striideE` or something similar?

tfalders: If you meant what happens if you make a typo on the command line, that should also be caught by the parser.

cgmb: That's great info, thanks!

> there are no checks for the strides, and giving a stride that is less than the data size gives a hip out of memory error when we do the memcheck() calls. This issue existed before my changes, though.

Ok. I think we can leave it as out-of-scope.

> If you'd like, I can change it so that both versions of get will produce an error if given an invalid name, however this will mean that strides have to be explicitly set in the gtest.cpp files.

You could invert the check and instead look to see that all things that were set were used.

If `Arguments::get(...)` is always the way these properties are accessed, then it could just remember every name that was queried. Give it a method to return the set of all names that were set but never accessed, and rocsolver-benchmark can check that at the end of its run. Print a warning at the bottom of the output if there's anything that was unused.

That would mean, `void testing_gesvd(Arguments argus)` would have to be `void testing_gesvd(Arguments& argus)` and `Arguments::get` would have to be non-const... but I don't think that would cause any problems in our current code.

We should definitely inform the user when they make that sort of mistake, but I'm not picky about how. Any way you'd like to handle it is fine with me.

cgmb: > If you meant what happens if you make a typo on the command line, that should also be caught by the parser.

Ah. Nevermind. I suppose my suggestion would catch some other classes of errors (e.g. where you pass something that would be a valid argument for a different benchmark), but the parser catches my main concern.

tfalders: Good idea! I'll try to put something together for the "inverted check", as using incorrect argument names is definitely a risk for us.

clients/include/rocsolver_arguments.hpp
```diff
@@ -5,14 +5,130 @@
...
+    template <typename T>
+    const T& get(const std::string& name, const T& default_value)
+    {
+        to_consume.erase(name);
+        auto val = find(name);
+        if(val != end() && !val->second.empty() && !val->second.defaulted())
+            return val->second.as<T>();
+        else
+            return default_value;
+    }
```
cgmb, line 64: I just wanted to mention before I forget. This is a video worth watching: https://www.youtube.com/watch?v=lkgszkPnV8g

tfalders: Fixed. I'll push the commit after Juan's feedback, since I have additional commits waiting in the wings and don't want to clutter the diff just yet.

clients/benchmarks/client.cpp
```diff
@@ -55,53 +89,72 @@ try
...
-        ("bsb",
-         value<rocblas_int>(&argus.bsb)->default_value(1024*1024),
-         "Specific stride of strided_batched matrix B, is only applicable to strided batched"
-         "BLAS-2 and BLAS-3: second dimension * leading dimension.")
+        ("ldv",
+         value<rocblas_int>(),
+         "Leading dimension of matrices V.")
+
+        // stride options
+        ("strideA",
```
doctorcolinsmith, line 119: Minor item - It would be nice to alphabetise the options in each section.  I think that will help maintain the options in the future.

clients/include/testing_getf2_getrf.hpp
```diff
@@ -319,7 +319,7 @@ void getf2_getrf_getPerfData(const rocblas_handle handle,
 }
 
 template <bool BATCHED, bool STRIDED, bool GETRF, typename T>
-void testing_getf2_getrf(Arguments argus)
+void testing_getf2_getrf(Arguments& argus)
```
doctorcolinsmith, line 322: Prefer const ref to plain ref (e.g., `const Arguments& argus`) to prevent accidental modification in the tests.  Same applies to the functions that follow.

cgmb: The `Arguments` class is doing some bookkeeping on property accesses to ensure that all arguments passed to `rocsolver-benchmark` are actually being used. The purpose is to warn the user if they pass an argument that goes unused (perhaps suggesting a mistake). It's not a beautiful solution, but I think it will be effective and low-maintenance.

There are a few ways we could adjust the design to pass `Arguments` as const (e.g. we could pass the bookkeeping separately, we could mark it as mutable, etc...), but I figured this was fine for now.

jzuniga-amd (commented):
It looks good to me in general. Great job Troy and Cory!. 
Below are some initial comments. I might have some more comments once the concept is applied to all functions...

My main concern now has to do with the error messages and the format they are printed. I have some suggestions to be more consistent with other messages, but it was easier to simply apply them in a PR to your branch than explain them here. So, please have a look and let me know what you think.

clients/benchmarks/client.cpp
tfalders: Sure, I can do that.

clients/benchmarks/client.cpp
```diff
@@ -250,27 +279,16 @@ try
     if(argus.storev != 'R' && argus.storev != 'C')
         throw std::invalid_argument("Invalid value for --storev");
```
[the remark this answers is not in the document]
tfalders: That's the plan, yes!

clients/benchmarks/client.cpp
```diff
@@ -250,27 +279,16 @@ try
...
 
-    // workmode
-    if(argus.workmode != 'O' && argus.workmode != 'I')
-        throw std::invalid_argument("Invalid value for --workmode");
-
-    // itype
-    if(argus.itype != '1' && argus.itype != '2' && argus.itype != '3')
-        throw std::invalid_argument("Invalid value for --itype");
+    argus.validate_workmode("fast_alg");
+    argus.validate_evect("jobz");
```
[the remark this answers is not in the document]
tfalders: Sounds good, I'll do that.

clients/gtest/orgbr_ungbr_gtest.cpp
```diff
@@ -25,7 +25,7 @@ typedef std::tuple<vector<int>, vector<int>> orgbr_tuple;
 // case when m = 0, n = 0 and storev = 'C' will also execute the bad arguments
 // test (null handle, null pointers and invalid values)
 
-const vector<vector<int>> store = {
+const vector<vector<int>> store_range = {
```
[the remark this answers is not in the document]
tfalders: Yes, I was getting a name conflict. It might be worth wrapping program_options.hpp in a "boost" namespace since I believe that's what the actual boost/program_options.hpp does. I'll add that in.

clients/include/rocsolver_arguments.hpp
```diff
@@ -5,14 +5,130 @@
...
+
+    std::set<std::string> to_consume;
+
 public:
+    ~Arguments()
+    {
+        if(to_consume.size() > 0)
+        {
+            std::stringstream ss;
+            ss << "WARNING: Not all arguments were consumed:";
```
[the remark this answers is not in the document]
tfalders: Ah, I missed that case. Shouldn't be too hard to fix.

clients/benchmarks/client.cpp
tfalders: I'll add it in as part of the final cleanup once all the tests have been updated.

clients/benchmarks/client.cpp
```diff
@@ -203,77 +320,63 @@ try
...
+            << "Note: When a required parameter/option is not provided, it will take the default "
+               "value as listed below.\n"
+            << "If no default value is defined, the program will try to calculate a suitable value "
+               "depending on the context\n"
+            << "of the problem and the tested function; if this is not possible, the program will "
+               "abort with error.\n\n"
+            << "Example: ./rocsolver-bench -f getf2_batched -m 30 --lda 75 --batch_count 350\n"
+            << "This will test getf2_batched with a set of 350 random 30x128 matrices. strideP "
+               "will "
+               "be set to be equal to 30.\n\n"
```
cgmb, line 340: This looks like it was a bit mangled by the auto-formatter, mixing multiple different ways of breaking the help text across multiple lines.

The `<<` for each line is unnecessary, since adjacent string literals are combined into a single string literal. You could just delete them, but I think it would be even better to replace entire thing with a [C++11 raw string literal](https://en.cppreference.com/w/cpp/language/string_literal) and write it how you want to see it. e.g.
```cpp
R"DELIMITER(
rocSOLVER benchmark client help.

Usage: ./rocsolver-bench <options>

In addition to some common general options, the following list of options corresponds to all the parameters
that might be needed to test a given rocSOLVER function. The parameters are named as in the API user guide.
)DELIMITER"
```

The `DELIMITER` can be replaced with a different string of characters. The important thing is just to ensure that it's not found within the quoted text.

clients/gtest/bdsqr_gtest.cpp
```diff
@@ -66,19 +66,20 @@ Arguments bdsqr_setup_arguments(bdsqr_tuple tup)
 
     Arguments arg;
 
-    arg.M = size[0]; // n
-    arg.N = size[1]; // nv
```
cgmb, line 70: I found this sort of thing very confusing. Nice work in removing them!

clients/include/rocsolver_arguments.hpp
```diff
@@ -5,67 +5,168 @@
...
+            for(std::string name : to_consume)
+                ss << ' ' << name;
+            rocsolver_cerr << ss.str() << std::endl;
+        }
+    }
```
cgmb, line 31: I would prefer a function for this, so you can more easily control if and when it occurs. That said, it's not a big deal either way.

Though, I would definitely make the class move-only (default the move constructor/assignment operator, delete the copy constructor/assignment operator). If you are doing bookkeeping, it's a bad idea to have a second set of books!

cgmb (commented):
There's still a couple details to adjust, but it looks like a good change. Nice job!

clients/include/rocsolver_arguments.hpp
tfalders: I think I didn't create a function for this because it would have to be added to every test - which, in hindsight, is a poor reason since I ended up editing every test anyway. Having it in the destructor did cause me a couple issues, so it might be worth it to make the change.

clients/rocblascommon/program_options.hpp
```diff
@@ -8,79 +8,151 @@
...
 #include <ostream>
 #include <regex>
-#include <set>
 #include <stdexcept>
 #include <string>
 #include <utility>
 #include <vector>
 
+namespace boost
+{
```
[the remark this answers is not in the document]
tfalders: Perhaps `roc::boost`? I don't want the name to be too different from the original, but I see why we should change it.

clients/benchmarks/client.cpp
tfalders: The raw string literal looks neat, but if I split a sentence across multiple lines, will the output force a newline in the middle of the sentence?

tfalders: On, never mind. The current format already forces some newlines in some sentences.

clients/include/rocsolver_arguments.hpp
cgmb: I figured it was only needed for the benchmark client, but if it's useful for the tests, that seems reasonable too.

tfalders: I wanted to add it to the tests too just to make sure one of us doesn't, e.g., copy the get for lda to ldb and forget to update the key.

clients/rocblascommon/program_options.hpp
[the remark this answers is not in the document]
cgmb: I don't think it's necessary to include the name `boost` at all. There's nothing about these classes that are inherently tied to boost.

It's common to have similar classes that vary only by namespace. When `boost::bind` was adopted by the C++ standard, it was adopted as `std::bind`. More recent examples include `boost::optional` => `std::optional`, `boost::filesystem` => `std::filesystem`, `boost::string_view` => `absl::string_view`, `fmt::string_view`, `std::string_view`, etc.

I'd prefer to drop the boost inner namespace, but getting it inside our namespace was the most important thing. That was the only part I would insist on, and you've addressed it.

cgmb (approved):
Looks like the CI is having some trouble. Definitely wait for green checkmarks before merging.

~I'm still not entirely behind the namespace choice, but that's small potatoes.~

clients/include/testing_orglx_unglx.hpp
```diff
@@ -193,14 +193,15 @@ void orglx_unglx_getPerfData(const rocblas_handle handle,
...
 {
     // get arguments
     rocblas_local_handle handle;
-    rocblas_int k = argus.K;
-    rocblas_int m = argus.M;
-    rocblas_int n = argus.N;
-    rocblas_int lda = argus.lda;
+    rocblas_int k = argus.get<rocblas_int>("k");
+    rocblas_int m = argus.get<rocblas_int>("m");
+    rocblas_int n = argus.get<rocblas_int>("n");
```
jzuniga-amd, line 202: To complete the example used in the comments on the default values, this should be:

`rocblas_int m = argus.get<rocblas_int>("m",128);`
`rocblas_int k = argus.get<rocblas_int>("k",m);`
 `rocblas_int n = argus.get<rocblas_int>("n",m);`

(same for other similar functions)

clients/include/testing_orgxl_ungxl.hpp
```diff
@@ -193,14 +193,15 @@ void orgxl_ungxl_getPerfData(const rocblas_handle handle,
...
 {
     // get arguments
     rocblas_local_handle handle;
-    rocblas_int k = argus.K;
-    rocblas_int m = argus.M;
-    rocblas_int n = argus.N;
-    rocblas_int lda = argus.lda;
+    rocblas_int k = argus.get<rocblas_int>("k");
+    rocblas_int m = argus.get<rocblas_int>("m");
+    rocblas_int n = argus.get<rocblas_int>("n");
```
jzuniga-amd, line 202: On the other hand, here the defaults should be determined as follows:

`rocblas_int n = argus.get<rocblas_int>("n",128);`
`rocblas_int m = argus.get<rocblas_int>("m",n);`
`rocblas_int k = argus.get<rocblas_int>("k",n);`

(same for other similar functions)

clients/include/testing_ormlx_unmlx.hpp
```diff
@@ -257,20 +257,21 @@ void ormlx_unmlx_getPerfData(const rocblas_handle handle,
...
-    rocblas_int m = argus.M;
-    rocblas_int n = argus.N;
-    rocblas_int lda = argus.lda;
-    rocblas_int ldc = argus.ldc;
-    rocblas_int hot_calls = argus.iters;
-    char sideC = argus.side_option;
-    char transC = argus.transA_option;
+    char sideC = argus.get<char>("side");
+    char transC = argus.get<char>("trans");
+    rocblas_int k = argus.get<rocblas_int>("k");
```
jzuniga-amd, line 266: Here the default of `k` depends on the value of `side`, etc...

**So in general, the idea is to look at the API guide and define the default values accordingly for each test. Only the parameters that can really take any value across all functions should have a global default (defined in client.cpp or rocsolver_arguments.hpp)**

jzuniga-amd (changes requested):
**Very nice Troy! The set and get mechanisms, and the mapping of all parameters is clean and functional...**

I just have a few more suggested changes to clarify things from the user perspective and to avoid further errors with the default values.

clients/benchmarks/client.cpp
```diff
@@ -22,178 +45,293 @@ try
...
+         value<rocblas_int>(&argus.norm_check)->default_value(0),
+            "Validate GPU results with CPU? 0 = No, 1 = Yes.\n"
+            "                           This will additionally print the relative error of the computations.\n"
+            "                           ")
+
+        // size options
+        ("k",
+         value<rocblas_int>()->default_value(128),
+            "Matrix/vector size parameter.\n"
+            "                           Typically, the number of Householder reflections in a transformation.\n"
```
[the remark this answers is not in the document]
tfalders: I don't think that calculating defaults for n and k in the test body will address this problem either. For example, `rocsolver-bench -f orglq -k 150` will still fail since m is not calculated based on k. We should probably require the user to provide either all size arguments or none, so `rocsolver-bench -f orglq -m 30 -n 30 -k 30` and `rocsolver-bench -f orglq` would be valid, but `rocsolver-bench -f orglq -m 30 -n 30` and `rocsolver-bench -f orglq -k 150` would not. What do you think?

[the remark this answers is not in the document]
jzuniga-amd: Yeah, you are right! mmm I was hoping this not to get more complicated, but I don't like us providing default values that fail; better not to have defaults and always require the user to pass all parameters. 

BTW what would happen if we call `argus.get<rocblas_int>("m");` and `m` has no value?
If this method returns NAN or something that we can detect, then we can ask the user to always provide the "fundamental" parameter...

So, for example, as `orglq` computes a matrix with orthonormal rows, `m` must be provided; `orgqr` computes a matrix with orthogonal columns, so `n` must be provided, etc. Everything else can be calculated. (Most functions will do well with m=128, n=128, anyways).

So, in the case of `rocsolver-bench -f orglq -k 150`, we can maybe return an error saying
"A default value for -m cannot be calculated; please provide a suitable value" or something like this...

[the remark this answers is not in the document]
cgmb: > BTW what would happen if we call argus.get<rocblas_int>("m"); and m has no value?
If this method returns NAN or something that we can detect, then we can ask the user to always provide the "fundamental" parameter...

It would throw an `std::out_of_range` exception, which would trigger the program to terminate. I'm not sure if the error message would specifically state what argument was missing, but it might actually be pretty good already. If it's not good enough, we could clean that up a bit without changing it much. We'd define our own exception, do an explicit throw in `get` if an argument is requested but not found, and catch the exception in `main()` to print a nicer error message.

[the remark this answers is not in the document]
tfalders: > So, in the case of `rocsolver-bench -f orglq -k 150`, we can maybe return an error saying
> "A default value for -m cannot be calculated; please provide a suitable value" or something like this...

I like this idea. I think we can still keep the defaults for most function, but for those you've mentioned I can calculate the secondary parameters and terminate the program if a secondary parameter is provided without the primary parameter.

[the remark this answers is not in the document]
jzuniga-amd: If you agree with the idea of only require the "fundamental parameter(s)" for those problematic functions, here is a list of 
[function - required parameter(s)]

laswp - k1
larft - n,k,storev
larf - m
larfb - m,n,k,side,storev
labrd - m,n
bdsqr - n,nu
latrd - n
steqr - n
orgxr/ungxr - m,n (yes, m is also required to avoid problems with lda)
orglx/unglx - m,n
orgxl/ungxl - m,n
orgbr/ungbr - m,n,storev
orgtr/ungtr - n
ormxr/unmxr - m,n,side
ormlx/unmlx - m,k,side
ormxl/unmxl - m,n,side
ormbr/unmbr - m,n,side,storev
ormtr/unmtr - m,n,side
potf2/potrf - n
getf2/getrf - m
geqr2/geqrf - m
geql2/geqlf - m
gelq2/gelqf - m
gebd2/gebrd - m
sytd2/sytrd/hetd2/hetrd - n
sygs2/sygst/hegs2/hegst - n
getri - n
getrs - n
gels - m,n
syev/heev - n
sygv/hegv - n
gesvd - m,n

So, rather almost always we will require the user to provide a size... but still we have some parameters that can default to something. It is mid-way between not requesting any parameter and requesting all parameters. 
If you want, you can check whether the required parameter is really needed. For example if in potf2, 'n' is not provided, but neither 'lda', then both can get defaults... Even fancier, you can see whether one parameter can be calculated if we know the other, and so on. This can be as elaborated as we want. To me, always requesting the minimal parameters in the list is just fine.  

As for the error, using the exceptions to provide the message is fine with me as long as it provides information on what is the missing required parameter.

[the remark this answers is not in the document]
tfalders: Oh, you want the check on lda, ldb, etc. as well? That makes it a bit more involved... I'll have to think about how to do this cleanly...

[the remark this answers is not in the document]
tfalders: Hmmm, I think we might be overengineering this a little bit. For now, adding calculations for the secondary parameters should be enough so that the client, in most cases, won't fail. I'll go ahead and do that. Perhaps we can discuss more advanced behaviours in a later meeting, as I also don't want things to become too complicated.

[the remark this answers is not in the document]
jzuniga-amd: > Oh, you want the check on lda, ldb, etc. as well? That makes it a bit more involved... I'll have to think about how to do this cleanly...

No, I am just saying that it could be done, but for me, it is fine if we just require the user to provide the primary parameters (as in the list), and calculate a default for the other parameters if not provided.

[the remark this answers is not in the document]
tfalders: Ah, okay, that's more doable. I can add that in.

jzuniga-amd (approved):
LGTM.
A few non-important comments... except for LASWP which I think needs to be fixed.
After that, let's merge this PR! Good job...

clients/include/testing_laswp.hpp
```diff
@@ -183,15 +183,16 @@ void laswp_getPerfData(const rocblas_handle handle,
...
 {
     // get arguments
     rocblas_local_handle handle;
-    rocblas_int n = argus.N;
-    rocblas_int lda = argus.lda;
-    rocblas_int k1 = argus.k1;
-    rocblas_int k2 = argus.k2;
-    rocblas_int inc = argus.incx;
+    rocblas_int n = argus.get<rocblas_int>("n");
+    rocblas_int lda = argus.get<rocblas_int>("lda", 1);
```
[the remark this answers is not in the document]
tfalders: Fixed.

Should we perhaps add a check for lda < k2 in laswp's argCheck as well? Maybe not in the PR, but if it's guaranteed to cause a memory access fault then that should probably be caught by the check.

clients/include/testing_lacgv.hpp
```diff
@@ -141,12 +141,13 @@ void lacgv_getPerfData(const rocblas_handle handle,
...
 
 template <typename T>
-void testing_lacgv(Arguments argus)
+void testing_lacgv(Arguments& argus)
 {
     // get arguments
     rocblas_local_handle handle;
-    rocblas_int n = argus.N;
-    rocblas_int inc = argus.incx;
+    rocblas_int n = argus.get<rocblas_int>("n");
```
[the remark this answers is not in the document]
tfalders: I'm uncomfortable with adding defaults that the user can't see. In cases where the default isn't a straightforward calculation, I'd prefer either no defaults (current situation) or global defaults (earlier situation, where the user can see these defaults in the help message). Defaulting to square matrices is a bit of an edge case, but we've now documented that in the help string.

clients/include/testing_latrd.hpp
```diff
@@ -243,19 +243,20 @@ void latrd_getPerfData(const rocblas_handle handle,
...
     rocblas_local_handle handle;
-    rocblas_int n = argus.N;
-    rocblas_int k = argus.K;
-    rocblas_int lda = argus.lda;
-    rocblas_int ldw = argus.ldb;
-    rocblas_int hot_calls = argus.iters;
-    char uploC = argus.uplo_option;
+    char uploC = argus.get<char>("uplo");
+    rocblas_int n = argus.get<rocblas_int>("n");
+    rocblas_int k = argus.get<rocblas_int>("k");
```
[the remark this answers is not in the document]
tfalders: Fixed

clients/include/testing_lacgv.hpp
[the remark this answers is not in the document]
jzuniga-amd: Sounds good.

clients/include/testing_laswp.hpp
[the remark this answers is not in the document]
jzuniga-amd: The values we use in the gtest ensure we don't have problems, but users can easily get memory access faults (in code or using rocsolver-bench) if they are not careful...
LAPACK does not have a check for lda, but we can certainly add it in a future PR.

tfalders:
If there are no objections, once CI passes I'll merge this PR.

PR #231, opened 2021-03-25
Remove compiler version parsing

cgmb:
We were parsing the output of `hipcc --version` to detect whether the compiler was ultimately going to be nvcc or hip-clang. However, the ROCmCC compiler is going to identify itself as "AMD clang" and `CXX_VERSION_STRING` is only populated with the first word, so the parsing code needs to be updated.

My proposal is to just delete all that code. Updating the check would be a smaller change, but the nvcc build doesn't actually work anyways so why bother maintaining code that exists to support a feature that wasn't implemented? This will also make it easier to use hip-clang directly, rather than relying on hipcc, as all the `if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$")` checks will go away with it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,69 +33,8 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
 # ########################################################################
-# NOTE:  CUDA compiling path
-# ########################################################################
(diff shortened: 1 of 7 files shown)


jzuniga-amd (approved):
LGTM.

tfalders (approved):
LGTM. CI failure looks unrelated.

PR #232, opened 2021-03-26
Add --clients-only to install.sh

cgmb:
This builds on #231 to add a `--clients-only` option to `./install.sh`. The CMake options for this feature are `-DBUILD_LIBRARY=<ON|OFF>` and the equivalent (but opposite) `-DSKIP_LIBRARY=<ON|OFF>`. I provided both CMake options, as `BUILD_LIBRARY` is more consistent with `BUILD_CLIENTS_<client>`, but `-DSKIP_LIBRARY` was the option that rocBLAS used.


CMakeLists.txt
```diff
@@ -67,15 +73,23 @@ endif( )
 set( ARMOR_LEVEL "${DEFAULT_ARMOR_LEVEL}" CACHE STRING "Enables increasingly expensive runtime correctness checks" )
 include( armor-config )
 
+# This option only works for make, nmake and ninja, but no reason it shouldn't be on all the time
+# It creates a compile_commands.json file for use with clang tooling or vim
+set( CMAKE_EXPORT_COMPILE_COMMANDS ON )
+
```
cgmb, line 79: Actually, do we care about this? I would not normally enable this in a CMakeLists.txt. It would be more appropriate for the user to specifically request it with `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON` if they want it.

I consolidated the two calls that enabled to this option into one call, but it might make more sense just to delete it entirely. This is not essential to the build.

CMakeLists.txt
```diff
@@ -55,7 +54,14 @@ set ( VERSION_STRING "3.13.0" )
...
 # NOTE:  workaround until llvm & hip cmake modules fixes symlink logic in their config files; remove when fixed
-list( APPEND CMAKE_PREFIX_PATH ${ROCM_PATH}/llvm ${ROCM_PATH}/hip /opt/rocm/llvm /opt/rocm/hip )
+list( APPEND CMAKE_PREFIX_PATH
+  ${ROCM_PATH}
+  ${ROCM_PATH}/llvm
+  ${ROCM_PATH}/hip
+  /opt/rocm
+  /opt/rocm/llvm
+  /opt/rocm/hip
+)
```
cgmb, line 64: It seems likely to me that this could be entirely deleted, but for now, I just consolidated two appends into one.

cgmb:
I figured this would be a quick change that could help delete more CMake code. I marked it as a draft briefly when I considered if improvements to the version number printing in rocsolver-test should be included in this PR (since this lets you mix-and-match even more component versions), but after further thought, I don't think that's essential.

TorreZuk (approved):
LGTM but added Daine as he did library only work

clients/CMakeLists.txt
```diff
@@ -2,41 +2,19 @@
...
-include( build-options )
-
-# This option only works for make/nmake and the ninja generators, but no reason it shouldn't be on all the time
-# This tells cmake to create a compile_commands.json file that can be used with clang tooling or vim
-set( CMAKE_EXPORT_COMPILE_COMMANDS ON )
+# Specify where to put the client binaries
+set( CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
 
+# The rocsolver target will exist if the library is being built along with the clients,
+# but if this is a clients-only build, we'll have to search for it.
```
daineAMD, line 11: In rocBLAS, there's an option to specify the location of the rocBLAS library to use, this can be combined with a default search in ${ROCM_PATH}/whatever. Not sure if it's useful for anyone, just thought I'd add it as a suggestion.

daineAMD (approved):

clients/CMakeLists.txt
cgmb: I think it's already possible to control the lookup through standard CMake options. CMake checks various [environment variables](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure) that control where find_package will search. I'll have to double-check, but `rocblas_DIR=/path/to/rocblas-install ./install.sh --clients-only` should work.

The nice thing is that it works in exactly the same way for third-party libraries. For example, you could do `GTest_DIR=/path/to/gtest cblas_DIR=/path/to/cblas rocblas_DIR=/path/to/rocblas-install ./install.sh --clients-only`

cgmb: It will actually work with everything except `rocblas_DIR`, since [the command line option will take priority, and we always add it to our CMake invokation](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/f0cbc734bdccc96f088287b7ebe16c2c3d6c469d/install.sh#L467). The `rocsolver_DIR`, `GTest_DIR`, and `cblas_DIR` environment variables do work with `install.sh`, but rocblas must be specified with the `--rocblas_dir` option.

tfalders (approved):
LGTM, though there appears to be a CI failure.

clients/CMakeLists.txt
jzuniga-amd: The environment variable rocblas_DIR also works for me:
` rocblas_DIR=/path/to/rocblas-install rocsolver_DIR=/path/to/rocsolver-install ./install.sh --clients-only`

I would prefer, anyway, to keep the `--rocblas_dir` option and to also add the `--rocsolver_dir` option to the `install.sh` script. I don't see why not; it just gives the user the exact same syntax as the one that it may be already used to... I mean, not all users are familiar with environment variables and define them at the command line before calling the script. It is confusing to have to do so for rocsolver but not for rocblas.

cgmb: I had a few improvements to the install.sh script in that old dependencies PR. I'll put up a separate PR that improves `--rocsolver_dir` and adds `--rocsolver_dir` as a new option. 

In general, I would like to avoid coding up custom logic in `install.sh` whenever it's reasonably possible. An option like that is only a dozen lines of code, but it's a dozen lines in rocblas, rocsolver, hipsolver, and hipblas. I would prefer to teach users how to do things using standard tools and methods, rather than writing and maintaining custom code to do those things for them.

I can see your point, though. `install.sh --help` is a form of documentation that makes these script options fairly obvious, but those other methods I suggested are not documented at all.

cgmb: @jzuniga-amd, I added `--rocsolver_dir` in 87b4dd705306f91beefa85c57015af6b76411e9b on this PR.

install.sh
```diff
@@ -393,6 +404,14 @@ else
   rm -rf -- "${build_dir}/debug"
 fi
 
+# resolve relative paths
```
jzuniga-amd, line 407: Nice!!!

jzuniga-amd (approved):
LGTM

PR #233, opened 2021-03-30
Use clang directly when building C sources

cgmb:
The C source files in rocSOLVER are all standard C code and do not need any particular compiler. The ROCm version of clang is just a convenient compiler that is guaranteed to be available wherever rocSOLVER is being built.

The hipcc wrapper is able to build both C and C++, but there are some bugs in its interpretation of command-line options, which can sometimes cause issues when building C sources. For more details, see  https://github.com/ROCm-Developer-Tools/HIP/issues/2246

In particular, this problem is a barrier for building rocSOLVER with Ninja, as some of the C configure checks will fail with hipcc. We do not need the extra ROCm-related flags that the hipcc wrapper would add, so we might as well keep things simple by cutting out the middleman.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -402,9 +402,9 @@ case "${ID}" in
   ;;
 esac
 
-cxx="hipcc"
-cc="clang"
-fc="gfortran"
+export CXX="hipcc"
+export CC="clang"
+export FC="gfortran"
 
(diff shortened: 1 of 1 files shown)


tfalders (approved):
LGTM

jzuniga-amd (approved):

PR #236, opened 2021-04-07
Replace rocsolver_ostream with libfmt

cgmb:
This change adds a dependency on [libfmt 7.1](https://fmt.dev/). By default when using `install.sh`, the parts we need will be statically linked into `rocsolver.so` and hidden using the `--exclude-libs` linker option.

Rather than managing a separate logging thread, we depend on the built-in locks for C and C++ print functions. Both Linux and Windows hold locks while printing using the standard library functions, so no additional synchronization is needed as long as the print is done with a single call. (That is the biggest problem with iostreams. Each use of `<<` is a seperate function call to `operator<<`, and so it's a pain to get fully synchronized output.)

Relying on the standard library for synchronization greatly simplifies the code, eliminating a lot of platform-specific logic related to IO and thread management.


clients/gtest/rocsolver_gtest_main.cpp
```diff
@@ -5,6 +5,8 @@
 #include "clientcommon.hpp"
 #include "internal/rocblas-version.h"
 #include "rocsolver-version.h"
+#include <cstdio>
+#include <fmt/core.h>
```
doctorcolinsmith, line 9: Super low priority, but as a best practice it is best to list standard library headers before third-party headers to minimize risk of symbol conflicts.

clients/gtest/rocsolver_gtest_main.cpp
```diff
@@ -14,17 +16,12 @@
 static void print_version_info()
 {
     // clang-format off
```
doctorcolinsmith, line 18: As a sanity check, do you still need the `clang-format off` after this change?

clients/gtest/rocsolver_gtest_main.cpp
cgmb: We aren't always doing that, but we should. It's also better for pre-compiled headers.

I actually made that fix globally on the headers I was changing as I was working on this a week or two ago, but I undid it for this initial draft just to keep the diff minimal when introducing the library to folks. Thanks for pointing that out. I'll be sure to include that change in the final version.

clients/gtest/rocsolver_gtest_main.cpp
cgmb: It is not needed. Removed.

clients/gtest/rocsolver_gtest_main.cpp
cgmb: I did that for a few files, but I think a separate PR might be the way to fix that globally.

cgmb:
I've squashed the changes, as merge conflicts were annoying to handle with my messy commits.

For more information about glibc's locking guarantees for stdio, see the [flockfile manual page](https://man7.org/linux/man-pages/man3/flockfile.3.html). In general, as long as you're making a single call to print, you will get a single output. It's worth noting that we never could make all output go through rocsolver_ostream even before. gtest, for example, was not using it when it printed its messages. Nor would standard library functions, like `assert`. We've always depended on a reasonable stdio implementation for the occasional message, even when using rocsolver_ostream for most of our output.

libfmt tends to have better performance and smaller code than iostreams, but there is one aspect of performance that we are giving up by dropping the threading: the waiting on IO is now done in the main thread. The use of a background logging thread is common in high-performance logging libraries and it is something that we would be giving up with this change. My recommendation would be that if we want to get that feature back, we should strongly consider relying on a mature library like [spdlog](https://github.com/gabime/spdlog),  rather than rolling our own.

common/include/common_host_helpers.hpp
```diff
@@ -30,45 +32,30 @@ double get_time_us_no_sync();
...
-inline void print_pairs(rocsolver_ostream& os, const char* sep)
-{
-    // do nothing
-}
 
 /** Set of helpers to print out data hosted in the CPU and/or the GPU **/
 /***********************************************************************/
 
-/*! \brief Print provided data into specified stream (real case)*/
+// TODO: Fixup these helper functions for the transition from rocsolver_ostream to libfmt
```
tfalders, line 54: It's unfortunate that these have all been commented out, but I guess it's not a blocking issue.

library/src/common/rocsolver_logger.cpp
```diff
@@ -128,14 +132,14 @@ rocblas_status rocsolver_log_write_profile(void)
 
     auto logger = rocsolver_logger::_instance;
 
-    // print profile logging results
+    // print and clear profile logging results
```
tfalders, line 135: Why did you add "and clear" here? rocsolver_log_write_profile shouldn't clear the profile results.

library/src/include/rocsolver_logger.hpp
```diff
@@ -328,3 +358,139 @@ class rocsolver_logger
...
+ */
+template <typename T>
+struct formatter<rocsolver_logvalue<T>> : formatter<T>
+{
+    template <typename FormatCtx>
+    auto format(rocsolver_logvalue<T> wrapper, FormatCtx& ctx)
+    {
+        return formatter<T>::format(wrapper.value, ctx);
+    }
+};
```
tfalders, line 374: Personally, I'd prefer to put these formatters in a separate file, but it's not a major issue.

tfalders (approved):
Looks good! Thanks for taking care of this.

library/src/common/rocsolver_logger.cpp
cgmb: Opps. I had done some refactoring on the logger to reduce duplication, but I backed it out to do in a later PR. I must have missed updating the comment when I re-duplicated things.

library/src/include/rocsolver_logger.hpp
cgmb: I think we need to ensure they're included whenever we're using the logger, so keeping them in a single file seemed like the way to go.

common/include/common_host_helpers.hpp
cgmb: I'm hoping to replace them with a more general solution that handles both printing and bounds-checked read-write access, but if you're just using them for debugging, I think you can uncomment the function and pass `std::cerr` as the `os`. The output might get interspersed if you're multi-threading, but I don't think that's the usual case when we're debugging.

I could actually just uncomment them and just state that limitation in the todo, instead.

cgmb: Nevermind. I'll just fix them properly. It's not that hard. I was just tired when I got to that point.

library/src/include/rocsolver_logger.hpp
cgmb: After some thought, I'm actually feeling a bit less comfortable with including specializations for public types directly in the library code. I don't think that specialization can escape, but the logger already uses a wrapper, so maybe I'll specialize the wrapped types instead, just to be on the safe side. Users can never directly refer to a `rocsolver_logvalue<T>`, so even if that specialization somehow is visible to library users, they can't have declared their own anyway.

cgmb: I've updated them all to be based on `rocsolver_logvalue`. If you'd like, I could create a new file for the `rocsolver_logvalue` and have the `rocsolver_logger` include that. Or, I could just leave it as-is. I'm fine either way. Do you have a preference?

tfalders: A separate file would be nice, but it's no biggie.

common/include/common_host_helpers.hpp
```diff
@@ -30,99 +37,88 @@ double get_time_us_no_sync();
...
-    os << std::endl;
+    s += '\n';
+    os << s;
+    os.flush();
 }
 
 /*! \brief Print data from a normal or strided_batched array on the GPU to screen*/
 template <typename T>
-void print_device_matrix(rocsolver_ostream& os,
+void print_device_matrix(std::ostream& os,
```
jzuniga-amd, line 121: So how can we use these helpers?
Passing std::cout as ostream here? 
I think you may need to `#include <iostream>` in this file so that we can call 
`print_device_matrix<T>(std::cout,...)` from other places in the library?

cgmb: > Passing std::cout as ostream here?

Yes. Or `std::cerr`.

> I think you may need to #include <iostream> in this file so that we can call print_device_matrix<T>(std::cout,...) from other places in the library?

iostream is a big header, and it would typically be better for compile times if the places in the library that want to use it include it themselves (to avoid having it included in places that don't need it), but... given that fstream is already included, maybe the cost/benefit for iostream favours including it for convenience...

cgmb:
CI is short on gfx900 nodes, so it hasn't finished building all jobs, but it appears to have worked as expected.

The header-only version of the library is slower to compile, but is simpler for the Linux and Windows shared library builds. We can optimize build times later (and the impact is rather modest anyway). I manually verified that the symbol visibility.

I think that approach also sufficiently addresses the needs of the static library build. libfmt follows good symbol versioning practices, so I don't think there will be conflicts. That said, I still have a lot to learn about the best practices for designing static libraries, so it's possible there's a potential problem I missed. This is the best I can do with the knowledge I have, so we'll just have to see if any tickets come back during integration testing.

I think this is ready to merge.

jzuniga-amd (approved):
LGTM. Good job Cory!!

PR #238, opened 2021-04-11
Remove unnecessary ROCSOLVER_COMMON namespace

cgmb:
Symbols marked as hidden can be repeated in another library / executable, and both sets of symbols will be considered distinct from each other. This namespace is therefore unnecessary.

See the algorithm for symbol resolution: https://www.airs.com/blog/archives/49

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -27,10 +27,6 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS )
   )
   prepend_path( "${CMAKE_CURRENT_SOURCE_DIR}/" common_source_files common_source_paths )
   target_sources( clients-common INTERFACE ${common_source_paths} )
-  target_compile_definitions( clients-common
-    INTERFACE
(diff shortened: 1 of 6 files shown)


tfalders (approved):
LGTM!

jzuniga-amd (approved):

cgmb:
The different namespaces were completely unnecessary for the shared library, but we need rocsolver prefixes on _all_ symbols for the static library. This PR removed the rocsolver prefix from the symbols in rocsolver-common. Whoops.

The fact that the symbols would have the same names in both the library and the clients is not a problem. Though, the backwards-compatibility promise that the static library provides requires that users follow one rule: never create symbols with the 'rocsolver' or 'rocblas' prefix. That's reserved for the rocsolver library. For the rocsolver-clients to benefit from the same backwards-compatibility guarantee when static linking, they would have to follow the same rule as user programs. So, well, while having a different namespace name isn't strictly necessary even for the static library, maybe it actually is a good idea.

This is a long way of saying that I'm basically going to revert this PR.

PR #239, opened 2021-04-12
add gfx1030 to target list

jzuniga-amd:
This is the only real difference in gfx1030 branch. The version string is not applicable, and the workaround for the compiler bug is not needed anymore (the bug was fixed by compiler team).


cgmb (approved):

tfalders (approved):
LGTM!

PR #240, opened 2021-04-13
Add gfx90a to AMDGPU_TARGETS if supported

cgmb:
The gfx90a target is not supported on all compilers, so we only add it
if compiler support is detected. Compiler support detection will be in
rocm-cmake for rocm-4.3, so we only try to add them if that is
available.

Once we start development for rocm-4.4, we can require rocm-cmake from
rocm-4.3 and drop the OPTIONAL / if(COMMAND) checks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ include( ROCMCreatePackage )
 include( ROCMInstallTargets )
 include( ROCMPackageConfigHelpers )
 include( ROCMInstallSymlinks )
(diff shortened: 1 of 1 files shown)


tfalders (approved):

jzuniga-amd (approved):

cgmb:
By the way, the failing tests were because I accidentally tested some CI configuration changes on this PR rather than #236 as I'd intended. All tests passed in http://math-ci.rocm.amd.com/blue/organizations/jenkins/compute-rocm-dkms-no-npi-hipclang%2Fprecheckin%2FrocSOLVER/detail/PR-240/2/

PR #241, opened 2021-04-16
Update to new rocBLAS internal API conventions

cgmb:
rocBLAS has renamed its C++ template functions to reflect that they are
not part of the stable API, and requires defining a macro to avoid
deprecation warnings when using them.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -177,7 +177,7 @@ if(OPTIMAL)
   target_compile_definitions(rocsolver PRIVATE OPTIMAL)
 endif( )
 
-target_compile_definitions( rocsolver PRIVATE ROCM_USE_FLOAT16 )
+target_compile_definitions( rocsolver PRIVATE ROCM_USE_FLOAT16 ROCBLAS_INTERNAL_API )
 
 add_armor_flags( rocsolver "${ARMOR_LEVEL}" )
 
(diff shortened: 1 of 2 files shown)


TorreZuk (approved):
LGTM if it links using my branch.  Is it okay if I merge my PR today if approved and break solver until this PR goes in?   Hoping we can have the builds cascade over the weekend so Monday solver and hipblas work again.

cgmb:
I've manually tested that this builds properly with your branch.

We discussed improving the process so this sort of rocBLAS/rocSOLVER internal breaking change will fail during the rocSOLVER CMake configure step, rather during the rocSOLVER CMake build step, but I didn't get a chance to implement that for this change.

cgmb:
@jzuniga-amd should be the one to give you go ahead. I would like to see your change go in ASAP, but he's the one that would know best.

jzuniga-amd (approved):

tfalders (approved):
Thanks for taking care of this, @cgmb.

TorreZuk:
Okay merged rocBLAS side but I guess you have to wait until it rebuilds in develop to get the new header file, not sure how long that will take.

cgmb:
There were CI changes between when the rocBLAS change was merged and when the rocBLAS package became available, so only some of the rocSOLVER builds succeeded. The debug build passed, as did the Ubuntu pre-checkin. I'm going to go ahead and merge, and this will all resolve itself by Monday.

PR #244, opened 2021-04-19
Remove gfx1030 from target list

cgmb:
Temporarily remove gfx1030 from the target list while some build
failures are investigated, and to reduce the load on the CI.

Thanks to @tfalders for identifying that the build issues were specific to gfx1030.


tfalders (approved):
LGTM!

I've managed to track the failure to the "apply pivots" section of getri_kernel_small. Still figuring out which line(s) cause the problem.

cgmb:
@jzuniga-amd has indicated that there's already a fix available in the compiler.

PR #247, opened 2021-04-23
Revert "Remove unnecessary ROCSOLVER_COMMON namespace (#238)"

cgmb:
This reverts commit 3e1da6cf216bfb2a7cc20df242591e63b7182035.


cgmb:
I opened this PR before testing the changes in order to get the CI going on them ASAP. And, while I feel very confident that these changes would be good for the static build, what I didn't realize is that rocBLAS is now broken in our static build containers (due to problems in `hipcc --version` with the static build).

The revert was not the ideal solution, nor a complete solution for symbol prefixing in the static build. Given that I don't have a rocBLAS build to test it with (and the static CI is disabled for the same reason), I'm just going to close this and we can implement a better solution later.

PR #248, opened 2021-04-26
Enable ccache on the CI

cgmb:
This PR makes use of ccache the default setting for rocSOLVER builds on
the CI. It can be disabled for PRs using the label `ci:no-ccache`.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -7,7 +7,6 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false)
 
     String compiler = 'hipcc'
     String hipClang = ''
-    String sles = platform.jenkinsLabel.contains('sles') ? '/usr/bin/sudo --preserve-env' : ''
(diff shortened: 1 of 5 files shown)


cgmb:
This change will turn it on by default, but it can also be enabled using `ci:ccache` even without this being merged.

saadrahim (approved):

tfalders (approved):
LGTM!

cgmb:
The use of sudo for the SLES build was preventing CMake 3.17 from being picked up in the PATH. While not ideal, that problem didn't lead to any errors. It just meant ccache didn't get used on SLES, and everything built as normal.

However, the use of sudo is not necessary. It appears to have been a workaround for a directory permissions problem that was fixed a long time ago.

cgmb:
@eidenyoshida, could you take a look? I'd like to get one more approval before merging. (The debug build failure is unrelated. It's also a problem on develop.)

eidenyoshida (approved):
Looks good to me

PR #249, opened 2021-04-28
Made gesvd's local templates reusable

tfalders:
gesvd has some local template functions designed to adapt non-batched arrays for batched function calls, similar to how we call a number of batched rocBLAS functions. I will be creating another such function for syevd/heevd, and for the sake of cleanliness and reusability I thought it made more sense to move these into their respective header files.


cgmb (approved):
Not my area of specialty, but looks good to me.

jzuniga-amd (approved):
LGTM! Nice...

PR #252, opened 2021-05-07
Updated memory model documentation

tfalders:
Addresses SWDEV-283307

### DIFF
--- a/docs/source/userguide_memory.rst
+++ b/docs/source/userguide_memory.rst
@@ -54,7 +54,7 @@ Minimum required size
 In order to choose an appropriate amount of memory to allocate, rocSOLVER can be queried to determine
 the minimum amount of memory required for functions to complete. The query can be started by calling
 ``rocblas_start_device_memory_size_query``, followed by calls to the desired functions with appropriate
-problem sizes (``nullptr`` can be passed to the device pointer arguments). A final call to
+problem sizes (a null pointer can be passed to the device pointer arguments). A final call to
 ``rocblas_stop_device_memory_size_query`` will return the minimum required size.
 
 For example, the following code snippet will return the memory size required to solve a 1024*1024 linear
--- a/docs/source/userguide_memory.rst
+++ b/docs/source/userguide_memory.rst
@@ -1,12 +1,119 @@
 .. _memory_label:
 
 *******************************
-Memory model
+Memory Model
 *******************************
 
 .. toctree::
    :maxdepth: 4
    :caption: Contents:
 
-More to come later...
+Almost all LAPACK and rocSOLVER routines require workspace memory in order to compute their results.
+In contrast to LAPACK, however, pointers to the workspace are not explicitly passed to rocSOLVER
+functions as arguments; instead, they are managed behind-the-scenes using a configurable device memory
+model.
 
+rocSOLVER makes use of and is integrated with `rocBLAS's memory model <https://rocblas.readthedocs.io/en/latest/device_memory.html>`_.
+Workspace memory, and the scheme used to manage it, is tracked on a per-``rocblas_handle`` basis, and
+the same functionality that is used to manipulate rocBLAS's workspace memory can and will also affect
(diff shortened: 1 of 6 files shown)


docs/source/userguide_memory.rst
```diff
@@ -1,12 +1,119 @@
...
+This allows the user to control when and if memory is reallocated and synchronization occurs. However,
+function calls will fail if there is not enough allocated memory.
+
+Minimum required size
+------------------------------
+
+In order to choose an appropriate amount of memory to allocate, rocSOLVER can be queried to determine
+the minimum amount of memory required for functions to complete. The query can be started by calling
+``rocblas_start_device_memory_size_query``, followed by calls to the desired functions with appropriate
+problem sizes (``nullptr`` can be passed to the device pointer arguments). A final call to
```
cgmb, line 57: I would probably say 'a null pointer' in the prose, as rocsolver provides a C API. IMO, it's fine if the examples are C++ / HIP, but the description of the API should default to C constructs (or generic terms applicable to both C and C++).

docs/source/userguide_memory.rst
```diff
@@ -1,12 +1,119 @@
...
    :maxdepth: 4
    :caption: Contents:
 
-More to come later...
+Almost all LAPACK and rocSOLVER routines require workspace memory in order to compute their results.
+In contrast to LAPACK, however, pointers to the workspace are not explicitly passed to rocSOLVER
+functions as arguments; instead, they are managed behind-the-scenes using a configurable device memory
+model.
 
+rocSOLVER makes use of and is integrated with `rocBLAS's memory model <https://rocblas.readthedocs.io/en/latest/device_memory.html>`_.
```
cgmb, line 16: It would be nice to send people to the corresponding rocBLAS version rather than just kicking them to the lastest rocblas docs. I thought for a second that [sphinx-version-ref](https://pypi.org/project/sphinx-version-ref/) could do it, but we want the readthedocs version (`rocm-*` or `latest`), not the `conf.py` version...

The actual solution is probably [intersphinx](https://docs.readthedocs.io/en/stable/guides/intersphinx.html), but that seems like a bit of a rabbit hole. Might just be a task to toss in our backlog.

cgmb (commented):
Looks good. Just a few minor comments. I enabled the tests so that this would build:
http://math-ci.rocm.amd.com/job/compute-rocm-dkms-no-npi-hipclang/job/Static%20Analysis/job/rocSOLVER/job/PR-252/Documentation/

docs/source/userguide_memory.rst
tfalders: I don't actually know if rocBLAS has docs for different versions. When I go to the readthedocs page for rocSOLVER, I can see different versions, but I don't see the same thing for rocBLAS.

cgmb: Huh. You're right. rocBLAS should have versioned docs, but it seems that they just got the docs build working again a month ago, after being broken for quite a while. I suppose it might be a while until my comment is relevant.

cgmb (approved):

tfalders:
> Looks good. Just a few minor comments. I enabled the tests so that this would build:
> http://math-ci.rocm.amd.com/job/compute-rocm-dkms-no-npi-hipclang/job/Static%20Analysis/job/rocSOLVER/job/PR-252/Documentation/

Do you think it would be possible to add a label to build the static analysis job only? Seems like a waste to do the full compilation and test suite when it's only a documentation change.

cgmb:
> Do you think it would be possible to add a label to build the static analysis job only? Seems like a waste to do the full compilation and test suite when it's only a documentation change.

I agree. It would be a good item for the CI backlog.

That said, if there's no actual changes to the code, the build is now basically a no-op due to ccache. The slowest part of the compile stage is downloading rocblas. The precheckin tests are fast enough that I don't worry about them too much, but the extended tests are expensive and unnecessary.

jzuniga-amd (approved):
LGTM

PR #253, opened 2021-05-18
Add STEDC (initial implementation)

jzuniga-amd:
Divide and conquer will be implemented in 4 phases:
1. Adapt STEQR to work with small independent blocks
2. Adpat GEMM to update eigenvector matrices
3. Prepare divide & conquer (split blocks, deflation.)
4. Implement secular equation and merge steps (to be done in a future PR).


library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -0,0 +1,436 @@
...
+
+/****************************************************************************
+  TODO:THIS IS BASIC IMPLEMENTATION. THE ONLY PARALLELISM INTRODUCED HERE IS
+  FOR THE BATCHED VERSIONS (A DIFFERENT THREAD WORKS ON EACH INSTANCE OF THE
+  BATCH). MORE PARALLELISM CAN BE INTRODUCED IN THE FUTURE IN AT LEAST TWO
+  WAYS:
+  1. the split diagonal blocks can be worked in parallel as they are
+  independent
+  2. for each block, multiple threads can accelerate some of the reductions
+  and vector operations of the DC algorithm (defaltion, solve and merge processes).
```
tfalders, line 26: Minor comment: spelling on deflation

tfalders (approved):
Nice job! LGTM.

The use of shiftT, ldt, and strideT with different arrays in the local_gemm functions feels a bit wrong to me, but I won't make it an issue.

clients/include/testing_stedc.hpp
```diff
@@ -0,0 +1,447 @@
...
+                    Th& hC,
+                    Th& hCRes,
+                    Uh& hInfo,
+                    Uh& hInfoRes,
+                    double* max_err)
+{
+    constexpr bool COMPLEX = is_complex<T>;
+    using S = decltype(std::real(T{}));
+
+    int lgn = floor(log(n - 1) / log(2)) + 1;
```
cgmb, line 153: As of C++20, the standard library includes a function for this called [`bit_ceil`](https://en.cppreference.com/w/cpp/numeric/bit_ceil). That's too new for us to use, but it would make a good name if we ever pull this out into our own utility function.

cgmb (approved):

library/src/auxiliary/rocauxiliary_stedc.hpp
jzuniga-amd: corrected

PR #254, opened 2021-05-18
Use new trsv substitution template

tfalders:
Updates rocSOLVER to reflect changes in https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/664.


library/src/include/rocblas.hpp
```diff
@@ -1449,49 +1449,42 @@ rocblas_status rocblasCall_symm_hemm(rocblas_handle handle,
...
+                                rocblas_int offset_A,
+                                rocblas_int lda,
+                                rocblas_stride stride_A,
+                                U B,
+                                rocblas_int offset_B,
+                                rocblas_int ldb,
+                                rocblas_stride stride_B,
+                                rocblas_int batch_count,
+                                rocblas_int* g_unique_row,
+                                rocblas_int* g_completed_sec,
```
cgmb, line 1472: That's an unfortunate name. `g_*` is a naming convention often used for global variables. Oh well. I suppose there's not much we can do given that's what it's called.

cgmb (approved):
Should wait for the tests, but LGTM. 👍

library/src/include/rocblas.hpp
TorreZuk: This is your template you should be able to call your variables whatever you want.   Ask @daineAMD and he could rename the rocBLAS side too, surprised I didn't notice it there as not sure what the g_ stands for if not global.

daineAMD: global as opposed to shared memory, but I'll make a change as I understand the mistake.

jzuniga-amd (approved):
LGTM!

cgmb:
I think we can merge once the tests finish. The name can be updated later if rocblas changes it to something else.

> A variable by any other name would be just as sweet, if it compiled.

~ William ShakeCPPeare

PR #255, opened 2021-05-20
Added syevd

tfalders:
Addresses SWDEV-275400.

I'm not seeing any performance improvement over syev/heev at this time, possibly because the optimization of stedc is not yet complete.

### DIFF
--- a/library/src/lapack/roclapack_syevd_heevd.hpp
+++ b/library/src/lapack/roclapack_syevd_heevd.hpp
@@ -42,7 +42,7 @@ void rocsolver_syevd_heevd_getMemorySize(const rocblas_evect evect,
 
     size_t unused;
     size_t w11 = 0, w12 = 0, w13 = 0;
-    size_t w21 = 0, w22 = 0;
+    size_t w21 = 0, w22 = 0, w23 = 0;
     size_t w31 = 0, w32 = 0;
     size_t t1 = 0, t2 = 0;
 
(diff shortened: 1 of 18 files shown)


library/src/lapack/roclapack_syevd_heevd.hpp
```diff
@@ -0,0 +1,178 @@
...
+    size_t t1 = 0, t2 = 0;
+
+    // requirements for tridiagonalization (sytrd/hetrd)
+    rocsolver_sytrd_hetrd_getMemorySize<T, BATCHED>(n, batch_count, size_scalars, &w11, &w21, &t1,
+                                                    &unused);
+
+    if(evect == rocblas_evect_original)
+    {
+        // extra requirements for computing eigenvalues and vectors (stedc)
+        rocsolver_stedc_getMemorySize<BATCHED, T, S>(evect, n, batch_count, &w12, &w21, &w31,
```
jzuniga-amd, line 56: w21 here is overwriting the w21 from the sytrd_hetrd call!
should not it be another variable and then take the maximum ?

jzuniga-amd (commented):
Looks good. 
There is just a minor error that I think must be corrected

jzuniga-amd:
> Addresses SWDEV-275400.
> 
> I'm not seeing any performance improvement over syev/heev at this time, possibly because the optimization of stedc is not yet complete.

The more independent blocks that we can split, the more noticeable the improvements in the performance of stedc over steqr (and thus syevd over syev).  We are not introducing any splits in the input matrix when testing syevd or syev (as we do for stedc/steqr), so both routines are essentially the same. 

Once the DC method is completed, the performance of syevd could improve again if the matrix has defective (or repeated) eigenvalues (ie. with algebraic multiplicity > 1), so that deflation can happen. At that point, we can think about adding splits and deflation to the tested matrix, so we can verify better performance, but for now, let's leave it as is.

library/src/lapack/roclapack_syevd_heevd.hpp
tfalders: Oh, thanks for catching that! I'll fix it asap.

jzuniga-amd (approved):
LGTM

cgmb (approved):
LGTM

PR #256, opened 2021-05-25
Added sygvd

tfalders:
Addresses SWDEV-275401.

This is essentially just a copy of sygv/hegv, but with d's added to many of the function calls and some additional workspace management for the vanilla LAPACK calls.

### DIFF
--- a/library/src/lapack/roclapack_sygvd_hegvd.hpp
+++ b/library/src/lapack/roclapack_sygvd_hegvd.hpp
@@ -1,0 +1,194 @@
+/************************************************************************
+ * Derived from the BSD3-licensed
+ * LAPACK routine (version 3.7.0) --
+ *     Univ. of Tennessee, Univ. of California Berkeley,
+ *     Univ. of Colorado Denver and NAG Ltd..
+ *     December 2016
+ * Copyright (c) 2021 Advanced Micro Devices, Inc.
+ * ***********************************************************************/
+
+#pragma once
+
+#include "roclapack_sygv_hegv.hpp"
+
+template <bool BATCHED, typename T, typename S>
+void rocsolver_sygvd_hegvd_getMemorySize(const rocblas_eform itype,
+                                         const rocblas_evect evect,
+                                         const rocblas_fill uplo,
+                                         const rocblas_int n,
+                                         const rocblas_int batch_count,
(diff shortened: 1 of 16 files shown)


library/src/lapack/roclapack_sygvd_hegvd.hpp
```diff
@@ -0,0 +1,192 @@
...
+ *     Univ. of Tennessee, Univ. of California Berkeley,
+ *     Univ. of Colorado Denver and NAG Ltd..
+ *     December 2016
+ * Copyright (c) 2021 Advanced Micro Devices, Inc.
+ * ***********************************************************************/
+
+#pragma once
+
+#include "roclapack_syevd_heevd.hpp"
+#include "roclapack_sygv_hegv.hpp"
```
jzuniga-amd, line 13: I would prefer to see all the includes list here (i.e. 
rocblas.hpp
rocsolver.h
roclapack_potrf.hpp
roclapack_sygst_hegst.hpp
roclapack_syevd_heevd.hpp
roclapack_sygv_hegv.hpp)
because to me, this tells me about the dependencies that this function has with other functions, etc.
It is also consistent with the hpp files of all other functions.

(just personal preference, not a big deal; can be left as is).

jzuniga-amd (approved):
LGTM

library/src/lapack/roclapack_sygvd_hegvd.hpp
tfalders: Sure. I'll do the same for syevd/heevd since I think I trimmed that include list too.

cgmb: I would second that. It's good practice to explicitly include any header you use functionality from. It tends to reduce unnecessary compilation failures during refactoring.

There are only two times (that I can think of) when it doesn't help to add an explicit include. (1) When another header is fundamentally required to be included by the one you used (e.g., the API provided in the header you included uses a type from the other header in a way that cannot be forward-declared), or (2) when it's explicitly promised that one header will include the other (e.g., `rocsolver.h` promises to include `rocblas.h`). But even then, you can always err on the side of explicitly including the transiently-included header and it will never do harm.

cgmb (approved):
LGTM

PR #257, opened 2021-05-26
Documentation revision 1/3

jzuniga-amd:
The main documentation revision will consist of 3 different PRs. 
1. Re-organization of documentation tree and current cross-references.
2. Revision of current documentation contents.
3. Revision of API descriptions (rocsolver-functions.h)
 
On this first PR:
- I re-organized the documentation structure and tree so that information can be found easier on the sidebar, and using the `previous` and `next`  buttons and cross-references. (Also I think that now is easier for developers to find information and add new information in the right files, according to the tree).
- I separated the API description into its own document. So now the documentation has 3 main documents: the user guide, the library design guide (for advanced users and developers), and the rocSOLVER API.
- Contents of all the different sections remain mostly unchanged. I will review them in the next PR. (So, don't focus too much on the contents but rather on the document structure and the pertinence of the different links and cross-references).


cgmb:
I really like the sidebar this adds, but the reorganization comes with major drawbacks.

1) It requires a lot more clicking to get to the relevant information. The hierarchy provides useful categories, but you now need to drill down a few layers before you get to the actual API details. It's particularly bad because you can't see the lower levels of the hierarchy until you open them, requiring the user to go through multiple rounds where they scan the page for the thing that appears most related to what they're looking for -> interact with a promising-looking option -> wait for page update -> scan new (or changed) page.
2) CTRL + F was great for finding things in the API docs, but it's a lot less useful when the API functions are spread across many pages. It was very easy to answer questions like, "what the heck is rocsolver_cheevd?" The hierarchy helps with queries that go from concept to function, but splitting it across pages hinders queries that go from function to concept.
3) Moving pages around breaks existing links to our 'latest' documentation. It's reasonable to do that on occasion, but it comes at a significant cost. It reduces the quality of third-party materials written about rocSOLVER, and will tend to lower the ranking of our documentation on Google and other search engines. The existence of our versioned documentation somewhat mitigates that effect, but not completely.

There's some other minor things, like "User's guide" should probably just be "User Guide" ¹ and the PDF version no longer has a useful Table of Contents, but issues 1 and 2 seem fundamental.

One thing I hadn't appreciated is how good the existing headings are. You've made them much more prominent and that aspect of this change is great. I wonder if we could still do that without such significant changes to the overall page structure? It would be awesome if we could somehow to get the sidebar to include links to headings within a page, and to add a table of contents to the top of each page like Wikipedia does (e.g. in [this article](https://en.wikipedia.org/wiki/Sally_Ride)).

¹ Don't ask my why it's an Owner's Manual but a User Guide.

jzuniga-amd:
- The integration of SPHINX and LATEX is not great, but I don't care much about the PDF as the main output (the one linked to the official AMD documentation) is the HTML -readtehdocs-. Anyways, I numbered the 3 main documents so that they appear as chapters in the PDF and this way I generated a main table of contents for the document.

- I deepened the tree one more level by collapsing the contents of the, now sections of chapters, into single files.  This way we keep the sidebar clean, and I think we find a sweet middle point between the initially proposed flat tree and our current super-deep, single-branch tree. (I don't know how to have links on the sidebar to headings within the same page; I guess you mean to always move them to the top of the page, right? I don't think it is possible, but if you find a solution let me know).

- I am also adding a table of contents for each section as suggested, and a expanded-tree contents at the beginning of each chapter to facilitate search at-a-glance.  I do believe that having the information (especially the API routines) organized by concept/functionality is very important. It not only helps, as you mentioned, with queries that go from concept to function, but it also shows care and a deeper understanding of the product that we offer; dumping all the information in a single place and having the users scrolling up and down a gigantic page is careless and does not look very clean in my opinion. CTRL+F is certainly useful, but ReadTheDocs has its own integrated search engine that is really powerful and can search throughout the entire tree no matter how complicated it is. It returns the information organized according to the document tree and so it is even more useful with searches going from specific to general.

- In addition, I enabled back-links from headings to the table of contents at the beginning of each section, and I also think that the organized tables in section 1.1.2 are great for finding functions at-a-glance.

- I don't worry too much about breaking links from third parties to our documentation (in case they exist), especially if it is for the sake of improving our documents. The important link is the one in the official AMD documentation, and they simply link to the main html index.

tfalders (approved):
Looks good! I looked through the entire documentation since I wanted to get a feel for the whole thing, and I did a spelling/grammar/consistency review while I did. I opened a PR in your fork with my suggested changes.

The only issue I have is that the table of contents for each page tends to look a little cluttered, since the headers for different levels are all lumped together in one long list. Is there a way to improve how they look?

jzuniga-amd:
I added the local table of contents as suggested by Cory, and in an effort to mitigate the merging of independent files into a big single file per section. Unfortunately, there are not many options to customize these local tables. What I am doing now is to place them one more level down in the tree (at the subsection level). It looks a bit cleaner. Let me know what you think.

tfalders:
> I added the local table of contents as suggested by Cory, and in an effort to mitigate the merging of independent files into a big single file per section. Unfortunately, there are not many options to customize these local tables. What I am doing now is to place them one more level down in the tree (at the subsection level). It looks a bit cleaner. Let me know what you think.

Oh, that looks much better. It's a shame we can't keep them all together at the top, but I'm happy with what you have now.

docs/source/api_auxiliaryfunc.rst
```diff
@@ -3,19 +3,27 @@
 LAPACK Auxiliary Functions
 ***************************
 
-These are functions that support more advanced LAPACK routines.
+These are functions that support more :ref:`advanced LAPACK routines <lapackfunc>`. 
+The auxiliaty functions are divided into the following categories:
```
tfalders, line 7: Minor comment: spelling error on auxiliary

jzuniga-amd: corrected

cgmb (approved):
Ok. Those seem like improvements. Let's roll with it.

PR #258, opened 2021-05-26
Add address sanitizer build option

eidenyoshida:


CMakeLists.txt
```diff
@@ -132,6 +133,12 @@ if( BUILD_LIBRARY )
   add_subdirectory( library )
 endif( )
 
+if(BUILD_ADDRESS_SANITIZER)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -shared-libasan")
```
cgmb, line 137: Should we be adding `-fno-omit-frame-pointer`?

CMakeLists.txt
```diff
@@ -132,6 +133,12 @@ if( BUILD_LIBRARY )
...
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address -shared-libasan")
+  add_link_options(-fuse-ld=lld)
```
cgmb, line 139: Why do we need to link using `lld`? Also, if users want to use a copy of the rocsolver library that was built with address sanitizer enabled, do they need to use lld when linking it into their program?

cgmb (commented):
Do we need to run with `LD_PRELOAD=$(clang -print-file-name=libclang_rt.asan-x86_64.so)` set? If so, should that be documented somewhere?

cgmb:
By the way, I originally tried adding enabling AddressSanitizer in #133, but I discovered that HIP did not support it at the time. I'm thrilled to have this tool. It is by far the best way of finding out-of-bounds memory accesses in existing code.

Chandler Carruth explains how to use AddressSanitizer in [The Care and Feeding of C++'s Dragons](https://youtu.be/JSjoCisIHcM?t=3338) around 56:40 in his excellent talk at Going Native 2013. (Incidentally, he has a number of other talks on writing efficient C++ which you can also find on YouTube, and they are pretty much all fantastic.)

CMakeLists.txt
eidenyoshida: This PR is in response to SWDEV-275532 in which we are asked to add the following flags to all math libs as a build option.
CFLAGS="-fsanitize=address -shared-libasan" CXXFLAGS="-fsanitize=address -shared-libasan" LDFLAGS="-fuse-ld-lld"

We can add further options such as -fno-omit-frame-pointer if that will assist in development and debugging of our libraries. I am not sure if we want it in the same build option or if we want to add additional build options

cgmb: Ok. One last question. These look like options that would typically be used for building a shared library. Does the static library build need to be handled differently?

jzuniga-amd (approved):
LGTM

tfalders (approved):
LGTM!

cgmb (approved):
There's still questions to answer and improvements that can be made, but even as-is, I'd rather have this change than not have it.

Note that address sanitizer conflicts with the `-D_FORTIFY_SOURCE` option enabled in the rocsolver debug build, so this only works in release mode. (Unless `-DARMOR_LEVEL=0` is specified to disable fortify source.)

eidenyoshida:
If it's ok, I'll merge as-is to give time for devops to test before 4.4 feature complete

cgmb:
I went to test the changes for the debug build and I thought of one more thing. AddressSanitizer is part of a suite of tools. The main tools are [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html), [ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html), [MemorySanitizer](https://clang.llvm.org/docs/MemorySanitizer.html), and [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html). Each is used for finding a different type of programming error.

I don't think you can generally combine the sanitizers, so each one would probably get its own build (eventually). Given that, do we want to use a flag more specific than `--sanitizer`?

eidenyoshida:
Would --address_sanitizer work?

cgmb:
How about `--address-sanitizer`, so I don't have to press Shift? :p
At least in rocSOLVER, we currently only have `_` in options for directories.

If that sounds good, I'll include that change that in the PR for conditionally disabling `_FORTIFY_SOURCE`.

eidenyoshida:
Sounds good to me, less shift pressing the better

cgmb:
The fixes for debug mode are now in, and the CI has passed. I think we can squash-and-merge any time.

cgmb:
Maybe we should have taken a target-based approach for this after-all. The flags are set after `add_library`, so this is only actually enabling it on the clients.

PR #259, opened 2021-05-29
Use standard library timer functions

cgmb:
This PR replaces [`clock_gettime`](https://linux.die.net/man/3/clock_gettime) with [`std::chrono::steady_clock`](https://en.cppreference.com/w/cpp/chrono/steady_clock) for portability. When running the rocsolver-bench, I'm seeing a modest change of ~10% in the reported time for the CPU implementation, but the GPU time remains unchanged.

As far as I can tell, the change in CPU time is an actual change in the time taken to perform the benchmarked operation, and not a result of measurement error. The CPU benchmaking time also varies when I make unrelated changes, so I suspect that our benchmarking of the CPU implementation is just not as robust as the GPU benchmarking.

### DIFF
--- a/clients/rocblascommon/utility.cpp
+++ b/clients/rocblascommon/utility.cpp
@@ -8,7 +8,6 @@
 #include <cstdlib>
 #include <cstring>
 //#include <regex>
-#include <sys/time.h>
 
 // Random number generator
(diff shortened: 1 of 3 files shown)


tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #260, opened 2021-06-01
Use -h to print install.sh help message

tfalders:
install.sh is very unusual in that it uses the -h flag to specify that the library should be compiled using hip-clang, instead of the standard behaviour of printing the help message. Since hip-clang is now the default compiler and the -h flag currently does nothing, I think we should change it to fit the standard behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,14 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
     - HEGVD (with batched and strided\_batched versions)
 
 ### Optimizations
+
 ### Changed
+- The -h option of install.sh now prints a help message, instead of doing nothing.
+
 ### Removed
+
 ### Fixed
+
 ### Known Issues
 
 
--- a/install.sh
+++ b/install.sh
@@ -14,7 +14,7 @@ Usage:
   $0 <options> (modify default behavior according to the following flags)
 
 Options:
-  --help                      Print this help message.
+  -h | --help                 Print this help message.
 
   --build_dir <builddir>      Specify path to the configure & build process output directory.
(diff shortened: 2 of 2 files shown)


cgmb (changes requested):
There's a reasonable chance that `-h` is still being used in old scripts, and this change will break them.

That's not the worst thing in the world. Compatibility between versions for the build script is just best-effort. We don't have hard compatibility requirements like with the actual code. And the broken scripts will be printing out a help message, so that's a bonus. This would probably help new users more than hinder old ones, so I agree with the change in principle. However, the docs need to updated to match the new behaviour.

tfalders:
> This would probably help new users more than hinder old ones, so I agree with the change in principle. However, the docs need to be updated to match the new behaviour.

One of the changes I made to https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/257 was to remove examples of `install.sh -h` from the build and install guide. I believe I got all the references to -h, but I can double check later to be sure.

cgmb:
Sounds good. I should take another look at that PR.

jzuniga-amd:
I wanted to make this change a long time ago, but at that time CI (CQA and devops) were still using the -h flag to build. If this is not the case anymore (either they are using --hip-clang or nothing), then I see no problems. Maybe @eidenyoshida can confirm?

eidenyoshida:
The rocm-ci job uses the command below
/install.sh -ricd --hip-clang --install_dir ${ROCM_PATH} --rocblas_dir ${ROCM_PATH}/rocblas | tee -a rocSOLVER.log 

Doesn't seem like there's a -h flag being used

jzuniga-amd (approved):

cgmb (approved):
👍

PR #262, opened 2021-06-03
fix getri-out-of-place tests

jzuniga-amd:
As per discussion in rocSOLVER's teams channel. 
(I tested it with the `compute-rocm-rel-4.2:21-STG1` container on a gfx900 machine and it worked)

### DIFF
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -13,8 +13,8 @@ find_package( GTest REQUIRED )
 
 set(roclapack_test_source
     # linear systems solvers
-    getrs_gtest.cpp
     getri_gtest.cpp
+    getrs_gtest.cpp
     trtri_gtest.cpp
     # least squares solvers
     gels_gtest.cpp


eidenyoshida:
I'll trigger a manual run on gfx90a for this PR

eidenyoshida:
Looks like I'll need to rerun rocBLAS. I'm assuming some breaking changes for rocBLAS/rocSOLVER came in yesterday or today?

jzuniga-amd:
Yes, rocblas recently changed trmv interface

tfalders (approved):
LGTM

getri_outofplace calls getrs, which has had similar issues. At some point, we should probably hunt down the source of this bug.

cgmb (approved):
This is more hiding the problem than fixing it, but it's ok for now. I'd agree with Troy that we should be circling back on this one at some point to dig a little deeper into the cause.

jzuniga-amd:
That's exactly what I said in the rocSOLVER channel. Just a workaround to buy us some time...

Merging this once CI passes. @eidenyoshida can you restart the test once we have a newer rocBLAS artifact?

jzuniga-amd:
CI still running on gfx900, but this is what I tested locally. CI gfx90a has passed. 
Merging now.

PR #265, opened 2021-06-04
Workaround for debug binary size on CI

cgmb:
Building for all architectures results in a binary that is too large
(>2GB) and will fail to link. However, the debug build is only tested
for a couple architectures. For now, limit the build to those tested
architectures, until we can find a general solution for the binary
size problem.

### DIFF
--- a/.jenkins/debug.groovy
+++ b/.jenkins/debug.groovy
@@ -20,7 +20,7 @@ def runCI =
     prj.defaults.ccache = true
 
     // customize for project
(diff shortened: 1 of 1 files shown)


tfalders (approved):

saadrahim (approved):

jzuniga-amd (approved):

cgmb:
Debug tests failed due to the same problem that #262 fixes, but it compiled correctly.

PR #266, opened 2021-06-07
Update client docs

tfalders:
Addresses SWDEV-283308.

### DIFF
--- a/docs/source/userguide_clients.rst
+++ b/docs/source/userguide_clients.rst
@@ -28,10 +28,10 @@ section <userguide_install>` of the User Guide for more information on building 
 Testing rocSOLVER
 ==========================
 
-``rocsolver-test`` executes a suite of `Google tests <https://github.com/google/googletest>`_ (*gtest*) that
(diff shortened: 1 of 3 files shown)


cgmb:
Not sure if this is in-scope, but it would be nice to include some example outputs (particularly for the logging).

docs/source/userguide_clients.rst
```diff
@@ -33,13 +42,31 @@ Calling the rocSOLVER gtest client with the --help flag
 
 returns information on different flags that control the behavior of the gtests.
 
+One of the most useful flags is the --gtest_filter flag, which allows the user to choose which tests to run
```
cgmb, line 45: Actually, I suppose the same is true here. That should be:
```
``--gtest_filter flag``
```
There's a bunch of other places where this problem occurs, too.

For consistency, I would probably just quote all command-line flags in that way, even if they're single-hyphen flags.

cgmb (approved):
LGTM

jzuniga-amd (approved):
LGTM. Just added some minor suggested changes in a PR to your branch. Have a look.

PR #267, opened 2021-06-07
Add mem-query option to benchmark client

tfalders:
Addresses SWDEV-279898.


clients/include/rocsolver_test.hpp
```diff
@@ -13,19 +13,24 @@
...
+    do                                                                                       \
+    {                                                                                        \
+        if(case == 3)                                                                        \
+            rocsolver_cout << arg << " bytes of device memory are required..." << std::endl; \
+        else if(case == 2)                                                                   \
+            rocsolver_cout << "Invalid value in arguments..." << std::endl;                  \
+        else if(case == 1)                                                                   \
+            rocsolver_cout << "Invalid size arguments..." << std::endl;                      \
+        else                                                                                 \
+            rocsolver_cout << "Quick return..." << std::endl;                                \
```
cgmb, line 26: Could we replace 0,1,2,3 with an enum? Maybe:

Before | After
--- | --- 
 `ROCSOLVER_BENCH_INFORM(0);` | `bench_inform(inform_quick_return);` 
 `ROCSOLVER_BENCH_INFORM(1);` | `bench_inform(inform_invalid_size);` 
 `ROCSOLVER_BENCH_INFORM(2);` | `bench_inform(inform_invalid_value);` 
  `ROCSOLVER_BENCH_INFORM_2(3, size);` | `bench_inform(inform_memory_query, size)`

cgmb: Actually, we could probably even just use separate functions. `bench_inform_quick_return()`, `bench_inform_invalid_size()`, `bench_inform_invalid_value()`, `bench_inform_memory_query(size_t)`.

It's not clear to me why it's handled by a switch at all.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 
 ### Changed
 - The -h option of install.sh now prints a help message, instead of doing nothing.
+- Added --mem_query option to rocsolver-bench, which will print the amount of device memory required
+  by a function.
```
jzuniga-amd, line 20: This is rather an addition... I would put it in the ### Added section.

jzuniga-amd (approved):
LGTM

clients/include/rocsolver_test.hpp
tfalders: There's a couple lines that always get output by `ROCSOLVER_BENCH_INFORM`, so I guess the switch is to avoid duplicating those lines. Personally, I think the first option looks a bit cleaner, so that's what I'll go with.

cgmb (approved):
Awesome. I could never remember what the numbers meant. This is a lot more readable, and adds a new feature too. :+1: 

Wait for the CI to give the all clear, but I approve.

tfalders:
Looks like CI aborted the tests on gfx900. rocsolver-bench isn't really tested anyway, so if there are no objections I think I'll merge the PR.

PR #268, opened 2021-06-09
Add profile option to rocsolver-bench

tfalders:
Addresses SWDEV-283304.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
     - HEGVD (with batched and strided\_batched versions)
 - Added --mem_query option to rocsolver-bench, which will print the amount of device memory required
   by a function.
+- Added --profile option to rocsolver-bench, which will print profile logging results for a function.
 
 ### Optimizations
 
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -94,6 +94,13 @@ try
             "                           Options are: s, d, c, z.\n"
             "                           ")
 
+        ("profile",
+         value<rocblas_int>(),
+            "Print profile logging results for the tested function.\n"
+            "                           The argument specifies the max depth of the nested output.\n"
+            "                           If the argument is unset or <= 0, profile logging is disabled.\n"
+            "                           ")
+
         ("singular",
          value<rocblas_int>(&argus.singular)->default_value(0),
             "Test with degenerate matrices? 0 = No, 1 = Yes\n"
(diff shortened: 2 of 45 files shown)


cgmb:
rocsolver-bench can't run trigger the bad-args cases, can it? When I was debugging `rocsolver_ostream` in the tests, I noticed that those will print a null byte to the log. We should probably change that so that the output is always printable characters, but if the bench client can't encounter that case, we can save it for another time.

tfalders:
> rocsolver-bench can't run trigger the bad-args cases, can it? When I was debugging `rocsolver_ostream` in the tests, I noticed that those will print a null byte to the log. We should probably change that so that the output is always printable characters, but if the bench client can't encounter that case, we can save it for another time.

Yeah, rocsolver-bench should not trigger bad-args type behaviours. We check the validity of enums before passing them to the tests, and we do not use any `nullptr`s in the tests called by rocsolver-bench.

cgmb (approved):

jzuniga-amd (commented):
It looks good, but what is the plan to solve the hangs?
Merge #236 first and then update this PR?

tfalders:
> It looks good, but what is the plan to solve the hangs?
> Merge #236 first and then update this PR?

Yes, I think that PR ought to be merged first. I'll make sure this PR doesn't hang any more before we merge it.

cgmb:
I rebased this branch on the current develop and tried it out. Seems to work. No hangs.
```
root@docker:/src/rocSOLVER# build/release/clients/staging/rocsolver-bench -f getf2_batched -m 30 --lda 75 --batch_count 350 --profile 2
Query device success: there are 1 devices
-------------------------------------------------------------------------------
Device ID 0 : Device 66af
with 17.2 GB memory, max. SCLK 1801 MHz, max. MCLK 1000 MHz, compute capability 9.0
maxGridDimX 2147483647, sharedMemPerBlock 65.5 KB, maxThreadsPerBlock 1024, warpSize 64
-------------------------------------------------------------------------------

============================================
Arguments:
============================================
m               n               lda             strideP         batch_c        
30              30              75              30              350            

============================================
Results:
============================================
cpu_time        gpu_time       
1936            3733.7         

------- PROFILE -------
rocsolver_getf2_template: Calls: 10, Total Time: 37.269 ms (in nested functions: 35.716 ms)
    rocblas_ger_template: Calls: 290, Total Time: 9.431 ms
    rocblas_scal_template: Calls: 300, Total Time: 7.000 ms
    rocblas_iamax_template: Calls: 300, Total Time: 12.062 ms
    rocsolver_laswp_template: Calls: 300, Total Time: 7.223 ms

```

jzuniga-amd (approved):
LGTM

PR #270, opened 2021-06-09
Workaround for static library size on CI

cgmb:
Building for all architectures results in a binary that is too large
to link into our test clients. However, the static build is only
tested for a couple architectures. For now, limit the build to those
tested architectures, until we can find a general solution for the
binary size problem.


cgmb:
This is essentially the same workaround as for the debug build in #265.

eidenyoshida (approved):

tfalders (approved):

jzuniga-amd (approved):

PR #272, opened 2021-06-11
Minor build fixes

cgmb:
Address sanitizer should be enabled before the rocsolver target is
created, so that the flags are enabled when building the library
itself.

fmt::fmt should not be on rocsolver-common. The library and the client
each handle linking it themselves. That snippet was added in error while
resolving a merge conflict.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,12 @@ option( BUILD_CLIENTS_BENCHMARKS "Build rocSOLVER benchmarks" OFF )
 option( BUILD_CLIENTS_SAMPLES "Build rocSOLVER samples" OFF )
 option(BUILD_ADDRESS_SANITIZER "Build with address sanitizer enabled" OFF)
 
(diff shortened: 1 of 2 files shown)


cgmb:
These might not make it in today, and that's probably fine. They won't cause any immediate problems.

tfalders (approved):
LGTM

jzuniga-amd:
> These might not make it in today, and that's probably fine. They won't cause any immediate problems.

Ok good!!
Otherwise, if everyone is happy with it, we can merge it today into master... (better than as a hotfix after the freeze)
It looks good to me.

eidenyoshida (approved):
Still building on my workstation but I think it should be fine to merge

cgmb:
The CI has finished compiling succeeded and the tests are running. This was purely a build change, so I think we can merge.

jzuniga-amd:
OK. Merged into master for 4.4

PR #273, opened 2021-06-14
Reformat CMake code

cgmb:
This change reformats all rocsolver CMake code into a consistent coding
style, with the exception of deps/* (which I hope can be replaced
entirely in the near future).

The rules followed were:
  - Indent with two spaces.
  - 100 character maximum line width.
  - Begin comments with a capital letter.
  - Do not include a space before or after function parentheses.
  - Any number of unrelated function arguments may go on the same line,
    but wrap when they would exceed the maximum line width.
  - Functions taking multi-value arguments are split across multiple
    lines under the following conditions:
    - A multi-valued argument is moved to new line and indented one
      level if there's more than one multi-valued argument being passed
      (e.g.  both a PRIVATE and a PUBLIC category).
    - A multi-valued arument's values are moved to a new line and
      indented if there's more than one value being passed for that
      argument.
  - Functions taking a list of items are treated like those taking
    multi-value arguments, i.e., if there are multiple values being
    passed, the values should each be on a seperate line.
  - Multi-value functions split across multiple lines are closed with
    a parenthesis at the same indentation level as the function name.
    (Unlike functions that were wrapped only due to length, which have
    the closing parenthesis immediately following the last argument.)


cgmb:
This is really a trivial matter, but our current style is both a bit weird and we're rather inconsistent about using it. Having explicit coding guidelines is nice to have (even if they're not strictly enforced). When there are rules about style, you can just shut your brain off and follow them.

We could adjust any of these choices if there's other styles you'd prefer. I picked rules I liked and was familiar with (e.g. from [Blender](https://github.com/blender/blender/blob/master/CMakeLists.txt)), but there's also plenty of projects using 4-space indentation levels or putting the closing parenthesis at the arguments' indentation level.

I tried a couple of auto-formatters, but I wasn't thrilled by any of them. So, these be more what you'd what you'd call guidelines than actual rules.

cmake/util.cmake
```diff
@@ -1,50 +1,45 @@
...
-# ########################################################################
-macro( option_opposite option opposite )
-  if( DEFINED "${opposite}" )
-    if( ${opposite} )
-      set( "${option}" OFF )
+macro(option_opposite option opposite)
+  if(DEFINED "${opposite}")
+    if(${opposite})
+      set("${option}" OFF)
     else( )
```
tfalders, line 41: This one still has a space.

tfalders (approved):
I do love format consistency. Looks good!

tfalders:
There also appears to be space in the root CMakeLists file: `get_imported_target_location( location roc::rocblas)`

jzuniga-amd (approved):

PR #274, opened 2021-06-16
Added potrs

tfalders:
Partially addresses SWDEV-291320.


library/src/lapack/roclapack_potrs.hpp
```diff
@@ -0,0 +1,140 @@
...
+    // constants to use when calling rocblas functions
+    T one = 1; // constant 1 in host
+
+    if(uplo == rocblas_fill_upper)
+    {
+        // solve U'*X = B, overwriting B with X
+        rocblasCall_trsm<BATCHED, T>(handle, rocblas_side_left, uplo,
+                                     rocblas_operation_conjugate_transpose, rocblas_diagonal_non_unit,
+                                     n, nrhs, &one, A, shiftA, lda, strideA, B, shiftB, ldb,
+                                     strideB, batch_count, optim_mem, work1, work2, work3, work4);
```
cgmb, line 115: This is again not specific to your work here in POTRS, but we should probably start considering the exception-safety of our interface with rocBLAS at some point in the near future.

Right now, if rocBLAS throws an exception from `rocblas_internal_trsm_template`, that exception will not be caught, and will propagate up though the rocSOLVER API. (Which, aside from being contrary to our API documentation, will also leave things like the pointer mode in an inconsistent state, because execution will not reach the end of the function where we restore the handle to its original state.)

It may be that `rocblas_internal_trsm_template` cannot throw an exception, so this is not something we have to worry about. However, if that is the case, rocBLAS should make the promise explicit by marking the functions as `noexcept`.

cgmb (approved):
I just had a couple comments about rocSOLVER in general. None of them are specific to POTRS.

library/src/lapack/roclapack_potrs.hpp
jzuniga-amd: rocblas internal template functions do not throw exceptions. If you want to open a PR in rocBLAS adding the `noexcept`, or bring this to the discussion in rocblas meetings, it would be OK.

jzuniga-amd (changes requested):
I open PR in your branch with updated tests according to the fix needed for POSV. Please have a look.

jzuniga-amd (approved):
LGTM

PR #275, opened 2021-06-17
Update to LAPACK v3.9.1

cgmb:
Switched to download tagged versions from the main lapack repository,
as the lapack-release repository has not been updated in several years.
    
There have been improvements in the LAPACK CMake such that special
handling for CBLAS_INCLUDE_DIRS is no longer needed.
    
It should not be necessary to specify lapack or blas if cblas has been
specified, but the lapack targets do not fully specify their library
dependencies. We build lapack with the reference blas, so we specify
that explicitly in the clients. (Explicitly specifying that blas should be
linked is not strictly necessary on Linux, but is required on Windows.)

(The Windows CI is using prebuilt copies of LAPACK 3.9.0.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Added
 ### Optimizations
 ### Changed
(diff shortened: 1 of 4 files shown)


tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #276, opened 2021-06-18
Upgrade to C++17

cgmb:
### Upgrade to C++17

It would be better if we could coordinate with rocBLAS to update the C++ standard version across both libraries, but I don't think it's required. I'm not aware of any issues with calling C++14 code from C++17 code, and it works in practice on both Windows and Linux in my testing. It would be possible for rocBLAS to break rocSOLVER by adding code to their internal headers that is valid in C++14 and invalid in C++17, but there's very few such constructs (and we could always send them a fix that's valid for both).

**(Update)** rocBLAS is now using C++17. Since we use rocBLAS internal headers, we will be forced to upgrade once they start using C++17 features in the headers we include.

### Add example uses of if constexpr

The code changes I made in `common_host_helpers.hpp` and `lib_device_helpers.hpp` were not necessary. I just wanted to use if constexpr in a couple places to show what sort of improvements can be made using C++17 features.

**(Update)** This part of the change has been removed. There's no code changes anymore.

### Suppress HIP warnings

C++17 enables a number of additional static checks that the HIP API uses to warn about places where we are not currently checking for errors. These warnings do not refer to new problems, so I'm just suppressing the warnings for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,10 +4,6 @@
 
 cmake_minimum_required(VERSION 3.13)
 
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 # This has to be initialized before the project() command appears
 # Set the default build type to Release
 if(NOT DEFINED CMAKE_CONFIGURATION_TYPES AND NOT DEFINED CMAKE_BUILD_TYPE)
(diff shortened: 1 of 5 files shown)


cgmb:
Torre made a good point that it's easier to back out the update if we don't actually start relying on any of the new features immediately. That seemed like a good argument to me, so perhaps I should drop the `if constexpr` commit from this PR.

jzuniga-amd (changes requested):
I agree!
Also, the places where you changed to use `if constexpr` are not the only cases in the library where this can be used. So good idea to move this to a different PR (where it can be applied everywhere) once we are sure that C++17  won't be backed out.

jzuniga-amd (approved):

doctorcolinsmith (approved):

PR #277, opened 2021-06-18
Consistent ordering for template arguments

tfalders:
Establishes a consistent ordering for typename S and bool BATCHED template arguments across the library. In summary:

- typename S always follows typename T.
- Test functions and _impl functions that do not have an argument of type S will use `using S = decltype(std::real(T{}));` instead of a template argument.
- bool BATCHED always precedes typename T in `getMemorySize` functions.

### DIFF
--- a/clients/common/lapack_host_reference.cpp
+++ b/clients/common/lapack_host_reference.cpp
@@ -2288,7 +2288,7 @@ void cblas_latrd<double, double>(rocblas_fill uplo,
 }
 
 template <>
-void cblas_latrd<float, rocblas_float_complex>(rocblas_fill uplo,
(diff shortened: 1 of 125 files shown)


cgmb (approved):

doctorcolinsmith:
I think this is OK, but please check that the template changes won't trigger a breaking ABI change on the public symbols.

tfalders:
> I think this is OK, but please check that the template changes won't trigger a breaking ABI change on the public symbols.

I don't think rocSOLVER exports any template methods, so we should be fine.

jzuniga-amd (approved):
LGTM.
Nice! Good to finally have this sorted out library-wide...

tfalders:
Looks like, other than a false failure, CI was passing. I'll resolve the conflicts and then merge the PR.

PR #278, opened 2021-06-22
Build option cleanup

cgmb:
These changes only affect users that build rocSOLVER from source
without using install.sh (such as Linux distro package maintainers).
The install.sh script avoids these cases.

- The default for OPTIMAL in install.sh is ON, but was OFF when
  calling cmake directly. We probably want these to match.
- fmt::fmt doesn't need to be PUBLIC. CMake propagates the linking
  dependency to library users even when set to PRIVATE. The PUBLIC
  specification would only be needed for propagating include
  directories (which is not needed). install.sh always specifies
  ROCSOLVER_EMBED_FMT=ON, which avoids this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Optimizations
 ### Changed
 - Raised reference LAPACK version used for rocSOLVER test and benchmark clients to v3.9.1
+- Minor CMake improvements for users building from source without install.sh:
+    - Removed fmt::fmt from rocsolver's public usage requirements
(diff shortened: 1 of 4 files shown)


jzuniga-amd (approved):
LGTM

tfalders (approved):
LGTM. Personally, I prefer to have the default for optimal to be off, but at least we'll be consistent.

cgmb:
> Personally, I prefer to have the default for optimal to be off, but at least we'll be consistent.

That would be my personal preference as a developer (and even as a light user). However, for a library like rocSOLVER, there's guaranteed to eventually be third-parties doing independent benchmarks, and they might not notice important build options like that one. The maximum performance option should be the default, to ensure we're always putting our best foot forward.

PR #280, opened 2021-06-23
Change to rocblas trsv template

daineAMD:
At some point I will remember to update you all when I make changes to the rocblas API, sorry again.

I haven't tested this so I'll keep an eye on CI.


tfalders (approved):

jzuniga-amd (approved):

jzuniga-amd (commented):
Wait, the changes look ok but there is a failing test that could be related: sygst calls sygs2 who calls trsv

daineAMD:
It looks like a Tensile problem to me, but I don't think we've seen this problem in rocBLAS as far as I know.

cgmb (approved):
If that fixes it, I think we're good.

PR #282, opened 2021-06-29
Initial Windows 10 Support

cgmb:
This PR includes all changes required to start building and testing rocSOLVER on Windows. Most tests pass, though there are still some failures on the Windows CI. However, Linux remains functional, and this will allow us to get useful feedback from the Windows CI on all our branches and PRs. The build changes are not exactly pretty, but we can improve that over time. Most of the Python code comes from rocBLAS, with some minimal adaptation for rocSOLVER.

### How to Build
```bat
:: Prerequisites: Visual Studio 2019, Python 3, vcpkg, HIP, LAPACK with cblas, rocBLAS
call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
set HIP_DIR=C:/hip
set HIP_PATH=C:/hip
set HIP_ROCCLR_HOME=C:/hip
set rocblas_DIR=C:/hipSDK
set cblas_DIR=C:/3rdparty/build
vcpkg install fmt:x64-windows
vcpkg install gtest:x64-windows
set VCPKG_PATH=C:/3rdparty/vcpkg
python rmake.py -ic
```

### How to Run Tests
```bat
cd build\release\clients\staging
python rtest.py -t psdb
python rtest.py -t osdb
```


clients/rocblascommon/rocblas_random.hpp
```diff
@@ -146,7 +146,7 @@ inline rocblas_bfloat16 random_generator<rocblas_bfloat16>()
 template <>
 inline int8_t random_generator<int8_t>()
 {
-    return std::uniform_int_distribution<int8_t>(1, 3)(rocblas_rng);
+    return int8_t(std::uniform_int_distribution<int>(1, 3)(rocblas_rng));
```
TorreZuk, line 149: unsigned short  or short may work as might be slightly better distribution than the low 8 bits.

TorreZuk (approved):
LGTM for first version, hopefully next week can see some CI output and feed into hipBLAS.
May need to evolve the packaging for what Danyi requires from us to build hipBLAS.

clients/rocblascommon/rocblas_random.hpp
cgmb: `std:::uniform_int_distribution<int>(1,3)` should be choosing an appropriate algorithm for the given range. Though, I don't think we use this code anyways. I was considering deleting it.

Incidentally, I think the compilation failure was because `int8_t` is a typedef for `char`, and the C++ standard specifies that `std::uniform_int_distribution` does not support char. Stefan T. Lavavej mentioned that he added the assertion enforcing `T != char / unsigned char` to the msvc standard library, and recommended just using `short` and casting in his talk, [rand() Considered Harmful](https://youtu.be/LDPMpc-ENqY?t=21m8s).

TorreZuk: If you don't use, no worries, I don't know where you were seeing differences but if you change the random inputs it could also lead to differences.  This is rocblas version now:
return static_cast<int8_t>(std::uniform_int_distribution<unsigned short>(1, 3)(t_rocblas_rng));
Seems like the committee screwed up leaving out 8bit ints but the developers added it anyway :-)   Just not the windows devs.

jzuniga-amd (approved):

tfalders (approved):

cgmb:
The Windows CI passes: http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/sandbox%2FrocSOLVER-win-build/detail/rocSOLVER-win-build/23/pipeline/

I'll wait for the Linux CI to pass, then merge.

PR #284, opened 2021-06-30
Add rtest.xml configuration for rtest.py

eidenyoshida:


lawruble13 (approved):
Looks good to me! As mentioned in other threads the `DOCTYPE` declaration as-is will not work for Windows. However, it isn't used for verification yet, so it should be safe to either remove or leave in place. As we progress further with this system, this should become a PUBLIC DTD, which should be safe for Windows and Linux.

cgmb (commented):
This change conflicts with the `rtest.xml` included in #282, which is used on Windows and was modeled on rocblas. It's unclear to me how the conflict should be resolved.

eidenyoshida:
I think this rtest.xml is compatible with changes in #282. If #282 is being merged soon then this PR can wait and any kinks can be worked out

cgmb:
Can I drop the DTD so it works on Windows?

lawruble13:
The DTD isn't used for anything yet, it can safely be dropped.

cgmb (approved):
I need to fix merge conflicts in #282 anyway, so I'm fine with this going in first.

tfalders (approved):
LGTM, though CI seems to be failing for some reason.

jzuniga-amd (approved):
LGTM

cgmb:
> LGTM, though CI seems to be failing for some reason.

I think now even our release mode binary is too large to link when we build for all supported architectures and include the small matrix optimizations. I like seeing green checkmarks, so I'll re-run the tests with noOptimizations.

PR #287, opened 2021-07-08
Added posv

tfalders:
Addresses SWDEV-291330.

### DIFF
--- a/clients/include/testing_posv.hpp
+++ b/clients/include/testing_posv.hpp
@@ -456,9 +456,9 @@ void testing_posv(Arguments& argus)
     }
 
     // validate results for rocsolver-test
-    // using m * machine_precision as tolerance
+    // using 5 * n * machine_precision as tolerance
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, n);
+        ROCSOLVER_TEST_CHECK(T, max_error, 5 * n);
 
     // output results for rocsolver-bench
     if(argus.timing)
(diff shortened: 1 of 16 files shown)


clients/include/testing_posv.hpp
```diff
@@ -0,0 +1,505 @@
...
+    rocblas_stride stB = argus.get<rocblas_stride>("strideB", ldb * nrhs);
+
+    rocblas_fill uplo = char2rocblas_fill(uploC);
+    rocblas_int bc = argus.batch_count;
+    rocblas_int hot_calls = argus.iters;
+
+    rocblas_stride stBRes = (argus.unit_check || argus.norm_check) ? stB : 0;
+
+    // check non-supported values
+    // N/A
```
jzuniga-amd, line 320: You need to verify that uplo != rocblas_fill_full as it is not supported...
This case is tested in badArgs( ) for the gtests, but a user can call rocsolver-bench with uplo=rocblas_fill_full and we must detect it and print corresponding message. See POTF2/POTRF for example. 
(Also, I forgot to add this comment to POTRS. You will need to change it as well)

jzuniga-amd (changes requested):
Just one little change that I forgot to add here and in POTRS

jzuniga-amd (approved):
LGTM

cgmb (approved):
LGTM

tfalders:
Some CI tests aren't passing because of the trsm bug, but others are so I'm going to go ahead and merge.

PR #288, opened 2021-07-08
Drop gfx803 from default build architectures

cgmb:
rocBLAS has removed gfx803 from its default architectures. As rocSOLVER
depends on rocBLAS functionality in many of it's functions, our
supported architectures are a subset of those specified by rocBLAS.

gfx803 is not officially supported by HIP anymore, and dropping the
architecture will reduce the binary size and compile times.

Note that users can still build for gfx803 by explicitly choosing
the architecture via the CMake AMDGPU_TARGETS variable,
or ./install.sh -a gfx803. Community contributions that fix or
improve rocSOLVER on gfx803 remain welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,7 @@ rocm_check_target_ids(OPTIONAL_AMDGPU_TARGETS
 
 # Set this before finding hip so that hip::device has the required arch flags
 # added as usage requirements on its interface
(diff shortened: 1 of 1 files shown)


cgmb:
Sadly, this is insufficient. Our binary size was 1.965 GiB before GESV was merged, and we're now well past the 2 GiB limit.

cgmb:
I've reopened this PR, as https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/e5eb1cfe4a6da5f70b59edf5fff2a2ddf42fa6d7 has made it relevant again.

tfalders (approved):

jzuniga-amd (approved):

PR #289, opened 2021-07-09
Reduce library size with optimizations enabled

tfalders:
Partially addresses SWDEV-291316.

Small-size kernels have been moved out of the hpp files and into their own cpp files. This should help with recompilation times when making changes and reduce the library size.

I also attempted to remove loop unrolling and consolidate the batched and non-batched versions of the kernels into one, but neither was successful.


tfalders:
I downloaded some CI artifacts from June 26 in order to get some performance results pre-loop rolling, and the performance appears to be comparable. I gathered the runtimes for getf2, getri, and trtri on 2 <= m, n <= 5, and they were similar pre- and post-loop rolling.

cgmb:
> Small-size kernels have been moved out of the hpp files and into their own cpp files. This should help with recompilation times when making changes, and to my surprise also reduced the library size to ~855 MB.

Awesome. I'd been wanting to do that for the first reason, and I wondered if it might also help with the latter. The normal, batched, and strided\_batched versions are split between multiple implementation files, so the template functions from the header are instantiated multiple times. IIRC, the normal and strided\_batched templates were often identical, so they would get the same instantiation twice. I'd figured the linker might de-duplicate them, but it seems not.

I imagine this will also result in faster builds, as the compiler has to redo less work.

cgmb (approved):
So, after your most recent commits, the functions themselves are unchanged (just reorganized)? That seems like a safe change.

I'd have probably only moved the kernels to a separate file. My expectation would be that moving the GPU code to a separate file can't possibly affect the compiler's optimization passes, as I don't think any inferences about parameter values can cross the boundaries between CPU and GPU code. On the other hand, moving some of the host code to a separate file definitely affects what information is available for applying optimizations to the host code at compile-time.

With that being said, it probably does not make much difference to the overall performance (if any). I just thought I'd share how I tend to think about inlining vs. de-inlining in HIP. So, I don't think you need to change anything. Explicitly instantiating those kernels would be a pain. As long as we're not seeing a performance regression, this looks good to me.

jzuniga-amd (approved):
LGTM!

tfalders:
Looks like one test aborted because of taking too much time. I don't think it's an issue with this PR, so I'll go ahead and merge.

cgmb:
Yeah, that's just a really slow CI node. It's the same one that prompted #286.

PR #291, opened 2021-07-14
Added potri

tfalders:
Partially addresses SWDEV-291321.


library/include/rocsolver-functions.h
```diff
@@ -9702,6 +9699,234 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zposv_strided_batched(rocblas_handle h
...
+              Specifies whether the factorization is upper or lower triangular.
+              If uplo indicates lower (or upper), then the upper (or lower) part of A is not used.
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The number of rows and columns of matrix A.
+    @param[inout]
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the factor L or U of the Cholesky factorization of A returned by
+              \ref rocsolver_spotrf "POTRF".
+              On exit, the inverses of A if info = 0; otherwise undefined.
```
jzuniga-amd, line 9731: If you are using the masked copy_mat workaround approach, then when A is singular, on exit it should remain unchanged. So, I would remove the "otherwise undefined"

library/src/lapack/roclapack_potri.hpp
```diff
@@ -0,0 +1,145 @@
...
+    // compute inv(U) * inv(U)' or inv(L)' * inv(L) and store in tmpcopy
+    rocblas_side side = (uplo == rocblas_fill_upper ? rocblas_side_right : rocblas_side_left);
+    rocblasCall_trmm<BATCHED, STRIDED, T>(handle, side, uplo, rocblas_operation_conjugate_transpose,
+                                          rocblas_diagonal_non_unit, n, n, &one, 0, A, shiftA, lda,
+                                          strideA, tmpcopy, 0, n, n * n, batch_count, workArr);
+
+    // copy elements of tmpcopy into A in cases where info is zero
+    hipLaunchKernelGGL(masked_copymat<T>, dim3(copyblocks, copyblocks, batch_count), dim3(32, 32),
+                       0, stream, copymat_from_buffer, n, n, A, shiftA, lda, strideA, tmpcopy, info,
+                       uplo, rocblas_diagonal_non_unit, true);
```
jzuniga-amd, line 141: A remains unchanged when it is singular!
(see my other comment in `rocsolver-functions.h`)

library/src/lapack/roclapack_potri.cpp
```diff
@@ -0,0 +1,118 @@
...
+    rocsolver_potri_getMemorySize<false, false, T>(n, batch_count, &size_work1, &size_work2,
+                                                   &size_work3, &size_work4, &size_tmpcopy,
+                                                   &size_workArr);
+
+    if(rocblas_is_device_memory_size_query(handle))
+        return rocblas_set_optimal_device_memory_size(handle, size_work1, size_work2, size_work3,
+                                                      size_work4, size_tmpcopy, size_workArr);
+
+    // always allocate all required memory for TRSM optimal performance
+    bool optim_mem = true;
```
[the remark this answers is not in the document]
tfalders: This change will affect a number of other functions, so I think creating a separate PR would be the way to go. It might not make it into 4.5.

jzuniga-amd (approved):

PR #294, opened 2021-07-21
Miscellaneous build and test changes

cgmb:
### Expand BDSQR tolerance
    
The errors seen with gfx1030 on Windows were slightly beyond the original tolerance:

```
../../clients/include\testing_bdsqr.hpp(593): error: Expected: ((max_errorv)) <= ((n)*get_epsilon<T>()), actual: 3.68428e-15 vs 3.33067e-15
[  FAILED  ] checkin_lapack/BDSQR.__double_complex/43, where GetParam() = ({ 15, 10, 10, 10 }, { 0, 0, 0, 0 }) (5 ms)
    
../../clients/include\testing_bdsqr.hpp(593): error: Expected: ((max_errorv)) <= ((n)*get_epsilon<T>()), actual: 3.41204e-15 vs 3.33067e-15
[  FAILED  ] checkin_lapack/BDSQR.__double_complex/47, where GetParam() = ({ 15, 10, 10, 10 }, { 0, 0, 0, 1 }) (4 ms)
```

### Fix build with CMake v3.21.0

The new ROCMClang compiler detection causes build failures, so disable it
until the problem has been resolved.

### Use cmake --build rather than make in install.sh
    
By using a generic cmake build call, the Ninja build system can be
used on Linux by calling `./install.sh --cmake-arg -GNinja`.

### Reduce timeout for math-ci checkin tests
    
The debug, precheckin and static library tests typically run in under
5 minutes, so a 45 minute timeout seems high enough to ensure it will
never be encountered unless there's a serious problem. Nevertheless,
this change should help conserve CI resources when tests stall, as the
default timeout is 5 hours.

### Disallow VLAs and other constructs on CI
    
vla, reorder and return-type warnings have been enabled as errors,
as these warnings often indicate problems and are easily avoidable.
    
-Wvla is enabled because VLAs can be used effectively for small arrays,
but a vector is more appropriate for unbounded input. This is
particularly noticeable on Windows, where the default stack size is
only 1MB. rocBLAS has built with this option for a while.
    
-Wreorder is enabled because out-of-order member initializer lists
have a misleading execution order, and I've only ever seen them
written that way by mistake.
    
-Wreturn-type is enabled because omitting a return statement from
a function that returns a value is only ever done by mistake in
modern code.

###  Promote Doxygen warnings to errors
All warnings in rocSOLVER Doxygen builds over the past few months have
indicated legitimate problems in the docs. Turning the warnings into
errors will make them more obvious.

`FAIL_ON_WARNINGS` would be preferable to `YES`, but requires at least
Doxygen 1.9.0, which is newer than the version provided in the
ReadTheDocs container.


cgmb:
The changes to the Doxygen configuration have caused the CI to report a failure. That is expected and desired. There's a legitimate error in the docs in the current develop branch.

The CI should pass once #291 is merged (which includes corrections to the docs for the `GESV` parameter list).

common/include/common_host_helpers.hpp
```diff
@@ -128,10 +129,10 @@ void print_device_matrix(std::ostream& os,
                          const rocblas_stride stride = 1,
                          const rocblas_int idx = 0)
 {
-    T hA[lda * n];
-    hipMemcpy(hA, A + idx * stride, sizeof(T) * lda * n, hipMemcpyDeviceToHost);
+    std::vector<T> hA(lda * n);
+    hipMemcpy(hA.data(), A + idx * stride, sizeof(T) * lda * n, hipMemcpyDeviceToHost);
 
-    print_to_stream<T>(os, name, m, n, hA, lda);
+    print_to_stream<T>(os, name, m, n, hA.data(), lda);
```
tfalders, line 135: Is this why our print statements were failing?

tfalders (commented):
Looks good, and I appreciate the explanations of what all the different changes are.

tfalders (approved):

common/include/common_host_helpers.hpp
cgmb: Probably not. I suspect I broke those debug prints when converting them over to libfmt. I'll take a look and see if I can add a fix.

jzuniga-amd (approved):
LGTM

common/include/common_host_helpers.hpp
cgmb: This works for me, but it may just be that I'm not testing under the right conditions to reproduce the problem.

cgmb: Just for posterity, I think it turned out that the answer was "yes." The use of VLAs was why the print statements were failing. After discussing it more with Juan, I learned that the problems were only encountered when printing very large matrices, so it probably was stack overflow.

PR #298, opened 2021-07-27
Support MSVC in public API

cgmb:
The rocSOLVER public headers currently only support Clang (on Windows). This change adds support for MSVC.


tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #301, opened 2021-07-29
Print the runtime version string during tests

cgmb:
rocsolver-test currently prints the version of rocSOLVER that it was built
against. However, it would be better to print the version that was
queried at runtime, as backwards compatibility allows those to
be different. It's currently a bit confusing when comparing two
implementations of the library via `LD_PRELOAD`.

There's currently no way to know what size of buffer to allocate for
the version string, aside from guessing and checking. In practice,
that works fine, but it's nice to have a more obvious method. For
that reason, I've added `rocsolver_version_string_size` to the API.

Depends on: https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/815


library/src/common/buildinfo.cpp
```diff
@@ -26,7 +26,18 @@ extern "C" rocblas_status rocsolver_get_version_string(char* buf, size_t len)
...
 }
+
+/*******************************************************************************
+ *! \brief   Returns size of buffer required for rocsolver_get_version_string
+ ******************************************************************************/
+extern "C" rocblas_status rocsolver_get_version_string_size(size_t* len)
+{
+    if(!len)
+        return rocblas_status_invalid_pointer;
+    *len = std::strlen(VERSION_STRING) + 1;
```
tfalders, line 41: What's with the +1?

docs/source/api_otherfunc.rst
```diff
@@ -0,0 +1,28 @@
...
+   :local:
+   :backlinks: top
+
+rocsolver_get_version_string()
+------------------------------------
+.. doxygenfunction:: rocsolver_get_version_string
+
+rocsolver_get_version_string_size()
+------------------------------------
+.. doxygenfunction:: rocsolver_get_version_string_size
```
tfalders, line 27: This is duplicated in api_loggingfunc. Only one should be kept.

docs/source/api_loggingfunc.rst
```diff
@@ -14,48 +17,67 @@ are divided in two categories:
 .. _initialize:
 
 Logging set-up and tear-down
-===============================
+-------------------------------
```
tfalders, line 20: I don't support changing all the header levels in order to accommodate the string functions as a separate section. This will make the API docs for logging inconsistent with all the other API docs.

That's not to say, though, that I oppose putting the new functions on this page. We could combine the logging sections into one. The filename will probably need changing, and I'd prefer the page be titled "Helper and Logging Functions" with the string functions at the top.

tfalders (changes requested):
Looks good, but I have concerns with the approach to the API docs in this PR.

library/src/common/buildinfo.cpp
cgmb: `std::strlen` returns the number of characters in the string, but the API provides the size of buffer to allocate (which should include the null character). For C users, it means the value returned can be passed directly to `malloc` and it will be the right size.

docs/source/api_otherfunc.rst
cgmb: I went back and forth on separating them. There's no links to this page... it was supposed to be deleted. Good catch.

docs/source/api_loggingfunc.rst
cgmb: I agree. I was trying to leave them unchanged, as I wasn't sure how attached you and Juan were to the existing logging docs layout. Merging them seems good to me.

cgmb: I partly addressed this. After some thought, I figured maybe we should keep the file name the same. After all, the queries are kinda related to logging (they exist so the information can be logged). When possible, I'd like to avoid renaming files since that changes URLs, so I figure that's close enough it might as well just stay as-is.

I'm also not a big fan of 'Helper' as a descriptor, as it's quite vague. I tried changing it to be more specific. The title might have to change again if we end up with any more oddball functions (unlikely any time soon), but we can cross that bridge when we come to it.

tfalders (approved):
LGTM!

jzuniga-amd (approved):
LGTM

PR #302, opened 2021-07-29
Replace pre-commit hook with post-commit

cgmb:
This post-commit hook is a bit more polite than the current pre-commit hook. It doesn't change what you committed and it asks before taking action. To apply the format changes, it creates a new commit.

Maybe take it for a test drive and let me know if you have any feedback.


tfalders (changes requested):
Personally, I prefer the pre-commit hook. I've taken to not worrying about spacing when I'm working and let the hook take care of making the commit look decent. If we could have the option of installing one or the other --- either with a flag in the install script or a separate install script --- that would work for me.

cgmb:
That's great feedback. I'll try updating it to see if I can address those concerns. If it doesn't meet your satisfaction, we can of course always keep the existing pre-commit hook.

---

Though, perhaps I should expand the motivation a bit more, including why I don't use the existing pre-commit hook. Let's start with a blank repository:

```
cgmb@ubuntu-18.04:~/rocSOLVER$ git status
On branch develop
Your branch is up to date with 'origin/develop'.

nothing to commit, working tree clean
```

Next create some changes by inserting a couple changes into a file, in this case a couple of comments:
```
cgmb@ubuntu-18.04:~/rocSOLVER$ sed -i '200i // first' library/include/rocsolver-functions.h
cgmb@ubuntu-18.04:~/rocSOLVER$ sed -i '400i // second' library/include/rocsolver-functions.h
```

Stage one of those changes for commit, but leave the other unstaged:
```
cgmb@ubuntu-18.04:~/rocSOLVER$ git add -p library/include/rocsolver-functions.h 
diff --git a/library/include/rocsolver-functions.h b/library/include/rocsolver-functions.h
index 7f4f67a..2ee6c3b 100644
--- a/library/include/rocsolver-functions.h
+++ b/library/include/rocsolver-functions.h
@@ -197,6 +197,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_slaswp(rocblas_handle handle,
 ROCSOLVER_EXPORT rocblas_status rocsolver_dlaswp(rocblas_handle handle,
                                                  const rocblas_int n,
                                                  double* A,
+// first
                                                  const rocblas_int lda,
                                                  const rocblas_int k1,
                                                  const rocblas_int k2,
Stage this hunk [y,n,q,a,d,j,J,g,/,e,?]? y
@@ -396,6 +397,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_dlarft(rocblas_handle handle,
                                                  const rocblas_int k,
                                                  double* V,
                                                  const rocblas_int ldv,
+// second
                                                  double* tau,
                                                  double* T,
                                                  const rocblas_int ldt);
Stage this hunk [y,n,q,a,d,K,g,/,e,?]? n
```

Reviewing the state of the repository, we see that some of the changes are listed as staged for commit, while others are listed as not (as expected). You could also get into this state by making a change, doing a `git add` then making another change:
```bash
cgmb@ubuntu-18.04:~/rocSOLVER$ git status
On branch develop
Your branch is up to date with 'origin/develop'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   library/include/rocsolver-functions.h

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   library/include/rocsolver-functions.h
```

Reviewing those staged changes:
```
cgmb@ubuntu-18.04:~/rocSOLVER$ git diff --staged
diff --git a/library/include/rocsolver-functions.h b/library/include/rocsolver-functions.h
index 7f4f67a..56563d4 100644
--- a/library/include/rocsolver-functions.h
+++ b/library/include/rocsolver-functions.h
@@ -197,6 +197,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_slaswp(rocblas_handle handle,
 ROCSOLVER_EXPORT rocblas_status rocsolver_dlaswp(rocblas_handle handle,
                                                  const rocblas_int n,
                                                  double* A,
+// first
                                                  const rocblas_int lda,
                                                  const rocblas_int k1,
                                                  const rocblas_int k2,
```

Then we call `git commit`. You can see what is being committed at the bottom of the message, since I have `git config --global commit.verbose true` set on my machine:
```bash
My commit message...
# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch develop
# Your branch is up to date with 'origin/develop'.
#
# Changes to be committed:
#   modified:   library/include/rocsolver-functions.h
#
# Changes not staged for commit:
#   modified:   library/include/rocsolver-functions.h
#
# ------------------------ >8 ------------------------
# Do not modify or remove the line above.
# Everything below it will be ignored.
diff --git a/library/include/rocsolver-functions.h b/library/include/rocsolver-functions.h
index 7f4f67a..56563d4 100644
--- a/library/include/rocsolver-functions.h
+++ b/library/include/rocsolver-functions.h
@@ -197,6 +197,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_slaswp(rocblas_handle handle,
 ROCSOLVER_EXPORT rocblas_status rocsolver_dlaswp(rocblas_handle handle,
                                                  const rocblas_int n,
                                                  double* A,
+// first
                                                  const rocblas_int lda,
                                                  const rocblas_int k1,
                                                  const rocblas_int k2,
```

Then, you look at what actually was committed. With the current pre-commit hook, you'll see something like this:
```
cgmb@ubuntu-18.04:~/rocSOLVER$ git show HEAD
commit 4e2910ed2a91431c9cfebb5de5964065ea9c4156 (HEAD -> develop)
Author: Cordell Bloor
Date:   Thu Jul 29 20:11:03 2021 -0600

    My commit message...

diff --git a/library/include/rocsolver-functions.h b/library/include/rocsolver-functions.h
index 7f4f67a..41fe170 100644
--- a/library/include/rocsolver-functions.h
+++ b/library/include/rocsolver-functions.h
@@ -197,6 +197,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_slaswp(rocblas_handle handle,
 ROCSOLVER_EXPORT rocblas_status rocsolver_dlaswp(rocblas_handle handle,
                                                  const rocblas_int n,
                                                  double* A,
+                                                 // first
                                                  const rocblas_int lda,
                                                  const rocblas_int k1,
                                                  const rocblas_int k2,
@@ -396,6 +397,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_dlarft(rocblas_handle handle,
                                                  const rocblas_int k,
                                                  double* V,
                                                  const rocblas_int ldv,
+                                                 // second
                                                  double* tau,
                                                  double* T,
                                                  const rocblas_int ldt);
```

Fundamentally, the problem is that the existing pre-commit hook operates on the wrong data: it processes the content found in the working tree, rather than processing the content that has been staged for commit.

To avoid operating on unstaged data, my solution is to allow the commit to complete, then checkout the commit into a new directory, run clang-format, commit the changes, and apply that commit back to the original repository. It would be easy to remove the prompts, so the whole process is automatic, but... I do wonder if I could fix the bug in the pre-commit hook without making it post-commit. There probably is a more efficient solution. I just already had a lot of the pieces of this one lying around already.

tfalders:
> To avoid operating on unstaged data, my solution is to allow the commit to complete, then checkout the commit into a new directory, run clang-format, commit the changes, and apply that commit back to the original repository. It would be easy to remove the prompts, so the whole process is automatic, but... I do wonder if I could fix the bug in the pre-commit hook without making it post-commit. There probably is a more efficient solution. I just already had a lot of the pieces of this one lying around already.

Perhaps we could add an option to toggle between doing a regular commit or a `commit --amend`? It doesn't matter to me if it's a pre- or post-commit hook, just as long as it automatically cleans up the commit (preferably without any prompting from me).

cgmb:
I've updated my approach. The new steps are:
1. commit as normal
2. stash any remaining modifications
3. apply reformatting
4. amend commit with updated formatting (if there were changes)
5. unstash modifications

This should work as normal for cases where all changes have been staged, and will correctly and automatically handle cases like the example I showed above. However, if the reformatting and the stashed modifications both affect the same lines, you will have a merge conflict on unstash. In that case, an error is reported and you will have to resolve the merge conflict manually. Fortunately, all the important information is committed, so it shouldn't be too bad to recover to your desired state.

When you're faced with a merge conflict:
- The original commit id is printed at the start of this process, and it contains everything you'd staged for commit, exactly as it was.
- The modifications that were not staged for commit remain in `stash@{0}` until you mark the conflict as resolved and `git stash drop` it.
- The commit after reformatting and amending is the current `HEAD`.

We could probably make the error recovery a bit friendlier, but this seems like an improvement to me already. It should just 'do the right thing' for most cases, and it only throws the problem back to you when the 'right thing' is actually unclear.

cgmb:
I also changed the hooks to install as files rather than as symlinks, so if you want to try it out, you can:
```
git remote add cgmb https://github.com/cgmb/rocSOLVER.git
git fetch cgmb update-commit-hook
git checkout cgmb/update-commit-hook -- tools
tools/install-hooks
git rm -rf -- tools
```
and continue with whatever you were working on.

cgmb:
Fixed a couple broken edge cases. On the bright side, they illustrate how robust this approach is. Every important state is saved up-front, so recovering from error was easy.

cgmb:
There should really be a warning when clang-format is skipped...

cgmb:
The hook now prints a warning (in yellow ) if clang-format is not found.

cgmb:
I'll do a bit more testing myself this weekend, but I'd like to ask you to take another look next week, @tfalders.

> I've taken to not worrying about spacing when I'm working and let the hook take care of making the commit look decent.

I think this script fulfills that criteria. If it somehow does not, please do let me know what's wrong.

tfalders (approved):
Looks good, seems to work. I had a bit of an issue where I accidentally created an empty commit by just changing the format, but other than this degenerate case my tests didn't have any issues.

Now that this post-commit does something similar to our existing pre-commit, I'm a little unclear on what we will gain by switching to a post-commit.

cgmb:
> I had a bit of an issue where I accidentally created an empty commit by just changing the format, but other than this degenerate case my tests didn't have any issues.

Interesting. I'd forgotten about that case. I've turned it into a warning when the amended commit is empty. The git command line will prevent you from creating empty commits by default, but only because it's usually done by mistake. An empty commit is harmless.

For people using the hooks, commits that differ only in formatting are probably also mistakes (e.g. forgot to `git add` the file that contained the meaningful changes). It would be better if the hook warned you of the problem before you wrote the commit message, but the empty commit preserves the commit message, so you can easily add whatever changes you missed and `git commit --amend`. If it actually was a useless commit, it can be removed with `git reset --soft HEAD~`.

The fact that the hook doesn't catch those sorts of mistakes is certainly a drawback to this method. It's fixable, but it would take some thought to address it properly. The impact seems minor, though, so perhaps that can be left as a future improvement.

> Now that this post-commit does something similar to our existing pre-commit, I'm a little unclear on what we will gain by switching to a post-commit.

It handles cases like https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/302#issuecomment-889590987 differently than the pre-commit hook did. It only reformats changes that were staged for commit.

cgmb:
We have finally specified which version of clang-format version is used during development: the ROCm math libraries are now expected to be formatted using the clang-format from the most recent ROCm release. At the moment, that is ROCm 4.3.1.

You can use a different version of clang-format and it will probably still work for most files, but there are occasionally differences in the output between versions. If you want to guarantee that your PR will pass the format check on the CI, you will need to use the clang-format from the ROCm release.

You should be able to do a [multi-version install](https://rocmdocs.amd.com/en/latest/Installation_Guide/Multiversion-Installation.html) of the llvm-amdgpu package (llvm-amdgpu4.3.1) using your package manager if you have the official ROCm repositories installed. However, it can sometimes be inconvenient to change your configured repositories and globally installed packages. For that reason, I've added `./tools/install-hooks --get-clang-format` to download and extract clang-format to your git hooks directory. If you are on a platform that does not have official ROCm repositories or packages for the most recent release, then perhaps instead use the `llvm-amdgpu@4.3.1` package from [Spack](https://spack.readthedocs.io/).

The change to use a fixed version of clang-format should be good for external contributors. Until now, there was no publicly documented place where it was specified what version to use (nor any publicly accessible copy of the clang-format binary). I was pushing to change that, as I had been tasked with writing the rocSOLVER Contributing Guidelines and it simply was not possible to provide useful instructions to contributors without resolving the matter. This PR had similar motivations; the pre-commit hook is too buggy for me to feel comfortable recommending it to new contributors.

Writing documentation can be a real motivator for fixing problems that were not getting the attention they deserved.

> Our habit of trying to document bugs and limitations visibly was enormously useful to the system. As we put out each edition, the presence of these sections shamed us into fixing innumerable things rather than exhibiting them in public. I remember clearly adding or editing many of these sections, then saying to myself "I can't write this," and fixing the code instead.

~ Dennis Richie, on the BUGS section of Unix manuals

jzuniga-amd (approved):

PR #307, opened 2021-08-12
Update for rocblas_internal_gemm_template change

cgmb:
The template arguments for rocblas_internal_gemm_template were changed
in rocBLAS [80df9950a6214c73b02aef31e38cb54dc9f7cffb](https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/80df9950a6214c73b02aef31e38cb54dc9f7cffb). This commit
adjusts how rocSOLVER calls the rocBLAS function, matching the style
used in rocBLAS itself (hopefully minimizing future breakages).

### DIFF
--- a/library/src/include/rocblas.hpp
+++ b/library/src/include/rocblas.hpp
@@ -4,12 +4,12 @@
 
 #pragma once
 
+#include <internal/rocblas-exported-proto.hpp>
+#include <internal/rocblas_device_malloc.hpp>
 #include <rocblas.h>
 
 #include "common_host_helpers.hpp"
 #include "init_scalars.hpp"
-#include "internal/rocblas-exported-proto.hpp"
-#include "internal/rocblas_device_malloc.hpp"
 #include "lib_device_helpers.hpp"
 #include "lib_host_helpers.hpp"
 #include "rocsolver_logger.hpp"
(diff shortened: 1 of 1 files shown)


tfalders (approved):
Ah, so that's what was causing the issue in hipSOLVER. Nice job!

jzuniga-amd (commented):
I don't understand what changed?
What is the effect of not specifying the type T?
`rocblas_internal_gemm_template` did not change (from what I can see). The second template parameter is still the type of the scalar alpha, ie. T  in our case.

cgmb:
That's a good point. It's quite possible I mixed something up during the testing of this change.

The problem is definitely the template parameters, and I'd thought this had somehow fixed the inference for the index type. Yet, this does seem like it should be insufficient to fix the problem. I'd expect to also need to match the way rocBLAS does its casts.

It's been a long day. I'm sure the reason will be obvious when I take a fresh look at it.

cgmb:
Ugh. I found the actual problem: hipSOLVER and rocSOLVER were being built with different versions of rocBLAS (with binary-incompatible changes to the rocblas_internal API). That's another mode of CI failure that I hadn't considered (and it only applies to hipBLAS / hipSOLVER).

I thought I'd fixed it when I rebuilt rocSOLVER with modifications and it worked, but I didn't go back to try removing my changes and rebuilding to check that it still failed without them. -_-;

You caught me programming through superstition. This change would have been harmless either way, but asking "why" brought attention to the real issue.

PR #308, opened 2021-08-13
GETRF optimizations

jzuniga-amd:
A series of different optimizations; focus on large matrices.


library/src/lapack/roclapack_getrf.hpp
```diff
@@ -57,43 +59,268 @@ rocblas_int getrf_get_blksize(rocblas_int dim)
...
+                if(m <= 32)
+                    blk = 48;
+                else if(m <= 64)
+                    blk = n;
+                else if(m <= 352)
+                    blk = 16;
+                else if(m <= 8960)
+                    blk = n;
+                else
+                    blk = 16;
```
tfalders, line 155: This is rather odd. Why does it switch between 16 and n?

tfalders (approved):
Looks good! I've opened a PR on your local fork with the requested changes to PIVOT.

library/src/lapack/roclapack_getrf.hpp
jzuniga-amd: The block sizes for the small cases will always be so that the specialized kernels can be used. 

Now, for larger panel matrices, I would probably expect the block size to always increase with the matrix size, but, in practice, there are many factors that could affect these values (tuned experimentally here)...
My initial guess is that there are more tuned cases for GEMMs with square matrices than panel matrices, although I did not investigate further.

cgmb (approved):
The code is cleaner.
The library is faster.
The binary is smaller (by 22%).
Beautiful.

The only thing I'll add is that we should wait for the extended tests to finish before merging.

Issue #230, opened 2021-03-24
Unable to dlopen librocsolver.so


deven-amd:
TensorFlow needs to dynamically load the rocSOLVER lib at runtime (TF does the same for all ROCm libs).

Attempting to do so results in the following error
```
2021-03-24 14:12:48.355140: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] 
Could not load dynamic library 'librocsolver.so'; 
dlerror: /opt/rocm-4.0.1/lib/librocsolver.so: undefined symbol: _Z20rocblas_ger_templateILb0EfPKfKPffE15rocblas_status_P15_rocblas_handleiiPKT3_lPKT1_iiiSC_iiiPT2_iiii

2021-03-24 14:12:48.355196: F tensorflow/stream_executor/lib/statusor.cc:34] 
Attempting to fetch value instead of handling error Failed precondition: 
Could not load dynamic library 'librocsolver.so'; 
dlerror: /opt/rocm-4.0.1/lib/librocsolver.so: undefined symbol: _Z20rocblas_ger_templateILb0EfPKfKPffE15rocblas_status_P15_rocblas_handleiiPKT3_lPKT1_iiiSC_iiiPT2_iiii
```

You can reproduce this error using the following simple standalone program
```cpp
// main.cpp
// /opt/rocm/bin/hipcc -std=c++11 main.cpp -ldl -o test_runner
// ./test_runner
#include <dlfcn.h>
#include <iostream>

#define ROCSOLVER_LIB_NAME "/opt/rocm-4.0.1/lib/librocsolver.so"


void* open_roc_lib(const char* lib_name) {
  void* handle = dlopen(lib_name, RTLD_NOW | RTLD_LOCAL);
  if (!handle) {
    std::cout << dlerror() << "\n";
    return nullptr;
  }
  std::cout << "loaded " << lib_name << "\n";
  return handle;
}

int main() {
  void* rocsolver_handle = open_roc_lib(ROCSOLVER_LIB_NAME);
  dlclose(rocsolver_handle);
  return 0;
}
```

It seems that the rocblas library is not in the list of dependencies for rocsolver library...could that be the cause of this error?
```
root@prj47-rack-15:/root# ldd /opt/rocm-4.0.1/lib/librocsolver.so
	linux-vdso.so.1 (0x00007fffdccd9000)
[...]

PR #309, opened 2021-08-19
Fix missing header in rocsolver.h

cgmb:
rocsolver-extra-types.h uses uint32_t, which requires stdint.h. The problem can be reproduced with a test program:

```
#include <rocsolver.h>
int main()
{
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- Added missing stdint.h include to rocsolver.h
+
 ### Known Issues
 ### Security
 
--- a/library/include/rocsolver-extra-types.h
+++ b/library/include/rocsolver-extra-types.h
@@ -5,6 +5,8 @@
 #ifndef ROCSOLVER_EXTRAS_H_
 #define ROCSOLVER_EXTRAS_H_
 
+#include <stdint.h>
+
 /*! \brief Used to specify the logging layer mode using a bitwise combination
  *of rocblas_layer_mode values.
  ********************************************************************************/


tfalders (approved):
Looks good, but why has this not caused any issues before? Surely a missing header should have caused CI to fail.

cgmb:
I could have sworn I'd checked for this when we added the typedef, but somehow I'd missed it.

In any case, this has been a defect since 4.2.0. The hip headers include stdint.h, so it wasn't noticed when building our samples. Fortunately, most users probably won't notice, either. And it's easy to workaround.

I think CMake has some functionality for compilation tests, but I don't quite have that integration setup yet. A quick and dirty way to catch these sorts of errors in the CI would be to change our samples to include rocsolver.h before anything else. Maybe I should add that change?

tfalders:
> A quick and dirty way to catch these sorts of errors in the CI would be to change our samples to include rocsolver.h before anything else. Maybe I should add that change?

The samples are user-facing and this change might look odd to them. Is there somewhere else we can change it where it will still have the intended effect? Maybe in rocsolver_gtest_main.cpp?

cgmb:
> Is there somewhere else we can change it where it will still have the intended effect? Maybe in rocsolver_gtest_main.cpp?

That's an interesting idea, but I just tried it now and I was surprised to discover that the answer is 'no'.

After moving the header, rocsolver-test compiled successfully. After some investigation, this appears to be because the file is compiled as HIP with the `-x hip` flag. I suspect the file gets automatically included when the compilation language is HIP because the `hipThread*`/ `hipBlock*` / `hipGrid*` variables are of type `uint32_t`.

The only code in our repository not compiled with `-x hip` are the C and Fortran samples.

TorreZuk (commented):
The sorting I think makes sense within include groupings (which don't seem to be used much), you could just leave a blank line after rocsolver.h at the top but I guess you are not encouraging this order.  Either way if this is a "hidden" test it could use some comment, or make a solver_only_include.c

cgmb:
If the header order in samples is causing concern, I'll just drop the test for now. I've been slowly getting towards integrating tests for #230 into the build, and this would be a very similar check.

cgmb:
A test had been added in #310.

jzuniga-amd (approved):

PR #311, opened 2021-08-30
Added sytf2

tfalders:
Addresses SWDEV-291556.

### DIFF
--- a/clients/common/lapack_host_reference.cpp
+++ b/clients/common/lapack_host_reference.cpp
@@ -6237,7 +6237,6 @@ void cblas_sytf2<rocblas_double_complex>(rocblas_fill uplo,
     zsytf2_(&uploC, &n, A, &lda, ipiv, info);
 }
 
-#include <iostream>
 // sytrf
 template <>
 void cblas_sytrf<float>(rocblas_fill uplo,
--- a/clients/common/lapack_host_reference.cpp
+++ b/clients/common/lapack_host_reference.cpp
@@ -1812,6 +1812,30 @@ void zhegvd_(int* itype,
              int* liwork,
              int* info);
 
+void ssytf2_(char* uplo, int* n, float* A, int* lda, int* ipiv, int* info);
+void dsytf2_(char* uplo, int* n, double* A, int* lda, int* ipiv, int* info);
+void csytf2_(char* uplo, int* n, rocblas_float_complex* A, int* lda, int* ipiv, int* info);
+void zsytf2_(char* uplo, int* n, rocblas_double_complex* A, int* lda, int* ipiv, int* info);
+
+void ssytrf_(char* uplo, int* n, float* A, int* lda, int* ipiv, float* work, int* lwork, int* info);
+void dsytrf_(char* uplo, int* n, double* A, int* lda, int* ipiv, double* work, int* lwork, int* info);
+void csytrf_(char* uplo,
+             int* n,
+             rocblas_float_complex* A,
+             int* lda,
+             int* ipiv,
+             rocblas_float_complex* work,
+             int* lwork,
+             int* info);
+void zsytrf_(char* uplo,
+             int* n,
+             rocblas_double_complex* A,
+             int* lda,
+             int* ipiv,
+             rocblas_double_complex* work,
+             int* lwork,
+             int* info);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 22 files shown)


clients/common/lapack_host_reference.cpp
```diff
@@ -1812,6 +1812,30 @@ void zhegvd_(int* itype,
              int* liwork,
              int* info);
 
+void ssytf2_(char* uplo, int* n, float* A, int* lda, int* ipiv, int* info);
+void dsytf2_(char* uplo, int* n, double* A, int* lda, int* ipiv, int* info);
+void csytf2_(char* uplo, int* n, rocblas_float_complex* A, int* lda, int* ipiv, int* info);
+void zsytf2_(char* uplo, int* n, rocblas_double_complex* A, int* lda, int* ipiv, int* info);
```
cgmb, line 1818: This is more a question for Juan, but is the hard-coded Fortran name mangling the reason why we try to default to gfortran in install.sh and CMakeLists.txt? As I understand it, some compilers use different schemes, like making symbols uppercase, rather than just appending an underscore.

library/include/rocsolver-functions.h
```diff
@@ -16987,6 +16987,305 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_ztrtri_strided_batched(rocblas_handle
...
+
+    \details
+    (This is the unblocked version of the algorithm).
+
+    The factorization has the form
+
+    \f[
+        \begin{array}{cl}
+        A = U D U^T & \: \text{or}\\
+        A = L D L^T &
```
jzuniga-amd, line 17002: It is important to describe how L (or D) is constructed. Especially because, when there are permutations or 2-by-2 pivots, L (or D) is not exactly what is returned in A.
(see lapack documentation for some ideas on how to write this)

library/include/rocsolver-functions.h
```diff
@@ -16987,6 +16987,305 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_ztrtri_strided_batched(rocblas_handle
...
+              Specifies whether the upper or lower part of the matrix A is stored.
+              If uplo indicates lower (or upper), then the upper (or lower)
+              part of A is not used.
+    @param[in]
+    n         rocblas_int. n >= 0.\n
+              The number of rows and columns of the matrix A.
+    @param[inout]
+    A         pointer to type. Array on the GPU of dimension lda*n.\n
+              On entry, the symmetric matrix A to be factored.
+              On exit, the block diagonal matrix D and the multipliers needed to
```
jzuniga-amd, line 17023: what are the multipliers? 
We must describe them. See comment above...

library/include/rocsolver-functions.h
```diff
@@ -16987,6 +16987,305 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_ztrtri_strided_batched(rocblas_handle
                                                                  const rocblas_int batch_count);
 //! @}
 
+/*! @{
+    \brief SYTF2 computes the factorization of a symmetric matrix \f$A\f$
```
jzuniga-amd, line 16991: I think we should add "symmetric indefinite matrix" as we are classifying this routine along with potf2/potrf as triangular factorization.

jzuniga-amd (changes requested):

clients/common/lapack_host_reference.cpp
jzuniga-amd: mmm I don't really understand the question. This is the only way I know to call fortran compiled lapack from c/c++ code. If it matters and there are better ways to do it, we can change it, no problem.

jzuniga-amd:
Some initial comments above...
I am still looking at the kernels.

library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -0,0 +1,571 @@
...
+                    j = i + sidx[0];
+                    rowmax = max(rowmax, aabs<S>(sval[0]));
+                    __syncthreads();
+                }
+
+                if(absakk >= alpha * colmax * (colmax / rowmax))
+                    // no interchange (1-by-1 block)
+                    kp = k;
+                else if(aabs<S>(A[i + i * lda]) >= alpha * rowmax)
+                    // interchange rows and columns k and i (1-by-1 block)
```
jzuniga-amd, line 392: " interchange rows and columns kp = i and kk = k (1-by-1 block)"

library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -0,0 +1,571 @@
...
+                    kp = i;
+                else
+                {
+                    // interchange rows and columns k+1 and i (2-by-2 block)
```
jzuniga-amd, line 396: "interchange rows and columns kp = i and kk = k+1 (2-by-2 block)"

library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -0,0 +1,571 @@
...
+        sytf2_iamax(tid, k, A + k * lda, 1, sval, sidx);
+        i = sidx[0] - 1;
+        colmax = aabs<S>(sval[0]);
+        __syncthreads();
+
+        if(max(absakk, colmax) == 0)
+        {
+            // singularity found
+            if(tid == 0 && info == 0)
+                info = k;
```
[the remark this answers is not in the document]
tfalders: Good catch! Adding the singular test in has revealed this bug, as well as a couple others that were not caught by the standard test. Working on them now.

library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -0,0 +1,571 @@
...
+    while(k >= 0)
+    {
+        kstep = 1;
+        absakk = aabs<S>(A[k + k * lda]);
+
+        // find max off-diagonal entry in column k
+        sytf2_iamax(tid, k, A + k * lda, 1, sval, sidx);
+        i = sidx[0] - 1;
+        colmax = aabs<S>(sval[0]);
+        __syncthreads();
```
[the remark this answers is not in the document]
tfalders: I've tried to remove one of these syncthreads but without success. All the threads in the block need to read sidx[0] and sval[0] only after thread 0 has written to them, but before thread 0 has changed them in the next call to iamax. One approach that could work is creating two separate sidx and sval arrays, but I haven't tried it yet.

library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -0,0 +1,571 @@
...
+                kp = k;
+            else
+            {
+                // find max off-diagonal entry in row i
+                sytf2_iamax(tid, i - k, A + i + k * lda, lda, sval, sidx);
+                j = k - 1 + sidx[0];
+                rowmax = aabs<S>(sval[0]);
+                __syncthreads();
+
+                if(i > 0)
```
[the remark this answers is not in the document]
tfalders: Good catch! Thanks. I must have missed it after I copied the upper kernel to the lower.

library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -0,0 +1,571 @@
...
+                        wk = d12 * (d22 * A[j + k * lda] - A[j + (k - 1) * lda]);
+                        for(i = j; i >= 0; i--)
+                            A[i + j * lda]
+                                = A[i + j * lda] - A[i + k * lda] * wk - A[i + (k - 1) * lda] * wkm1;
+                    }
+                    __syncthreads();
+
+                    // update columns k and k-1
+                    for(j = k - 2 - tid; j >= 0; j -= SYTF2_MAX_THDS)
+                    {
```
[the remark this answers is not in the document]
tfalders: Splitting the loop is necessary, since the second loop is updating entries that the first loop reads. We need to make sure all the threads finish their reads before the updates begin.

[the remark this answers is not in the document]
jzuniga-amd: oh, I see, you are right!
Thread j read data that will be updated by other threads.

jzuniga-amd (approved):
Looks great now...

cgmb (approved):
LGTM. I had some trivial comments, but that's it.

I'm also happy to see that our internal docs build has improved. The LaTeX is now properly rendered. Hurrah!

PR #316, opened 2021-09-22
Added lasyf

tfalders:
Addresses SWDEV-291555.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_lasyf.hpp
+++ b/library/src/auxiliary/rocauxiliary_lasyf.hpp
@@ -16,53 +16,6 @@
 // number of threads for the lasyf kernel (currently not tunable)
 #define LASYF_MAX_THDS 256
 
-/** Device function to compute y = y - Ax (gemv) **/
-template <typename T>
-__device__ void lasyf_gemv(const rocblas_int tid,
-                           const rocblas_int m,
-                           const rocblas_int n,
-                           T* A,
-                           const rocblas_int lda,
-                           T* x,
-                           const rocblas_int incx,
-                           T* y,
-                           const rocblas_int incy,
-                           T* temp)
-{
-    for(int i = tid; i < m; i += LASYF_MAX_THDS)
-    {
-        temp[tid] = y[i * incy];
-        for(int j = 0; j < n; j++)
-            temp[tid] -= A[i + j * lda] * x[j * incx];
-        y[i * incy] = temp[tid];
-    }
-}
-
-/** Device function to compute C = C - A B' (gemm) **/
-template <typename T>
-__device__ void lasyf_gemm(const rocblas_int tid,
-                           const rocblas_int m,
-                           const rocblas_int n,
-                           const rocblas_int k,
-                           T* A,
-                           const rocblas_int lda,
-                           T* B,
-                           const rocblas_int ldb,
-                           T* C,
-                           const rocblas_int ldc,
-                           T* temp)
-{
-    for(int e = tid; e < m * n; e += LASYF_MAX_THDS)
-    {
-        int i = e % m;
-        int j = e / m;
-        temp[tid] = C[i + j * ldc];
-        for(int l = 0; l < k; l++)
-            temp[tid] -= A[i + l * lda] * B[j + l * ldb];
-        C[i + j * ldc] = temp[tid];
-    }
-}
-
 template <typename T, typename U>
 ROCSOLVER_KERNEL void __launch_bounds__(LASYF_MAX_THDS)
     lasyf_kernel_upper(const rocblas_int n,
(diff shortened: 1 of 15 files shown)


library/src/auxiliary/rocauxiliary_lasyf.hpp
```diff
@@ -0,0 +1,652 @@
...
+                       const rocblas_stride strideA,
+                       rocblas_int* ipivA,
+                       const rocblas_stride strideP,
+                       rocblas_int* infoA,
+                       T* WA)
+{
+    // select batch instance
+    rocblas_int bid = hipBlockIdx_y;
+
+    int tid = hipThreadIdx_x;
```
cgmb, line 83: This is kind of a tangent, but [apparently tid is a built-in variable](https://godbolt.org/z/eTvPa7qWz) for the thread index. Related: One of the Google LLVM devs added HIP language support to Compiler Explorer in July/August. It still needs HIP library support to be useful, but... one step closer.

This code is fine. No changes required (nor desired). I just thought that was an interesting fact about the HIP language.

library/src/auxiliary/rocauxiliary_lasyf.hpp
```diff
@@ -0,0 +1,652 @@
...
+                    kp = k;
+                else if(aabs<S>(W[imax + (kw - 1) * ldw]) >= alpha * rowmax)
+                {
+                    // interchange rows and columns kk = k and kp = imax (1-by-1 block)
+                    kp = imax;
+
+                    // copy column kw-1 of W to column kw of W
+                    for(i = tid; i <= k; i += LASYF_MAX_THDS)
+                        W[i + kw * ldw] = W[i + (kw - 1) * ldw];
+                    __syncthreads();
```
cgmb, line 188: I don't understand the full context of this algorithm so it's probably just my ignorance, but I don't see how it's guaranteed that all threads will end up at this sync point. Isn't rowmax based on matrix contents?

library/src/auxiliary/rocauxiliary_lasyf.hpp
tfalders: The code you linked has tid as an argument for the square method, and removing it gave a compiler error. Are you sure it's built-in? It would be neat if it is.

library/src/auxiliary/rocauxiliary_lasyf.hpp
tfalders: All of the variables other than thread and loop indices should be the same across all the threads working on a problem. rowmax in particular is a shared variable, which should guarantee that all the threads use the same value.

library/src/auxiliary/rocauxiliary_lasyf.hpp
cgmb: You're right. I appear to be mistaken about tid. This was provided as a modified version of a CUDA example, and I didn't notice that aspect had been changed. They'd just mentioned that the indexing variables had different names in HIP than CUDA, and I suppose I misunderstood. Now I'm not actually sure what they were referring to.

That was confusing, as the HIP indexing variables I'm familiar with (like hipThreadIdx_{x,y,z} and hipBlockIdx_{x,y,z} or threadIdx.{x,y,z} and blockIdx.{x,y,z}) are all defined at the library level, and they're actually turned into function calls. I created [an example with the relevant portions of the HIP headers included](https://godbolt.org/z/r9zxT414f), and you can see how these HIP variables ultimately compile to functions, such as `__ockl_get_group_id` for the block id. (If the device libraries were available, the call to `__ockl_get_group_id` would not be in the resulting assembly, as it would be inlined.)

Anyway. It was a tangent, but I hope it was interesting.

library/src/auxiliary/rocauxiliary_lasyf.hpp
```diff
@@ -0,0 +1,652 @@
...
+#include "lapack_device_functions.hpp"
+#include "rocblas.hpp"
+#include "rocsolver.h"
+
+// number of threads for the lasyf kernel (currently not tunable)
+#define LASYF_MAX_THDS 256
+
+/** Device function to compute y = y - Ax (gemv) **/
+template <typename T>
+__device__ void lasyf_gemv(const rocblas_int tid,
```
jzuniga-amd, line 21: Is there any particularity with this kernel? It looks like a generic gemv so I wonder why not use `gemv_kernel` in `lapack_device_functions.hpp` ? (or a consolidated version that we can put in the header for code re-using)

library/src/auxiliary/rocauxiliary_lasyf.hpp
```diff
@@ -0,0 +1,652 @@
...
+        temp[tid] = y[i * incy];
+        for(int j = 0; j < n; j++)
+            temp[tid] -= A[i + j * lda] * x[j * incx];
+        y[i * incy] = temp[tid];
+    }
+}
+
+/** Device function to compute C = C - A B' (gemm) **/
+template <typename T>
+__device__ void lasyf_gemm(const rocblas_int tid,
```
jzuniga-amd, line 43: Same here, it might be a good idea to move this to `lapack_device_functions.hpp` and have a single consolidated gemm

PR #317, opened 2021-09-23
Print version information in rocsolver-bench

cgmb:
This prints version information when rocsolver-bench is invoked. The version information is not printed when `--perf 1` is added, as that option has a specific output format.

I've also slightly changed the error messages when no devices are found.

Example output:
```
$ ROCR_VISIBLE_DEVICES='' build/release/clients/staging/rocsolver-bench -n 10
rocSOLVER version 3.16.0.2e3d8f64-dirty (with rocBLAS 2.42.0.c98b458c)
Query device error: cannot get device count
No devices found
```

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -408,6 +408,8 @@ try
         print_version_info();
 
         rocblas_int device_count = query_device_property();
+        if(device_count <= 0)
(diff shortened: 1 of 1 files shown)


tfalders (approved):
LGTM!

jzuniga-amd (approved):
Looks good, thanks Cory.

Issue #295, opened 2021-07-23
Segfault with batched syev/heev from ROCm 4.2.0


leofang:
Hello, I am a CuPy core contributor working on supporting the new eigensolvers in CuPy, but I found that the batched version would just segfault (the non-batched version works fine and passes our tests). Below is a super minimal reproducer:
```C++
#include <iostream>
#include <algorithm> // for std::min
#include <stddef.h>  // for size_t
#include <vector>
#include <hip/hip_runtime_api.h> // for hip functions
#include <rocsolver.h> // for all the rocsolver C interfaces and type declarations

int main() {
  int batch_size = 2;
  rocblas_int M = 3;
  rocblas_int N = 3;
  rocblas_int lda = 3;

  rocblas_handle handle;
  rocblas_create_handle(&handle);

  size_t size_A = size_t(lda) * N;

  // A = [ [1, 2, 3,
  //        2, 5, 4,
  //        3, 4, 9],
  //       [2, 2, 3,
  //        2, 6, 4,
  //        3, 4, 10]]
  std::vector<float> hA = {1, 2, 3, 2, 5, 4, 3, 4, 9, 2, 2, 3, 2, 6, 4, 3, 4, 10}; // creates array for matrix in CPU

  float *dA, *dW, *dE;
  hipMalloc(&dA, sizeof(float)*size_A*batch_size);
  hipMalloc(&dW, sizeof(float)*M*batch_size);
  hipMalloc(&dE, sizeof(float)*M*batch_size);
  int *info;
  hipMalloc(&info, sizeof(int)*batch_size);

  hipMemcpy(dA, hA.data(), sizeof(float)*size_A*batch_size, hipMemcpyHostToDevice);
  rocblas_status status;

  #ifdef DO_BATCH
  // this would segfault
  status = rocsolver_ssyev_batched(handle, rocblas_evect_original, rocblas_fill_upper,
                                   N, reinterpret_cast<float* const*>(dA), lda,
                                   dW, N,
                                   dE, N, info, batch_size);
  if (status != 0) {
    std::cout << "got error: " << status << std::endl;
    exit(1);
  }
  #else
  // but this would not
  status = rocsolver_ssyev(handle, rocblas_evect_original, rocblas_fill_upper,
                           N, dA, lda,
                           dW,
                           dE, info);
  if (status != 0) {
    std::cout << "got error: " << status << std::endl;
    exit(1);
  }
  #endif

[...]

Issue #312, opened 2021-08-31
Low performance for medium size matrices


gh6576:
I use the example code to do QR factorization for matrices of different sizes and I get the following results:

QR factorization of a rank 3 matrix took 34701 microseconds
QR factorization of a rank 30 matrix took 1316 microseconds
QR factorization of a rank 300 matrix took 5766996 microseconds
QR factorization of a rank 3000 matrix took 71824 microseconds
QR factorization of a rank 30000 matrix took 473530 microseconds

It looks like calculating a medium-size matrix is really slow. Is it a bug or intended?

jzuniga-amd:
Hi, could you please provide more information on how you are calling the functions? Maybe a sample code? How are you timing the calls? What do you mean by rank 3, 30, 300... ? Are these the sizes of square matrices?

Assuming you are calling `rocsolver_dgeqrf` in double precision for sizes m=n=3,30,300,..., I got the following execution times using the **rocsolver-bench** client: 

m=n=3, 171 microseconds
m=n=30, 1287 microseconds
m=n=300, 14655 microseconds
m=n=3000, 201572 microseconds
m=n=30000, 19750572 microseconds

(QR is not a rank-revealing method so the rank of the matrix does not really matter for the execution time, but we use a random input matrix for the test that we scale to be full-rank anyways.)

So, I don't observe any weird behaviour.

gh6576:
Thanks for your reply. 
I append the code at the end. Nothing much is changed from the example code. I just adjust the matrix and use the chrono package for timing  (I don't have the rocsolver-bench client on the cluster and probably I am not allowed to install it.)
By 'rank', I mean m and n for the square matrices. Sorry for causing confusion. 
And I have more results here (by a for loop):

QR factorization of m=n=2 matrix took 347516 microseconds
QR factorization of m=n=4 matrix took 1367 microseconds
QR factorization of m=n=8 matrix took 463 microseconds
QR factorization of m=n=16 matrix took 791 microseconds
QR factorization of m=n=32 matrix took 1479 microseconds
QR factorization of m=n=64 matrix took 3761 microseconds
QR factorization of m=n=128 matrix took 6394 microseconds
QR factorization of m=n=256 matrix took 5772862 microseconds
QR factorization of m=n=512 matrix took 38466 microseconds
QR factorization of m=n=1024 matrix took 84774 microseconds
QR factorization of m=n=2048 matrix took 166706 microseconds
[...]

PR #323, opened 2021-10-11
Expand documentation samples

cgmb:
I have tried to improve the [samples in the rocsolver user guide](https://rocsolver.readthedocs.io/en/rocm-4.3.1/userguide_examples.html). There are four main motivations for this change:
- Reduce copy/pasted code.
  - The sphinx [literalinclude](https://www.sphinx-doc.org/en/1.5.1/markup/code.html#includes) directive exists for the purpose of including code examples from elsewhere in the tree.
- Expand the sample code so it can be compiled and executed.
  - I think a full, working example using a batched function would have helped avoid the confusion in #295.
- Correct a minor defect in the sample code.
  - There appears to be a bug in the batched sample. The allocation for `dA` should be `sizeof(double*) * batch_count`
```cpp
    size_t size_A = size_t(lda)*N;                 // size of the array that holds one matrix
    ...
    double **dA, *dIpiv;
    hipMalloc(&dA,sizeof(double*) * size_A);    // array on the GPU of pointers on the GPU
    ...
    hipMemcpy(dA, A, sizeof(double*) * batch_count, hipMemcpyHostToDevice);
```
- I'm embarrassed that I missed one of the URLs in #321.
  - I didn't want to make a one-line 'oops' PR.


docs/source/userguide_examples.rst
```diff
@@ -212,74 +61,6 @@ The following code snippet uses rocSOLVER to compute the QR factorization of a s
...
-        for(int b=0;b<batch_count;++b)
-            hipFree(A[b]);
-        hipFree(dA);                      // de-allocate GPU memory
-        hipFree(dIpiv);
-        rocblas_destroy_handle(handle);   // destroy handle
-
-        return 0;
-    }
+.. literalinclude:: ../../clients/samples/example_batched.c
+    :language: c
```
tfalders, line 65: I think it would be best if all the code examples on this page used the same language. We don't necessarily need to get rid of the batched C example (as more C examples are nice to have), but I'd appreciate a separate batched C++ example for inclusion in the user guide.

jzuniga-amd (commented):
Also, can you add `rocblas_initialize()` to the samples (or some of them) and comment on the rocblas initialization overhead that may occur when gemms are called?

(see my comments on the issue #312 Low performance for medium size matrices. A quick comment could help other users in the future)

PR #325, opened 2021-10-15
Added sytrf

tfalders:
Partially addresses SWDEV-291558.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_lasyf.hpp
+++ b/library/src/auxiliary/rocauxiliary_lasyf.hpp
@@ -56,7 +56,7 @@ __device__ void lasyf_device_upper(const rocblas_int tid,
         if(k < n - 1)
         {
             gemv<MAX_THDS>(tid, k + 1, n - k - 1, &minone, A + (k + 1) * lda, lda,
-                                 W + k + (kw + 1) * ldw, ldw, &one, W + kw * ldw, 1);
+                           W + k + (kw + 1) * ldw, ldw, &one, W + kw * ldw, 1);
             __syncthreads();
         }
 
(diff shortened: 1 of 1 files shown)


library/src/lapack/roclapack_sytf2.hpp
```diff
@@ -14,34 +14,20 @@
...
-    using S = decltype(std::real(T{}));
     const S alpha = S((1.0 + std::sqrt(17.0)) / 8.0);
 
-    // get array pointers
-    T* A = load_ptr_batch<T>(AA, bid, shiftA, strideA);
-    rocblas_int* ipiv = ipivA + (bid * strideP);
-
     // local and shared variables
-    __shared__ rocblas_int info;
+    __shared__ rocblas_int _info;
```
cgmb, line 30: I'm curious about this name change. What is the purpose?

cgmb (approved):
LGTM

library/src/lapack/roclapack_sytf2.hpp
tfalders: I changed the name of input parameter infoA to info, so I changed this to avoid a name conflict. And the reason I changed infoA to info is because it's now coming in as a properly offset pointer, i.e. info = infoA + bid.

cgmb: It's not directly related, but I wonder if a naming convention for shared variables might be nice. A trailing underscore (or something like that) would make shared variables easier to distinguish without scrolling back to look up the declarations.

I like the idea, but you and Juan write way more kernels than me. We have more important things to do regardless, but I'm curious what you think.

PR #327, opened 2021-10-20
Separate small-size kernels by precision

tfalders:
This PR distributes the small-size kernel instantiations between several cpp files, allowing the small-size kernels to be compiled in parallel and thereby reduce compilation times.

In order to keep things organized, I moved these cpp files into a new "optimized" directory.


cgmb (commented):
I like this idea in principle and I'm not too fussed about the details, either. We can always change it if we find a better way.

Could you provide some information on how this affects binary size or compile time on your machine? I expect binary size is more or less the same. Maybe slightly larger. Compile time is probably slightly increased overall (in CPU time) but it's vastly outweighed by the greater parallelism that this enables (for a much lower wall-clock time building on a modern workstation).

jzuniga-amd:
I think is better to stop calling these kernels "optimized" because it seems that nothing else is optimized. So I like better "specialized"; these kernels are specialized to deal with small sizes, panel matrices, combining functions (like ger + scale) etc.

PR #329, opened 2021-10-25
Add test suite for tests new to ROCm 5.0

lawruble13:


jzuniga-amd (commented):
Is this still required?
and, for 5.0? Should it be treated as a hotfix and thus targeting this PR to master?
Looks good to me although I am not sure what is this about.

lawruble13:
QA wanted specific filters for new features. We had originally intended to provide these through rtest.xml, however due to the overhead required elected to use another method. This is no longer required.

PR #337, opened 2021-10-28
Add vcpkg manifest

cgmb:
This change adds a [vcpkg manifest](https://vcpkg.readthedocs.io/en/latest/users/manifests/). The vcpkg tool is an alternative way to install dependencies when building rocSOLVER, and the manifest specifies what packages that rocSOLVER depends on. The package versions are specified indirectly, through the "builtin-baseline".

The vcpkg.json file just seems like a good, standardized way of documenting what third-party dependencies we use. The fact that it can install them is more of a bonus. The LAPACK package is a bit weird, though. I needed to change our CMake to get it to work, so I put that behind an option.

To give it a try:
## Ubuntu
```bash
# install vcpkg
sudo apt install build-essential tar curl zip unzip git 
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg
git checkout 603748b5dbed0bcd6aee02025c80260362e7d0df
./bootstrap-vcpkg.sh -disableMetrics

# note your vcpkg install directory
export VCPKG_ROOT="$(pwd)"

# build rocSOLVER
cd ~/rocSOLVER
./install.sh -cna gfx906:xnack- \
  --cmake-arg -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
  --cmake-arg -DVCPKG_MANIFEST_FEATURES=tests \
  --rocblas_dir=$HOME/rocBLAS/build/release/rocblas-install
```

## Windows
```bash
# install vcpkg
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg
git checkout 603748b5dbed0bcd6aee02025c80260362e7d0df
.\bootstrap-vcpkg.bat -disableMetrics

# note your vcpkg install directory
set "VCPKG_PATH=%CD%"

# build rocSOLVER
cd C:\Workspace\rocSOLVER
python rmake.py -cna "gfx906:xnack-;gfx1030" ^
  --rocblas_dir=C:/hipSDK
```

### DIFF
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -2,10 +2,11 @@
   "registries": [
     {
       "kind": "git",
-      "baseline": "21cdaaa75fea30ef19a943d0c5fcc83cfad9525a",
+      "baseline": "c0ed6c5c2828f6dd1fbf2f554445308fce84ae89",
       "repository": "https://github.com/cgmb/rocm-vcpkg",
       "packages": [
         "cblas",
+        "lapack-reference",
         "rocm-cmake"
       ]
     }
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -2,8 +2,8 @@
   "registries": [
     {
       "kind": "git",
-      "baseline": "c0ed6c5c2828f6dd1fbf2f554445308fce84ae89",
-      "repository": "https://github.com/cgmb/rocm-vcpkg",
+      "baseline": "d9ab59cf053aafec41b990793c1e9a704d0efceb",
+      "repository": "git@github.com:ROCmSoftwarePlatform/rocm-vcpkg-registry.git",
       "packages": [
         "cblas",
         "lapack-reference",
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -2,7 +2,7 @@
   "registries": [
     {
       "kind": "git",
-      "baseline": "d665e72c1561a98a28036f1f7f985f9f16dd5458",
+      "baseline": "21cdaaa75fea30ef19a943d0c5fcc83cfad9525a",
       "repository": "https://github.com/cgmb/rocm-vcpkg",
       "packages": [
         "cblas",
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -1,0 +1,12 @@
+{
+  "registries": [
+    {
+      "kind": "git",
+      "baseline": "f9d6e76bda1110cd8612881454c06dcedfc4ceb6",
+      "repository": "https://github.com/cgmb/rocm-vcpkg",
+      "packages": [
+        "cblas"
+      ]
+    }
+  ]
+}
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -2,10 +2,11 @@
   "registries": [
     {
       "kind": "git",
-      "baseline": "f9d6e76bda1110cd8612881454c06dcedfc4ceb6",
+      "baseline": "d665e72c1561a98a28036f1f7f985f9f16dd5458",
       "repository": "https://github.com/cgmb/rocm-vcpkg",
       "packages": [
-        "cblas"
+        "cblas",
+        "rocm-cmake"
       ]
     }
   ]
--- a/bump_rocsolver_version.sh
+++ b/bump_rocsolver_version.sh
@@ -3,9 +3,9 @@
 # run this script in develop after merging develop/staging into master at the feature-complete date
 # Edit script to bump versions for new development cycle/release.
 
-OLD_ROCSOLVER_VERSION="3.17.0"  
-NEW_ROCSOLVER_VERSION="3.18.0"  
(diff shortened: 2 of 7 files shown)


TorreZuk (commented):
Looks good.  Have you tested on all our linux CI OS?
A ticket should be made if you can try and pull out all all the install.sh deps installation and replace with vcpkg

cgmb:
> Have you tested on all our linux CI OS?

No. I should probably do so.

However, I did test on Windows. The fmt and gtest packages work fine, but there are some problems with cblas. It appears to be some sort of problem with the find_package path. I've opened up https://github.com/microsoft/vcpkg/issues/20952.

cgmb:
I'll expand a bit more on why I'm only using `builtin-baseline` for [vcpkg versioning](https://vcpkg.readthedocs.io/en/latest/users/versioning/).

Support for versioning is very new to vcpkg. My experience is that the `version>=` field does not work well yet. There's three mechanisms for selecting the version, which I'll explain; follow along with [this version diagram](https://devblogs.microsoft.com/cppblog/take-control-of-your-vcpkg-dependencies-with-versioning-support/#how-versioning-works-in-vcpkg). The first version mechanism is the `builtin-baseline`, which is mandatory. You cannot use versions without specifying a value for it. The second mechanism is `version>=`, which can only specify versions that are more recent than those specified by the `builtin-baseline`. Finally, there's `override`, which can specify an exact version for a dependency which will be used regardless of what is specified for `builtin-baseline` or `version>=`.

So, you have to specify `builtin-baseline` but `version>=` and `override` are optional. I would not recommend using `version>=` at all, because there are multiple types of versions that are not ordered with respect to each other. Some vcpkg ports will switch back and forth between those incomparable version identifiers, which causes a great deal of trouble with `version>=`. For example, gtest in vcpkg 2021-05-12 uses a version-date of [2021-05-07](https://github.com/microsoft/vcpkg/blob/2021.05.12/ports/gtest/vcpkg.json#L3). Thus, you can't specify gtest as `"version>=": "1.11.0"` if the `builtin-baseline` is set to the 2021-05-12 commit, because `version>=` can only specify versions newer than the baseline and the two versions specified are not comparable. Given that the gtest package in vcpkg went from "1.10.0" to "2021-05-12" and is now "1.11.0", it's clear that using `version>=` will often cause unnecessary problems when you change the baseline.

I would recommend that if you wish to explicitly specify versions, you should use `override`. This is the same vcpkg.json as in my PR, but using explicit versions:
```json
{
  "name": "rocsolver",
  "version": "3.17.0",
  "description": "AMD GPU implementation of LAPACK",
  "homepage": "https://github.com/ROCmSoftwarePlatform/rocSOLVER",
  "documentation": "https://rocsolver.readthedocs.io",
  "builtin-baseline": "5568f110b509a9fd90711978a7cb76bae75bb092",
  "dependencies": [
    "fmt"
  ],
  "features": {
    "tests": {
      "description": "The rocsolver-test client",
      "dependencies": [
        "gtest",
        {
          "name": "lapack-reference",
          "default-features": false,
          "features": [
            "cblas"
          ]
        }
      ]
    }
  },
  "overrides": [
    {
      "name": "fmt",
      "version": "7.1.3",
      "port-version": 4
    },
    {
      "name": "gtest",
      "version-date": "2021-05-07"
    },
    {
      "name": "lapack-reference",
      "version": "3.8.0",
      "port-version": 4
    }
  ]
}
```

In this case, the versions are the same as those specified by the `builtin-baseline`, so it's redundant. Still, I do find it more clear about what versions you're installing.

jzuniga-amd (approved):

tfalders (approved):

cgmb:
> Looks good. Have you tested on all our linux CI OS?

The cblas feature stopped working once I started trying other Linux distros. I'm suspicious that it never actually worked, and that I just didn't have a properly clean environment when I tested on Ubuntu.

I've converted to a draft because I'm changing tack somewhat. Instead, I've created my own package for cblas. I took the same opportunity to create a rocm-cmake. This was made possible via the [vcpkg registries](https://devblogs.microsoft.com/cppblog/registries-bring-your-own-libraries-to-vcpkg/) feature.

vcpkg-configuration.json
```diff
@@ -0,0 +1,14 @@
+{
+  "registries": [
+    {
+      "kind": "git",
+      "baseline": "c0ed6c5c2828f6dd1fbf2f554445308fce84ae89",
+      "repository": "https://github.com/cgmb/rocm-vcpkg",
```
TorreZuk, line 6: Could this just be a subdirectory in rocSOLVER repo?   Each library provides it's own overlay?

cgmb: Technically it probably could be. However, I don't think it would be a good idea. These packages are complex to develop and maintain, so it would be better to centralize it. In particular, package versioning works by directly interacting with git internals (like [git-tree objects](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects)).

cgmb:
There are still some blockers to moving forward with this. I'll close it for now, and reopen it when I'm ready to pick it up again...

PR #338, opened 2021-10-28
Update external-gtest.cmake by bumping googletest to release-1.11.0

searlmc1:
Bump googletest to release-1.11.0 to account for newer clang compiler, which emits warnings that turn to errors ( 'PolymorphicAction<testing::internal::ReturnNullAction>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy] )


cgmb:
The update has been merged for rocBLAS and the math-ci. I think we can go ahead and merge this.

PR #344, opened 2021-11-07
syrkx offset hot fix

amcamd:
- This PR should be merged after rocBLAS-internal PR 957
- It is a hot-fix to master branch because rocBLAS-internal PR 957 fixes a bug
- Tests have been run on this PR together with rocBLAS PR 957. rocSOLVER builds correctly, and all rocSOLVER tests pass
- The rocBLAS-internal PR 957 changes the following rocblas_internal APIs
  - rocblas_internal_syr2k_template
  - rocblas_internal_her2k_template
  - rocblas_internal_gemm_template
- There are two changes to the APIs    
  - A template parameter BATCHED is added for the two functions rocblas_internal syr2k and her2k functions
  - For the three listed functions the offset argument is changed from rocblas_int datatype to rocblas_stride datatype
- The reason for the rocBLAS-internal changes is because if matrix size gets large, offset can exceed the rocblas_int maximum size of 2147483647. rocblas_int is typedef to int32_t (rocblas_stride is typedef to int64_t)


library/src/lapack/roclapack_sytrd_hetrd.cpp
```diff
@@ -14,6 +14,7 @@ rocblas_status rocsolver_sytrd_hetrd_impl(rocblas_handle handle,
                                           S* E,
                                           T* tau)
 {
+    constexpr bool BATCHED = false;
```
tfalders, line 17: To maintain consistency with our other files, I would ask for this line to be removed and for true/false to be passed directly to the rocsolver_sytrd_hetrd_template function. This also applies to the batched and strided_batched cpp files.

tfalders (approved):
I have one nitpicky comment, but other than that it looks good.

jzuniga-amd (approved):

library/src/lapack/roclapack_sytrd_hetrd.cpp
amcamd: Thanks for review, In below commit 7f0e8ca constexpr bool BATCHED template argument is removed and replaced by true/false.

cgmb (approved):

PR #345, opened 2021-11-12
Add attributions

tfalders:
Addresses SWDEV-308187.

I've added copyright notices for LAPACK and MAGMA to our license file, and included the license file in our readthedocs documentation. I've also added a copyright notice to the file that was flagged by our scan.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2018-2020 Advanced Micro Devices, Inc.
+Copyright (c) 2018-2021 Advanced Micro Devices, Inc.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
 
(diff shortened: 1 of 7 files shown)


jzuniga-amd (approved):
As per our last discussion, it looks good to me.

doctorcolinsmith (approved):

cgmb (approved):

saadrahim (approved):
adding m2r2 should be checked. Not sure if readthedocs is including that.

tfalders:
> adding m2r2 should be checked. Not sure if readthedocs is including that.

The readthedocs appears to be building correctly. We've tested m2r2 previously with the hipSOLVER documentation too.

cgmb:
> adding m2r2 should be checked. Not sure if readthedocs is including that.

ReadTheDocs uses that requirements file to determine what packages we need. We configured it that way in [.readthedocs.yaml](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/fec75b57a438e6695c61b055cce508559bcf653b/.readthedocs.yaml#L14).

cgmb:
This will needed in master for 5.0.

PR #351, opened 2021-11-19
ROCm package file/folder reorganization(SWDEV-291455)

arvindcheru:
Summary of proposed changes:

- HIP Path updated to rocm path (/opt/rocm)
- Package file/folder reorg applied for lib (rocmpath/lib) (removed component name prefix)
- Package file/folder reorg applied for include (rocmpath/include/<compnm>)
- Dep Changes for header inclusions updated to <compnm>/.h
- Dep Changes for rocblas header inclusions (reorg applied rocblas)


jzuniga-amd (commented):
Is this a hotfix for 5.0?, a late feature for 5.0? or a normal feature for the next release (5.1)?
If it is a hotfix, shouldn't it be targeting the master branch?

CMakeLists.txt
```diff
@@ -61,17 +61,15 @@ get_os_id(OS_ID)
 message(STATUS "OS detected is ${OS_ID}")
 
 # Versioning via rocm-cmake
-set(VERSION_STRING "3.16.0")
+set(VERSION_STRING "3.17.0")
```
jzuniga-amd, line 64: Following the discussion on teams, if this is a hotfix (late feature) for 5.0, you shouldn't be changing rocSOLVER versions here (at least not if we want to merge into master before the branching date)
So, I guess you should drop commits 
91ab703, 95d308f, and 1c6be9f from this PR

arvindcheru: Thanks for comments. We have just now dropped it, Please take a look.

jzuniga-amd (approved):
Looks good to me.
Just wait for @cgmb and @tfalders approvals

jzuniga-amd:
tests are not passing, though. Seems that the project cannot build, but @cgmb can maybe say more.

cgmb:
> tests are not passing, though. Seems that the project cannot build, but @cgmb can maybe say more.

SWDEV-291455 involves a breaking change in source compatibility in every math library. This PR  depends on the rocBLAS change, which hasn't been merged. The math-ci will fail until that happens, so the real testing will be done manually.

At least our users have version numbers to identify which version is which and `#ifdef` as needed... speaking of which, I don't think the `ROCSOLVER_VERSION_{MAJOR,MINOR,PATCH,TWEAK}` macros are documented. Those will be important for our users in handling this change.

Actually, to that point, maybe we can actually handle this transition better. The version number we originally chose for this release is not valid. IMO, this is a breaking change, so it should be rocSOLVER 4.0.0 rather than 3.16.0. The same is true of rocBLAS. We could actually check the rocBLAS version and keep the CI build working... rather pointless for end users, but perhaps useful for us.

cgmb (changes requested):
We need to sort out the version number before this can be merged.

saadrahim:
Good point on changing version number @cgmb. We should request that in all the libraries.

cgmb:
> `ROCSOLVER_VERSION_{MAJOR,MINOR,PATCH,TWEAK}` [...] will be important for our users in handling this change.

I suppose that the ones that are undocumented (in `rocsolver-version.h`) and the ones that are useful for this purpose (in `rocsolver-config.cmake`) are different variables.

If we were to use them to allow the CI to build with either version of rocBLAS, our clients (including sample programs) would depend on variables defined through the build system. That doesn't sound very appealing to me. It's probably not worth jumping through hoops to fix the CI builds. This sort of breaking change happens internally a few times an iteration anyway.

PR #355, opened 2021-11-30
Fix memory model tests

jzuniga-amd:
The optimized GETRF requires less memory than before and our memory_model tests were failing. Here is a quick fix.

### DIFF
--- a/.jenkins/debug.groovy
+++ b/.jenkins/debug.groovy
@@ -40,7 +40,7 @@ def runCI =
     {
         platform, project->
 
-        def gfilter = '*checkin_lapack*'
+        def gfilter = 'checkin*'
         commonGroovy.runTestCommand(platform, project, gfilter)
     }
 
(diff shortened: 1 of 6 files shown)


cgmb:
SWDEV-313490.

This is a fragile test and we're not running it on CI. Ideally we'd fix both those problems, but the latter is the easier of the two.

I'm thinking of changing the test name to `checkin_misc_MANAGED_MEMORY` and our psdb test filter to `checkin*` rather than `*checkin_lapack*`.

tfalders (approved):

jzuniga-amd:
Yes, sounds good to me...
I was wondering why we did not catch the error with the PR...
So, enabling this test on CI with the new naming is fine. 
If you can do the modifications in this PR we can merge them all together

cgmb:
I've opened https://github.com/jzuniga-amd/rocSOLVER/pull/7 with my suggested changes.

cgmb (approved):
LGTM, but let's just be sure the tests work on CI before merging.

PR #352, opened 2021-11-19
Fixing trsv rocblas template call

daineAMD:

PR #356, opened 2021-12-03
rocblas trsm memory interface change

daineAMD:
See https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/995.

Ended up being more changes than expected. Let me know if you see anything off. Will also need successful CI and CQE approval.


library/src/include/lapack_host_functions.hpp
```diff
@@ -58,8 +58,9 @@ void rocsolver_trsm_mem(const rocblas_side side,
         //  blk cannot be multiple of 128.)
         //        rocblas_int mm = (blk % 128 != 0) ? blk : blk + 1;
         rocblas_int mm = (m % 128 != 0) ? m : m + 1;
-        rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, mm, n, batch_count, size_work1,
-                                         size_work2, size_work3, size_work4);
+        // rocblas_operation doesn't matter here as we're ensuring m is not a multiple of 128 and isn't used in this case
```
tfalders, line 61: It looks like the rocsolver_trsm method that this mem function is associated with always uses `rocblas_operation_none` for the operation, so this comment can probably be removed.

library/src/lapack/roclapack_potrs.hpp
```diff
@@ -66,8 +66,9 @@ void rocsolver_potrs_getMemorySize(const rocblas_int n,
     }
 
     // workspace required for calling TRSM
-    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, n, nrhs, batch_count, size_work1,
-                                     size_work2, size_work3, size_work4);
+    // rocblas_operation_none will always allocate at least as much as the transpose
```
tfalders, line 69: I'm a little uncomfortable building this assumption into our code. Even if it's true now, this could change on the rocBLAS side and introduce a bug without our knowledge. I'd prefer that we take the maximum of the transpose and no-op cases.

library/src/lapack/roclapack_sygst_hegst.hpp
```diff
@@ -59,10 +59,11 @@ void rocsolver_sygst_hegst_getMemorySize(const rocblas_eform itype,
...
             // extra requirements for calling TRSM
-            rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, n - kb, kb, batch_count, &temp1,
-                                             &temp2, &temp3, &temp4);
-            rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_right, n - kb, kb, batch_count, &temp5,
-                                             &temp6, &temp7, &temp8);
+            // rocblas_operation_none will always allocate at least as much as the transpose
+            rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, rocblas_operation_none, n - kb, kb,
+                                             batch_count, &temp1, &temp2, &temp3, &temp4);
+            rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_right, rocblas_operation_none, n - kb, kb,
+                                             batch_count, &temp5, &temp6, &temp7, &temp8);
```
tfalders, line 66: Same comment as for potrs. I would suggest the following alternative:

```
// extra requirements for calling TRSM
if (uplo == rocblas_fill_upper)
{
    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, rocblas_operation_conjugate_transpose, n - kb, kb, batch_count, &temp1, &temp2, &temp3, &temp4);
    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_right, rocblas_operation_none, n - kb, kb, batch_count, &temp5, &temp6, &temp7, &temp8);
}
else
{
    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_left, rocblas_operation_none, n - kb, kb, batch_count, &temp1, &temp2, &temp3, &temp4);
    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_right, rocblas_operation_conjugate_transpose, n - kb, kb, batch_count, &temp5, &temp6, &temp7, &temp8);
}
```

tfalders (changes requested):
Looks good! I have just a couple suggested changes.

daineAMD:
Thanks for the review! Just as a note: the rocBLAS/rocSOLVER changes have been approved by Mani, rocBLAS has passed CI, and I have locally built rocSOLVER and ran rocSOLVER tests successfully.

Would we potentially be ok with these changes going through as-is and making the suggested changes in a PR to develop after? I'm asking since these changes have been through the approval process successfully, and as long as users aren't mixing rocBLAS/rocSOLVER versions then the issue you mentioned shouldn't be relevant. That being said, I definitely agree with your comments.
If not, I'm happy to make the changes, but it should probably go through the approval process again. Let me know what you think!

library/src/lapack/roclapack_potrs.hpp
cgmb: I agree. That's exactly the sort of thing I created [`ROCSOLVER_ASSUME_X`](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/e93586ccfabfe089c58a089030eac59e3e5dd197/library/src/lapack/roclapack_gels.hpp#L81) for. Though, a conceptually simpler alternative would be to put the call and an assert behind `#ifndef NDEBUG`.

cgmb (approved):
IMO, it's ok to merge as-is. I agree with Troy's suggestions, but perhaps we can target those improvements at 5.1.

tfalders (approved):
Sure, we can merge this now, and fix it up later.
If there are any other hotfixes going into 5.0 on the rocSOLVER side, perhaps we could even sneak in the change along with that hotfix.

jzuniga-amd (approved):

jzuniga-amd:
What about the changes in #352 ?
Are you pushing the corresponding changes in rocBLAS to the master branch as well?

daineAMD:
https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/995 is directed towards master in rocBLAS. The changes in #352 shouldn't overlap (much?) with the changes here.

daineAMD:
Thanks for all the comments everyone, I'll merge this in about half an hour unless I hear otherwise, then will create a PR into develop later today with the suggested changes.

PR #357, opened 2021-12-10
Pin 'mistune' to last known good version

cgmb:
This change fixes the error:

    Module 'mistune' has no attribute 'BlockGrammar'

The mistune module is used by m2r2, but the version was not specified
tightly, so the release of mistune 2.0 caused a breakage due to
interface changes.

SWDEV-314585


cgmb:
See also https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/70

tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #360, opened 2022-01-07
Revert "Workaround for debug binary size on CI"

cgmb:
Reverts ROCmSoftwarePlatform/rocSOLVER#265. That was a temporary workaround and it has outlived its usefulness.

Troy addressed the binary size problem in #289. The debug build is our early warning system for library size, so we should keep it at parity with the Release build.

### DIFF
--- a/.jenkins/debug.groovy
+++ b/.jenkins/debug.groovy
@@ -21,7 +21,7 @@ def runCI =
     prj.defaults.ccache = true
 
     // customize for project
-    prj.paths.build_command = './install.sh -a "gfx900;gfx906:xnack-" -c'
+    prj.paths.build_command = './install.sh -c'
 
     // Define test architectures, optional rocm version argument is available
     def nodes = new dockerNodes(nodeDetails, jobName, prj)


cgmb:
There's interest from a number of parties for building with debug symbols (e.g., SWDEV-309024). The Debian project is also now starting to package our libraries, and they build with debug symbols by default. We'll want to be sure our library is small enough that it continues to compile successfully despite the addition of debug symbols.

tfalders (approved):

jzuniga-amd (approved):

cgmb:
I'm a little concerned about the CI test failures on Linux with gfx90a, though they're certainly not associated with this change:
```
[  FAILED  ] daily_lapack/GESVD.batched__float_complex/19, where GetParam() = ({ 300, 120, 1 }, { 0, 0, 0, 1, 3 })
[  FAILED  ] daily_lapack/GESVD.strided_batched__double_complex/19, where GetParam() = ({ 300, 120, 1 }, { 0, 0, 0, 1, 3 })
```

The test failures on Windows are because the new HIP package is missing `hipinfo.exe`. One day it will pass...

PR #364, opened 2022-01-18
Add gfx1010 support

cgmb:
I picked up an RX 5700 XT to give gfx1010 a try. rocBLAS has supported it since ROCm 4.2, and I figured we might as well enable it in rocSOLVER if it works.

The full rocsolver-test suite passes after a small fix to `sytrf_kernel_lower`. The lack of a sync resulted in incorrect pivot indices returned for the sytrf family of functions when called with `rocblas_fill_lower`. In theory, this sync is needed on all architectures, but the bug doesn't seem to manifest except on gfx1010. Thanks to @tfalders for identifying the cause of this problem.

### DIFF
--- a/library/src/lapack/roclapack_sytrf.hpp
+++ b/library/src/lapack/roclapack_sytrf.hpp
@@ -5,7 +5,7 @@
  *     Univ. of Colorado Denver and NAG Ltd..
  *     December 2016
  *
- * Copyright (c) 2019-2021 Advanced Micro Devices, Inc.
+ * Copyright (c) 2019-2022 Advanced Micro Devices, Inc.
  * ***********************************************************************/
 
 #pragma once
@@ -118,6 +118,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(SYTRF_MAX_THDS)
             sytf2_device_lower<SYTRF_MAX_THDS>(tid, n - k, A + k + k * lda, lda, ipiv + k, &iinfo,
                                                sidx, sval);
             ktemp = n;
+            __syncthreads();
         }
 
         if(tid == 0 && iinfo != 0 && info[bid] == 0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- Fixed missing synchronization in SYTRF with `rocblas_fill_lower` that could potentially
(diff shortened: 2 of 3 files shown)


tfalders (approved):
Looks good, but won't this increase our binary size?

library/src/lapack/roclapack_sytrf.hpp
```diff
@@ -118,6 +118,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(SYTRF_MAX_THDS)
             sytf2_device_lower<SYTRF_MAX_THDS>(tid, n - k, A + k + k * lda, lda, ipiv + k, &iinfo,
                                                sidx, sval);
             ktemp = n;
+            __syncthreads();
```
jzuniga-amd, line 121: I don't see why this sync is needed. Is there evidence that this is failing?

tfalders: Apparently it fails on gfx1010. The last `ipiv` value can get mangled because there is no `syncthreads` between where `sytf2_device_lower` sets the last `ipiv` on thread 0 and where `sytrf_kernel_lower` updates all `ipiv` values on all threads.

cgmb:
Yes, though you'll see that I've enabled the debug build for this PR and it has completed successfully.

In fact, we could get away with adding gfx1011 too, as that would bring us up to 1.9 GB (and I'm tempted to do so). Adding gfx1012 to match rocBLAS, however, would bring us up to 2.1 GB, which is too close for comfort to the 2.147 GB limit.

library/src/lapack/roclapack_sytrf.hpp
cgmb: [sytrf_gfx1010.txt](https://github.com/ROCmSoftwarePlatform/rocSOLVER/files/7891301/sytrf_gfx1010.txt)
Yes. This is the log before the fix was applied (with some additional failure information added).

TorreZuk:
You probably shouldn't just add default architectures without wider approval, but nor do I know why you don't match the rocblas set of arch even after this addition, see https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/a5502eeb7333d1974ef871b7ab6d299e54fe3dff 
I would suggest you continue that discussion in a channel with Daine and Juan.

cgmb:
> nor do I know why you don't match the rocblas set of arch even after this addition

We've always tried to match the rocBLAS architectures, but we couldn't due to binary size limitations. The size optimizations we've made just allow us to get closer.

jzuniga-amd (approved):

PR #365, opened 2022-01-18
Remove logic for installing LICENSE.md

cgmb:
The license installation is handled by rocm-cmake as of https://github.com/RadeonOpenCompute/rocm-cmake/pull/63.

PR #369, opened 2022-01-24
Hotfix/rpm-license

lawruble13:
Change requested by legal to specify the license type in RPM packages.

Also only install the license using the mechanism in rocm-cmake.


cgmb (approved):
I noticed that the value `CPACK_RPM_PACKAGE_LICENSE` doesn't match the SPDX identifier for our license, `BSD-2-Clause`, so I thought I'd dig into that.

It's not documented in CMake how to choose the value for `CPACK_RPM_PACKAGE_LICENSE`, but ultimately it ends up as the `License` field of the RPM specfile. So, I suppose that [the Fedora packaging guidelines](https://fedoraproject.org/wiki/Packaging:LicensingGuidelines?rd=Packaging/LicensingGuidelines) are the thing to follow. That document explains that the license field should be set with the short name from the [Fedora Software License List](https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Software_License_List), which is `BSD`. So, LGTM!

This change invalidates #365, which will be addressed afterwards.

tfalders (approved):

lawruble13:
> So, I suppose that the Fedora packaging guidelines are the thing to follow.

Yes, we have been using the Fedora packaging guidelines for this field, as RH and SUSE also generally copy Fedora's short name. Thanks for bringing that up, good to clarify!

jzuniga-amd (approved):

cgmb:
Please use "squash and merge" when merging.

PR #370, opened 2022-01-29
Fix multi-level logging file output

cgmb:
It is not valid to open a stream using both `std::ios::app` and `std::ios::trunc`. The flags are contradictory, and were added by mistake in 659b64c8e0fbfa5c360dc25d411f80003f47f84c. The default behaviour of `std::ofstream` is equivalent to `std::ios::out | std::ios::trunc`. This change reverts to using the same flags as when the feature was introduced in b193d895b9b43f797b9d66afa973cdb80ee135f6.

See SWDEV-319538.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Fixed
 - Fixed missing synchronization in SYTRF with `rocblas_fill_lower` that could potentially
   result in incorrect pivot values.
+- Fixed multi-level logging output to file with the `ROCSOLVER_LOG_PATH`,
+  `ROCSOLVER_LOG_TRACE_PATH`, `ROCSOLVER_LOG_BENCH_PATH` and `ROCSOLVER_LOG_PROFILE_PATH`
+  environment variables.
 
 
 ## rocSOLVER 3.16.0 for ROCm 5.0.0
--- a/library/src/common/rocsolver_logger.cpp
+++ b/library/src/common/rocsolver_logger.cpp
@@ -229,8 +229,10 @@ try
     logger->trace_os = logger->open_log_stream("ROCSOLVER_LOG_TRACE_PATH");
     logger->bench_os = logger->open_log_stream("ROCSOLVER_LOG_BENCH_PATH");
     logger->profile_os = logger->open_log_stream("ROCSOLVER_LOG_PROFILE_PATH");
-
-    return rocblas_status_success;
+    if(logger->trace_os->good() && logger->bench_os->good() && logger->profile_os->good())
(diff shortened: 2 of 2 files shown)


clients/gtest/logging_gtest.cpp
```diff
@@ -0,0 +1,338 @@
...
+        ".*getrf.*Calls: 1, Total Time: .+ .+ .in nested functions: .+ .+.",
+        "\\s*",
+        ".*PROFILE.*",
+        ".*getrf.*Calls: 1, Total Time: .+ .+ .in nested functions: .+ .+.",
+        "\\s*",
+    };
+
+    verify_file(profile_filepath, expected_lines);
+
+    ASSERT_EQ(rocsolver_log_end(), rocblas_status_success);
```
tfalders, line 231: Shouldn't this be called before the file is verified?

clients/gtest/logging_gtest.cpp
```diff
@@ -0,0 +1,338 @@
...
+        ".*getrf.*Calls: 1, Total Time: .+ .+ .in nested functions: .+ .+.",
+        "\\s*",
+        ".*PROFILE.*",
+        ".*getrf.*Calls: 2, Total Time: .+ .+ .in nested functions: .+ .+.",
+        "\\s*",
+    };
+
+    verify_file(profile_filepath, expected_lines);
+
+    ASSERT_EQ(rocsolver_log_end(), rocblas_status_success);
```
tfalders, line 273: Same comment as above

clients/gtest/logging_gtest.cpp
```diff
@@ -0,0 +1,338 @@
...
+        "\\s*",
+    };
+
+    verify_file(trace_filepath, expected_lines);
+
+    if(!HasFailure())
+        EXPECT_TRUE(fs::remove(trace_filepath));
+}
+
+TEST_F(checkin_misc_LOGGING, rocsolver_log_path)
```
tfalders, line 113: Incredibly minor, but it feels like this test belongs above the rocblas_layer_mode_log_trace test.

tfalders (approved):
Looks good! Just some minor comments.

clients/gtest/logging_gtest.cpp
cgmb: No. This test is checking that `rocsolver_log_write_profile` has actually written the file. This line is to close the log file before I delete it. That's unnecessary on Linux, but I believe it's required on Windows.

It's a good question, though. That's definitely not obvious from looking at the code. I'll see if I can make that more clear.

cgmb: I've moved all tests to #372 as per discussion. It's still marked as draft, as I've not yet addressed your comments (or Juan's).

tfalders (approved):

PR #373, opened 2022-02-09
Hotfixes from release/rocm-rel-5.0 at release 5.0

ROCmMathLibrariesBot:
This is an autogenerated PR.
This is intended to pull any hotfixes for ROCm release 5.0 (including changelogs and documentation) back into develop.


cgmb:
This should probably actually be going to master, so we don't regress by dropping commits between 5.0 and 5.1.

jzuniga-amd (approved):

PR #378, opened 2022-02-23
Minor rocsolver-test improvements

cgmb:
### Remove redundant unit_check variable
The code can check if it's being built for rocsolver-test or rocsolver-bench by testing for the `ROCSOLVER_CLIENTS_TEST` macro.

### Use printable_char to improve gtest messages
    
With `char`:

    [  FAILED  ] checkin_lapack/SYTRF.batched__float/10, where GetParam() = ({ 70, 100, 1 }, 'L' (76, 0x4C))
    
With `printable_char`:

    [  FAILED  ] checkin_lapack/SYTRF.batched__float/10, where GetParam() = ({ 70, 100, 1 }, L)


tfalders:
I'm a fan of the `printable_char` change, but I don't know how I feel about removing `unit_check`. I does seem like a rather pointless variable, but it's not expensive to keep and highlights the difference (such as it is) between the rocsolver-test error checks and the rocsolver-bench error checks.

clients/include/testing_gebd2_gebrd.hpp
```diff
@@ -592,10 +592,9 @@ void testing_gebd2_gebrd(Arguments& argus)
                                                        hot_calls, argus.profile, argus.perf);
     }
 
-    // validate results for rocsolver-test
-    // using m*n * machine_precision as tolerance
-    if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, m * n);
+#ifdef ROCSOLVER_CLIENTS_TEST
+    ASSERT_LE(max_error, m * n * machine_precision<T>());
```
jzuniga-amd, line 596: this `machine_precision()` function here should be returning 1/2*epsilon.
Alternatively, use `1/2*machine_epsilon()`

common/include/common_host_helpers.hpp
```diff
@@ -22,6 +22,13 @@
  * ===========================================================================
  */
 
+template <typename T>
+constexpr double machine_precision()
```
jzuniga-amd, line 26: this function should be called machine_epsilon or get_epsilon as before

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -560,7 +560,7 @@ rocblas_status rocsolver_bdsqr_template(rocblas_handle handle,
 
     // set tolerance and max number of iterations:
     // machine precision (considering rounding strategy)
-    S eps = get_epsilon<S>() / 2;
+    S eps = machine_precision<S>() / 2;
```
jzuniga-amd, line 563: Here we need the machine precision, ie. `machine_epsilon()/2`, or simply `machine_precision()`  if you add such function.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -391,7 +391,7 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
         rocblas_int blocks = (n - 1) / 32 + 1;
 
         // constants
-        S eps = get_epsilon<S>();
+        S eps = machine_precision<S>();
```
jzuniga-amd, line 394: this should be calling `machine_epsilon()` or `get_epsilon()`

library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -417,7 +417,7 @@ rocblas_status rocsolver_steqr_template(rocblas_handle handle,
                                 stream, n, n, C, shiftC, ldc, strideC);
     }
 
-    S eps = get_epsilon<S>();
+    S eps = machine_precision<S>();
```
jzuniga-amd, line 420: this should be calling `machine_epsilon()` or `get_epsilon()`

library/src/auxiliary/rocauxiliary_sterf.hpp
```diff
@@ -354,7 +354,7 @@ rocblas_status rocsolver_sterf_template(rocblas_handle handle,
     if(n <= 1)
         return rocblas_status_success;
 
-    T eps = get_epsilon<T>();
+    T eps = machine_precision<T>();
```
jzuniga-amd, line 357: this should be calling `machine_epsilon()` or `get_epsilon()`

library/src/include/lib_host_helpers.hpp
```diff
@@ -19,18 +19,11 @@
...
-    using S = decltype(std::real(T{}));
-    return std::numeric_limits<S>::epsilon();
-}
-
 template <typename T>
 constexpr double get_safemin()
 {
     using S = decltype(std::real(T{}));
-    auto eps = get_epsilon<S>();
+    auto eps = machine_precision<S>();
```
jzuniga-amd, line 26: this should be calling `machine_epsilon()` or `get_epsilon()`

jzuniga-amd (changes requested):
a. The printable_char changes are very good!

b. Unit_check and norm_check were not really redundant because, at the beginning, there was no difference between the google-test client and the bench client (if you built clients, you built both). Now that you can build one or the other (or both) the proposed solution makes more sense… but you still need to verify whether you are running the test client to assert the error value, right? Whether you do it with macros or with if(), does not matter to me, I just want to confirm that when running the bench  client there is not verification (assert) on the error value.
 
c. epsilon and machine precision are, strictly speaking, not the same thing!! Epsilon (or machine epsilon) is the distance between 1 and the next real number representable in the floating point system in the given machine (for IEEE double precision, epsilon = 2^(-52). Machine precision is the max relative error that you could have when representing any real number in your floating point system (so, considering rounding to the nearest by default, machine_precision = 1/2 * epsilon = 2^(-53)).  Sometimes people use machine_epsilon and machine_precision as the same thing but this is not correct! Again, epsilon refers to a distance between numbers in your system while precision has to do with the error that you have when “rounding” a general number so that it can be “represented” in your system. 

**So please don’t have just one function returning the same value for epsilon and machine precision. We need both. We need epsilon for the library code (in steqr, stedc, etc.) and we need machine_precision = 1/2 * epsilon to validate the error in our computations in the test and bench clients.** 

The original macro ROCSOLVER_TEST_CHECK in rocsolver_test.hpp should compare against tol*get_epsilon()/2, i.e. tol*machine_precision. I removed the /2 just to have the original rocsolver routines (getf2 and getrf) to pass the tests when I did the very first library and clients refactoring years ago. I was planning to halve epsilon again once I worked out getf2 and getrf, but I suppose I forgot. So, we have been testing all our functions with a slightly bigger tolerance since then…. Not a big deal; in theory, the error bounds that we use accept a “small constant factor c”; c = 2 is just fine, but anyways, to make this correctly:

**Please do required changes and verify that all our tests pass with machine_precision = 1/2*epsilon. We have been very careful with our code and the code in our unit tests (implementing implicit tests when needed, etc.) so I am sure that there should not be any problem, but if you see any failing tests, let me know… we may need to review the routine, or it could be that we simply need c=2 in that case.**

clients/include/testing_bdsqr.hpp
```diff
@@ -584,14 +584,11 @@ void testing_bdsqr(Arguments& argus)
...
 
-    // validate results for rocsolver-test
-    // using 2 * n * machine_precision as tolerance
-    if(argus.unit_check)
-    {
-        ROCSOLVER_TEST_CHECK(T, max_error, 2 * n);
-        if(nv || nu || nc)
-            ROCSOLVER_TEST_CHECK(T, max_errorv, 2 * n);
-    }
+#ifdef ROCSOLVER_CLIENTS_TEST
```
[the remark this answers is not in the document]
cgmb: Regardless of what build options you've selected, the `ROCSOLVER_CLIENTS_TEST` macro is only defined in rocsolver-test source files. In both my changes and the current develop, all asserts are removed from rocsolver-bench by the pre-processor, as the asserts are provided by Google Test (which is not linked with rocsolver-bench).

cgmb:
> So please don’t have just one function returning the same value for epsilon and machine precision. We need both. We need epsilon for the library code (in steqr, stedc, etc.) and we need machine_precision = 1/2 * epsilon to validate the error in our computations in the test and bench clients.

Absolutely. I have no intention to change any of the values that are being used. I wasn't ready to return to this yesterday, but my question for you was just going to be what we wanted to name these functions. My intention is to define two functions, give them clear names, and include a comment on each function that defines the value.

It is definitely is not ready to merge as-is. I probably should have converted it to a draft after renaming `get_epsilon` to `machine_precision` (as per the existing comments in rocsolver-test) and realizing that I had two different functions with the same name. It shouldn't even compile except that I moved the renamed function to a different header and forgot to update my include statements.

clients/include/testing_bdsqr.hpp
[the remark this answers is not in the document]
jzuniga-amd: Ok, I understand now...

jzuniga-amd:
If you want to have the two functions, then the names `machine_epsilon()` (that returns numeric_limits::epsilon) and `machine_precision` (that returns 1/2 * epsilon) are good to me

cgmb:
> Please do required changes and verify that all our tests pass with machine_precision = 1/2*epsilon

I've completed the renaming, but I don't want to change the pass/fail criteria. My goal here was to make it more clear what the code was doing, and I think that has been achieved. We've discovered that we've been using wider bounds than we thought, but perhaps we could defer the task of narrowing them into the backlog? We may need to adjust our error bounds anyway if we're to consider comparing our results against another reference, like OpenBLAS.

jzuniga-amd:
Yes, we can do that in a future PR (not super urgent).

jzuniga-amd (approved):

tfalders:
The changes to `argus.norm_check` still bug me, but I think I would be happier if we also removed `argus.timing` and used ifs in normal code (with defined constants rather than `argus` variables) rather than preprocessor ifs.

Probably too big of a change to make it into this release, but I'm reluctant to approve in the current state.

cgmb:
This is not fundamentally necessary. It can wait.

cgmb:
I'm going to give the `unit_check` change more thought and maybe come back to it later.

PR #379, opened 2022-02-25
File/Folder Reorg Changes using rocm-cmake wrapper functions

arvindcheru:
- File/Folder Reorg Changes
- Backward Compatibility enabled using rocm-cmake wrapper functions.


cgmb:
This looks a lot cleaner!

One thing that was discussed with regards to https://github.com/RadeonOpenCompute/rocm-cmake/pull/70 was that we could remove the need for a lot of these one-off options if we just forwarded install.sh arguments that begin with `-D` through to the underlying CMake.

I'm fine with `--rm-legacy-include-dir`, but you might find a general solution that allows `./install.sh -DBUILD_FILE_REORG_BACKWARD_COMPATIBILITY=ON` to give you more options in the future.

library/src/CMakeLists.txt
```diff
@@ -281,22 +289,29 @@ endif()
...
-  PREFIX rocsolver
   DEPENDS
     PACKAGE hip
     PACKAGE rocblas
   NAMESPACE roc::
 )
 
-rocm_install_symlink_subdir(rocsolver)
+if(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)
+  install(
```
lawruble13, line 311: Make sure this is `rocm_install` to ensure that the headers are installed to the `devel` package, rather than the `runtime` package.

arvindcheru:
> This looks a lot cleaner!
> 
> One thing that was discussed with regards to [RadeonOpenCompute/rocm-cmake#70](https://github.com/RadeonOpenCompute/rocm-cmake/pull/70) was that we could remove the need for a lot of these one-off options if we just forwarded install.sh arguments that begin with `-D` through to the underlying CMake.
> 
> I'm fine with `--rm-legacy-include-dir`, but you might find a general solution that allows `./install.sh -DBUILD_FILE_REORG_BACKWARD_COMPATIBILITY=ON` to give you more options in the future.

OK will note this

library/src/CMakeLists.txt
arvindcheru: OK, Will update this

cgmb:
Please take a look at the ROCmMathLib-Windows-CI/pr-merge check. The build is failing on Windows:
```
In file included from ../../library/src/lapack/roclapack_sytrf_strided_batched.cpp:5:
In file included from ../../library/src/lapack/roclapack_sytrf.hpp:13:
In file included from ../../library/src/lapack/../auxiliary/rocauxiliary_lasyf.hpp:12:
In file included from ../../library/src/include\lapack_device_functions.hpp:7:
In file included from ../../library/src/include/lib_device_helpers.hpp:11:
In file included from ../../library/src/include/libcommon.hpp:7:
In file included from ../../library/src\rocblascommon/utility.hpp:9:
In file included from ../../library/include\rocsolver.h:13:
../../library/include/rocsolver-aliases.h:17:10: fatal error: 'rocsolver-export.h' file not found
#include "rocsolver-export.h"
         ^~~~~~~~~~~~~~~~~~~~
```

arvindcheru:
> 
Will check this and update

library/src/CMakeLists.txt
```diff
@@ -281,22 +289,26 @@ endif()
...
 # Installation
-
+execute_process(COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/library/include ${PROJECT_BINARY_DIR}/include/rocsolver)
 rocm_install_targets(
   TARGETS ${package_targets}
   INCLUDE
-    ${CMAKE_SOURCE_DIR}/library/include
     ${CMAKE_BINARY_DIR}/include
-  PREFIX rocsolver
 )
```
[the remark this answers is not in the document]
cgmb: An example of an incremental build that I'm unsure about would be if you did something like:

```bash
./install.sh -d
echo "// changed rocsolver-functions" >> library/include/rocsolver-functions.h
make -C build/release
make -C build/release install
make -C build/release package
```

Would the package that gets built have the edited rocsolver.h file, or would it still contain the original?

PR #380, opened 2022-02-25
Consistency check for syev and sygv

tfalders:
Introduces a bunch of minor changes to make syev/syevd and sygv/sygvd more consistent with each other.

The most important part of this change (as it's user-facing) is the update to the documentation so that `j` is consistently used as the batch index and `i` is consistently used as the info index. This is a change I would like to make across the other functions in the library as well, since it's currently a hodgepodge of which letter is used for which index, but that's for a later date.


jzuniga-amd (approved):
LGTM

cgmb (approved):
LGTM.

PR #383, opened 2022-03-03
Getrf tuning3 (tuning for npvt and batch cases)

jzuniga-amd:


library/src/include/lapack_host_functions.hpp
```diff
@@ -139,7 +272,7 @@ void rocsolver_trsm(rocblas_handle handle,
         grid = dim3(1, blocks, batch_count);
         threads = dim3(dimx, dimy, 1);
         lmemsize = dimy * sizeof(T);
-        ROCSOLVER_LAUNCH_KERNEL(trsm2_kernel<T>, grid, threads, lmemsize, stream, jb, n, MM,
+        ROCSOLVER_LAUNCH_KERNEL(trsm2L_kernel<T>, grid, threads, lmemsize, stream, jb, n, MM,
```
cgmb, line 275: [The convention in rocBLAS](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/rocm-5.0.1/library/src/blas2/rocblas_hemv_symv_kernels.cpp#L1198) seems to be `*_kernel_lower`. I think I slightly prefer it that way, but it's a pretty minor cosmetic difference. If you like `*L_kernel` better, that's fine with me too.

library/src/lapack/roclapack_getf2.hpp
```diff
@@ -174,6 +174,208 @@ inline rocblas_int getf2_get_checksingularity_blksize(const rocblas_int n)
     return singular_thds;
 }
 
+/** This funcitons test if one of the specialized kernels can be used.
+    Returns 1 if the small kernel can be used.
+    Returns 2 if the panel kernel can be used.
+    Returns 0 otehrwise **/
```
cgmb, line 180: `s/otehrwise/otherwise/`

library/src/lapack/roclapack_getf2.hpp
```diff
@@ -174,6 +174,208 @@ inline rocblas_int getf2_get_checksingularity_blksize(const rocblas_int n)
...
+template <bool ISBATCHED, typename T, std::enable_if_t<!is_complex<T>, int> = 0>
+inline int select_spkernel(const rocblas_int m, const rocblas_int n, const bool pivot)
```
cgmb, line 182: The `inline` is unnecessary. As a template, it's already implicitly inline.

library/src/lapack/roclapack_getf2.hpp
```diff
@@ -174,6 +174,208 @@ inline rocblas_int getf2_get_checksingularity_blksize(const rocblas_int n)
...
+    {
+        ker = 2;
+    }
+
+    return ker;
+}
+
+/** Complex type version **/
+template <bool ISBATCHED, typename T, std::enable_if_t<is_complex<T>, int> = 0>
+inline int select_spkernel(const rocblas_int m, const rocblas_int n, const bool pivot)
```
cgmb, line 280: `s/inline //`

library/src/lapack/roclapack_getf2.hpp
```diff
@@ -174,6 +174,208 @@ inline rocblas_int getf2_get_checksingularity_blksize(const rocblas_int n)
...
+            {
+                ker = (m <= 256) ? 1 : 2;
+            }
+            else if((n > 60 && n <= 68 && m >= n && m <= 656 && (m <= 148 || m > 328))
+                    || (n > 68 && n <= 76 && m >= n && m <= 896 && (m <= 148 || m > 204))
+                    || (n > 76 && n <= 124 && m >= n && (m <= 128 || m > 328))
+                    || (n > 124 && m >= n && m > 228))
+            {
+                ker = 2;
+            }
```
cgmb, line 311: That's a lot of conditions! At some point, we might want to use lcov to check that our benchmarks have full coverage...

library/src/lapack/roclapack_getf2.hpp
```diff
@@ -174,6 +174,208 @@ inline rocblas_int getf2_get_checksingularity_blksize(const rocblas_int n)
     return singular_thds;
 }
 
+/** This funcitons test if one of the specialized kernels can be used.
```
cgmb, line 177: `s/funcitons test/function tests/`

library/src/lapack/roclapack_getf2.hpp
tfalders: We might also want to have a conversation about maintainability at some point. The more such conditions are added, the less confident I feel in our ability to maintain them going forward, especially as we transition to newer hardware.

cgmb: Two questions to that end:
- Is there a pattern to this that we might be able to describe these conditions more concisely in a high-level manner? If there's a straightforward interpretation of these conditions, we might be able to do something like put the threshold values into a table. That might be easier to understand and maintain.
- Do these numbers need to be kept in sync with a corresponding set of numbers somewhere else?

library/src/include/lapack_host_functions.hpp
```diff
@@ -15,26 +15,150 @@
...
+    {1,  1,  1,  1,  1, 80, 80, 80, 56, 56, 40, 40, 40, 32, 32, 32},    \
+    {1,  1,  1,  1, 80, 80, 80, 80, 80, 48, 48, 48, 40, 32,  0,  0},    \
+    {1,  1,  1, 80, 80, 80, 80, 80, 56, 56, 32, 32, 32, 32,  0,  0},    \
+    {1,  1,  1,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0},    \
+    {1,  1,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0},    \
+    {1,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0},    \
+    {0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0}
+
+#define TRSM_NUMROWS_COMPLEX 10
+#define TRSM_NUMCOLS_COMPLEX 12
```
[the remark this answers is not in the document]
jzuniga-amd: it would be nice if you can figure this out, for sure.

library/src/include/lapack_host_functions.hpp
jzuniga-amd: I like to have the `kernel` at the end of the name string so that it easily identifies kernels. So how about `trsm2_lower_kernel` and `trsm2_upper_kernel`?

library/src/lapack/roclapack_getf2.hpp
jzuniga-amd: These conditions have to do with the occupancy, number of registers, amount of memory, data flow, etc. of the generated kernels (the small kernel stores data on registers; the panel kernel stores data on shared memory)... 
Even if we had the compiler returning the number of used registers and an occupancy calculation tool (as in cuda), finding analytic meaningful expressions for these conditions is not an easy task. I have tried before, but unfortunately never got results as good as with the heuristic/experimental approach. 

I don't see a special problem with maintainability. These conditions should not change much unless you need to ensure the best performance in a substantially different hardware... but in that case you would also need to review the grid sizes and thread-block sizes off all your other kernels, the use of memory hierarchy, gemms need to be re-tuned, etc. It is part of the game, I guess. 

Ideally, I wanted to express these conditions in a table as I do with the block sizes in getrf, but I did not find a way to express the `m>=n` condition as both values are only known at run time.

library/src/include/lapack_host_functions.hpp
```diff
@@ -43,65 +167,77 @@ void rocsolver_trsm_mem(const rocblas_side side,
...
+        if(blk > 0)
+        {
+            *size_work1 = 0;
+            *size_work2 = 0;
+            *size_work3 = 0;
+            *size_work4 = 0;
+            return;
+        }
+        else
+            mm = m;
```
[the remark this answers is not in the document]
jzuniga-amd: Not necessarily. This is only strictly required when in blocked mode (`inblocked == true`) because we need to ensure that the workspace allocated for the largest trsm will be enough for all the smaller sizes when doing the block recursion. 

If not in blocked mode, I prefer to return the real size requirement. This way we can also use the internal rocsolver trsm in other routines like getrs, etc.

tfalders (approved):
LGTM!

PR #386, opened 2022-03-08
Packages for test and benchmark executables on all supported OSes using CPack.

lawruble13:


cgmb (commented):
Sweet. It will be nice to have client packages.

CMakeLists.txt
```diff
@@ -142,9 +143,58 @@ if(BUILD_LIBRARY)
...
+    rocm_package_setup_client_component(tests)
+  endif()
+  if(BUILD_CLIENTS_BENCHMARKS)
+    rocm_package_setup_client_component(benchmarks)
+  endif()
   add_subdirectory(clients)
 endif()
 
+# Package-specific CPACK vars
+rocm_package_add_dependencies("rocblas >= 2.44" "rocblas < 2.45")
```
tfalders, line 157: The bump_rocsolver_version script ought to be updated to reflect that this has been moved from library/CMakeLists.txt

cgmb (approved):

jzuniga-amd (approved):

tfalders (approved):
Whoops, I thought I'd already approved this. LGTM!

cgmb:
@lawruble13, you can squash-and-merge when ready.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 
 ## (Unreleased) rocSOLVER
 ### Added
+- Packages for test and benchmark executables on all supported OSes using CPack.
```
jzuniga-amd, line 7: Minor detail: we are not adding the packages but the cmake/cpack infrastructure to generate them.
Also, it would be good to update the install.sh script help for the `-p` and `-i` flags to say: "... library and client packages..."

CMakeLists.txt
```diff
@@ -143,12 +143,23 @@ if(BUILD_LIBRARY)
...
+    set(GFORTRAN_PKG "devtoolset-7-gcc-gfortran")
+  endif()
   rocm_package_setup_component(clients)
   if(BUILD_CLIENTS_TESTS)
-    rocm_package_setup_client_component(tests)
+    rocm_package_setup_client_component(tests DEPENDS DEB "gfortran" RPM "${GFORTRAN_PKG}")
   endif()
   if(BUILD_CLIENTS_BENCHMARKS)
-    rocm_package_setup_client_component(benchmarks)
+    rocm_package_setup_client_component(benchmarks DEPENDS DEB "gfortran" RPM "${GFORTRAN_PKG}")
```
cgmb, line 162: I don't think this is correct. Your Ubuntu 18.04 and Ubuntu 20.04 packages are not compatible.

The clients don't actually depend on gfortran; they depend on the gfortran runtime library. On Ubuntu 18.04, you're getting to the real dependency via `gfortran`->`gfortran-7`->`libgfortran-7-dev`->`libgfortran4`. On Ubuntu 20.04, you go `gfortran`->`gfortran-9`->`libgfortran-9-dev`->`libgfortran5`. In both cases, you're pulling in a bunch of extra packages, but that's not the real problem.

The issue is that there is only one repository for ROCm packages on Ubuntu. We install the same package for Ubuntu 18.04 as for Ubuntu 20.04. They're all built on Ubuntu 18.04 and designed to be installed on either OS. However, when you build the client package on Ubuntu 18.04 and install it on Ubuntu 20.04, you're linking to a different major version of libgfortran than the one that you're asking apt to install.

If you change it to depend on `libgfortran4` when the package is built on Ubuntu 18.04 and `libgfortran5` when the package is built on Ubuntu 20.04, I think that addresses the problem. The package built on Ubuntu 18.04 can still be installed on Ubuntu 20.04, because there's a `libgfortran4` in the Ubuntu 20.04 repositories (it's just not the default version anymore).

cgmb (changes requested):

CMakeLists.txt
cgmb: CentOS 7 and SLES should probably also depend on `libgfortran4`. On CentOS 8, the package is named `libgfortran`.

cgmb (approved):
LGTM. If you've tested these packages on all platforms and found that they function as expected, then I think it's ready to merge.

cgmb:
A related, but different problem:

```
$ sudo dpkg -r rocsolver rocsolver-dev
(Reading database ... 249085 files and directories currently installed.)
Removing rocsolver-dev (3.18.0-6065cc7c) ...
Removing rocsolver (3.18.0-6065cc7c) ...
rm: cannot remove '/etc/ld.so.conf.d/rocsolver.conf': No such file or directory
Processing triggers for libc-bin (2.27-3ubuntu1.4) ...
```

I think that error is because both `rocsolver` and `rocsolver-dev` get ldconfig stuff added to the Debian `postinst` and `prerm`. With these new test packages, will we see that error repeatedly? It seems like the ldconfig scripts should be limited to the main library package.

jzuniga-amd (approved):

tfalders (approved):

PR #387, opened 2022-03-09
Disable rtest logfile parsing

cgmb:
By default, rtest.py will parse the test log and mark the build as a
failure if it contains the strings "fail" or "error". This causes the
test suite to be marked as failure even when all tests pass when
executing the checkin tests, due to tests added in 3c3c865 with
problem names like 'checkin_misc_LOGGING.profile_file_open_failure'.

This log-parsing behaviour is unnecessary, so we can disable it by
setting a regex for an impossible pattern that can never match
any string.

### DIFF
--- a/clients/gtest/logging_gtest.cpp
+++ b/clients/gtest/logging_gtest.cpp
@@ -364,7 +364,7 @@ TEST_F(checkin_misc_LOGGING, rocblas_layer_mode_log_bench_tree)
     verify_file(log_filepath, expected_lines);
 }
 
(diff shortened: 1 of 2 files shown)


tfalders:
Personally, I'd prefer changing the test names. For example, changing "trace_file_open_failure" to "invalid_trace_file_open". I don't know if there's any reason why someone would want to reenable the log-parsing behaviour, but I'd like to be covered just in case.

cgmb:
why_not_both

tfalders (approved):

jzuniga-amd (approved):

PR #388, opened 2022-03-11
Reorg Changes with rocm-cmake wrapper usage

arvindcheru:
- File/Folder Reorg Changes
- Backward Support enabled (rocm-cmake wrapper)
- Replacement of https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/379 with all comments

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -8,8 +8,6 @@ set(rocsolver_SOVERSION 0.1)
 # Create version header from templated .in file using CMake info
 configure_file(include/rocsolver-version.h.in "${PROJECT_BINARY_DIR}/include/rocsolver/rocsolver-version.h")
 
-
-
 set(rocsolver_headers_public
   include/rocsolver.h
   include/rocsolver-extra-types.h
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -6,15 +6,17 @@
 set(rocsolver_SOVERSION 0.1)
 
 # Create version header from templated .in file using CMake info
-configure_file(include/rocsolver-version.h.in "${PROJECT_BINARY_DIR}/include/rocsolver-version.h")
+configure_file(include/rocsolver-version.h.in "${PROJECT_BINARY_DIR}/include/rocsolver/rocsolver-version.h")
+
+
 
 set(rocsolver_headers_public
   include/rocsolver.h
   include/rocsolver-extra-types.h
   include/rocsolver-aliases.h
   include/rocsolver-functions.h
-  ${PROJECT_BINARY_DIR}/include/rocsolver-version.h
-  ${PROJECT_BINARY_DIR}/include/rocsolver-export.h
+  ${PROJECT_BINARY_DIR}/include/rocsolver/rocsolver-version.h
+  ${PROJECT_BINARY_DIR}/include/rocsolver/rocsolver-export.h
 )
 
 # For IDE integration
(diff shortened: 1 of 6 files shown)


library/CMakeLists.txt
```diff
@@ -6,15 +6,17 @@
 set(rocsolver_SOVERSION 0.1)
 
 # Create version header from templated .in file using CMake info
-configure_file(include/rocsolver-version.h.in "${PROJECT_BINARY_DIR}/include/rocsolver-version.h")
+configure_file(include/rocsolver-version.h.in "${PROJECT_BINARY_DIR}/include/rocsolver/rocsolver-version.h")
+
```
lawruble13, line 10: excess newlines

library/src/CMakeLists.txt
```diff
@@ -281,22 +289,39 @@ endif()
...
+rocm_install(
+	DIRECTORY
+	${PROJECT_SOURCE_DIR}/library/include
+	DESTINATION
+	${PROJECT_BINARY_DIR}/include/rocsolver
+	FILES_MATCHING
+	PATTERN *.h
+	PATTERN *.hpp
+	PERMISSIONS OWNER_WRITE OWNER_READ GROUP_READ WORLD_READ
+	)
```
lawruble13, line 303: Copying files from the source directory to the binary directory should not be done with `rocm_install`. This will attempt to install those header files to the exact build location within the packages as well.

lawruble13 (changes requested):
Does the changelog need to be updated?

library/src/CMakeLists.txt
arvindcheru: Thanks for comments, But all source header files are same to be installed and packaged. So this is installing, packaging as expected.

Meantime, this change was also  added to address  incremental changes, build scenario as review comments from @cgmb.

arvindcheru: https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/379#pullrequestreview-897072157

cgmb: I'll look into this issue today. It might take a bit of digging, but I'm sure we can find a solution that satisfies everyone.

lawruble13: It is still incorrect, by specifying the full destination path all header files will be installed to that location in the package (e.g. if the build directory is `/home/amd/rocSOLVER/build/release`, then this will cause the package to install the header files to `/home/amd/rocSOLVER/build/release/include/rocsolver`.

library/src/CMakeLists.txt
```diff
@@ -281,22 +289,28 @@ endif()
 
 ############################################################
 # Installation
+execute_process(COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/library/include ${PROJECT_BINARY_DIR}/include/rocsolver)
```
cgmb, line 292: We haven't yet resolved the issues with this line, but they're not relevent when you're doing a clean build (which is always the case if you're using `./install.sh`).

cgmb (approved):
There remain some issues to resolve for incremental builds, but we can fix those in the next pass. The release and CI always do clean builds, so the remaining issues are only a problem during development. We can work around the problem by re-running the CMake configure step manually.

That's not ideal, but let's get this merged in so the the integration testing can get going...

jzuniga-amd (approved):

tfalders (approved):

PR #389, opened 2022-03-15
Correct print_host_matrix

tfalders:
For some reason, the print methods that show a side-by-side comparison of the CPU and GPU results take `m` to be the number of columns and `n` the number of rows, which is flipped from every other print method. This PR fixes that, and prints the name like the other print methods.


cgmb (approved):
I don't use these functions much and they're not used in rocSOLVER itself, so this has little effect on me. LGTM.

PR #392, opened 2022-03-16
Add stebz

jzuniga-amd:
range=index added in 3060420


tfalders (commented):
I've opened a PR in your local fork that should address the main merge conflicts that could occur between this PR and my upcoming PRs.

Also, there appear to be two tests that are failing due to a not_implemented status being returned by STEBZ.

library/include/rocsolver-functions.h
```diff
@@ -3901,6 +3901,123 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zstedc(rocblas_handle handle,
...
+    nsplit      pointer to a rocblas_int on the GPU.\n
+                The number of split off blocks in the matrix.
+    @param[out]
+    W           pointer to real type. Array on the GPU of dimension n.\n
+                The first nev elements contain the computed eigenvalues. (The remaining elements
+                can be used as workspace for internal computations).
+    @param[out]
+    iblock      pointer to rocblas_int. Array on the GPU of dimension n.\n
+                The block indices corresponding to each eigenvalue. When matrix T has
+                split off blocks (nsplit > 1) then, if iblock[i] = k, the
```
tfalders, line 3967: Typo: the comma should come before "then"

clients/include/testing_stebz.hpp
```diff
@@ -0,0 +1,494 @@
...
+    double max_error = 0, gpu_time_used = 0, cpu_time_used = 0;
+
+    size_t size_WRes = (argus.unit_check || argus.norm_check) ? size_W : 0;
+    size_t size_iblockRes = (argus.unit_check || argus.norm_check) ? size_iblock : 0;
+    size_t size_isplitRes = (argus.unit_check || argus.norm_check) ? size_isplit : 0;
+
+    // check invalid sizes
+    bool invalid_size = (n < 0) || (range == rocblas_erange_value && vl >= vu)
+        || (range == rocblas_erange_index && (il < 1 || iu < 0))
+        || (range == rocblas_erange_index && (iu > n || (n > 0 && il > iu)));
```
tfalders, line 356: Incredibly minor, but (stylistically) I would prefer:
```
        || (range == rocblas_erange_index && (il < 1 || iu < 0 || iu > n))
        || (range == rocblas_erange_index && n > 0 && il > iu);
```

library/src/auxiliary/rocauxiliary_stebz.hpp
```diff
@@ -0,0 +1,1001 @@
...
+
+    // 2. invalid size
+    if(n < 0)
+        return rocblas_status_invalid_size;
+    if(range == rocblas_erange_value && vlow >= vup)
+        return rocblas_status_invalid_size;
+    if(range == rocblas_erange_index && (ilow < 1 || iup < 0))
+        return rocblas_status_invalid_size;
+    if(range == rocblas_erange_index && (iup > n || (n > 0 && ilow > iup)))
+        return rocblas_status_invalid_size;
```
tfalders, line 883: As with the test, I'd prefer
```
    if(range == rocblas_erange_index && (ilow < 1 || iup < 0 || iup > n))
        return rocblas_status_invalid_size;
    if(range == rocblas_erange_index && n > 0 && ilow > iup)
        return rocblas_status_invalid_size;
```

library/src/auxiliary/rocauxiliary_stebz.hpp
```diff
@@ -0,0 +1,1001 @@
...
+                                        rocblas_int* work,
+                                        T* pivmin,
+                                        T* Esqr,
+                                        T* bounds,
+                                        T* inter,
+                                        rocblas_int* ninter)
+{
+    ROCSOLVER_ENTER("stebz", "range:", range, "order:", order, "n:", n, "vlow:", vlow, "vup:", vup,
+                    "ilow:", ilow, "iup:", iup, "abstol:", abstol, "shiftD:", shiftD,
+                    "shiftE:", shiftE, "bc:", batch_count);
```
tfalders, line 932: I missed this in my PR, but the strings for vlow, vup, etc... should probably be vl, vu, etc... instead since they can be visible to the user, and we should probably match the API.

library/src/auxiliary/rocauxiliary_stebz.hpp
```diff
@@ -0,0 +1,1001 @@
...
+        // have eigenvalues outside of the desired indices. Thus, the following
+        // code also discards those extra values)
+        for(int b = 0; b < nofb; ++b)
+        {
+            bin = (b == 0) ? 0 : IS[b - 1];
+            for(int bb = 0; bb < tmpnev[b]; ++bb)
+            {
+                tmp = W[bin + bb];
+                ntmp = index ? sturm_count(n, D, Esqr, pmin, tmp + pmin + bnorm * eps * n) : 0;
+                if(!index || (ntmp >= ilow && ntmp <= iup))
```
tfalders, line 782: I'm confused as to why we check ntmp instead of (bin + bb) against the index range, since (bin + bb) appears to be the index of the eigenvalue we're considering.

tfalders (approved):
The algorithm doesn't entirely make sense to me, but it looks good. I just have a couple suggestions and a question.

library/src/auxiliary/rocauxiliary_stebz.hpp
```diff
@@ -0,0 +1,1001 @@
...
+            // find lower bound for the set of indices
+            j = -1;
+            for(rocblas_int i = 0; i < 2 * n; ++i)
+            {
+                if(ninter[i] < ilow)
+                    j++;
+                else
+                    break;
+            }
+            vl = inter[j];
```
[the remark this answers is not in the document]
jzuniga-amd: Yes, it is guaranteed because ilow >= 1 (otherwise you get rocblas_invalid_size)...
but good suggestion,  I'll change it anyways to avoid confusions.

library/src/auxiliary/rocauxiliary_stebz.hpp
```diff
@@ -0,0 +1,1001 @@
...
+
+            // find upper bound for the set of indices
+            j = -1;
+            for(rocblas_int i = 0; i < 2 * n; ++i)
+            {
+                j++;
+                if(ninter[i] >= iup)
+                    break;
+            }
+            vu = inter[j];
```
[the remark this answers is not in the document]
jzuniga-amd: Same here, the j++ happens at least once unless n=0 (in which case you have a quick return)... I'll change as well anyways.

library/src/auxiliary/rocauxiliary_stebz.hpp
jzuniga-amd: bin+bb is the index of the found eigenvalue w in the corresponding split block; it does not correspond necessarily to its position in the entire spectrum. What we need to verify is that its position in the spectrum is within ilow and iup.

cgmb (approved):
Apologies for holding this up. I didn't notice it was waiting on me.

cgmb:
There are, however, test failures on Windows that look legitimate. :(
```
[  FAILED  ] checkin_lapack/STEBZ.__float/70, where GetParam() = ({ 64, 0, -1 }, { 2, 0, 0, 1, 15 })
[  FAILED  ] checkin_lapack/STEBZ.__double/70, where GetParam() = ({ 64, 0, -1 }, { 2, 0, 0, 1, 15 })
```

jzuniga-amd:
This should be fixed in latest commits... I think windows CI haven't run recently on this PR

cgmb:
> This should be fixed in latest commits... I think windows CI haven't run recently on this PR

Sadly, no. The Windows CI was down for a few days, but it's back up now and #400 is failing with the STEBZ issue.

PR #393, opened 2022-03-17
Added stein

tfalders:


library/src/auxiliary/rocauxiliary_stein.hpp
```diff
@@ -0,0 +1,384 @@
...
+                        xj = xjm + pertol;
+                }
+
+                rocblas_int iters = 0;
+                rocblas_int nrmchk = 0;
+
+                // initialize starting eigenvector
+                // TODO: how to make it random?
+                for(i = tid; i < blksize; i += MAX_THDS)
+                    work[i] = (i == j - j1 ? 1 : 0);
```
cgmb, line 141: Does the algorithm we use to make the initial eigenvector guesses need to be stable between rocSOLVER versions?

library/src/include/rocsolver_logvalue.hpp
```diff
@@ -154,6 +154,24 @@ struct formatter<rocsolver_logvalue<rocblas_eform>> : formatter<char>
     }
 };
 template <>
+struct formatter<rocsolver_logvalue<rocblas_erange>> : formatter<char>
+{
+    template <typename FormatCtx>
+    auto format(rocsolver_logvalue<rocblas_erange> wrapper, FormatCtx& ctx)
```
cgmb, line 160: This is missing `ROCSOLVER_FMT_CONST`.

library/src/auxiliary/rocauxiliary_stein.hpp
tfalders: I don't know if that's a hard requirement, but it certainly can't hurt. It looks like in vanilla LAPACK, the random generator used to make the initial guess is seeded with constant values, so it seems that LAPACK's guesses ought to be stable.

clients/include/testing_stein.hpp
```diff
@@ -0,0 +1,536 @@
...
+        // scale matrix
+        for(rocblas_int i = 0; i < n; i++)
+        {
+            hD[0][i] += 10;
+            hE[0][i] -= 5;
+        }
+
+        // compute a subset of the eigenvalues
+        S vl = 0.0;
+        S vu = 10.0;
```
jzuniga-amd, line 144: The matrix, as scaled above, will have all its eigenvalues within the interval [1, 30] --an very close to each other specially for large sizes --
Asking for a subset in (0, 10] might be quite reduced in some cases; and we never know exactly how many eigenvalues will be found... we may be running cases that do not test anything (with nev = 0).

Better to use range=index and ask exactly for nev eigenvalues... the last ones so we have better numerics.

jzuniga-amd (commented):
This is not quite ready yet. 

The test coverage is limited (see comments below), and the main concern is that we have no control on the number of eigenvectors; it depends on the result of stebz (that we don't verify).

It is important that we can test for a given number of eigenvalues (nev), specially for benchmarking purposes: we should be able to say to the rocsolver-bench, compute 5, 10, or 50 eigenvectors...

(I prepared a PR to your fork with my suggested solution. Unfortunately, once I extended the tests, more failures appeared... please have a look.)

jzuniga-amd:
I still have to take a closer look at the algorithm implementation... I may be able to help with the error then.

library/include/rocsolver-functions.h
```diff
@@ -3901,6 +3901,122 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zstedc(rocblas_handle handle,
...
+    \brief STEIN computes the eigenvectors associated with a set of 
+    provided eigenvalues of a symmetric tridiagonal matrix.
+
+    \details
+    The eigenvectors of the symmetric tridiagonal matrix are computed using
+    inverse iteration.
+
+    The matrix is not represented explicitly, but rather as the array of
+    diagonal elements D and the array of symmetric off-diagonal elements E.
+    The eigenvalues must be provided in the array W, as returned by \ref rocsolver_sstebz "STEBZ".
```
cgmb, line 3914: Is this going in after sstebz? Sphinx is complaining that you're referring to something that doesn't exist.

tfalders: That's what I was thinking. I'll do a check tomorrow on my local machine to make sure the docs build properly when both stebz and stein are merged.

clients/include/testing_stein.hpp
```diff
@@ -0,0 +1,537 @@
...
+    size_t size_isplit = size_D;
+    size_t size_Z = ldz * n;
+    size_t size_ifail = size_D;
+    double max_error = 0, gpu_time_used = 0, cpu_time_used = 0;
+
+    size_t size_ZRes = (argus.unit_check || argus.norm_check) ? size_Z : 0;
+    size_t size_ifailRes = (argus.unit_check || argus.norm_check) ? size_ifail : 0;
+
+    // check invalid sizes
+    bool invalid_size = (n < 0 || ldz < n);
```
jzuniga-amd, line 398: you should check that nev is valid too (0 <= nev <= n)

common/include/common_host_helpers.hpp
```diff
@@ -22,6 +23,28 @@
...
+{
+    using S = decltype(std::real(T{}));
+    return std::numeric_limits<S>::epsilon();
+}
+
+template <typename T>
+constexpr double get_safemin()
+{
+    using S = decltype(std::real(T{}));
+    auto eps = get_epsilon<S>();
```
jzuniga-amd, line 40: depending on the order in which we merge this, and #378, somebody will need to modify get_epsilon (same goes for #392 )

jzuniga-amd (approved):
As per discussions, if all tests are passing, I think this is ready to go (we can always revisit in the future if necessary)

common/include/common_host_helpers.hpp
cgmb: Mine is the least important. Do whatever's best for stein/stebz, and I will adjust.

library/src/auxiliary/rocauxiliary_stein.hpp
cgmb: It seems like your updated initialization is fine for now. We can probably revisit it later if necessary. Out of curiosity, what was the missing property of this 'outdated' version?

common/include/common_host_helpers.hpp
jzuniga-amd: I merged stebz, I think this can go next, and finally we can update all new files in #378

clients/include/testing_stein.hpp
tfalders: nev is on the device when it comes to stein. I don't know if there's a good way to check it outside of a kernel.

library/src/auxiliary/rocauxiliary_stein.hpp
tfalders: In LAPACK, the initial eigenvector entries have an average of 0 (since they're uniformly distributed), whereas my outdated eigenvector entries had an average of 1. According to Juan, I guess that privileges certain entries and makes it harder to converge. The initial eigenvector entries now have an average of 0, and that seems to fix the test failures we were experiencing.

cgmb (approved):
LGTM.

clients/include/testing_stein.hpp
jzuniga-amd: right! I forgot this detail...

tfalders:
Due to time constraints, I don't think there's enough time to finish running all the CI tests, so I'm going to go ahead and merge. Feel free to blame if something breaks.

PR #395, opened 2022-03-22
Add CONTRIBUTING.md

cgmb:
I wasn't certain of the best way to split the document between the docs and the repository. There's a few benefits of using a contributing file in the project root, but one of the main ones is that this is a special file that [GitHub understands and displays to new contributors](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors).

There's still lots more we can do to make the project more welcoming for new contributors. I hope this is a useful start that we can continue to build upon.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,62 @@
+# Contributing
+
+## Philosophy
+
+AMD welcomes contributions from the community. Whether those contributions are bug reports,
+bug fixes, documentation additions, performance notes, or other improvements, we value
+collaboration with our users. We can build better solutions together.
+
+## License agreement
+
+Contributors must agree to the following terms:
+
+1. The code I am contributing is mine, and I have the right to license it.
+
+2. By submitting a pull request for this project I am granting you a license to distribute
+   said code under the BSD License for the project.
+
+# Submitting a Pull Request
+
+To contribute changes to rocSOLVER, open a pull request targeting the `develop` branch. Pull
+requests will be tested and reviewed by the AMD development team. AMD may request changes or
+modify the submission before acceptance.
+
+## Interface requirements
+
+The public interface must be:
+
+- C99 compatible
+- Source and binary compatible with previous releases
+- Fully documented with Doxygen and Sphinx
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+bug fixes, documentation additions, performance notes, or other improvements, we value
+collaboration with our users. We can build better solutions together.
+
+## License agreement
+
+Contributors must agree to the following terms:
+
+1. The code I am contributing is mine, and I have the right to license it.
+
+2. By submitting a pull request for this project I am granting you a license to distribute
```
tfalders, line 15: Instead of "you" should we put "AMD"?

CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+be removed using `scripts/uninstall-hooks`.
+
+## Tests
+
+To run the rocSOLVER test suite, first build the rocSOLVER test client following the instructions in
+[Building and Installation][1]. Then, run the `rocsolver-test` binary. For a typical build, the test
+binary will be found at `./build/release/clients/staging/rocsolver-test`.
+
+The full test suite is quite large and may take a long time to complete, so passing the
+[`--gtest_filter=<pattern>`][2] option to rocsolver-test may be useful during development.
```
tfalders, line 53: It might be handy to mention the checkin_lapack and daily_lapack tests as possible options here.

tfalders (commented):
Couple small suggestions. Looks good otherwise.

CONTRIBUTING.md
cgmb: Aside from changing 'MIT' to 'BSD', this is directly from the [rocBLAS contributing guide](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/rocm-5.0.2/CONTRIBUTING.rst#license-agreement). I've been asking about updating them both, but for now I would recommend using exactly the same wording as rocBLAS.

jzuniga-amd (approved):

CONTRIBUTING.md
TorreZuk: This agreement section is being removed from rocBLAS ask Andrew if you want details.

tfalders (approved):

PR #397, opened 2022-03-23
Add tests for rocsolver-bench client

cgmb:
This is a really basic set of tests for the rocsolver-bench client. It is not comprehensive, but it should exercise a large number of code paths, and it provides a foundation that we can build upon. If nothing else, it should be very easy to add a test to this script whenever we find a bug in the benchmark client.

In these tests, I'm checking the inputs and outputs of the rocsolver-bench for various sets of parameters. That can check a lot of important behaviours of the benchmark client, but it doesn't catch them all. For example, it doesn't ensure that we actually passed the right arguments to the rocsolver library (only that we reported that we did). To extend the testing to the actual function invoked, my suggestion would be to enable trace logging in rocsolver-bench, and parse that output as well.

I'm in the process of filling in this the pattern I applied to GELS for all the various other functions in the `rocsolver_dispatcher`. It's tedious, but simple. It could probably be refactored to something more concise. Je n’ai fait celle-ci plus longue que parce que je n’ai pas eu le loisir de la faire plus courte.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -48,6 +48,8 @@ def runTestCommand (platform, project, gfilter)
                   ./test-rocsolver-dlopen --gtest_color=yes
                 fi
                 ${hmmTestCommand}
+                cd ${project.paths.project_build_prefix}/build/${buildType}
+                CTEST_OUTPUT_ON_FAILURE=1 ctest -R '^test-rocsolver-bench'
                 """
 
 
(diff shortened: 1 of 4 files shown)


tfalders (approved):
I'm going to approve this now. Looking through all the functions is making me go bug-eyed and I think the tests are set-up well enough that any errors should be caught by test failures. As long as you stick to the same test pattern and all tests pass, we should be golden.

cgmb:
The rest of the input/output tests are in. @JeremyAdamHart gave me a hand with the last bunch. This was definitely the brute-force approach to checking the args and results on the rocsolver-bench functions, but it works and it should be pretty simple to extend when we add new rocsolver-bench functionality.

jzuniga-amd (approved):

PR #402, opened 2022-03-30
Added syevx

tfalders:
This PR adds SYEVX/HEEVX to rocSOLVER. ~~It should be followed later on by a PR to improve performance.~~

~~When `erange` is `rocblas_erange_value`, the current implementation will read `nev` from the device onto the host and then iterate over each batch instance to call ORMTR/UNMTR. Ideally, we should create a kernel for ORMTR/UNMTR or allow ORMTR/UNMTR to read size parameters from the device instead, but I'm saving this task for later.~~

Edit: Per Juan's advice, ORMTR/UNMTR will now be called with `n` instead of `nev` when `erange = rocblas_erange_value`.


jzuniga-amd (changes requested):

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+    }
+
+    // gpu-lapack performance
+    hipStream_t stream;
+    CHECK_ROCBLAS_ERROR(rocblas_get_stream(handle, &stream));
+    double start;
+
+    if(profile > 0)
+    {
+        rocsolver_log_set_layer_mode(rocblas_layer_mode_log_profile);
```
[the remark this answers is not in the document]
cgmb: It also needs to go into the benchmark tests.

clients/include/rocsolver.hpp
```diff
@@ -5625,6 +5625,246 @@ inline rocblas_status rocsolver_syevd_heevd(bool STRIDED,
...
+                                            float* Z,
+                                            rocblas_int ldz,
+                                            rocblas_stride stZ,
+                                            rocblas_int* ifail,
+                                            rocblas_stride stF,
+                                            rocblas_int* info,
+                                            rocblas_int bc)
+{
+    return STRIDED
+        ? rocsolver_ssyevx_strided_batched(handle, evect, erange, uplo, n, A, lda, stA, vl, vu, il,
```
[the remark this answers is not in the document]
tfalders: I figured `abstol` was safe to remove since it has an optimal value and cuSOLVER doesn't include it either, but since you requested it I will add it in.

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+                }
+
+                // error is ||hZ - hZRes|| / ||hZ||
+                // using frobenius norm
+                err = norm_error('F', n, hNev[b][0], ldz, hZ[b], hZRes[b]);
+                *max_err = err > *max_err ? err : *max_err;
+            }
+            else
+            {
+                // check ifail
```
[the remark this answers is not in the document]
tfalders: I'm not sure what you mean. There's a test for ifail when info = 0 and a separate test when info != 0.

clients/include/testing_syevx_heevx.hpp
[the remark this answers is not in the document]
tfalders: Thanks for catching that! I'll add those in.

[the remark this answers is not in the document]
tfalders: @cgmb I don't see tests for syev and sygv and their batched versions in the benchmark tests. Is that intentional?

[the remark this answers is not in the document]
cgmb: I was working from the function list in `run_function`. It seems there's also the `run_function_limited_precision` to add.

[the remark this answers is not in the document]
cgmb: Don't worry about it, then. I'll add the whole lot when I get the chance.

clients/include/rocsolver.hpp
```diff
@@ -5625,6 +5625,246 @@ inline rocblas_status rocsolver_syevd_heevd(bool STRIDED,
...
+                                            rocblas_stride stW,
+                                            rocblas_double_complex* Z,
+                                            rocblas_int ldz,
+                                            rocblas_stride stZ,
+                                            rocblas_int* ifail,
+                                            rocblas_stride stF,
+                                            rocblas_int* info,
+                                            rocblas_int bc)
+{
+    return STRIDED
```
EdDAzevedo: Do you want to use the "if" form to make sure only one branch is evaluated (as done is SYGV_HEGV)?

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+                          Th& hA,
+                          std::vector<T>& A,
+                          bool test = true)
+{
+    if(CPU)
+    {
+        rocblas_init<T>(hA, true);
+
+        // scale A to avoid singularities
+        for(rocblas_int b = 0; b < bc; ++b)
```
EdDAzevedo, line 192: Have you considered adding OpenMP directives on CPU, something  such as "#pragma omp parallel for"?

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+                          Th& hZRes,
+                          Ih& hIfail,
+                          Ih& hIfailRes,
+                          Ih& hinfo,
+                          Ih& hinfoRes,
+                          double* max_err)
+{
+    constexpr bool COMPLEX = is_complex<T>;
+
+    int lwork = !COMPLEX ? 8 * n : 2 * n;
```
EdDAzevedo, line 263: The lapack documentation mentions for optimal efficiency, use a larger value of lwork >= (NB+3)*N. Would the value of lwork affect the performance in this implementation?

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+    }
+
+    // CPU lapack
+    for(rocblas_int b = 0; b < bc; ++b)
+        cblas_syevx_heevx<T>(evect, erange, uplo, n, hA[b], lda, vl, vu, il, iu, abstol, hNev[b],
+                             hW[b], hZ[b], ldz, work.data(), lwork, rwork.data(), iwork.data(),
+                             hIfail[b], hinfo[b]);
+
+    // Check info for non-convergence
+    *max_err = 0;
```
EdDAzevedo, line 299: Perhaps you might consider using  a local variable, say "lmax_err" instead of "*max_err", then just before return assign " *max_err = lmax_err;"?

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+                             Ih& hinfo,
+                             double* gpu_time_used,
+                             double* cpu_time_used,
+                             const rocblas_int hot_calls,
+                             const int profile,
+                             const bool perf)
+{
+    constexpr bool COMPLEX = is_complex<T>;
+
+    int lwork = !COMPLEX ? 8 * n : 2 * n;
```
EdDAzevedo, line 411: lapack documentation mentions using a larger value of lwork >= (NB+3) * N for optimal performance. If the code is trying to compare performance, you might consider a larger value of lwork.

clients/include/testing_syevx_heevx.hpp
```diff
@@ -0,0 +1,751 @@
...
+    std::vector<int> iwork(liwork);
+    std::vector<T> A;
+    S abstol = 2 * get_safemin<S>();
+
+    if(!perf)
+    {
+        syevx_heevx_initData<true, false, T>(handle, evect, n, dA, lda, bc, hA, A, 0);
+
+        // cpu-lapack performance (only if not in perf mode)
+        *cpu_time_used = get_time_us_no_sync();
```
EdDAzevedo, line 428: You might consider using OpenMP "#pragma omp parallel for" to use all CPU cores. Is get_time_us_no_sync() returning the elapsed wall clock time or the total cpu time used? The variable name "cpu_time_used" suggests the cpu time used. Perhaps the elapsed time is more relevant, especially if multiple CPU cores are used (by OpenMP or internally by vendor BLAS).

library/include/rocsolver-functions.h
```diff
@@ -15003,6 +15003,794 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zheevd_strided_batched(rocblas_handle
...
+    ifail       pointer to rocblas_int. Array on the GPU (the size depends on the value of strideF).\n
+                If info[j] = 0, the first nev_j elements of ifail_j are zero.
+                Otherwise, contains the indices of those eigenvectors that failed
+                to converge. Not referenced if evect is rocblas_evect_none.
+    @param[in]
+    strideF     rocblas_stride.\n
+                Stride from the start of one vector ifail_j to the next one ifail_(j+1).
+                There is no restriction for the value of strideF. Normal use case is strideF >= n.
+    @param[out]
+    info        pointer to rocblas_int. Array of batch_count integers on the GPU.\n
```
EdDAzevedo, line 15342: Should there be checking for valid arguments and return info = -i, if the i-th argument is invalid as performed in lapack?

clients/include/testing_syevx_heevx.hpp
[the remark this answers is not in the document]
jzuniga-amd: I meant that if you wanted to have this similar code only once, you could compare `hIfail` directly with `hIfailRes`; their entries must be equal no matter the value of info... 
But now, I think I like better the two separate cases... up to you.

clients/include/testing_syevx_heevx.hpp
jzuniga-amd: Yes, we have been thinking on adding OpenMP directives not only here but specially in the implicit tests that we have for some functions... the objective would be of course to reduce the execution time of the tests suite (which is growing fast)... 
We want to apply this to all the unit tests, though. So, this should be part of a dedicated separate PR.

clients/include/testing_syevx_heevx.hpp
jzuniga-amd: We have not paid much attention to comparing the execution times of our GPU routines and the legacy LAPACK on CPU. We use the LAPACK time just as a reference. But it may be a good idea to see if the LAPACK performance change much if lwork is larger.

clients/include/testing_syevx_heevx.hpp
jzuniga-amd: get_time_us_no_sync() returns the time since "epoch". We captured before the API call and then after; then we take the difference to have the wall time (elapsed time) required by the API call. So no, it is not cpu time...

(We don't want to use openMP here as we want legacy LAPACK performance as reference. It may be interesting to compare in the future the performance of an optimized implementation of LAPACK on the CPU, although we know in advance that CPU will be faster in some problems, especially for small matrices).

clients/include/rocsolver.hpp
cgmb: Only one branch will be evaluated. From the C++17 Standard § 8.16 [expr.cond]:

> The first expression is contextually converted to bool. It is evaluated and if it is true, the result of the conditional expression is the value of the second expression, otherwise that of the third expression. Only one of the second and third expressions is evaluated.

With that being said, whatever way we want to structure this is fine by me.

clients/include/testing_syevx_heevx.hpp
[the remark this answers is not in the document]
tfalders: I think there might actually be a bug in how LAPACK's syevx handles its ifail array. The way it updates ifail when sorting the eigenvalues & vectors doesn't look correct to me, so I did ours differently. If info != 0, `hIfail` and `hIfailRes` won't be equal.

library/include/rocsolver-functions.h
tfalders: Currently, the library convention is to not do this, and adding it in would be a library-wide change that would require some method signature changes (and thus deprecations). If an argument is invalid, we use the returned status to indicate what kind of invalid argument was detected. Unfortunately, the index of the argument is not provided.

clients/include/testing_syevx_heevx.hpp
tfalders: If we were to make this change, for the sake of consistency I would want to make the same change across all of our tests. That goes beyond the scope of this PR, and I'm not sure the effort would really be worth it.

cgmb (approved):
I'm not sure why the Windows tests didn't trigger, but I've manually started them. Given that there are known test failures on Windows right now, I've downgraded the Required tests to just the build stage.

tfalders:
> Perhaps the code in "getPerfData" (around line 412) should be updated too?

Ah, you're right! Silly me. Thanks for catching that!

PR #404, opened 2022-04-01
Add profile_kernels argument to rocsolver-bench

tfalders:
This PR follows-up on the comment at https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/324#issuecomment-953169500. Used in conjunction with the --profile argument, a non-zero --profile_kernels argument will indicate that kernel calls should be included in the profile log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 
 ## (Unreleased) rocSOLVER
 ### Added
+- Added --profile_kernels option to rocsolver-bench, which will include kernel calls in the
+  profile log (if profile logging is enabled with --profile).
+
 ### Optimized
 ### Changed
 ### Deprecated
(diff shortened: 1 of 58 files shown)


jzuniga-amd (approved):
LGTM. Thanks Troy!!

cgmb (approved):
I love the functionality. There's room for improvement in the rocsolver-bench command-line interface, but I suppose I shouldn't let perfect be the enemy of good.

This discussion goes well beyond the changes here, but a few things I would like to eventually change for ease of use:
- `--profile_kernels 1` should enable profile logging if not otherwise specified.
- `--profile_kernels` should mean `--profile_kernels 1`. If `--profile_kernels 0` is really needed, maybe there should be a `--no-profile-kernels`.
- The default layer depth should be unlimited.
- It would be nice if the layer depth command-line argument were a string, rather than an integer, so you could specify a layer depth of `[0-9]+k?`.
- It would be nice to be able to enable other logging in rocsolver-bench (like trace logging). 

A couple style things that bug me about our command-line:
- The use of `1` and `0` for `true`/`false` or `on`/`off` feels lazy.
- The use of underscore defies [GNU conventions](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html). I'm torn on matching Google Test, because it's simply very weird.

PR #411, opened 2022-04-18
Fixed incorrect thread indexing in sygs2

tfalders:
While investigating unrelated sygs2/sygst test issues, I discovered that sygs2's `set_diag` kernels do not calculate their thread index `b` correctly. This will negatively impact the results of sygs2, sygst, sygv, and sygvd when `batch_count > 32`.

In addition to fixing this issue, I've also renamed some of the parameters to rocblas_trsv to be more consistent with existing conventions.


jzuniga-amd (approved):

cgmb:
This is a good catch! Could you please add a test case for this behaviour?

When I was working on high reliability software, they drilled into me: if a defect is found in the software, you should write a test for it. The test serves a number of purposes over its lifetime. Before the fix is applied, the test is a reproducible example of the bug. After the fix is applied, the transition from failing to passing is proof that the fix worked. Finally, the test becomes a protection against regressions in the future.

Not all of those uses are applicable at this point and I'm a bit more relaxed about that guideline on the projects I work on these days. Nevertheless, this seems like an important hole in our test coverage.

tfalders:
> This is a good catch! Could you please add a test case for this behaviour?

The reason I didn't add a test for this is because it's probably something we should do for all of our tests, which would be better for a different PR. Right now, we only test `batch_count <= 3`. We should have a discussion on how to introduce larger batch counts into our test suite and how to do it without increasing the daily_lapack test time.

tfalders:
In the absence of a dedicated test, this change can be validated using `rocsolver-bench`. The following two calls to `rocsolver-bench` should have similar (small) reported error values:
```
./build/release/clients/staging/rocsolver-bench -f sygs2_batched -n 100 --batch_count 32 --verify 1
./build/release/clients/staging/rocsolver-bench -f sygs2_batched -n 100 --batch_count 33 --verify 1
```

cgmb (approved):
LGTM! :+1:

PR #412, opened 2022-04-18
Added sygvx

tfalders:
This PR adds sygvx to rocSOLVER, as well as adjustments to the test initialization for syevx and sygs2/sygst.

Many thanks to @jzuniga-amd for help with debugging and test refactoring.


clients/include/testing_sygsx_hegsx.hpp
```diff
@@ -124,20 +124,66 @@ void sygsx_hegsx_initData(const rocblas_handle handle,
...
+                cblas_trmm<T>(rocblas_side_left, rocblas_fill_upper,
+                              rocblas_operation_conjugate_transpose, rocblas_diagonal_non_unit, n,
+                              n, one, U[b], ldu, hA[b], lda);
+                cblas_trmm<T>(rocblas_side_right, rocblas_fill_upper, rocblas_operation_none,
+                              rocblas_diagonal_non_unit, n, n, one, U[b], ldu, hA[b], lda);
+            }
+            else
+            {
+                // form A = inv(U) M inv(U')
+                cblas_trtri<T>(rocblas_fill_upper, rocblas_diagonal_non_unit, n, U[b], ldu, &info);
```
EdDAzevedo, line 180: If U is (upper) triangular, I wonder whether TRSM triangular solve may be better compared to computing the explicit inverse, then multiplying with the explicit inverse.

library/src/lapack/roclapack_syevx_heevx.hpp
```diff
@@ -31,7 +31,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(BS1) syevx_sort_eigs(const rocblas_int n
 {
     // select batch instance
     rocblas_int bid = hipBlockIdx_y;
-    rocblas_int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+    rocblas_int tid = hipThreadIdx_x;
```
EdDAzevedo, line 34: Perhaps consider adding some comments about the change with tid and the use of bubble sort, say  nev is expected to be small.  If nev can be large, perhaps consider other alternatives such as shell sort algorithm.

clients/include/testing_sygvx_hegvx.hpp
```diff
@@ -0,0 +1,964 @@
...
+                              T dZ,
+                              const rocblas_int ldz,
+                              const rocblas_stride stZ,
+                              rocblas_int* dIfail,
+                              const rocblas_stride stF,
+                              rocblas_int* dInfo,
+                              const rocblas_int bc)
+{
+    // handle
+    EXPECT_ROCBLAS_STATUS(rocsolver_sygvx_hegvx(STRIDED, nullptr, itype, evect, erange, uplo, n, dA,
```
EdDAzevedo, line 44: I wonder whether the checks can be table driven as well with the expected error code as part of the table values. Just a thought.

clients/include/testing_sygvx_hegvx.hpp
```diff
@@ -0,0 +1,964 @@
...
+        // check bad arguments
+        sygvx_hegvx_checkBadArgs<STRIDED>(handle, itype, evect, erange, uplo, n, dA.data(), lda,
+                                          stA, dB.data(), ldb, stB, vl, vu, il, iu, abstol,
+                                          dNev.data(), dW.data(), stW, dZ.data(), ldz, stZ,
+                                          dIfail.data(), stF, dInfo.data(), bc);
+    }
+}
+
+template <bool CPU, bool GPU, typename T, typename Td, typename Th>
+void sygvx_hegvx_initData(const rocblas_handle handle,
```
EdDAzevedo, line 211: I wonder whether the lapack testing code can be leveraged to generate the test matrices with known eigen values and eigen vectors.  Just a thought.

clients/include/rocsolver.hpp
```diff
@@ -6273,6 +6273,286 @@ inline rocblas_status rocsolver_sygvd_hegvd(bool STRIDED,
 }
 /********************************************************/
 
+/******************** SYGVX/HEGVX ********************/
+// normal and strided_batched
+inline rocblas_status rocsolver_sygvx_hegvx(bool STRIDED,
```
EdDAzevedo, line 6278: Is there a performance reason to inline these routines? Is there a concern about too much replicated code and too many registers used with very aggressive inlining? Just a thought.

clients/common/lapack_host_reference.cpp
```diff
@@ -1881,7 +1990,7 @@ void ssygv_(int* itype,
             int* lda,
             float* B,
             int* ldb,
-            float* W,
+            float* D,
```
EdDAzevedo, line 1993: what is the reason for the change from "W" to "D"? The lapack ssygv() uses "W".

EdDAzevedo (approved):

clients/common/lapack_host_reference.cpp
tfalders: Oh, that's a change I forgot to revert. Thanks for catching it! Originally, I thought we were going to have D and E be parameters to sygvx like we have for sygv, but that turned out to not be the case.

clients/include/testing_sygsx_hegsx.hpp
tfalders: Sounds like a good idea; I'll give it a try!

clients/include/testing_sygvx_hegvx.hpp
tfalders: Right! I was thinking of trying that out.

library/src/lapack/roclapack_syevx_heevx.hpp
tfalders: nev can be large, as large as n. Optimizing this sort algorithm is something we should probably do eventually, but I'd prefer to do that in a separate PR and it's not that high on my priority list right now.

clients/include/rocsolver.hpp
tfalders: Not to my knowledge, no. The file is only used by the tests, not the actual library, so performance is less of a concern. But every function in rocsolver.hpp is marked as inline, so I'm just sticking to convention.

clients/include/testing_sygvx_hegvx.hpp
tfalders: That would be a library-wide change, and I'm unclear on what benefits it would offer. Do we get much out of it other than a stylistic change?

clients/include/testing_sygvx_hegvx.hpp
tfalders: Unfortunately, I'm having trouble getting the LAPACK testing code to build properly (it's not included in our dependency install by default). I'll probably hold off on this for now, since the test matrix generation is working as-is.

library/include/rocsolver-functions.h
```diff
@@ -17331,6 +17331,1100 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zhegvd_strided_batched(rocblas_handle
                                                                  const rocblas_int batch_count);
 //! @}
 
+/*! @{
+    \brief SYGVX computes a set of the eigenvalues and (optionally) eigenvectors of
```
jzuniga-amd, line 17335: "... a set of the eigenvalues and optionally the corresponding eigenvectors..."
(to be consistent with the syevx documentation)

library/src/lapack/roclapack_sygvx_hegvx.hpp
```diff
@@ -0,0 +1,325 @@
...
+    rocblas_get_stream(handle, &stream);
+
+    rocblas_int blocksReset = (batch_count - 1) / BS1 + 1;
+    dim3 gridReset(blocksReset, 1, 1);
+    dim3 threads(BS1, 1, 1);
+
+    // info=0 (starting with no errors)
+    ROCSOLVER_LAUNCH_KERNEL(reset_info, gridReset, threads, 0, stream, info, batch_count, 0);
+
+    // quick return with m = 0
```
jzuniga-amd, line 254: "with n = 0"

jzuniga-amd (commented):
Some minor comments below:

clients/include/testing_sygsx_hegsx.hpp
```diff
@@ -446,9 +493,9 @@ void testing_sygsx_hegsx(Arguments& argus)
     }
 
     // validate results for rocsolver-test
-    // using n * machine_precision as tolerance
+    // using 3 * n * machine_precision as tolerance
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, n);
+        ROCSOLVER_TEST_CHECK(T, max_error, 3 * n);
```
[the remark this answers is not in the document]
tfalders: Without the increased tolerance, the complex types fail some daily_lapack tests when itype = 1 (only the 640x640 and 1024x1024 sizes). I'll take another look, but I haven't been able to find any bugs so far (other than those fixed in #411) and the errors seemed relatively small.

clients/include/testing_sygvx_hegvx.hpp
```diff
@@ -0,0 +1,962 @@
...
+
+    // output results for rocsolver-bench
+    if(argus.timing)
+    {
+        if(!argus.perf)
+        {
+            rocsolver_bench_header("Arguments:");
+            if(BATCHED)
+            {
+                rocsolver_bench_output("itype", "evect", "uplo", "n", "lda", "ldb", "vl", "vu",
```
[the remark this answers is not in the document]
tfalders: Thanks for catching that!

clients/include/testing_sygsx_hegsx.hpp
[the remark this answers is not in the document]
jzuniga-amd: oh, maybe you just have to ensure that A is strictly diagonally dominant (ie. the element in the diagonal must be larger than the sum of all other elements in the row/column)... 
I just had a look, and we are not making A tri-diagonal here, so when there are many columns (like for n=1000), there is no guarantee that A is diagonally dominant.
So,  try changing line 143 and 149 by:
`hA[b][i + j * lda] = std::real(hA[b][i + j * lda]) + 100;` and
`hA[b][i + j * lda] = (hA[b][i + j * lda] - 5) / 100;` 
for example...

This just work with minimum tolerance...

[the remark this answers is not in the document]
jzuniga-amd: The other option to consider for a future change maybe, is to simply forget trying to match lapack results (this is rather a problem of numerical reproducibility... not the purpose of our tests) and do the test  completely implicit; ie. as you already know the matrix `M` from where you constructed `(A,B)`, the result of sygst must be close to `M` for the machine precision... no need to compare with lapack.

jzuniga-amd (approved):
LGTM

cgmb (approved):

PR #415, opened 2022-04-21
Update client package dependencies

cgmb:
Update client package dependencies for the release of CentOS Stream 9 and Ubuntu 22.04.

I tested manually by installing the packages built by the CentOS Stream 9 job on this PR. The hipSOLVER jobs do that for develop, so once this is merged, then the hipSOLVER cs9 jobs should start passing.


tfalders (approved):

jzuniga-amd (approved):

saadrahim:
Can this be merged to unblock CS9 testing on for other libraries?

saadrahim:
I think this will solve the RHEL8 builds as well. Triggered job at http://math-ci.rocm.amd.com/job/hipclang-bleeding-edge/job/precheckin/job/rocSOLVER/view/change-requests/job/PR-415/

cgmb:
Ah. That's nice. I was thinking that further changes may be required. I wonder if it was showing up as version "8.6" and thus failing to compare equal to "8"? In any case, I'll retarget at master to address SWDEV-333401.

saadrahim:
I need it on develop. Please target what makes sense to you.

cgmb:
Since it's needed for 5.2, it will first go to master then get merged back to develop.

cgmb:
Jenkins seems to still be installing dependencies from develop.

cgmb:
The process of bringing it through master will be too slow. We'll get it on develop and pick it into the release, I guess.

PR #416, opened 2022-04-22
Generic CMake plumbing for install.sh and rmake.py

cgmb:
This change adds generic options to install.sh and rmake.py that will hopefully prevent the need for bespoke options that just forward a particular value to CMake (e.g., [`ROCM_PKGTYPE`](https://github.com/RadeonOpenCompute/rocm-cmake/pull/70#issuecomment-1021622532), [`--rm-legacy-include-dir`](https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/388)).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -81,7 +81,14 @@ Options:
                               (Default build type is Release)
 
   --cmake-arg <argument>      Forward the given argument to CMake when configuring the build.
+
   --rm-legacy-include-dir     Remove legacy include dir Packaging added for file/folder reorg backward compatibility.
+
+    -D <arg>                         Pass '-D <arg>' to CMake during configuration.
+
+    -G <arg>                         Pass '-G <arg>' to CMake during configuration.
+
+    -C <arg>                         Pass '-C <arg>' to CMake during configuration.
 EOF
 }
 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -81,7 +81,14 @@ Options:
...
 
   --cmake-arg <argument>      Forward the given argument to CMake when configuring the build.
+
   --rm-legacy-include-dir     Remove legacy include dir Packaging added for file/folder reorg backward compatibility.
+
+    -D <arg>                         Pass '-D <arg>' to CMake during configuration.
+
+    -G <arg>                         Pass '-G <arg>' to CMake during configuration.
+
+    -C <arg>                         Pass '-C <arg>' to CMake during configuration.
```
tfalders, line 91: Does `cmake-arg` not already provide this functionality?

tfalders: Okay. I think it would make the most sense to put the new args together with `--cmake-arg` instead of after `--rm-legacy-include-dir`; please move them. Also, are the new arg names and descriptions supposed to be misaligned with the others?

cgmb:
I'm just going to close this. This is kind of a half-measure, but it would take a more energy than I have right now to get all teams on the same page.

PR #418, opened 2022-04-26
Added syevx_inplace

tfalders:
This PR adds syevx_inplace to rocSOLVER's private API, for use by hipSOLVER.

The method and its test are largely the same as for syevx, with the following changes:
- The matrix Z is removed; eigenvectors are instead returned in the matrix A
- The array ifail is removed
- The array nev, which contains the number of returned eigenvalues/vectors, is now a host pointer rather than a device pointer. A device synchronization is required to read nev from the device (where it is calculated) to the host.

### DIFF
--- a/clients/gtest/syevx_heevx_gtest.cpp
+++ b/clients/gtest/syevx_heevx_gtest.cpp
@@ -4,6 +4,7 @@
  * ************************************************************************ */
 
 #include "testing_syevx_heevx.hpp"
+#include "testing_syevx_heevx_inplace.hpp"
 
 using ::testing::Combine;
 using ::testing::TestWithParam;
@@ -46,7 +47,7 @@ const vector<vector<int>> large_size_range = {{192, 192, 192, 390, 420, 100, 170
                                               {300, 300, 330, 405, 420, 200, 300}};
 
 template <typename T>
-Arguments syevx_heevx_setup_arguments(syevx_heevx_tuple tup)
+Arguments syevx_heevx_setup_arguments(syevx_heevx_tuple tup, bool inplace)
 {
     using S = decltype(std::real(T{}));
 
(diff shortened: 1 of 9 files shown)


clients/include/rocsolver.hpp
```diff
@@ -5873,6 +5937,206 @@ inline rocblas_status rocsolver_syevx_heevx(bool STRIDED,
 }
 /********************************************************/
 
+/******************** SYEVX/HEEVX_INPLACE ********************/
+// normal and strided_batched
+inline rocblas_status rocsolver_syevx_heevx_inplace(bool STRIDED,
```
EdDAzevedo, line 5942: Is "inline static" needed to not expose the name of the function in case the compiler ignores the hint for "inline"?  Just a thought.

cgmb: This is in the `clients` subdirectory, so it will only be going into the rocsolver-test and rocsolver-bench executables. We only need to worry about symbol visibility for code in the `library` or `common` subdirectories.

EdDAzevedo (approved):

cgmb (approved):

clients/gtest/syevx_heevx_gtest.cpp
```diff
@@ -46,7 +46,7 @@ const vector<vector<int>> large_size_range = {{192, 192, 192, 5, 15, 100, 170},
                                               {300, 300, 330, -15, -5, 200, 300}};
 
 template <typename T>
-Arguments syevx_heevx_setup_arguments(syevx_heevx_tuple tup)
+Arguments syevx_heevx_setup_arguments(syevx_heevx_tuple tup, bool inplace)
```
jzuniga-amd, line 49: I think you can remove all the "inplace" stuff from this file now, right?

jzuniga-amd (approved):
LGTM!

clients/gtest/syevx_heevx_gtest.cpp
tfalders: Oh, thanks for catching that!

PR #419, opened 2022-04-27
Use consistent naming for rocblas_erange and rocblas_eorder variables

tfalders:
For ease of use, the arguments to `rocsolver-bench` are designed to have the exact same names as the function arguments in `rocsolver-functions.h`. However, this is currently broken due to the existence of two different names for arguments of type `rocblas_erange`: "range" and "erange", only one of which is represented in client.cpp.

This PR adopts a consistent naming strategy for arguments of type `rocblas_erange`, and in order to be consistent with arguments of type `rocblas_evect`, I've adopted use of the "erange" name. For additional consistency I've also changed the name of arguments of type `rocblas_eorder` from "order" to "eorder".

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -361,7 +361,7 @@ try
             "                           ")
 
         // stebz options
-        ("order",
+        ("eorder",
          value<char>()->default_value('E'),
             "E = entire matrix, B = by blocks.\n"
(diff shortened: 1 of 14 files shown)


cgmb (approved):
Sure. I think the naming confused me a bit when I was working on the rocsolver-bench tests. Speaking of which, you'll need to update them for the change to the flags.

jzuniga-amd (approved):

PR #421, opened 2022-04-28
Remove header wrapper

urmbista:


saadrahim:
This is a packaging change only. Surprisingly, it has CI failures. Can someone assist to check if the CI failure can be ignored? Alternatively, can a fix be merged?

cgmb (approved):
A more standard way of doing this would be to use [`CMakeDependentOption`](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) to set a different default value on Windows. It will also [mark the option as advanced](https://cmake.org/cmake/help/latest/command/mark_as_advanced.html) so it doesn't show up in the list of available options when `WIN32` is true.
```cmake
include(CMakeDependentOption)
cmake_dependent_option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY "Build with file/folder reorg with backward compatibility enabled" ON "NOT WIN32" OFF)
```

With that said, you have a lot of these PRs and we're not at the point on Windows where we're trying to optimize the experience of external developers building from source. So, the difference doesn't really matter that much. I'm fine with leaving it as-is. We'll be deleting this option when ROCm 6 rolls around anyways.

cgmb (changes requested):
There seems to be the same problem here as in https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/99. There's multiple places where `BUILD_FILE_REORG_BACKWARD_COMPATIBILITY` is checked (e.g., in [library/src/CMakeLists.txt](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/bd0acd0f573e42385ce89707662089fb0d7f0280/library/src/CMakeLists.txt#L259), but you've only changed one. Perhaps it's worth reconsidering that dependent option approach I'd mentioned, as it would keep the platform logic in one place.

Sorry to turn this around, but I don't usually look for correctness matters in my code reviews, so I missed this earlier. Verifying that your proposed change functions correctly is _your_ responsibility when you submit a PR. That may be difficult, especially on Windows. Still, it is necessary. The process of building on Windows will never get easier if we let people skip it just because it's hard.

I wouldn't suggest doing it too often, but one thing that can make it easier is opening a PR as a draft to indicate that it's not ready for review, and letting the CI build a package for you. Looking through the packages built on the Windows CI for this PR, I see that there appears to still be header wrappers on Windows. If you'd like, send me a message privately and I can show you how to review the Windows CI build artifacts.

lawruble13:
@cgmb I had suggested having the option be dependent on whether we're building on Windows, the reasoning behind not doing so was that we didn't want to allow Windows developers to use the header wrappers. Personally, I agree with you that making it a dependent option would be a much cleaner solution, but given that this is a temporary measure anyway I don't think putting too much effort into the experience is worth it.

cgmb:
> the reasoning behind not doing so was that we didn't want to allow Windows developers to use the header wrappers

I know this is not your argument, but this is nonsense. If they're building our software, they have the source code. They can change whatever they want and it's not possible to stop them.

Any user advanced enough to figure out how to build our code from source on Windows that purposefully chooses to use a hidden option that's explicitly marked as platform-dependent... is not worth spending effort fighting against.

> given that this is a temporary measure anyway I don't think putting too much effort into the experience is worth it.

Agreed. Once the Windows build passes, I'll verify the resulting artifacts. Assuming they look correct, I'll approve the PR as-is.

cgmb:
```
../../library/src/auxiliary/rocauxiliary_aliases.cpp:5:10: fatal error: 'rocsolver-aliases.h' file not found
```
This PR fails but #426 passed... that is very weird.

cgmb (approved):
> Personally, I agree with you that making it a dependent option would be a much cleaner solution, but given that this is a temporary measure anyway I don't think putting too much effort into the experience is worth it.

@lawruble13, I think the lesson to take from this PR is: it's usually _less effort_ to do things right than to slap something in there. This PR has been open and broken for 54 days. It took several hours of my time to debug the problems in this set of changes because things were missed multiple times (and setting up our Windows environment is a pain).

I've updated with my recommended solution.

cgmb:
@jzuniga-amd, @tfalders, I'd like to merge this next. In #430, we introduced a case where rocSOLVER depends on the backwards-compatibility layer. I've fixed that in this change.

Once this has been merged, the Windows CI will build with `BUILD_FILE_REORG_BACKWARD_COMPATIBILITY` set to `OFF` and thereby ensure the CI will report a build failure if there are any places where we are depending on the backwards-compatibility layer to build rocSOLVER itself.

jzuniga-amd (approved):

PR #422, opened 2022-04-28
Enable Origin header file dump option for wrapper

arvindcheru:
Summary of proposed changes:

- Original Header File Dump enabled for version header wrapper to support backward Compatibility for PT/TF
- ROCM-CMAKE Version required https://github.com/RadeonOpenCompute/rocm-cmake/commit/d108dbf05e029996d5d7bcbe258abb1166547a30


jzuniga-amd (approved):

cgmb:
There was a merge and a revert. The revert undoes the merge by adding a new commit that does the opposite change. However, if we merged it like that, we'd end up with all the commit messages getting added to the squashed commit.

I've rolled the PR branch back to before the mistaken merge using `git reset --hard`.

PR #427, opened 2022-05-19
Added syevj

tfalders:
syevj calculates eigenvalues and eigenvectors by the iterative Jacobi method.


tfalders:
It seems there is a bug with the complex case. I will close for now and re-open once it's fixed.

clients/include/testing_syevj_heevj.hpp
```diff
@@ -323,7 +323,7 @@ void syevj_heevj_getPerfData(const rocblas_handle handle,
                              const bool profile_kernels,
                              const bool perf)
 {
-    constexpr bool COMPLEX = is_complex<T>;
+    constexpr bool COMPLEX = rocblas_is_complex<T>;
 
     int lwork = (COMPLEX ? 2 * n - 1 : 0);
```
EdDAzevedo, line 328: Should lwork == 0? Perhaps it should be  say something like

int const lwork = (COMPLEX) ? 2*n-1 : 3*n-1 ; 


The lapack documentation also says setting a larger value  lwork >= (NB+2)*N for higher performance.

library/src/include/lib_device_helpers.hpp
```diff
@@ -37,6 +37,30 @@ __device__ S aabs(T val)
...
+{
+    return z * z;
+}
+
+template <typename S, typename T, std::enable_if_t<rocblas_is_complex<T>, int> = 0>
+__device__ S lengthsq(const T& z)
+{
+    return z.real() * z.real() + z.imag() * z.imag();
+}
+
```
cgmb, line 63: What do you think about `magnitude` and `magnitude_sq`? Or `abs` and `abs_sq`? I will freely admit that 90% of my math background is working in real vector spaces, so I don't have a great feel for complex vocabulary. Still, `length` feels weird.

EdDAzevedo: If this is not performance critical, you might consider scaling the real part and complex part  to avoid numerical overflow. 

Maybe something like

auto  hypot = [](double const  a, double const b) -> double {
  // evaluate  sqrt( a*a + b*b)  as  abs(a) * sqrt( 1 + (b/a)*(b/a) ) or
  //   abs(b) * sqrt( (a/b)*(a/b) + 1 )

   double ans = 0;
   if ((a == 0) && (b == 0)) { return( 0 ) };

   double const abs_a = (a > 0) ? a : (-a); 
   double const abs_b = (b > 0) ? b : (-b);

   if (abs_a > abs_b) {
      double const c = abs_b/abs_a;  
       ans = ( abs_a * sqrt( 1 + c * c ) );
      };
   else {
      double const c = abs_a/abs_b;
       ans = ( abs_b * sqrt( 1 + c * c ) );
    };
 return( ans );
}

clients/benchmarks/client.cpp
```diff
@@ -408,7 +402,20 @@ try
             "                           Note: the used random input matrices have all eigenvalues in [-20, 20].\n"
             "                           ")
 
+        // iterative Jacobi options
+        ("max_sweeps",
+         value<rocblas_int>()->default_value(100),
+            "Maximum number of sweeps/iterations.\n"
+            "                           Used in iterative Jacobi functions.\n"
+            "                           ")
```
cgmb, line 410: This should probably be added to the benchmark tests.

tfalders: Aren't the parameters in the benchmark tests all `char`s? I don't see any tests for numeric parameters like `max_sweeps`.

library/src/include/lib_device_helpers.hpp
tfalders: Sure, I can change it to `magnitude` and `magnitude_sq`. I'll also take a look at doing the proposed scaling - if it doesn't have a noticeable impact on the performance, I'll make the change.

jzuniga-amd: Yes, `abs` is the right name for these helpers, although I don't see why you are re-writing this functionality. rocblas_complex have `abs` implementations which guards against overflow as Ed suggested (see https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/f6730d8614fa2a23b4c1402eb6c626197bf36cbf/library/include/internal/rocblas-complex-types.h#L481)

I think rocBLAS is injecting its `abs` into std, but anyways, since we switched to c++11 `std::abs` (being rocblas or not) should give you the magnitude or modulo of any number (real or complex)

tfalders: Oh, interesting! I didn't realize rocBLAS had this functionality. Originally, I was using rocSOLVER's `aabs` function, which does not call `abs` but `asum`. I guess I can call `abs` directly and it should work.

cgmb: > I think rocBLAS is injecting its abs into std, but anyways, since we switched to c++11 std::abs (being rocblas or not) should give you the magnitude or modulo of any number (real or complex)

We really should get them to change that, because

> It is undefined behavior to declare a full specialization of any standard library function template. (since C++20)

https://en.cppreference.com/w/cpp/language/extending_std

Frankly, I find it very confusing anyways. What even provides the `__device__` implementations of `std::abs` and `std::norm` for real values?

clients/benchmarks/client.cpp
cgmb: Ah. Yeah, that's fine. I suppose the function itself is in the limited precision list too, which is still on the todo list.

library/src/include/lib_device_helpers.hpp
cgmb: Looking closer, those aren't even specializations. They're just new definitions added to the std namespace, so I think that falls under:
> It is undefined behavior to add declarations or definitions to namespace std or to any namespace nested within std

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -0,0 +1,449 @@
...
+                }
+
+                // apply J' from the left
+                syevj_lasr_left(n, i, j, c, s1, s2, Acpy, n);
+
+                // update eigenvectors
+                if(evect != rocblas_evect_none)
+                    syevj_lasr_right(n, i, j, c, s1, s2, A, lda);
+            }
+            __syncthreads();
```
jzuniga-amd, line 231: This barrier and all the other syncs in this kernel only synchronizes the threads within the same thread-group. In other words, when n>512 (as BS1=256), at this point you will have threads belonging to different groups and they wont be synchronized; there is no guarantee that this kernel will work... (see two consecutives runs below)

>> $$./clients/staging/rocsolver-bench -f syevj -n 515 -v 1 --iters 1
============================================
Arguments:
============================================
evect           uplo            n               lda             abstol          max_sweeps
N               U               515             515             0               100
============================================
Results:
============================================
cpu_time_us     gpu_time_us     error
29503           1367509         _**1.4387098860657362e+16**_

>> $$./clients/staging/rocsolver-bench -f syevj -n 515 -v 1 --iters 1
============================================
Arguments:
============================================
evect           uplo            n               lda             abstol          max_sweeps
N               U               515             515             0               100
============================================
Results:
============================================
cpu_time_us     gpu_time_us     error
29547           21505279        _**0**_

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -0,0 +1,449 @@
...
+                i = top[tid - 1];
+            if(tid == half_n - 1)
+                j = top[half_n - 1];
+            else
+                j = bottom[tid + 1];
+            __syncthreads();
+
+            top[tid] = i;
+            bottom[tid] = j;
+            __syncthreads();
```
jzuniga-amd, line 255: threads from different groups are not synchronized. One group could start working with a column/row that is still being modified by another group! 
This design needs to be modified...

jzuniga-amd (changes requested):

library/src/include/lib_device_helpers.hpp
jzuniga-amd: This may be a work to be done in conjunction with the rocBLAS team. We have been using `std::abs` in all other rocsolver functions... I don't know whether it is calling the original `abs` in the standard library, or the rocBLAS implementation, but it's working OK. For now let's just don't add more helpers for the same functionality.

library/src/lapack/roclapack_syevj_heevj.hpp
tfalders: How in the world did I forget that? Thanks, I'll take care of it right away.

jzuniga-amd (approved):
LGTM. Thanks Troy!
(Performance gain of this first implementation was not as good as predicted but it produces correct results. We'll merge as feature complete is approaching. A round of optimizations will follow in a different PR later)

cgmb (approved):

PR #430, opened 2022-05-25
Added sygvdx_inplace

tfalders:
Adds sygvdx_inplace to rocSOLVER's private API, for use by hipSOLVER.


library/src/lapack/roclapack_sygvdx_hegvdx_inplace.hpp
```diff
@@ -0,0 +1,308 @@
...
+
+    // perform Cholesky factorization of B
+    rocsolver_potrf_template<BATCHED, T, S>(handle, uplo, n, B, shiftB, ldb, strideB, info,
+                                            batch_count, scalars, work1, work2, work3, work4,
+                                            (T*)work7_workArr, iinfo, optim_mem);
+
+    /** (TODO: Strictly speaking, computations should stop here if B is not positive definite.
+        A should not be modified in this case as no eigenvalues or eigenvectors can be computed.
+        Need to find a way to do this efficiently; for now A will be destroyed in the non
+        positive-definite case) **/
```
cgmb, line 252: Can we extend the `info_mask` pattern used by [`copy_mat`](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/7bfa7f500143097f937b49b3c94e1f47b3b1c3d3/library/src/include/lib_device_helpers.hpp#L133-L138) into our other functions?

tfalders: It would have to be extended into rocBLAS. It's only because `copy_mat` is general-use that we need to tell it when to ignore info - I think pretty much any other rocSOLVER kernel is specific enough that the masking or lack thereof can be hard-coded.

cgmb: That sounds reasonable. Which rocBLAS functions would it need to be extended to?

tfalders: Based on the profile log for this function (ignoring potrf), those would be dot, syr2, symv, scal, gemv, syr2k, symm, trsv, and trsm (though trsm will most likely be replaced by our internal trsm, which we'd have to add the mask to as well).

jzuniga-amd (approved):
LGTM

cgmb (approved):

library/src/lapack/roclapack_sygvdx_hegvdx_inplace.hpp
cgmb: I will open a feature request. I have a number of tickets to make... today will be a day of Jira. Yay!

tfalders:
I'm going to hold off on merging this until #437 is merged in. This function calls trsm and will need to be updated to use our internal kernels.

tfalders:
> @tfalders with the new lazy CI scheme your PR now won't re-run unless you manually re-run it or merge develop into it, your last fails seem to be 4 days ago so if fixed by external commits you can rerun.

@TorreZuk I triggered a windows build that failed 2 days ago, and I believe it's the last required test for our automatic merge. I haven't been re-running the extended test in order to reduce the CI load and since it isn't required. Is there a way to see which version of rocBLAS the CI is going to use before I trigger a re-run? I hate to waste CI resources when I know there will be a failure.

TorreZuk:
If you go to the classic mode on develop build the artifacts links are there, you can also browse artifactory.  The last passing develop rocSOLVER should have grabbed the one you'll get, but if it wasn't run latesly just go to artifactory and take the last artifact from the same location.  e.g. http://mlsejenkinsvm.amd.com/job/ROCmMathLib/job/rocSOLVER/job/develop/476/

PR #431, opened 2022-05-26
rocblas_is_complex instead of is_complex

daineAMD:
Search-and-replace of `is_complex` to `rocblas_is_complex` (see https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1216).

Don't mean to step on any toes, just noticed an abundance of warnings when trying to build rocSOLVER against some rocBLAS changes, feel free to disregard/close if you have other plans.

### DIFF
--- a/clients/include/norm.hpp
+++ b/clients/include/norm.hpp
@@ -64,7 +64,7 @@ inline void xaxpy(int* n,
 
 /* Norm of error functions */
 
-template <typename T, std::enable_if_t<!is_complex<T>, int> = 0>
+template <typename T, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
 double norm_error(char norm_type,
                   rocblas_int M,
                   rocblas_int N,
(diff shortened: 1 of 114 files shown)


TorreZuk:
@daineAMD this is included in their PR#428 which included other change as well so think you can just close.  Although that other one has auto-merge set so not clear when it will get merged.

daineAMD:
Ah, not sure how I missed the obvious title. Closing then.

cgmb:
> @daineAMD this is included in their PR#428 which included other change as well so think you can just close. Although that other one has auto-merge set so not clear when it will get merged.

We need to update for the https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/07656cac30bb6f79904b715b1d097d80b8ec944a before any rocSOLVER tests will pass. One all tests pass, the #428 will automatically merge itself.

daineAMD:
If no one has started on the rocBLAS herk/syrk interfaces changes, I can make the changes this morning and open a PR.

cgmb:
Thanks @daineAMD. That would be great!

PR #435, opened 2022-05-31
Fix stebz bug for range = index, and failing tests in Windows

jzuniga-amd:
This PR fixes failing tests on Windows CI (https://ontrack-internal.amd.com/browse/SWDEV-330297).
Also, it standardizes the tests matrices as in other partial-eigenproblem routines.  This revealed further bugs in the actual algorithm for range = index. This PR also includes the fix.


cgmb (approved):
This is why I like testing on a variety of different platforms. It's a bunch of extra work, but it helps in catching bugs that affect all platforms.

PR #437, opened 2022-06-02
Move trsm kernels to specialized folder

tfalders:
The internal trsm kernels are now used extensively throughout the library, and would benefit from explicit instantiation in order to reduce our binary size.

Without optimizations, the release binary size before this change is 28.25 MB. After, it is 25.12 MB.


jzuniga-amd (approved):
Cool!
It looks good to me.

PR #438, opened 2022-06-03
Fix git safe.directory error on dependency install

cgmb:
## Fix git safe.directory error

When installing LAPACK and GTest with `./install.sh -cd` as a normal user, the [new git safe.directory feature](https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9) causes a build failure. This is because the first build (`make`) as a normal user just downloads the repos. The second build (`make install`) as root tries to build the libraries, but `ExternalProject_Add` first attempts to query the status of the cloned git repos.

In rocBLAS, this was fixed with https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/46b04c6de0e3e406dca339a6f0d447152909acec, and I first ported that change over to rocSOLVER. However, there's some really subtle issues with the CMake target dependencies and dropping `-j$(nproc)` like rocBLAS did makes the build rather slow.

My solution was to just stop using git and move the dependency build into the shell script rather than doing it in CMake. This was much less code and is conceptually simpler.

## Installing dependencies in one call

This is basically a port of https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/478. It's faster and simpler to just call the installer once.

## Remove llvm7.0

I think this was Tensile-related, but it was removed from rocBLAS ages ago.

## Remove check_exit_code

```bash
<command>
check_exit_code "$?"
```
is equivalent to
```
<command> || exit
```

## Remove `id -u`

The built-in bash variable `$EUID` is equivalent to `$(id -u)`.


cgmb:
This is a draft as I still need to manually verify it on all platforms.

cgmb:
The new behaviour for the install script is now:

> For each dependency:
> 1. Download the tarball to `deps/<name>-<version>.tar.gz` (skip if already exists)
> 2. Extract source to `deps/<name>-<version>-src` (skip if already exists)
> 3. Build source in `deps/<name>-<version>-bld`

Anyways, this is ready for review so please take a look.

jzuniga-amd (approved):
LGTM. Thanks Cory!
Adding an option to build dependencies without any sudo privileges can follow in a future PR. Maybe now that all the packages build instructions are in the bash script, it could be easier/cleaner.

tfalders (approved):
LGTM!

cgmb:
> Adding an option to build dependencies without any sudo privileges can follow in a future PR. Maybe now that all the packages build instructions are in the bash script, it could be easier/cleaner.

That sounds like a good idea to me. There's probably some things we can draw from #209.

PR #439, opened 2022-06-07
Using rocblas header file for block sizes

daineAMD:
Companion to [rocBLAS #1289](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1289). Moving rocBLAS block sizes to a header file.

### DIFF
--- a/library/src/include/rocblas.hpp
+++ b/library/src/include/rocblas.hpp
@@ -12,7 +12,7 @@
 #include "lib_host_helpers.hpp"
 #include "rocblas/internal/rocblas-exported-proto.hpp"
 #include "rocblas/internal/rocblas_device_malloc.hpp"
-#include "rocblas/internal/rocblas_block_sizes.hpp"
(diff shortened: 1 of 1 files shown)


jzuniga-amd (approved):

tfalders (approved):
LGTM, though I'm not sure how much this will actually help cut down on maintenance. Changes in block size often seem to be accompanied by new or renamed constants, which we'll still have to account for in rocSOLVER.

cgmb:
> LGTM, though I'm not sure how much this will actually help cut down on maintenance. Changes in block size often seem to be accompanied by new or renamed constants, which we'll still have to account for in rocSOLVER.

We don't really seem to get much out of having them as templates given that only certain values are actually supported anyway. Is rocSOLVER actually going to do anything with this blocksize information besides just passing it back to rocBLAS? If we don't actually need those details, maybe hiding them behind a C API would make the interface between rocBLAS and rocSOLVER more stable?

PR #442, opened 2022-06-15
Updating herk syrk rocblas block sizes

daineAMD:
Changed specificity of herk/syrk block sizes in rocblas.

Needs rocBLAS commit [63ec9562](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/commit/63ec9562d0d0817e61182bf5db5ad22c4ee0b0a1)

### DIFF
--- a/library/src/include/rocblas.hpp
+++ b/library/src/include/rocblas.hpp
@@ -1065,9 +1065,8 @@ rocblas_status rocblasCall_syrk_herk(rocblas_handle handle,
 
     using S = decltype(std::real(T{}));
 
-    constexpr rocblas_int NB = BATCHED  ? ROCBLAS_SDSYRK_BATCHED_NB
-        : std::is_same<T, float>::value ? ROCBLAS_SSYRK_NB
-                                        : ROCBLAS_DCZSYRK_NB;
+    constexpr rocblas_int NB = BATCHED ? ROCBLAS_SDSYRK_BATCHED_NB : ROCBLAS_SDZSYRK_NB;
+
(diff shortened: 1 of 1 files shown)


cgmb (approved):
Thanks @daineAMD! Just repeating [my question from the last thread](https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/439#issuecomment-1149426229):

> Is rocSOLVER actually going to do anything with this blocksize information besides just passing it back to rocBLAS?

That's an open question to all rocBLAS or rocSOLVER developers. I'm not sure what we might be able to do with it in rocSOLVER, but perhaps I just lack imagination.

TorreZuk:
Just to add to Daine's reply to you @cgmb, these aren't passed back to rocblas so not overhead, they are just a required part of the templated function name that you want called. We could hide these from the names by adding an extra layer in our API but don't think it is trivial, or could try pushing these constants down one level on our side.  For historic reasons I think they percolated up as they are chosen by datatype, so it was natural to move up to where the type based dispatching occurs early.  We might try to force them down a level by by using type inspection at some point.

tfalders (approved):
> As I remember, the original reason rocSOLVER opted to use the internal C++ templates of rocBLAS was to have access to the offset parameters, at least I think.

We also wanted to bypass the overhead of argument checking and logging, AFAIK.

I think at some point it might be good if rocBLAS defines a block size for each precision, even if they're identical, so that these can be changed on the rocBLAS side without affecting rocSOLVER.

PR #448, opened 2022-06-27
syevj Nonsorting option

jzuniga-amd:
Adds non-sorting option to the Jacobi eigensolvers.

I think this option could be actually useful and, as these are not legacy lapack APIs, I see no problem with adding it.
This also will give full support to the functionality in hipSOLVER https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/114

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -409,6 +409,13 @@ try
             "                           Used in iterative Jacobi functions.\n"
             "                           ")
 
+        ("esort",
+         value<char>()->default_value('A'),
+            "N = no sorting, B = ascending order.\n"
+            "                           Indicates whether the computed eigenvalues are sorted in ascending order.\n"
+            "                           Used in iterative Jacobi functions.\n"
+            "                           ")
+
         // other options
         ("abstol",
          value<double>()->default_value(0),
@@ -511,6 +518,7 @@ try
     argus.validate_evect("evect");
     argus.validate_erange("erange");
     argus.validate_eorder("eorder");
+    argus.validate_esort("esort");
     argus.validate_itype("itype");
 
     // prepare logging infrastructure and ignore environment variables
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -411,7 +411,7 @@ try
 
         ("esort",
          value<char>()->default_value('A'),
-            "N = no sorting, B = ascending order.\n"
+            "N = no sorting, A = ascending order.\n"
             "                           Indicates whether the computed eigenvalues are sorted in ascending order.\n"
             "                           Used in iterative Jacobi functions.\n"
             "                           ")
--- a/clients/gtest/syevj_heevj_gtest.cpp
+++ b/clients/gtest/syevj_heevj_gtest.cpp
@@ -52,6 +52,9 @@ Arguments syevj_heevj_setup_arguments(syevj_heevj_tuple tup)
     arg.set<char>("evect", op[0]);
     arg.set<char>("uplo", op[1]);
 
+    // only need to test the sorted case
+    arg.set<char>("esort", 'A');
+
     arg.set<double>("abstol", 0);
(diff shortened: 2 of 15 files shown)


clients/benchmarks/client.cpp
```diff
@@ -409,6 +409,13 @@ try
             "                           Used in iterative Jacobi functions.\n"
             "                           ")
 
+        ("esort",
+         value<char>()->default_value('A'),
+            "N = no sorting, B = ascending order.\n"
```
tfalders, line 414: Should that be 'A' instead of 'B'?

tfalders (commented):
As noted in https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/114#discussion_r907619784, I would personally prefer to create a separate method that has this functionality, kind of like our npvt methods. This would also have the advantage of allowing us to add non-sorting versions of the other eigensolver methods too, which would otherwise require deprecations and would disrupt the resemblance to LAPACK's API.

That said, the PR looks good as-is, so if I'm out-voted I will approve it.

clients/benchmarks/client.cpp
jzuniga-amd: Yes, should be A :)

jzuniga-amd:
I think we all agree that having a sorting option is nice. I don't see to much use of having a different API (almost identical) and repeating all the code, testing, etc. just to introduce an additional parameter that only acts as a quick return before doing the sorting...

We do introduce different APIs when the base API is legacy LAPACK so we don't diverge. But this is not the case; the Jacobi method is not in LAPACK so I think we can  define it as we want. 
We would need to add a different API (let's say syev_nosort) if we want to extend this to the classic eigensolver, but this can be done later in other PR.

jzuniga-amd:
> though I'm a bit concerned about the way the test won't check the eigenvalues in the `evect = N` case.

Yes, I though about this...
But the google test always test eigenvalues because we always pass esort=ascending.  This could be a problem for the bench client if we use `-v 1` and `--esort N`, but if the eigenvalues are wrong, then the sorted case should fail too (the values are the same), so we better be detecting the problem  in the google test anyways.  

That said, if we want to check the eigenvalues when they are not sorted and the eigenvectors are not computed, we need to sort them on the CPU before comparing with lapack... I think too much work that is really not needed. What do you think?

cgmb (approved):
I don't have strong feelings about whether this should be together or separate. It kind of makes sense to me for it to be a parameter, as the functionality seems to be in keeping with the other things that can be set via parameters. However, I'm happy with whatever you two decide.

cgmb:
> That said, if we want to check the eigenvalues when they are not sorted and the eigenvectors are not computed, we need to sort them on the CPU before comparing with lapack... I think too much work that is really not needed. What do you think?

I'm not sure we want rocsolver-bench failing for `-v 1 --esort N` over the long term, but I'm fine with that for now. We could put a task for that into the backlog and deal with it later.

jzuniga-amd:
No, it won't fail! It will just not compute the numerical error on the eigenvalues (because we cannot compare with lapack as its returned eigenvalues are always sorted)... so the reported error would be 0 unless there are convergence or residual problems.

tfalders (approved):
Very well, I'll approve.

> We would need to add a different API (let's say syev_nosort) if we want to extend this to the classic eigensolver, but this can be done later in other PR.

I believe in using a consistent approach, so if this PR is the approach we're going to take, I will oppose creating _nosort methods for the classic eigensolvers in the future. I'd expect us to add esort as an argument to any method we want to have nonsorting behaviour, even if that requires deprecations.

cgmb:
> I believe in using a consistent approach, so if this PR is the approach we're going to take, I will oppose creating _nosort methods for the classic eigensolvers in the future. I'd expect us to add esort as an argument to any method we want to have nonsorting behaviour, even if that requires deprecations.

I think that's acceptable. With some care, I think we can find a strategy to transition the other functions. There's lots of tools in our toolbox that we've never used.

PR #450, opened 2022-06-28
Use explicit template instantiation in tests for faster compilation

qjojo:
This forward declares all the template instantiations used in client/include/rocsolver_dispatcher.hpp and explicitly instantiates them in client/common/real_template_inst.cpp and client/common/complex_template_inst.cpp. Since this stops us from doing lots of redundant instantiations and lets us do it in parallel it speeds up compilation by a noticeable amount, on my machine it saves ~26s each when compiling rocsolver-bench and the gtest client.

In the future it would be a good idea to follow this pattern for any other functions we add to the maps in rocsolver_dispatcher.hpp

Ran rocsolver-test locally with no issues.


clients/common/complex_template_inst.cpp
```diff
@@ -0,0 +1,114 @@
...
+template void testing_syevd_heevd<true, true, rocblas_double_complex>(Arguments& argus);
+template void testing_syevd_heevd<true, true, rocblas_float_complex>(Arguments& argus);
+template void testing_syevx_heevx<false, true, rocblas_double_complex>(Arguments& argus);
+template void testing_syevx_heevx<false, true, rocblas_float_complex>(Arguments& argus);
+template void testing_sytxx_hetxx<false, true, 0, rocblas_double_complex>(Arguments& argus);
+template void testing_sytxx_hetxx<false, true, 0, rocblas_float_complex>(Arguments& argus);
+template void testing_sytxx_hetxx<false, true, 1, rocblas_double_complex>(Arguments& argus);
+template void testing_sytxx_hetxx<false, true, 1, rocblas_float_complex>(Arguments& argus);
+template void testing_ormbr_unmbr<rocblas_double_complex>(Arguments& argus);
+template void testing_ormbr_unmbr<rocblas_float_complex>(Arguments& argus);
```
TorreZuk, line 46: I just finished breaking up such a coding pattern in rocBLAS so will comment on this.  Files are cheap so instantiating all these based on function name (so one file for say sygsx both real and complex) makes more sense to me.  Compiling this mega unit is a pain for the compiler even if parallel options are turned on.  If you split it into function based intantiated .cpp all that compilation will be in parallel as well.

TorreZuk (commented):
Suggestions made

clients/common/complex_template_inst.cpp
qjojo: Good point, I'll split these off into a file for each testing_*.hpp

clients/common/testing_sygsx_hegsx.cpp
```diff
@@ -0,0 +1,26 @@
...
+template void testing_sygsx_hegsx<false, false, 1, double>(Arguments& argus);
+template void testing_sygsx_hegsx<false, false, 1, float>(Arguments& argus);
+template void testing_sygsx_hegsx<false, true, 1, double>(Arguments& argus);
+template void testing_sygsx_hegsx<false, true, 1, float>(Arguments& argus);
+template void testing_sygsx_hegsx<false, false, 0, double>(Arguments& argus);
+template void testing_sygsx_hegsx<false, false, 0, float>(Arguments& argus);
+template void testing_sygsx_hegsx<true, true, 0, double>(Arguments& argus);
+template void testing_sygsx_hegsx<true, true, 0, float>(Arguments& argus);
+template void testing_sygsx_hegsx<false, true, 0, rocblas_double_complex>(Arguments& argus);
+template void testing_sygsx_hegsx<false, true, 0, rocblas_float_complex>(Arguments& argus);
```
tfalders, line 26: This is rather nitpicky, but I'd appreciate it if a consistent ordering were used for the template instantiations (not limited to this file). An ordering I'd suggest is like so:

- Unblocked, non-batched, real functions (false, false, 0, float then double)
- Unblocked, strided batched, real functions (false, true, 0, float then double)
- Unblocked, batched, real functions (true, true, 0, float then double)
- Blocked, non-batched, real functions (false, false, 1, float then double)
- Blocked, strided batched, real
- Blocked, batched, real
- If applicable, the complex functions, in the same order

(Same comment applies to the extern declarations)

clients/common/testing_geqr2_geqrf.cpp
```diff
@@ -0,0 +1,16 @@
...
+template void testing_geqr2_geqrf<true, true, 1, double>(Arguments& argus);
+template void testing_geqr2_geqrf<true, true, 1, float>(Arguments& argus);
+template void testing_geqr2_geqrf<false, true, 1, double>(Arguments& argus);
+template void testing_geqr2_geqrf<false, true, 1, float>(Arguments& argus);
+template void testing_geqr2_geqrf<true, true, 0, double>(Arguments& argus);
+template void testing_geqr2_geqrf<true, true, 0, float>(Arguments& argus);
+template void testing_geqr2_geqrf<false, true, 0, double>(Arguments& argus);
+template void testing_geqr2_geqrf<false, true, 0, float>(Arguments& argus);
+template void testing_geqr2_geqrf<false, false, 1, double>(Arguments& argus);
+template void testing_geqr2_geqrf<false, false, 1, float>(Arguments& argus);
```
tfalders, line 16: Many of the functions appear to be missing instantiations for the complex types. Keep in mind that almost all of our functions have support for both real and complex types. AFAIK, the only exceptions are sterf and stebz (real types only) and lacgv (complex types only).

tfalders (changes requested):
The approach looks solid, though it seems like many of the functions are missing instantiations of the complex types.

clients/include/testing_gels_outofplace.hpp
```diff
@@ -655,3 +655,5 @@ void testing_gels_outofplace(Arguments& argus)
     // ensure all arguments were consumed
     argus.validate_consumed();
 }
+
+//forward declared explicit instantiations
```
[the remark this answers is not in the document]
tfalders: IIRC, those functions are part of the private API and so I did not make them callable from the benchmark client.

cgmb:
This is what I was thinking for the macros...

```c
#define APPLY_STAMP(STAMP, ...)            STAMP(__VA_ARGS__)
#define FOREACH_BOOLEAN_INT(STAMP, F, ...) F(STAMP, __VA_ARGS__, 0)                     \
                                           F(STAMP, __VA_ARGS__, 1)
#define FOREACH_BOOLEAN_0(STAMP, F, ...)   F(STAMP, __VA_ARGS__, false)                 \
                                           F(STAMP, __VA_ARGS__, true)
#define FOREACH_BOOLEAN_1(STAMP, F, ...)   F(STAMP, __VA_ARGS__, false)                 \
                                           F(STAMP, __VA_ARGS__, true)
#define FOREACH_SCALAR_TYPE(STAMP, F, ...) F(STAMP, __VA_ARGS__, float)                 \
                                           F(STAMP, __VA_ARGS__, double)                \
                                           F(STAMP, __VA_ARGS__, rocblas_float_complex) \
                                           F(STAMP, __VA_ARGS__, rocblas_double_complex)
#define INSTANTIATE(STAMP, F, ...)         F(STAMP, __VA_ARGS__)

#define TESTING_SYGSX_HEGSX(...) template void testing_sygsx_hegsx<__VA_ARGS__>(Arguments&);
INSTANTIATE(TESTING_SYGSX_HEGSX, \
    FOREACH_SCALAR_TYPE,         \
    FOREACH_BOOLEAN_0,           \
    FOREACH_BOOLEAN_1,           \
    FOREACH_BOOLEAN_INT,         \
    APPLY_STAMP)

#define TESTING_SYGVD_HEGVD(...) template void testing_sygvd_hegvd<__VA_ARGS__>(Arguments&);
INSTANTIATE(TESTING_SYGVD_HEGVD, \
    FOREACH_SCALAR_TYPE,         \
    FOREACH_BOOLEAN_0,           \
    FOREACH_BOOLEAN_1,           \
    FOREACH_BOOLEAN_INT,         \
    APPLY_STAMP)
```

The same FOREACH can't be used more than once in a pipeline, which is why I created `FOREACH_BOOLEAN_0` and `FOREACH_BOOLEAN_1`.

cgmb:
I suppose that we don't actually want all permutations. The first two booleans are for `BATCHED` and `STRIDED`. I can never keep this straight in my mind, but if I'm understanding this correctly, the mapping is

BATCHED | STRIDED | Function
| - | - | - |
false | false | rocsolver_\<fn>
true | false | rocsolver_\<fn>_batched
false | true | rocsolver_\<fn>_strided_batched
true | true | rocsolver_\<fn>_strided_batched

thus, there is a a redundant permutation we don't want to instantiate. I would argue that the pair of `BATCHED` and `STRIDED` are actually a single property, and should really be defined by something like
```c
enum matrix_data {
  single,
  batched,
  strided_batched,
};
```
...but that's a story for a different time. In any case, naming this concept provides a way to prevent the duplication problem by combining two of the `FOREACH` macros into one. To be consistent, we might as well name all of the `FOREACH` constructs after the concepts they're used for:
```c
#define FOREACH_BLOCKED_VARIANT(STAMP, F, ...) F(STAMP, __VA_ARGS__, 0)                 \
                                               F(STAMP, __VA_ARGS__, 1)
#define FOREACH_MATRIX_DATA_LAYOUT(STAMP, F, ...) F(STAMP, __VA_ARGS__, false, false)   \
                                                  F(STAMP, __VA_ARGS__, true, false)    \
                                                  F(STAMP, __VA_ARGS__, false, true)

INSTANTIATE(TESTING_SYGSX_HEGSX, \
    FOREACH_SCALAR_TYPE,         \
    FOREACH_MATRIX_DATA_LAYOUT,  \
    FOREACH_BLOCKED_VARIANT,     \
    APPLY_STAMP)
```

cgmb:
It's nice to see that it works. The names I used when I suggested the macros were just placeholders, so feel free to adjust the naming as you see fit.

clients/include/testing_gels_outofplace.hpp
tfalders, line 659: This comment should probably be removed now that the approach has changed.

clients/include/testing_gesv_outofplace.hpp
```diff
@@ -590,3 +590,5 @@ void testing_gesv_outofplace(Arguments& argus)
     // ensure all arguments were consumed
     argus.validate_consumed();
 }
+
+//forward declared explicit instantiations
```
tfalders, line 594: This comment can be removed too.

clients/include/testing_managed_malloc.hpp
```diff
@@ -338,3 +338,5 @@ void testing_managed_malloc(Arguments& argus)
     // ensure all arguments were consumed
     argus.validate_consumed();
 }
+
+//forward declared explicit instantiations
```
tfalders, line 342: Same here

clients/include/testing_syevdx_heevdx_inplace.hpp
```diff
@@ -682,3 +682,5 @@ void testing_syevdx_heevdx_inplace(Arguments& argus)
     // ensure all arguments were consumed
     argus.validate_consumed();
 }
+
+//forward declared explicit instantiations
```
tfalders, line 686: Same here

clients/include/testing_syevj_heevj.hpp
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include "client_util.hpp"
```
tfalders, line 7: Why is there no extern declaration for syevj?

clients/include/testing_sygvj_hegvj.hpp
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include "client_util.hpp"
```
tfalders, line 7: Why is there no extern declaration for sygvj?

tfalders (commented):
Looks good! Just a couple comments to resolve before I approve.

cgmb:
> Looks good, though I don't entirely understand how it works

The implementation depends heavily on functions that get passed as arguments to other functions, but it's surprisingly elegant. Here's a brief summary of how these macro functions work together:

1. There's a `STAMP()` macro function that actually emits the text. Every other macro function in this technique just makes function calls to other macros. The stamp is the only thing that directly creates any text. All other functions take the stamp function as their first argument, because the stamp function needs to be forwarded through the entire chain of calls down to `APPLY_STAMP()`.
2. The final function in the pipeline is always `APPLY_STAMP()`, which is the function that actually calls/expands the stamp macro. It takes the stamp function as the first argument and it just calls the stamp function.
3. For the rest of the macro functions, it's easiest to understand how they work by examining their arguments. I'll skip the first argument for a moment and discuss it at the end. The second argument is the next function in the pipeline. All remaining arguments, `__VA_ARGS__`, are the arguments that should be forwarded to the next function. So, each function just calls the next function, appending whatever arguments it wants to add. By calling the next function twice, it can double the number of times the stamp is instantiated. By appending different values in each call to the next function, it can thereby instantiate the stamp with multiple different values. Circling back, the first argument is just the stamp function, which is always just forwarded as the first argument to the next function.

In other words, at the start, the list of arguments to each function is basically a list of functions. Each function pops a function from the front and then calls that function, forwarding all remaining arguments and appending some arguments of its own. This continues with the number of functions at the head shrinking and the number of values at the tail growing until the call to `APPLY_STAMP()`, by which point there should be no functions remaining and all arguments to apply stamp are values for the stamp.

A description to this effect could probably be added to the source code in the comments.


5.  temporarily ignoring the first argument (the stamp function). Each function takes as its arguments a function and the arguments that should be passed to that function (`__VA_ARGS__`).

clients/include/testing_gels_outofplace.hpp
cgmb: The macros are just a way to generate the same text that was otherwise written out manually. I agree the comment should be omitted if there's no explicit instantiations being declared, but it's not actually because anything changed in this PR.

cgmb:
I will open a PR to this PR with some further suggestions.

cgmb:
I've opened https://github.com/qjojo/rocSOLVER/pull/1 with my suggestions.

qjojo:
Big thanks to Cory for helping me clean things up, this should be good to merge.

jzuniga-amd (approved):
LGTM

tfalders (approved):

PR #452, opened 2022-06-30
Fix compatibility with older versions of LAPACK

tfalders:
After #445 was merged, the rocSOLVER build fails when both LAPACK 3.7.1 (previously required by rocSOLVER) and 3.9.1 are present on a system. This is because rocSOLVER's cmake system picks up 3.7.1, in which LAPACK_LIBRARIES is undefined. This PR ensures <strike>3.9.1 will be picked up instead</strike> LAPACK_LIBRARIES will be defined.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -20,9 +20,16 @@ endif()
 
 if(BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
   if(ROCSOLVER_FIND_PACKAGE_LAPACK_CONFIG)
-    find_package(LAPACK 3.9.1 REQUIRED CONFIG)
+    find_package(LAPACK REQUIRED CONFIG)
   else()
-    find_package(LAPACK 3.9.1 REQUIRED)
+    find_package(LAPACK REQUIRED)
+  endif()
+
+  if(NOT LAPACK_LIBRARIES)
+    set(LAPACK_LIBRARIES
+      ${LAPACK_blas_LIBRARIES}
+      ${LAPACK_lapack_LIBRARIES}
+    )
   endif()
 
   add_library(clients-common INTERFACE)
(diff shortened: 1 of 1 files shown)


clients/CMakeLists.txt
```diff
@@ -25,6 +25,13 @@ if(BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
     find_package(LAPACK REQUIRED)
   endif()
 
+  if(NOT LAPACK_LIBRARIES)
+    set(LAPACK_LIBRARIES
```
jzuniga-amd, line 29: Yes, good idea. Much better solution!

jzuniga-amd (approved):

clients/CMakeLists.txt
jzuniga-amd: Should we (in the future) enforce the use of a lapack version anyways?

cgmb (approved):
LGTM, though a better description might be "Fix compatibility with LAPACK 3.8 and earlier". BTW, you can control where `find_package` looks for LAPACK by setting the environment variables `LAPACK_DIR` or `CMAKE_PREFIX_PATH` as documented in step three of the [config mode search procedure](https://cmake.org/cmake/help/v3.22/command/find_package.html#config-mode-search-procedure).

clients/CMakeLists.txt
cgmb: Actually, FindLAPACK and FindBLAS don't seem to actually do a version check, so I suppose we might as well put it for both.

clients/CMakeLists.txt
```diff
@@ -20,9 +20,16 @@ endif()
 
 if(BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
   if(ROCSOLVER_FIND_PACKAGE_LAPACK_CONFIG)
-    find_package(LAPACK REQUIRED CONFIG)
+    find_package(LAPACK 3.7 REQUIRED CONFIG)
   else()
-    find_package(LAPACK REQUIRED)
+    find_package(LAPACK 3.7 REQUIRED)
```
cgmb, line 25: I've checked now. It makes no difference whether you specify `3.7` or `3.7.0`. Feel free to use whichever you prefer.

cgmb (approved):
LGTM. This will (correctly) ignore LAPACK 3.6 and works with LAPACK 3.7.0 through 3.10. Packages found through FindLAPACK (such as OpenBLAS) are unaffected by the version constraint.

PR #454, opened 2022-07-14
Include Path updated

arvindcheru:
Change proposed:
- To update the reorg path of include headers.


arvindcheru:
> I think we can just delete the Fortran sample. We should probably be directing Fortran users to [hipfort](https://github.com/ROCmSoftwarePlatform/hipfort) instead.

OK will close this PR

arvindcheru:
Closing this PR, as per comment this change will not be required since the Fortran sample is getting deleted.

PR #455, opened 2022-07-18
Add benchmark scripts for performance tracking

cgmb:
To generate benchmark data, first build `rocsolver-bench`, then
```bash
pip3 install pandas
export PATH="/path/to/rocsolver-bench/dir:$PATH"
./scripts/perf/benchmark-suite
```

The resulting output can be found in `scripts/perf/rocsolver-benchmarks`.

### DIFF
--- a/scripts/perf/benchmark-suite
+++ b/scripts/perf/benchmark-suite
@@ -14,18 +14,6 @@ mkdir -p "$temp_dir"
 
 # run benchmarks
 for p in s d c z; do
-  for f in geqrf geqrf_strided_batched; do
-    python3 rocsolver-bench-suite.py -v $p $f -o "$temp_dir/$p${f}_benchmarks.csv"
-  done
-done
-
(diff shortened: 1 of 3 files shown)


RobsonRLemos:
I have tested the benchmark-suite script and the generated datasets for the specified functions are working fine in the Performance Tracking System (PTS). The code is fine and it is written in such a way that more functions can be added in the near future. Thanks Cory for the first version of the implementation for the integration of rocSolver with the PTS system.

RobsonRLemos (approved):
The current implementation for the benchmark-suite is generating the needed datasets for the PTS system. With the current functionality the benchmark-suite can also be integrated in the near future with the CI system. Also the way the code is organized new functions can be added in the next versions of the implementation.

qjojo (approved):
Looks correct and maintainable 👍

tfalders (approved):
Nice job!

PR #459, opened 2022-08-11
Documentation update

jzuniga-amd:
Prepares documentation to be consistent with the new APIs that will be introduced soon

### DIFF
--- a/docs/source/api_auxiliaryfunc.rst
+++ b/docs/source/api_auxiliaryfunc.rst
@@ -17,14 +17,22 @@ The auxiliary functions are divided into the following categories:
 .. note::
     Throughout the APIs' descriptions, we use the following notations:
 
+    * i, j, and k are used as general purpose indices. In some legacy LAPACK APIs, k could be
+      a parameter indicating some problem/matrix dimension.
+    * Depending on the context, when it is necessary to index rows, columns and blocks or submatrices, 
+      i is assigned to rows, j to columns and k to blocks. l is always used to index 
+      matrices/problems in a batch. 
     * x[i] stands for the i-th element of vector x, while A[i,j] represents the element
       in the i-th row and j-th column of matrix A. Indices are 1-based, i.e. x[1] is the first
       element of x.
+    * To identify a block in a matrix or a matrix in the batch, k and l are used as sub-indices 
+    * x_i :math:`=x_i`; we sometimes use both notations, :math:`x_i` when displaying mathematical
+      equations, and x_i in the text describing the function parameters.
     * If X is a real vector or matrix, :math:`X^T` indicates its transpose; if X is complex, then
       :math:`X^H` represents its conjugate transpose. When X could be real or complex, we use X' to
       indicate X transposed or X conjugate transposed, accordingly.
-    * x_i :math:`=x_i`; we sometimes use both notations, :math:`x_i` when displaying mathematical
-      equations, and x_i in the text describing the function parameters.
+    * When a matrix `A` is formed as the product of several matrices, the following notation is used:
+      `A=M(1)M(2)...M(t)`.
 
 
 
--- a/library/include/rocsolver/rocsolver-functions.h
+++ b/library/include/rocsolver/rocsolver-functions.h
@@ -334,8 +334,8 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfg(rocblas_handle handle,
 
     \f[
         \begin{array}{cl}
-        H = H_1H_2\cdots H_k & \: \text{if direct indicates forward direction, or} \\
-        H = H_k\cdots H_2H_1 & \: \text{if direct indicates backward direction}
+        H = H(1)H(2)\cdots H(k) & \: \text{if direct indicates forward direction, or} \\
+        H = H(k)\cdots H(2)H(1) & \: \text{if direct indicates backward direction}
         \end{array}
     \f]
 
@@ -346,13 +346,13 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfg(rocblas_handle handle,
         H = I - VTV'
     \f]
 
-    where the i-th column of matrix V contains the Householder vector associated with \f$H_i\f$. If storev is row-wise, then
+    where the \f$j\f$th column of matrix V contains the Householder vector associated with \f$H(j)\f$. If storev is row-wise, then
 
     \f[
         H = I - V'TV
     \f]
 
-    where the i-th row of matrix V contains the Householder vector associated with \f$H_i\f$.
+    where the \f$i\f$th row of matrix V contains the Householder vector associated with \f$H(i)\f$.
 
     @param[in]
     handle      rocblas_handle.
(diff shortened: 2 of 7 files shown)


library/include/rocsolver/rocsolver-functions.h
```diff
@@ -4223,39 +4224,40 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetf2_npvt(rocblas_handle handle,
     could be executed with small and mid-size matrices if optimizations are enabled (default option). For more details, see the
     "Tuning rocSOLVER performance" section of the Library Design Guide).
 
-    The factorization of matrix \f$A_j\f$ in the batch has the form
+    The factorization of matrix \f$A_l\f$ in the batch has the form
 
     \f[
-        A_j = L_jU_j
+        A_l = L_lU_l
```
EdDAzevedo, line 4230: Should it be A(l) = L(l) U(l)?       Note the lower case  "el" (l) can look like "one" (1) in some fonts.

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -5853,54 +5856,54 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgeql2(rocblas_handle handle,
...
 
-    where the last m-i elements of the Householder vector \f$v_{j_i}\f$ are zero, and \f$v_{j_i}[i] = 1\f$.
+    where the last m-i elements of the Householder vector \f$v_{l_i}\f$ are zero, and \f$v_{l_i}[i] = 1\f$.
 
     @param[in]
     handle      rocblas_handle.
     @param[in]
     m           rocblas_int. m >= 0.\n
-                The number of rows of all the matrices A_j in the batch.
+                The number of rows of all the matrices A_l in the batch.
```
EdDAzevedo, line 5887: Should these be \f$A_l\f$  with escape to latex  instead of just A_l ?  Just checking.

EdDAzevedo (approved):

library/include/rocsolver/rocsolver-functions.h
cgmb: If you click on "Details" next to docs/readthedocs.org:rocsolver in the GitHub checks for this PR, you can see how the documentation will be rendered. Here's a direct link to this function: https://rocsolver--459.org.readthedocs.build/en/459/api_lapacklike.html#c.rocsolver_zgetf2_npvt_batched

It mostly looks ok, but the varying subscript height in `L_lL'_l` bugs me. That gets rendered as $L_lL'_l$. This is because the first variable has a superscript and the second does not. In those equations, I'd suggest giving all variables in the equation an empty superscript. `L^{}_lL'_l` renders as $L^{}_lL'_l$.

docs/source/api_auxiliaryfunc.rst
```diff
@@ -17,14 +17,22 @@ The auxiliary functions are divided into the following categories:
 .. note::
     Throughout the APIs' descriptions, we use the following notations:
 
+    * i, j, and k are used as general purpose indices. In some legacy LAPACK APIs, k could be
+      a parameter indicating some problem/matrix dimension.
+    * Depending on the context, when it is necessary to index rows, columns and blocks or submatrices, 
+      i is assigned to rows, j to columns and k to blocks. l is always used to index 
+      matrices/problems in a batch.
```
cgmb, line 24: Would it be ridiculous to suggest using $A_b$ instead of $A_l$? I feel like with small fonts, $A_l$ may be difficult to distinguish from $A_1$ or $A_i$.

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -4223,39 +4224,40 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgetf2_npvt(rocblas_handle handle,
     could be executed with small and mid-size matrices if optimizations are enabled (default option). For more details, see the
     "Tuning rocSOLVER performance" section of the Library Design Guide).
```
cgmb, line 4225: It's beyond the original scope of your change, but this should probably be a link.

docs/source/api_auxiliaryfunc.rst
tfalders: I'd also prefer a different subscript to `l` for the batch index. Using `b` makes sense to me.

library/include/rocsolver/rocsolver-functions.h
jzuniga-amd: The problem is that, for some reason, because of the newline (\n) that we use to separate the type of the parameter from the description of the parameter, doxygen introduces newlines before the latex formulas as well.
So, it will look like this:
>rocblas_int. m>=0.
>The number of rows of all matrices
>$A_l$ in the batch.

If we remove the \n it will look like this:
>rocblas_int. m>=0. The number of rows of all matrices $A_l$ in the batch

Ideally, what I would like is this:
>rocblas_int. m>=0.
>The number of rows of all matrices $A_l$ in the batch.

but I don't know how to make it work... Any ideas?
Do you mind having everything in a single line (i.e. remove \n)?
It would be good to have math formulas in the parameter description as well because the `_l` definitely  looks ugly.

docs/source/api_auxiliaryfunc.rst
jzuniga-amd: Normally the first letters of the alphabet are never used as indices in math, but I don't really see a problem; I can change `_l` for `_b` if we agree... `_l` looks bad when it is not in latex. 
Now, if we can find a way to have latex formulas everywhere (see my comment above), do you still want to use `_b`? I personally don't see a problem with $A_l$ $A_1$ or $A_i$.

library/include/rocsolver/rocsolver-functions.h
EdDAzevedo: Have you tried "\linebreak" command in latex?

You can also experiment with "A_\ell" for "A_l" to see which one do you prefer.

library/include/rocsolver/rocsolver-functions.h
jzuniga-amd: Will change this once the description of the related parameters is completed so we have the correct anchor for the link

library/include/rocsolver/rocsolver-functions.h
jzuniga-amd: Problem is not in latex. The problem is doxygen introducing line breaks where it shouldn't (or where I don't want them)

docs/source/api_auxiliaryfunc.rst
jzuniga-amd: >You can also experiment with "A_\ell" for "A_l" to see which one do you prefer.

Yeah, I like better `A_\ell`. It is more distinguishable from `A_i` and `A_1` , I think
$A_\ell$, $A_1$, $A_i$

I would go with `A_\ell` instead of `A_b` if we need to change, but no strong preference.

tfalders: Yeah, now that I think about it, using `A_b` might cause confusion in cases where `b` is the name of the right-hand solution vector. Hmmm....

cgmb: I like $A_\ell$. I see that in Fortran, variables starting with the letters i, j, k, l, m, or n are of type integer by default, which I suppose is strong cultural precedent.

jzuniga-amd:
I added empty superscripts when needed. It looks good on my local build, but apparently it has no effect on the build in readthedocs :(

I also removed the line breaks in the parameter descriptions; it looks OK. If you don't mind we can leave it; this will allow us to introduce latex (and `_\ell`) in the parameter descriptions without the undesired line break. Have a look at the `ipiv` description in `laswp` function, for example.

Let me know what you think.

cgmb (approved):
I didn't review the whole thing after the recent updates, but I don't have any further suggestions.

qjojo (approved):

PR #460, opened 2022-08-15
Improved syevj performance

tfalders:
This PR improves the performance of syevj. My tests show ~~a 4x to 5x speed-up~~ an over 10x speed-up for larger sizes, and is now ~~comparable to~~ better in many cases than the current performance of syevd.

The small-size kernel is now a 2D kernel and uses shared memory.

~~The large-size kernel has been divided into 3 smaller kernels, with the largest implementing the iterative loop. This kernel is a 2D kernel spread across multiple thread-blocks, which are synced using the new `get_task_id` method. `get_task_id` is designed to function even if not all thread-blocks are currently active on the GPU.~~
The single large-size kernel has been replaced by a blocked algorithm relying on many small kernels, with the CPU now implementing the loop for iterations/sweeps. In order to terminate the sweep loop early, I am reading a value from the GPU that indicates if all the batch instances have reached the termination condition. AFAIK, this is the first time in rocSOLVER that we're reading a value from the GPU during the execution of a function.

Performance is not yet up to the level of cuSOLVER's syevj. Some suggestions for future improvements:
- Papers indicate that the off-diagonal residual can be updated using the values of the `a_ij`, rather than recalculating it each iteration. I've not been able to get this to work yet, however.
- Papers indicate that using an approximate rotation can improve performance by cutting down on the number of `sqrt` calls.
- ~~I've not yet tried the blocked algorithm strategy.~~


library/src/include/lib_device_helpers.hpp
```diff
@@ -86,6 +86,85 @@ __device__ void scale_tridiag(const rocblas_int start, const rocblas_int end, T*
...
+                            rocblas_int* counters,
+                            const rocblas_int task_count)
+{
+    // obtain lock
+    while(atomicCAS(counters, 0, 1) != 0)
+    {
+        // do nothing
+    }
+
+    int current_phase = atomicAdd(counters + 1, 0);
```
EdDAzevedo, line 116: Perhaps consider using  something like  the following to make the code easier to read?

rocblas_int * const mutex = counters; 
rocblas_int * const phase_counter = counters + 1;
rocblas_int * const task_counter = counters + 2;
rocblas_int * const active_block_counter = counters + 3;

library/src/include/lib_device_helpers.hpp
```diff
@@ -86,6 +86,90 @@ __device__ void scale_tridiag(const rocblas_int start, const rocblas_int end, T*
...
+    rocblas_int* const mutex = counters;
+    rocblas_int* const phase_counter = counters + 1;
+    rocblas_int* const task_counter = counters + 2;
+    rocblas_int* const active_block_counter = counters + 3;
+
+    // obtain lock
+    while(atomicCAS(mutex, 0, 1) != 0)
+    {
+        // do nothing
+    }
```
cgmb, line 119: A spin-lock like this is very dependent on the behaviour of the scheduler. On the CPU, this construct would be trouble. On the GPU... I'm not sure.

How do you guarantee that the thread holding the mutex makes forward progress? In particular, how does this interact with Compute Wave Save/Restore (CWSR)? If the scheduler ever suspends the wave holding the mutex, the waves waiting on the mutex cannot make progress. So, you need some sort of guarantee that the scheduler will always wake the wave that holds the mutex in a reasonable amount of time, even if no other computation has completed in any other thread/block.

It may be perfectly fine to do this on our GPUs. It's just a dangerous construct to use in the implementation of parallel CPU algorithms, and I don't know enough about GPU thread scheduling to say whether the same problems exist here.

cgmb: For a discussion of spinlocks in the context of CPU thread scheduling, see the articles linked from [_Beware Spinlocks in User Space_](https://mjtsai.com/blog/2020/01/06/beware-spinlocks-in-user-space/).

qjojo (approved):

tfalders:
The blocked algorithm is now implemented and ready for review. My tests show an improvement even over the multi-block synchronization approach I was previously using.

Due to concerns over correctness, I've commented out `get_task_id` but not deleted it in case we need a multi-block synchronization in the future.

qjojo (approved):
LGTM

library/src/include/lib_device_helpers.hpp
```diff
@@ -86,6 +86,90 @@ __device__ void scale_tridiag(const rocblas_int start, const rocblas_int end, T*
     }
 }
 
+// /** GET_TASK_ID uses task stealing to synchronize multiple thread blocks.
```
cgmb, line 89: I'd prefer not to include this, because I don't like commented-out code in the repo. In my experience, it's rare for it to get used and it ends up just being a distraction. For the sake of keeping a codebase simple and minimal, I prefer to only include things that are actually used.

With that said, I do appreciate that something like this is difficult to produce and we don't want to lose the work that was put into it. My suggestion would be to put the code in [a gist](https://gist.github.com/) and put a link in this PR. We'll be able to find it again if we need it. If you still would prefer adding it as-is, though, please feel free. A bit here or there is no big deal.

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -19464,7 +19464,7 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_dsygvj_strided_batched(rocblas_handle
     @param[in]
     abstol      real type.\n
                 The absolute tolerance. The algorithm is considered to have converged once the residual
-                is <= abstol. If abstol <= 0, then the tolerance will be set to machine precision.
+                is <= n * abstol. If abstol <= 0, then the tolerance will be set to machine precision.
```
cgmb, line 19467: Interesting. Is that a bugfix or a docs fix?

cgmb (approved):
It's nice to see you found such big improvements.

> The single large-size kernel has been replaced by a blocked algorithm relying on many small kernels, with the CPU now implementing the loop for iterations/sweeps. In order to terminate the sweep loop early, I am reading a value from the GPU that indicates if all the batch instances have reached the termination condition. AFAIK, this is the first time in rocSOLVER that we're reading a value from the GPU during the execution of a function.

Interesting. I'm still processing the implications of that.

> Due to concerns over correctness, I've commented out get_task_id

Just to be clear, I'm not sure there's anything wrong with it. Also, I wonder if there's anything in rocPRIM that can help us with this sort of thing?

library/include/rocsolver/rocsolver-functions.h
tfalders: This is more a case of me (mis)remembering something from the code sample @jzuniga-amd supplied to me and adding it in (to both the code and documentation). It's actually supposed to be `norm(A)` instead of `n`, but that's a bit more difficult to get and test against so I haven't fixed it yet. I'll try and have it updated before this PR gets merged.

library/src/include/lib_device_helpers.hpp
tfalders: That sounds like a good idea. I felt pretty uncomfortable with leaving such a large body of commented-out code, so that works for me.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -745,29 +1435,134 @@ rocblas_status rocsolver_syevj_heevj_template(rocblas_handle handle,
...
+        ROCSOLVER_LAUNCH_KERNEL(syevj_init<T>, grid, threads, lmemsizeInit, stream, evect, uplo,
+                                half_blocks, n, A, shiftA, lda, strideA, atol, residual, Acpy, top,
+                                bottom, completed);
+
+        while(h_sweeps < max_sweeps)
+        {
+            // if all instances in the batch have finished, exit the loop
+            hipMemcpyAsync(&h_completed, completed, sizeof(rocblas_int), hipMemcpyDeviceToHost,
+                           stream);
+            hipStreamSynchronize(stream);
```
cgmb, line 1503: We should mention in the docs that these functions may synchronize the stream.

I was talking with Yvan about using Cooperative Groups for this synchronization rather than synchronizing on the host. Unfortunately, while CG were added in ROCm 4.1, you're very limited in the number of cooperative groups that you get. If multiple different libraries use CG at the same time, it won't result in a resource error but it will result in a large performance degradation. He's been working on an iterative solver as well and it seems he concluded that he couldn't really use CG. His solution was also just to do a synchronize with the CPU each iteration.

I wonder if HIP Graph might reduce launch overhead enough that we could do `max_sweeps` every time (with all kernels short-circuited to return immediately when it's already converged). However, that is not a particularly great solution anyway.

My conclusion is that this hipStreamSynchronize is reasonable. There does not seem to be a better block-level synchronization primative.

library/src/include/lib_device_helpers.hpp
tfalders: Here is the gist: https://gist.github.com/tfalders/c2d3d8cd079d7a7b4e43d906c72f4d4d

library/src/lapack/roclapack_syevj_heevj.hpp
tfalders: > We should mention in the docs that these functions may synchronize the stream.

Good idea! Done.

jzuniga-amd (commented):
I like this approach much better and if it is accompanied by a substantial improvement, it should be good to go. Good job!!!
(I still want to try a more classic blocked approach based on level-3 blas, but we can try/investigate that once we incorporate thread-level gemms, for example)

Just a few minor comments below...

library/src/include/ideal_sizes.hpp
```diff
@@ -166,6 +166,16 @@
...
 
+/************************** syevj/heevj ***************************************
+*******************************************************************************/
+/*! \brief Determines the size at which rocSOLVER switches from
+    the small-size kernel to the blocked algorithm when executing SYEVJ. It also applies to the
+    corresponding batched and strided-batched routines. Must be <= 64.
+
+    \details If the size of the matrix is not greater than SYEVJ_SWITCHSIZE, the eigenvalues
+    and eigenvectors will be computed with a single kernel call. */
+#define SYEVJ_SWITCHSIZE 58
```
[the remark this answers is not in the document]
tfalders: Done! I went with `SYEVJ_BLOCKED_SWITCH` and added the documentation to `design_tuning.rst`. Let me know if you have any comments for the documentation.

jzuniga-amd (approved):
LGTM

tfalders:
We're having some issues with the managed malloc tests on gfx90a CI, so I'm going to go ahead and force the merge.

PR #461, opened 2022-08-23
Always rebuild dependencies with ./install.sh -cd

cgmb:
* Always rebuild dependencies with ./install.sh -cd

  The attempts to skip already completed steps caused problems when a
  step was only half-completed. For example, if only half a file was
  downloaded before the script aborted, then on the next run of
  ./install.sh -cd, the download would be skipped. This would require
  user intervention to correct.

  Starting from scratch on each call to ./install.sh -cd allows the
  script to recover from failure on its own by always starting from
  a known state.

* Use -H instead of -S

  The `-S <src-dir> -B <bld-dir>` options were added in CMake 3.13, but the
  `-H<src-dir> -B<bld-dir>` options are equivalent and have existed for
  much longer.

  rocSOLVER itself requires at least CMake 3.13 to build, but we can
  avoid having to do a redundant version check in bash if we stick
  to using older command line flags.

* Print CMake executable information from install.sh

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -159,7 +159,7 @@ install_fmt_from_source( )
   rm -rf "$fmt_srcdir" "$fmt_blddir"
   tar xzf fmt-$fmt_version.tar.gz --one-top-level="$fmt_srcdir" --strip-components 1
   ${cmake_executable} \
-    -S "$fmt_srcdir" -B "$fmt_blddir" \
+    -H$fmt_srcdir -B$fmt_blddir \
     -DCMAKE_BUILD_TYPE=Release \
     -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
     -DCMAKE_CXX_STANDARD=17 \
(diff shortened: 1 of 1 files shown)


cgmb:
@YvanMokwinski, I still can't reproduce the build issues you're having, but I put together a few changes to improve the robustness of the script and to provide additional debug info.

tfalders (approved):
LGTM! We should have @YvanMokwinski test these changes before merging to see if they fix the issue.

jzuniga-amd (approved):

cgmb:
> LGTM! We should have @YvanMokwinski test these changes before merging to see if they fix the issue.

It seems the script was using a different copy of CMake than Yvan was expecting. He was getting CMake 3.10 (the system version on Ubuntu 18.04). The extra print statements I added should help make that kind of problem more obvious, though the build will still fail because that version does not meet the minimum requirements for googletest (when built with Clang 14) or rocsolver.

PR #464, opened 2022-08-29
Reduce duplication within debug code

cgmb:
This change reduces duplication in the common host helpers by providing a fmt printer for `rocblas_float_complex` and `rocblas_double_complex`. If you have low-level functions specialized appropriately for both real and complex (like `format` or `abs` are), the higher level functions (like `print_to_stream` and `print_host_matrix`) don't need to be specialized themselves.

There were a couple fmtlib-related problems that I fixed at the same time.
1. rocSOLVER is potentially affected by https://github.com/fmtlib/fmt/issues/2357. I've never actually seen it in rocSOLVER, so we may have gotten lucky with the way our headers are arranged, but it's possible to encounter this bug in fmtlib if <fmt/ostream.h> is only included in a subset of headers used in the library. This bug was fixed in fmt v9.0.0 but I'm having some trouble updating to that version. In previous versions, there is a simple workaround: either always include <fmt/ostream.h> whenever you include another fmtlib header, or never include <fmt/ostream.h>. The first of those two options was the easier one.
2. It was a mistake for me to set a different default for DROCSOLVER_EMBED_FMT when building with CMake vs. install.sh.
3. There are two different files named `rocblascommon/utility.hpp` and `common/include/common_host_helpers.hpp` was therefore including a different one when it was used in the library than when it was used in the client. This resulted in some very confusing behaviour. I've moved/renamed both headers and trimmed some unused stuff.

This whole thing is basically just me cleaning up some of the current debug code before submitting another PR with some additions.


tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #463, opened 2022-08-26
Updates for rocBLAS syr2k changes.

daineAMD:
Corresponding changes for https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1408

PR #465, opened 2022-08-30
Add option to build rocblas to install.sh

cgmb:
This change adds an option to build and install a known good version of
rocBLAS when building rocSOLVER. The rocBLAS version hash should be
updated whenever changes are made to rocSOLVER that require a newer
version of rocBLAS. I expect it to require an update in any change that
fixes the rocSOLVER build after rocBLAS makes a change to its private
API.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -6,31 +6,44 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false)
     project.paths.construct_build_prefix()
 
     String compiler = 'hipcc'
-    String hipClang = ''
     String debug = project.buildName.contains('Debug') ? '-g' : ''
     String centos = platform.jenkinsLabel.contains('centos') ? 'source scl_source enable devtoolset-7' : ''
     String noOptimizations = ''
+    boolean usePrebuiltDependencies = true
 
     if (env.BRANCH_NAME ==~ /PR-\d+/)
     {
+        usePrebuiltDependencies = false
+
         pullRequest.labels.each
         {
             if (it == "noOptimizations")
             {
                 noOptimizations = "-n"
             }
+            else if (it == "ci:prebuilt-deps")
+            {
+                usePrebuiltDependencies = true
+            }
         }
     }
 
-    def getRocBLAS = auxiliary.getLibrary('rocBLAS-internal',platform.jenkinsLabel, null, sameOrg)
+    String getRocBLAS = ''
+    String deps = ''
+    if (usePrebuiltDependencies)
+    {
+        getRocBLAS = auxiliary.getLibrary('rocBLAS-internal',platform.jenkinsLabel, null, sameOrg)
+    }
+    else
+    {
+        deps = '-b'
+    }
     def command = """#!/usr/bin/env bash
-                set -x
+                set -ex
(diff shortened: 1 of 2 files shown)


jzuniga-amd (commented):
I really like the idea. 
We should think on better ways to keep the `rocblas_version` updated, though (other than by hand, ideally). Otherwise we could be giving users an option that fails. As of today, for example, I think we need at least rocblas @ 2e659b6ce7b87ee30767ebbe34c6161b2922901c

tfalders:
> We should think on better ways to keep the `rocblas_version` updated, though (other than by hand, ideally).

This is my concern as well. IIRC, rocBLAS has a bot to keep its Tensile tag up-to-date - perhaps something similar could work for us.

qjojo:
> I really like the idea. We should think on better ways to keep the `rocblas_version` updated, though (other than by hand, ideally).

Could we just clone develop and provide an option to grab a specific hash as an override in case HEAD is broken?

cgmb:
> We should think on better ways to keep the `rocblas_version` updated, though (other than by hand, ideally). Otherwise we could be giving users an option that fails.

I don't think updating by hand is that bad. We only need to update `rocblas_version` when we make a we make a change to rocSOLVER such that we require a newer version of rocBLAS. IMO, that's useful information to have anyway when we're evaluating a change. If there were a test that exercised this option, it could run on PRs and help keep us informed.

With that said, you're right that the version is a maintenance burden. Among other things, it's a potential source of merge conflicts. This sort of often-changing dependency version information shouldn't live within the repo. As such, this not a long-term solution to the problem of building rocSOLVER together with rocBLAS... just a stopgap. I've been trying to coordinate better solutions for a long time, but it's been a very slow process.

One thing that would help is reducing the rate of churn on the rocBLAS internal ABI. A lot of the recent breakages have to do with changes to things that probably did not actually have to be part of the ABI.

> This is my concern as well. IIRC, rocBLAS has a bot to keep its Tensile tag up-to-date - perhaps something similar could work for us.

The requirement for a newer version of rocBLAS will always be introduced in a rocSOLVER PR. I'd want to update it in the same PR that introduced the new requirement, otherwise there will be  some rocSOLVER commits in which the option is broken.

> Could we just clone develop and provide an option to grab a specific hash as an override in case HEAD is broken?

No. That would make this option basically unusable for building anything besides the latest commits on the rocSOLVER develop branch.

cgmb:
> I think we need at least rocblas @ 2e659b6ce7b87ee30767ebbe34c6161b2922901c

Updated. This PR was prompted by #463 and I was anticipating it would need to change. In the future, those sorts of PRs would update the `rocblas_version`...

...and in the more distant future, perhaps we can look at the adjusting the rocBLAS private API to reduce or eliminate the need for those sorts of PRs.

cgmb:
I've changed the CI to use `./install.sh -b` for rocSOLVER by default on PRs. In addition to verifying our dependency installation script, it will have a side-effect of eliminating the problem that our PRs are blocked when rocBLAS is broken.

I left the old behaviour as default when building branches and the old behaviour can be used on PRs by applying the label `ci:prebuilt-deps`.

qjojo (approved):

cgmb:
I think we're going to end up addressing this in other ways.

PR #466, opened 2022-08-30
invalid enum test use 0, rocblas invalid_data 255

TorreZuk:
* fixes switch enum warnings
* matches fixes from rocblas for SWDEV-352481 for llvm16 use of -1 for invalid enum value


cgmb (approved):
The initialization with `-1` was not undefined behaviour as I initially had thought, but it doesn't surprise me that it's causing trouble (via `-Wenum-constexpr-conversion`). That's a weird construct and I'll be happy to see it gone.

jzuniga-amd (approved):

PR #468, opened 2022-08-31
Added bdsvdx

tfalders:
This PR adds bdsvdx to rocSOLVER.

Since this is an SVD routine and not an eigensolver, for the sake of consistency I created a new `rocblas_srange` enum rather than reuse the existing `rocblas_erange`, even though they're functionally identical. I've also added V as a character mapping to `rocblas_svect_singular`, which is otherwise mapped to S. This is to match the LAPACK arguments to bdsvdx without creating a new enum similar to `rocblas_svect`.


library/src/auxiliary/rocauxiliary_bdsvdx.hpp
```diff
@@ -110,14 +238,95 @@ rocblas_status rocsolver_bdsvdx_template(rocblas_handle handle,
...
 
-    return rocblas_status_not_implemented;
+    // quick return
+    if(batch_count == 0)
+        return rocblas_status_success;
+
+    hipStream_t stream;
+    rocblas_get_stream(handle, &stream);
+
+    // info = 0
```
qjojo, line 268: Can we remove this or add a comment explaining why it's here?

tfalders: Are you referring to the stream or to the info comment?

tfalders: The info comment is saying that we're setting info to zero, which occurs when we call the `reset_info` kernel two lines down. There are similar comments in other files, but I suppose I can clarify it a little bit.

qjojo: If its a convention already thats probably fine then.

clients/common/lapack_host_reference.cpp
```diff
@@ -2255,6 +2353,41 @@ void zsytrf_(char* uplo,
              int* lwork,
              int* info);
 
+void sbdsvdx_(char* uplo,
+              char* svect,
```
EdDAzevedo, line 2357: Just a minor suggestion to name argument to be similar to lapack routine, so consider "jobz" instead of "svect".

clients/common/lapack_host_reference.cpp
```diff
@@ -2255,6 +2353,41 @@ void zsytrf_(char* uplo,
...
+              int* il,
+              int* iu,
+              int* nsv,
+              float* S,
+              float* Z,
+              int* ldz,
+              float* work,
+              int* iwork,
+              int* info);
+void dbdsvdx_(char* uplo,
```
EdDAzevedo, line 2373: Note there is LAPACKE to provide C interface to lapack, such as LAPAPACKE_dbdsvdx(). If you wish to use cblas, perhaps you might also wish to use lapacke as well?

clients/gtest/bdsvdx_gtest.cpp
```diff
@@ -0,0 +1,137 @@
...
+// each size_range vector is a {n, ldz, vect}
+// if vect = 0, then don't find singular vectors
+// if vect = 1, then find singular vectors
+
+// each ops_range vector is a {rng, vl, vu, il, iu}
+// if rng = 0, then find all singular values
+// if rng = 1, then find singular values in (vl, vu]
+// if rng = 2, then find the il-th to the iu-th singular value
+
+// Note: all tests are prepared with diagonally dominant matrices that have random diagonal
```
EdDAzevedo, line 24: Would the test code under lapack (TESTING/dchkbd.f) that calls dbdsvdx() be helpful?

clients/common/lapack_host_reference.cpp
cgmb: rocSOLVER no longer uses CBLAS. There was actually only one function that used CBLAS and I removed it recently, because dropping CBLAS made the build system simpler (especially when using an alternative LAPACK implementation).

There are many functions with a `cblas` prefix in rocSOLVER, but the naming is misleading. They do not actually use CBLAS; they invoke BLAS directly. I mean to fix the naming. Since they're C++ wrappers of BLAS functions and are overloaded on the parameter type (rather than using `s`, `d`, `c` and `z` prefixes), I was thinking maybe renaming them with `x` as the prefix. For example, renaming `cblas_trsm` to `xtrsm` (since that one function has `float`, `double`, `rocblas_float_complex` and `rocblas_double_complex` overloads).

cgmb: I would like to stick with invoking BLAS and LAPACK functions directly for the moment, but I do agree that CBLAS and LAPACKE could be useful. The build system and package manager problems that were sidestepped by avoiding those libraries will probably eventually be fixed. The issues were long-standing, but I've seen some progress made recently. I could very well have a different opinion in six months.

clients/gtest/bdsvdx_gtest.cpp
cgmb: It would be interesting to have a wrapper for rocSOLVER that is ABI-compatible with LAPACK. The performance might not be ideal, but that would allow us to run the reference test suite without modifications.

jzuniga-amd (commented):
Some initial comments below. I am still reviewing the algorithm implementation.

clients/gtest/bdsvdx_gtest.cpp
```diff
@@ -0,0 +1,137 @@
...
+    BDSVDX() {}
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+
+    template <typename T>
+    void run_tests()
+    {
+        Arguments arg = bdsvdx_setup_arguments(GetParam());
+
+        if(arg.peek<rocblas_int>("n") == 0 && arg.peek<char>("svect") == 'E'
```
EdDAzevedo, line 109: Are the arguments case insensitive so that passing, say, upper case "V" and lower "v" are both valid? lapack uses "lsame()" routine but the code may also simply check for both "V" and "v". Just a suggestion.

clients/include/testing_bdsvdx.hpp
```diff
@@ -0,0 +1,540 @@
...
+                                         vu, il, iu, abstol, dNsv.data(), dS.data(), dZ.data(), ldz,
+                                         dIfail.data(), dInfo.data()));
+    CHECK_HIP_ERROR(hNsvRes.transfer_from(dNsv));
+    CHECK_HIP_ERROR(hSRes.transfer_from(dS));
+    CHECK_HIP_ERROR(hZRes.transfer_from(dZ));
+    CHECK_HIP_ERROR(hIfailRes.transfer_from(dIfail));
+    CHECK_HIP_ERROR(hInfoRes.transfer_from(dInfo));
+
+    // CPU lapack
+    // WORKAROUND: For some test cases, LAPACK's bdsvdx is returning incorrect singular values
```
EdDAzevedo, line 194: when lapack bdsvdx is returning incorrect singular values, is the "INFO" value == 0 (successful exit) or "INFO" is non-zero to indicate something is not right. Perhaps one may use lapack dbsvdx first, and check INFO, then use the work around in computing all singular values as the  fall back option? If there is a bug in lapack bdsvdx, perhaps the lapack developers would be interested in a matrix that can reproduce the problem?

EdDAzevedo (approved):

clients/include/testing_bdsvdx.hpp
```diff
@@ -0,0 +1,540 @@
...
+    // when srange is rocblas_srange_index. In this case, we use rocblas_srange_all to get
+    // all the singular values and offset and use il as an offset into the result array.
+    rocblas_int ioffset = 0;
+    if(srange == rocblas_srange_index)
+    {
+        cblas_bdsvdx<T>(uplo, rocblas_svect_none, rocblas_srange_all, n, hD[0], hE[0], vl, vu, il,
+                        iu, hNsv[0], hS[0], hZ[0], ldz, work.data(), iwork.data(), hInfo[0]);
+        ioffset = il - 1;
```
jzuniga-amd, line 202: If the singular values are in descending order, this offset is not correct!
`il` stands for the index of the smallest singular value to be computed. 
The offset should be something like:
`ioffset =  n - iu`

EdDAzevedo: I think the lapack dbdsvdx() routine returns singular values in array "S" in  ascending order. 
https://netlib.org/lapack/explore-html/dc/dd2/group__double_o_t_h_e_reigen_ga63829fd8c0a1ae89779a35c9c1093fa6.html

clients/gtest/bdsvdx_gtest.cpp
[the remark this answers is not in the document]
tfalders: I copied the test from STEBZ and put `svect` in place of `eorder`. Apparently I missed changing the E; thanks for catching it!

clients/gtest/bdsvdx_gtest.cpp
```diff
@@ -0,0 +1,137 @@
...
+    arg.set<char>("svect", (size[2] == 0 ? 'N' : 'V'));
+
+    arg.set<char>("srange", (op[0] == 0 ? 'A' : (op[0] == 1 ? 'V' : 'I')));
+    arg.set<double>("vl", op[1]);
+    arg.set<double>("vu", op[2]);
+    arg.set<rocblas_int>("il", op[3]);
+    arg.set<rocblas_int>("iu", op[4]);
+
+    // always use the best accuracy
+    arg.set<double>("abstol", 0);
```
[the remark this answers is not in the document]
tfalders: Sounds good. I'll remove it.

clients/include/testing_bdsvdx.hpp
jzuniga-amd: The workaround is just for the test; our API definition is as close to the lapack API as any other rocsolver function... nothing to be reconciled.
But yes, it would be good to open an issue to report the bug. I also found issues in lapack gesvdx that will require a similar workaround in the upcoming rocsolver_gesvdx tests.

clients/include/testing_bdsvdx.hpp
tfalders: Yes, the singular values are returned in ascending order. The tests are passing with `ioffset = il - 1`, which suggests the offset is correct.

clients/include/testing_bdsvdx.hpp
```diff
@@ -0,0 +1,540 @@
...
+    CHECK_HIP_ERROR(hNsvRes.transfer_from(dNsv));
+    CHECK_HIP_ERROR(hSRes.transfer_from(dS));
+    CHECK_HIP_ERROR(hZRes.transfer_from(dZ));
+    CHECK_HIP_ERROR(hIfailRes.transfer_from(dIfail));
+    CHECK_HIP_ERROR(hInfoRes.transfer_from(dInfo));
+
+    // CPU lapack
+    // WORKAROUND: For some test cases, LAPACK's bdsvdx is returning incorrect singular values
+    // when srange is rocblas_srange_index. In this case, we use rocblas_srange_all to get
+    // all the singular values and offset and use il as an offset into the result array.
```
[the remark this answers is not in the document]
tfalders: I believe using il as the offset is the correct approach. See comment below.

clients/common/lapack_host_reference.cpp
jzuniga-amd: This is just for internal use in the test suite (no user visibility). We use the same names as in the rocsolver API to easily identify the arguments, but I am OK with any name.

clients/include/testing_bdsvdx.hpp
jzuniga-amd: mmm, this part in lapack's documentation is misleading then:
"Given a TGK matrix, one can either a) compute -s,-v and change signs
  so that the singular values (and corresponding vectors) are already in
  descending order (as in DGESVD/DGESDD) or b) compute s,v and reorder
  the values (and corresponding vectors). DBDSVDX implements a) by
  calling DSTEVX..."

I haven't looked carefully at the algorithm implementation yet, but if you are asking stebz to compute the negative eigenvalues of TGK, then when you negate the values, your singular values should be in descending order.

tfalders: Oh wait, you're right. I just checkout our output, and they are in descending order. The LAPACK documentation for the parameters is wrong then.

In this case, I guess it must be that `il` stands for the index of the _largest_ singular value, and `iu` is the index of the smallest.

clients/gtest/bdsvdx_gtest.cpp
tfalders: I believe that our argument handling is case-sensitive, and the convention is to always use capital letters. This only applies to the tests, however, and will only affect our users if they use the benchmark client. AFAIK, none of our library functions take characters as arguments.

clients/include/testing_bdsvdx.hpp
tfalders: @EdDAzevedo The incorrect singular values are returned with INFO == 0, yes. Our tests always verify that LAPACK and rocSOLVER agree on the value of INFO.

I'll look into opening a bug report for LAPACK. I think the smallest size that had this issue was a 12x12 bidiagonal matrix, which should be easy to provide.

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -4134,6 +4134,129 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zstein(rocblas_handle handle,
...
+    nsv         pointer to a rocblas_int on the GPU. \n
+                The total number of singular values found. If srange is rocblas_srange_all, nsv = n.
+                If srange is rocblas_srange_index, nsv = iu - il + 1. Otherwise, 0 <= nsv <= n.
+    @param[out]
+    S           pointer to real type. Array on the GPU of dimension n.\n
+                The first nsv elements contain the computed singular values in descending order.
+    @param[out]
+    Z           pointer to real type. Array on the GPU of dimension ldz*nsv.\n
+                If info = 0, the first nsv columns contain the computed singular vectors corresponding to the
+                singular values in S. The first n rows of Z contain the matrix U, and the next n rows contain
```
jzuniga-amd, line 4204: Just don't forget to mention that when `nsv` is not known (`srange_value case`), `Z` must have enough columns... recommended value is `n`

jzuniga-amd (approved):
LGTM. Just a couple of suggestions below...

clients/include/testing_bdsvdx.hpp
jzuniga-amd: Ok good. That makes sense now (in math we usually refer to `\sigma_1` as the largest singular vector and `\sigma_n` as the smallest)...
And with this notation `il <= iu` as required in the argument list. 

If we report an issue with the bug when range = index, we should also point out the inconsistency in the documentation:
1. `S` contains the singular values in DESCENDING order,
2. `il` refers to the LARGEST computed value, and
3. `iu` refers to the SMALLEST computed value.

library/include/rocsolver/rocsolver-functions.h
EdDAzevedo: I tried a simple program to call lapack dbdsvdx()  and see that entries in S are in descending order so I was wrong to assume entries in S are in ascending order. Please ignore my previous comments related to ascending order in array "S".

jzuniga-amd: Yes, see the comment thread above. There is a number of inconsistencies in lapack bdsvdx's documentation

library/src/auxiliary/rocauxiliary_bdsvdx.hpp
```diff
@@ -0,0 +1,331 @@
...
+    // select batch instance
+    rocblas_int bid = hipBlockIdx_y;
+    rocblas_int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+
+    // local variables
+    rocblas_int nsv = nsvA[bid];
+    T* S = SS + (bid * strideS);
+    T* Stmp = StmpA + (bid * 2 * n);
+
+    if(nsv > n)
```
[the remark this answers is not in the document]
tfalders: This is more of a safety measure just in case.

Also, there is currently no check in place to make sure `vl, vu >= 0` and I was thinking of adding it. Do you agree?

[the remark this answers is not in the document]
jzuniga-amd: Right, if you accept negative values for `vl` and `vu`, then `nsv` could be larger than `n`.
Of course, there are no negative singular values so I think we should add the check `vl,vu >= 0` to protect us from uninformed users. With this, `if(nsv > n)` is redundant... (can leave it if you prefer, no big deal)

qjojo (approved):

PR #470, opened 2022-09-09
Add partial SVD routine gesvdx

jzuniga-amd:
Adds GESVDX, based on BDSVDX, with batched and strided_batched versions.

This PR also fixes some errors in BDSVDX that we did not catch during the review process. 

As with lapack bdsvdx_, a number of problems with gesvdx_ were found. This makes the unit test of GESVDX a bit convoluted but functional. 
(We may change to a full implicit test for the SVDs in the future (in another PR); right now we only use reference cpu lapack to compare/validate the singular values, the singular vectors are already tested implicitly).


library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+                The upper bound of the search interval [vl, vu). Ignored if srange indicates to look
+                for all the singular values of B or the singular values within a set of indices.
+    @param[in]
+    il          rocblas_int. il = 1 if n = 0; 1 <= il <= iu otherwise.\n
+                The index of the largest singular value to be computed. Ignored if srange indicates to look
+                for all the singular values of B or the singular values in a half-open interval.
+    @param[in]
+    iu          rocblas_int. iu = 0 if n = 0; 1 <= il <= iu otherwise.\n
+                The index of the smallest singular value to be computed. Ignored if srange indicates to look
+                for all the singular values of B or the singular values in a half-open interval.
```
tfalders, line 11824: B -> A

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
```
tfalders, line 12016: B -> A_j
(Same for strided_batched)

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+    strideS     rocblas_stride.\n
+                Stride from the start of one vector S_j to the next one S_(j+1).
+                There is no restriction for the value of strideS. Normal use case is strideS >= min(m,n).
+    @param[out]
+    U           pointer to type. Array on the GPU (the size depends on the value of strideU).\n
+                The matrix U_j of left singular vectors stored as columns. Not
+                referenced if left_svect is set to none.
+    @param[in]
+    ldu         rocblas_int. ldu >= m if left_svect singular; ldu >= 1 otherwise.\n
+                The leading dimension of U.
```
tfalders, line 12035: U_j
(Same for strided_batched)

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+                Stride from the start of one matrix U_j to the next one U_(j+1).
+                There is no restriction for the value of strideU. Normal use case is strideU >= ldu*min(m,n).
+    @param[out]
+    V           pointer to type. Array on the GPU (the size depends on the value of strideV). \n
+                The matrix V_j of right singular vectors stored as rows (transposed / conjugate-transposed).
+                Not referenced if right_svect is set to none.
+    @param[in]
+    ldv         rocblas_int. ldv >= min(m,n) if right_svect is
+                set to singular; or ldv >= 1 otherwise.\n
+                The leading dimension of V.
```
tfalders, line 12047: V_j
(Same for strided_batched)

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+
+    where the m-by-n matrix \f$S_j\f$  is zero except, possibly, for its min(m,n)
+    diagonal elements, which are the singular values of \f$A_j\f$ . \f$U_j\f$  and \f$V_j\f$  are orthogonal
+    (unitary) matrices. The first min(m,n) columns of \f$U_j\f$  and \f$V_j\f$  are the left and
+    right singular vectors of \f$A_j\f$ , respectively.
+
+    The computation of the singular vectors is optional and it is controlled by
+    the function arguments left_svect and right_svect as described below. When
+    computed, this function returns the transpose (or transpose conjugate) of the
+    right singular vectors, i.e. the rows of \f$V_j'\f$ .
```
tfalders, line 11967: There's a space between V_j' and the period.
(Same for strided_batched)

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+    srange      #rocblas_srange.\n
+                Specifies the type of range or interval of the singular values to be computed.
+    @param[in]
+    m           rocblas_int. m >= 0.\n
+                The number of rows of matrix A_j.
+    @param[in]
+    n           rocblas_int. n >= 0.\n
+                The number of columns of matrix A_j.
+    @param[inout]
+    A           pointer to type. Array on the GPU (the size depends on the value of strideA)\n
```
tfalders, line 12223: Missing a period

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+                On entry, the matrices A_j.
+                On exit, the contents of A_j are destroyed.
+    @param[in]
+    lda         rocblas_int. lda >= m.\n
+                The leading dimension of A_j.
+    @param[in]
+    strideA     rocblas_stride.\n
+                Stride from the start of one matrix A_j to the next one A_(j+1).
+                There is no restriction for the value of strideS. Normal use case is strideA >= lda*n.
```
tfalders, line 12232: strideS -> strideA

tfalders (commented):
I still have to review the implementation, but here are some initial comments.

tfalders:
Looks like there are a number of CI compiler failures because `rocblas_int offset[bc]` is a variable-length array. Might be good to replace it with a `std::vector`.

clients/gtest/gesvdx_gtest.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+const vector<vector<int>> opt_range = {
+    // always invalid
+    {0, 0, 1, 2, 1, 0, 0},
+    {0, 0, 1, -1, 1, 0, 0},
+    {0, 0, 2, 0, 0, 2, 1},
+    {0, 0, 2, 0, 0, 10, 80},
+    // valid only when n=0
+    {0, 0, 2, 0, 0, 1, 0},
+    // valid only when n>0
```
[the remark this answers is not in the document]
jzuniga-amd: No, it should be `min(m,n) > 0` actually. 
For the sizes used in the tests, all these samples are valid as long as `min(m,n) > 0`. If `m` or `n` are zero, then `il` must be 1 and `iu` must be zero.

clients/include/testing_gesvdx.hpp
```diff
@@ -0,0 +1,1017 @@
...
+    EXPECT_ROCBLAS_STATUS(rocsolver_gesvdx(STRIDED, handle, rocblas_svect_all, right_svect, srange,
+                                           m, n, dA, lda, stA, vl, vu, il, iu, dNsv, dS, stS, dU,
+                                           ldu, stU, dV, ldv, stV, difail, stF, dinfo, bc),
+                          rocblas_status_invalid_value);
+    EXPECT_ROCBLAS_STATUS(rocsolver_gesvdx(STRIDED, handle, left_svect, rocblas_svect_all, srange,
+                                           m, n, dA, lda, stA, vl, vu, il, iu, dNsv, dS, stS, dU,
+                                           ldu, stU, dV, ldv, stV, difail, stF, dinfo, bc),
+                          rocblas_status_invalid_value);
+    EXPECT_ROCBLAS_STATUS(rocsolver_gesvdx(STRIDED, handle, left_svect, right_svect,
+                                           rocblas_srange(-1), m, n, dA, lda, stA, vl, vu, il, iu,
```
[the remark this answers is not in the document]
jzuniga-amd: corrected in both: gesvdx and bdsvdx

clients/include/testing_gesvdx.hpp
```diff
@@ -0,0 +1,1017 @@
...
+        rocblas_int nn = std::min(m, n);
+
+        // construct non singular matrix A such that all singular values are in (0, 20]
+        for(rocblas_int b = 0; b < bc; ++b)
+        {
+            for(rocblas_int i = 0; i < m; i++)
+            {
+                if(i == nn / 4 || i == nn / 2 || i == nn - 1 || i == nn / 7 || i == nn / 5
+                   || i == nn / 3)
+                    hA[b][i + i * lda] = 0;
```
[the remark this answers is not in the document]
jzuniga-amd: No, that's intentional. As I am working directly with a bidiagonal matrix, I want to ensure that the singular values cover all the range (0,20], but I don't want exact zeros in the diagonal as it will make the matrix singular.

clients/include/testing_gesvdx.hpp
```diff
@@ -0,0 +1,1017 @@
...
+                            else
+                                hA[b][i + j * lda] = 0;
+                        }
+                        else
+                        {
+                            if(i == j + 1)
+                                hA[b][i + j * lda] = (hA[b][i + j * lda] - 5) / 10;
+                            else
+                                hA[b][i + j * lda] = 0;
+                        }
```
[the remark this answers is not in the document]
jzuniga-amd: It does not really matter. The code does not know that my matrix is already bidiagonal, so the column/row compression (QR/LQ factorizations) and the bidiagonalization (GEBRD) steps are executed anyways... besides, functions GEQRF/GELQF and GEBRD are also tested in their own test suites. 

I could "disguise" the matrix by multiplying it with orthogonal matrices, but I don't think it is worth the effort.

clients/gtest/gesvdx_gtest.cpp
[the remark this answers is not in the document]
tfalders: I was thinking that in the case of the 1x1 size, these test cases will be invalid because `iu > n`.

library/src/lapack/roclapack_gesvdx.hpp
```diff
@@ -0,0 +1,498 @@
...
+
+    // init sizes
+    size_t a[3] = {};
+    size_t b[3] = {};
+    size_t c[3] = {};
+    size_t d[3] = {};
+    size_t e[5] = {};
+    size_t f[5] = {};
+    size_t g[5] = {};
+    size_t w[5] = {};
```
tfalders, line 156: I'd feel safer if these were explicitly initialized to all zero, or if each unused element were set to zero in the following code.

clients/gtest/gesvdx_gtest.cpp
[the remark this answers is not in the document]
jzuniga-amd: oh, I get it!
You are right. Comment in the code changed.

clients/include/testing_gesvdx.hpp
```diff
@@ -0,0 +1,1017 @@
...
+                     const rocblas_int nT,
+                     Td& dUT,
+                     const rocblas_int lduT,
+                     const rocblas_stride stUT,
+                     Td& dVT,
+                     const rocblas_int ldvT,
+                     const rocblas_stride stVT,
+                     Wh& hA,
+                     Ih& hNsv,
+                     Ih& NsvRes,
```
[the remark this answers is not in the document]
jzuniga-amd: done

library/src/lapack/roclapack_gesvdx.hpp
```diff
@@ -0,0 +1,498 @@
...
+    *size_WS_svdx2_lqrf1_brd1 = *std::max_element(std::begin(a), std::end(a));
+    *size_WS_svdx3_lqrf2_brd2 = *std::max_element(std::begin(b), std::end(b));
+    *size_WS_svdx4_lqrf3_brd3 = *std::max_element(std::begin(c), std::end(c));
+    *size_WS_svdx5_brd4 = *std::max_element(std::begin(d), std::end(d));
+    *size_WS_svdx10_mlqr1_mbr1 = *std::max_element(std::begin(e), std::end(e));
+    *size_WS_svdx11_mlqr2_mbr2 = *std::max_element(std::begin(f), std::end(f));
+    *size_WS_svdx12_mlqr3_mbr3 = *std::max_element(std::begin(g), std::end(g));
+
+    // size of extra array of pointers to workspace
+    if(BATCHED && thinSVD && ((leftvS && row) || (rightvS || !row)))
```
tfalders, line 252: Should that be `rightvS && !row`?

clients/include/testing_gesvdx.hpp
```diff
@@ -0,0 +1,1017 @@
...
+                     double* max_errv)
+{
+    /** As per lapack's documentation, the following workspace size should work:
+        rocblas_int minn = std::min(m,n);
+        rocblas_int maxn = std::max(m,n);
+        rocblas_int lwork = minn * minn + 6 * minn + maxn;
+        rocblas_int lrwork = 17 * minn * minn;
+        std::vector<T> work(lwork);
+        std::vector<S> rwork(lrwork);
+        HOWEVER, gesvdx_ GIVES ILEGAL VALUE FOR ARGUMENT lwork.
```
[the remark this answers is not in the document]
jzuniga-amd: corrected

library/src/lapack/roclapack_gesvdx.hpp
```diff
@@ -0,0 +1,498 @@
...
+                                         T** workArr,
+                                         T** workArr2)
+{
+    ROCSOLVER_ENTER("gesvdx", "leftsv:", left_svect, "rightsv:", right_svect, "srange:", srange,
+                    "m:", m, "n:", n, "shiftA:", shiftA, "lda:", lda, "vl:", vl, "vu:", vu,
+                    "il:", il, "iu:", iu, "ldu:", ldu, "ldv:", ldv, "bc:", batch_count);
+
+    // quick return
+    if(n == 0 || m == 0 || batch_count == 0)
+        return rocblas_status_success;
```
tfalders, line 313: If `m == n == 0` but `batch_count > 0`, I think we still need to set `info` and `nsv`, right?

clients/include/testing_gesvdx.hpp
```diff
@@ -0,0 +1,1017 @@
...
+    if(!perf)
+    {
+        //gesvdx_initData<true, false, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A, 0);
+
+        // cpu-lapack performance (only if not in perf mode)
+        *cpu_time_used = get_time_us_no_sync();
+        //for(rocblas_int b = 0; b < bc; ++b)
+        //    cblas_gesvdx<T>(left_svect, right_svect, srange, m, n, hA[b], lda, vl, vu, il, iu, hNsv[b], hS[b], hU[b], ldu, hV[b], ldv,
+        //                   work.data(), lwork, rwork.data(), hifail[b], hinfo[b]);
+        *cpu_time_used = get_time_us_no_sync() - *cpu_time_used;
```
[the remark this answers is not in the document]
jzuniga-amd: Good idea, setting it to quite NAN

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -11750,6 +11750,662 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zgesvd_strided_batched(rocblas_handle
...
+    @param[in]
+    m           rocblas_int. m >= 0.\n
+                The number of rows of matrix A.
+    @param[in]
+    n           rocblas_int. n >= 0.\n
+                The number of columns of matrix A.
+    @param[inout]
+    A           pointer to type. Array on the GPU of dimension lda*n.\n
+                On entry, the matrix A.
+                On exit, the contents of A are destroyed.
```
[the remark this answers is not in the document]
jzuniga-amd: Yes, in our case A cannot be constant as the factorizations modify its contents. 
I guess the const qualifier will go only on the compatibility API (as for all the other functions) for now.

tfalders (commented):
Implementation looks good! Just a few minor comments.

library/src/lapack/roclapack_gesvdx.hpp
jzuniga-amd: I think the default is initialized to zero, but sure, why not... Doing the same for gesvd test as well.

library/src/lapack/roclapack_gesvdx.hpp
jzuniga-amd: Correct, good catch!!

library/src/lapack/roclapack_gesvdx.hpp
jzuniga-amd: Correct, I forgot!

tfalders (approved):

library/src/lapack/roclapack_gesvdx.hpp
```diff
@@ -0,0 +1,509 @@
...
+        return rocblas_status_invalid_size;
+    if(srange == rocblas_srange_index && (iu > min(m, n) || (min(m, n) > 0 && il > iu)))
+        return rocblas_status_invalid_size;
+
+    // skip pointer check if querying memory size
+    if(rocblas_is_device_memory_size_query(handle))
+        return rocblas_status_continue;
+
+    // 3. invalid pointers
+    if((n * m && !A) || (min(m, n) && !S) || (min(m, n) && !ifail) || (batch_count && !info)
```
tfalders, line 76: One small thing: I think it makes more sense for ifail to be ignored when both svects are `rocblas_svect_none`. Up to you though.

library/src/lapack/roclapack_gesvdx.hpp
```diff
@@ -0,0 +1,511 @@
...
+        (TT*)WS_svdx5_brd4, WS_svdx6, WS_svdx7, WS_svdx8, WS_svdx9, (TT*)WS_svdx10_mlqr1_mbr1,
+        (TT*)WS_svdx11_mlqr2_mbr2, (TT*)WS_svdx12_mlqr3_mbr3);
+
+    /***** 3. compute/update left vectors *****/
+    /******************************************/
+    if(leftvS)
+    {
+        // For now we work with the k columns in tmpZ, i.e. k vectors
+        // (TODO: explore other options like transfering nsv to the host, or having
+        //  template functions accepting dimensions as pointers)
```
tfalders, line 416: This is a bit last minute but we can calculate `nsv` exactly when `srange` is `rocblas_srange_index` (in the other cases we can leave it as `k` for now). This will help in hipSOLVER, because my workaround for gesvda always uses `rocblas_srange_index`, so I would be able to always have U and V be the exactly size needed.

PR #471, opened 2022-09-14
add missing batch_count to gesvdx documentation

jzuniga-amd:


tfalders (approved):

cgmb (approved):
This is why I harp on spurious failures in CI... the static analysis has an explicit check for this sort of mistake, but the warning got lost in the noise of all the other failures.

I won't be at the retrospective, but my complaint is that the new base image that rocSOLVER was being built on was rolled out without testing. There should never need to be fixes like #469 after a rollout. Those issues should be caught beforehand, so we don't have _everything_ failing for a few days. Once that happens enough times, [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue) begins to set in.

PR #473, opened 2022-09-21
HOTFIX: Remove backwards compatibility headers on Windows

lawruble13:
The backwards compatibility headers should not be generated or installed on Windows.


tfalders:
I don't think we've ever had an identical hotfix targeting two releases before. We like to maintain a linear history by merging the hotfix back into develop, so I wonder if we should merge the 5.3 branch into the 5.4 branch after the hotfix is applied (rather than merging two separate PRs) so that the commit hashes will stay the same.

lawruble13:
Opened to incorrect branch.

lawruble13:
@tfalders Sorry for the confusion! I should have closed this PR prior to opening the other one.

PR #477, opened 2022-10-14
Cherry Pick Fix for old version rocm include/lib folders not removed on upgrade.

arvindcheru:
Summary of proposed changes:
- Cherry Pick Fix for old version rocm include/lib folders not removed on upgrade commit


cgmb (approved):
LGTM. For this sort of change in the future, you might want to consider using [`git cherry-pick -x`](https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt--x). It includes some information about the original commit in the log message, which makes it easier track these sorts of changes back to their origin when you're browsing the git history.

jzuniga-amd (approved):

tfalders (approved):

PR #478, opened 2022-11-03
Update include file path of rocblas fortran file

arvindcheru:
Proposed Change:
- Update the cmake file of samples with path to rocblas fortran program according to latest structure <include/rocblas/rocblas_module.f90>


cgmb:
@jzuniga-amd, @tfalders, @qjojo: Do we just want to drop the Fortran sample instead? I introduced the Fortran sample long ago as a cheap check for our binary interface, but I feel like it's now causing more problems than it prevents.

qjojo (approved):

cgmb:
All references to rocblas_module.f90 have been removed from rocSOLVER, so this change is now moot.

PR #481, opened 2022-11-16
Remove Fortran sample

cgmb:
The rocSOLVER Fortran sample was created mostly to test that rocSOLVER was provided a suitable ABI. For developers like myself who were new to the rocSOLVER project, it was a living piece of documentation of the library requirements. It's been a long time since the sample has provided much value, but it still has maintenance costs. Fortran is a pain on Windows and it interacts poorly with Address Sanitizer on all platforms. Worse, the sample isn't even particularly good! With hipfort replacing the library-provided modules, we can take this opportunity to drop Fortran from the repo and save ourselves some future maintenance.


tfalders (approved):

cgmb:
This is an alternative to #478.

qjojo (approved):

PR #482, opened 2022-11-17
Rename cblas_* functions to cpu_*

cgmb:
These function names are a bit misleading as the actual cblas library has not been used in rocSOLVER since 7a280668b939604cbdad8cf5110076697a6e905e. A different prefix may help make that more clear.

The x prefix was already used by some blas functions in the tests and it seems an appropriate stand-in for s, d, c, or z when the function call could dispatch to any of those.

This is a follow-up to #445, which removed all use of the actual cblas library from rocSOLVER.


cgmb:
I suppose functions like `cblas_symv_hemv` should probably be `xsymv_xhemv`. I'll make that change before marking this as ready for review.

cgmb:
I also dropped the unnecessary explicit template type from the function call. In nearly all cases, it was superfluous.

There were four cases where it was used to resolve ambiguity and I fixed those in d391db2824e9ecb7f67fbfe0646ba1ba1272c67d. It's a bit of a matter of taste which way is better, but I mildly prefer explicitly converting the arguments so that they exactly match the function parameter types vs. specifying the function type explicitly and relying on implicit conversion for the arguments.

If you want me to revert that change (or use a different prefix), it shouldn't be difficult. I converted everything with a couple commands like this:

    find -name '*pp' -exec sed -ri 's/\<cblas_([[:alnum:]]+)(<[ST]>)?/x\1/g' {} \+

tfalders:
I'm not a big fan of the x* prefix for this, especially since we've already been using x as a wildcard in some of our test file names. Perhaps a different prefix? Like cpu_*?

cgmb:
`cpu_` would be fine by me. Another alternative would be to drop the prefix entirely.

cgmb:
It sounds like `x` is unpopular, so I've dropped it. I'm happy to go with whatever you guys want, as long as it's not `cblas_`. Perhaps we can take a few minutes to discuss it next Wednesday.

clients/include/testing_ormtr_unmtr.hpp
```diff
@@ -138,7 +138,7 @@ void ormtr_unmtr_initData(const rocblas_handle handle,
         }
 
         // compute sytrd/hetrd
-        cblas_sytrd_hetrd<T>(uplo, nq, hA[0], lda, D.data(), E.data(), hIpiv[0], hW.data(), size_W);
+        cpu_sytrd_hetrd<T>(uplo, nq, hA[0], lda, D.data(), E.data(), hIpiv[0], hW.data(), size_W);
```
tfalders, line 141: It looks like you've been removing all the template parameters, but missed those functions with two possible names.

clients/common/lapack_host_reference.cpp
```diff
@@ -5041,7 +4779,7 @@ void cblas_trsm<rocblas_double_complex>(rocblas_side side,
 /*
 // trsv
 template <>
-void cblas_trsv<float>(rocblas_fill uplo,
+void cpu_trsv<float>(rocblas_fill uplo,
                        rocblas_operation transA,
                        rocblas_diagonal diag,
                        rocblas_int n,
```
tfalders, line 4785: Not terribly important, but it looks like some of the whitespace on the args didn't get updated.

tfalders (approved):
LGTM!

qjojo (approved):

clients/common/lapack_host_reference.cpp
cgmb: I suppose they didn't get autoformatted since those lines are commented out.

clients/include/testing_ormtr_unmtr.hpp
cgmb: Thanks, I didn't notice that.

PR #483, opened 2022-11-18
Remove rpath/runpath from ROCm binaries and libraries

raramakr:
SWDEV-310152:

- Single version package: add ldconfig, no RPATH/RUNPATH in either binaries or libraries
- Multi version package: use RPATH for binaries, no RPATH/RUNPATH for libraries
- amdclang/clang/hipcc compilers shall not add RPATH/RUNPATH to produced binaries or libraries unless explicitly requested by an option
- Versioning scripts will take care of adding rpath to binaries for multi version package

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,7 +59,7 @@ Options:
   -s | --static               Pass this flag to build rocsolver as a static library.
                               (rocsolver must be built statically when the used companion rocblas is also static).
 
-  -r | --relocatable          Pass this to add RUNPATH(based on ROCM_RPATH) and remove ldconf entry.
+  -r | --relocatable          Pass this flag to disable creation of ldconfig files in generated packages.
 
   -n | --no-optimizations     Pass this flag to disable optimizations for small sizes.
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -582,12 +583,7 @@ else
...
 fi
 
-rocm_rpath=""
 if [[ "${build_relocatable}" == true ]]; then
-    rocm_rpath=" -Wl,--enable-new-dtags -Wl,--rpath,/opt/rocm/lib:/opt/rocm/lib64"
-    if ! [ -z ${ROCM_RPATH+x} ]; then
-        rocm_rpath=" -Wl,--enable-new-dtags -Wl,--rpath,${ROCM_RPATH}"
-    fi
     cmake_common_options+=('-DROCM_DISABLE_LDCONFIG=ON')
 fi
```
cgmb, line 588: Please update the help text for the relocatable option from:

> Pass this to add RUNPATH(based on ROCM_RPATH) and remove ldconf entry.

to

> Pass this flag to disable creation of ldconf entry in generated packages.

cgmb (commented):
It looks good, but you need to update the help text for `-r` to match the new behaviour.

cgmb (approved):

raramakr:
> It looks good, but you need to update the help text for `-r` to match the new behaviour.

Updated help text

qjojo (approved):

PR #489, opened 2022-12-12
Completes implementation of divide & conquer method in STEDC

jzuniga-amd:
This PR also refactors STEDC unit test so that all the aspects of the divide & conquer algorithm (such as deflation) can be fully tested.

### DIFF
--- a/clients/include/testing_stedc.hpp
+++ b/clients/include/testing_stedc.hpp
@@ -95,13 +95,13 @@ void stedc_initData(const rocblas_handle handle,
         for(rocblas_int i = 0; i < n; i++)
         {
             hD[0][i] += 400;
-            hE[0][i] -= 5;
+//            hE[0][i] -= 5;
         }
 
         // add fixed splits in the matrix to test split handling
-        rocblas_int k = n / 2;
-        hE[0][k] = 0;
-        hE[0][k - 1] = 0;
+//        rocblas_int k = n / 2;
+//        hE[0][k] = 0;
+//        hE[0][k - 1] = 0;
 
         // initialize C to the identity matrix
         if(evect == rocblas_evect_original)
--- a/clients/include/testing_stedc.hpp
+++ b/clients/include/testing_stedc.hpp
@@ -96,8 +96,8 @@ void stedc_initData(const rocblas_handle handle,
             rocblas_init<S>(hE, true);
         }
 
-        // otherwise, the marix will be divided in exactly 2 independent blocks, if the size is even, 
-        // or 3 if the size if odd. The 2 main indepedent blocks will have the same eigenvalues. 
+        // otherwise, the marix will be divided in exactly 2 independent blocks, if the size is even,
+        // or 3 if the size if odd. The 2 main indepedent blocks will have the same eigenvalues.
         // The last block, if the case, will have eigenalue equal 1.
         else
         {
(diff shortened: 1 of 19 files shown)


library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -330,7 +330,7 @@ void rocsolver_steqr_getMemorySize(const rocblas_evect evect,
     if(evect == rocblas_evect_none)
         *size_work_stack = sizeof(rocblas_int) * (2 * 32) * batch_count;
     else
-        *size_work_stack = sizeof(S) * (2 * n - 2) * batch_count;
+        *size_work_stack = sizeof(S) * (2 * n) * batch_count;
```
tfalders, line 333: Was there a memory access fault caused by this array being too small? Do we not also need to update `strideW` in `steqr_kernel`?

clients/include/testing_stedc.hpp
```diff
@@ -438,7 +529,7 @@ void testing_stedc(Arguments& argus)
             if(argus.norm_check)
             {
                 rocsolver_bench_output("cpu_time_us", "gpu_time_us", "error");
-                rocsolver_bench_output(cpu_time_used, gpu_time_used, max_error);
+                rocsolver_bench_output(cpu_time_used, gpu_time_used, max_err);
```
tfalders, line 532: Should we not print `max(max_err, max_errv)`?

tfalders (commented):
I still have to look over the changes in rocauxiliary_stedc.hpp, but here are my comments so far.

Also, there seems to be a CI failure in heevd on Centos 7.

qjojo (approved):
The algorithm is a bit beyond my reach but the glue looks good and I trust the tests. Nice job.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -15,79 +15,1308 @@
...
+        bb *= aa;
+        fx += bb;
+
+        // calculate tolerance er for convergence test
+        er += 8 * (hx - gx) + 2 * pinv + 3 * abs(bb) + abs(tau) * fdx;
+
+        // from now on, further step corrections will be calculated either with fixed weights method
+        // or with normal interpolation depending on the value of boolean fixed
+        cc = up ? -1 : 1;
+        fixed = (cc * fx > abs(oldfx) / 10) ? true : false;
```
cgmb, line 278: I would omit the redundant `? true : false`. IMO, it's easy enough to read something like

    fixed = (cc * fx) > (abs(oldfx) / 10);

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -15,79 +15,1308 @@
...
+    }
+
+    *ev = x;
+    return converged ? 0 : 1;
+}
+
+/** STEDC_NUM_LEVS returns the ideal number of times/levels in which a matrix (or split block)
+    will be divided during the divide phase of divide & conquer algorithm.
+    i.e. number of sub-blocks = 2^levels **/
+__host__ __device__ inline rocblas_int stedc_num_levs(const rocblas_int n)
```
cgmb, line 576: Is the "lev" abbreviation coming from somewhere? I'm more familiar with "lvl". Though, given how long the function name is already, it seems like "level" would be fine too. It's only saving two characters after all.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -15,79 +15,1308 @@
...
+        levels = 8;
+    }
+    else if(n < 16)
+    {
+        levels = 0;
+    }
+    else
+    {
+        levels = n / 8;
+        levels = floor(log(levels) / log(2));
```
cgmb, line 595: I don't think this is in a hot path, so it doesn't really matter, but this pains me nonetheless. I've been meaning to ask about some of the solutions to this sort of thing... I suppose this is a good prompt to finally get around to it.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -15,79 +15,1308 @@
...
+                        }
+                        __syncthreads();
+                    }
+                    nrm = sqrt(nrm);
+
+                    // multiply by C (row by row)
+                    for(int ii = 0; ii < tsz; ++ii)
+                    {
+                        rocblas_int i = in + ii;
+                        go = go && (ii < sz);
```
cgmb, line 1248: Is the branch here intentional? The short-circuiting of logical operators introduces a branching point in this logic. If that's not intentional, I would probably suggest using the bitwise, `go &= (ii < sz)`.

cgmb (commented):
Youch. I thought I was getting along well, but I clearly had more than two weeks remaining. Thanks so much for taking this over.

I have only minor comments.

library/src/auxiliary/rocauxiliary_steqr.hpp
jzuniga-amd: I added the 2 extra elements just for simplicity when combining workspaces with stedc. 

Good catch! I corrected strideW.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1,5 +1,5 @@
 /************************************************************************
- * Derived from the BSD3-licensed
+  Derived from the BSD3-licensed
```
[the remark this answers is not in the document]
jzuniga-amd: I don't know. I did not touch this line.
I reversed the change. Let's see if the formatting removes the '*' again

library/src/auxiliary/rocauxiliary_stedc.hpp
jzuniga-amd: Yeah, does not really matter. This function is only called once. Will leave it as it is for readability. Besides, this code will disappear once the tuning of the number of levels is done

clients/include/testing_stedc.hpp
```diff
@@ -88,20 +88,107 @@ void stedc_initData(const rocblas_handle handle,
...
+            // The second sub-block will have some repeated eigenvalues in order to test the deflation process
+            S d;
+            rocblas_int NN1 = N1 / 2;
+            rocblas_int NN2 = N1 - NN1;
+            rocblas_int s1 = NN1 * NN1;
+            rocblas_int s2 = NN2 * NN2;
+            rocblas_int sw = NN2 * 32;
+            std::vector<S> A1(s1);
+            std::vector<S> A2(s2);
+            for(rocblas_int i = 0; i < NN1; ++i)
```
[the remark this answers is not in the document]
cgmb: Unfortunately, this is the iteration order used in the initialization loop in pretty much every test. I'm fine with fixing it on a case-by-case basis if @jzuniga-amd and @tfalders are. If not, I suppose we should finally create a ticket for fixing it across all the tests.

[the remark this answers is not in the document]
jzuniga-amd: Yeah, I have been waiting until we introduce openMP and other optimizations to the clients to take a look into these details across all tests... we have never had the time yet :(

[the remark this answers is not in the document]
cgmb: I was a bit worried that OpenMP would be a pain for packaging, but it's pretty easy to make it optional. rocSPARSE does it nicely, I think.

cgmb (approved):

tfalders (approved):
Unfortunately, my mind is too scattered today to give a thorough review of the algorithm, but what I've checked looks good!

library/src/auxiliary/rocauxiliary_stedc.hpp
cgmb: For posterity, a better solution would be built on [`__builtin_clz`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html). I linked the GCC documentation, but it's also available in clang.

PR #490, opened 2022-12-13
Implement lauum

qjojo:
This adds lauum to rocSOLVER, which computes U*U^T or L^T*L for real triangular matrices, or U*U^H or L^H*L for complex triangular matrices. 

Currently the complex tests are notched out because the trmm call in rocauxiliary_lauum.hpp is producing invalid results for the complex case for reasons that I can't figure out, the same call works in the real case and the requested operation is `rocblas_operation_conjugate_transpose`, so there's either a bug in rocBLAS or I'm missing a gotcha in the way trmm is being called.


clients/include/lapack_host_reference.hpp
```diff
@@ -461,6 +461,18 @@ void cpu_ormlq_unmlq(rocblas_side side,
...
+void cblas_lauum(rocblas_fill uploR, rocblas_int n, T* A, rocblas_int lda, rocblas_int* info);
+
+template <typename T>
+void cblas_laswp(rocblas_int n,
+                 T* A,
+                 rocblas_int lda,
+                 rocblas_int k1,
+                 rocblas_int k2,
+                 rocblas_int* ipiv,
+                 rocblas_int inc);
```
cgmb, line 474: s/cblas/cpu/

clients/include/testing_lauum.hpp
```diff
@@ -0,0 +1,310 @@
+/* ************************************************************************
+ * Copyright (c) 2020-2022 Advanced Micro Devices, Inc.
```
cgmb, line 2: Is there any code in here from 2020? If not, just 2022.

clients/include/testing_lauum.hpp
```diff
@@ -0,0 +1,310 @@
...
+ * ************************************************************************ */
+
+#pragma once
+
+#include <iostream>
```
cgmb, line 7: This looks unnecessary.

cgmb (changes requested):
This is a public API function, right? You've added the Doxygen descriptions, but you'll need to edit the Sphinx docs in `docs/sources` to add it to the human-readable documentation.

clients/include/testing_lauum.hpp
qjojo: Leftover from debugging, removed

library/src/auxiliary/rocauxiliary_lauum.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    void* work;
+    rocblas_device_malloc mem(handle, size_work);
+    if(!mem)
+        return rocblas_status_memory_error;
+
+    work = mem[0];
+
+    // execution
+    return rocsolver_lauum_template<T>(handle, uplo, n, A, shiftA, lda, info, strideA, batch_count,
```
tfalders, line 47: Why is `strideA` after `info`? Though I suppose it doesn't matter if we're removing `info`.

library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -0,0 +1,109 @@
...
+    // quick return
+    if(n == 0 || batch_count == 0)
+        return rocblas_status_success;
+
+    hipStream_t stream;
+    rocblas_get_stream(handle, &stream);
+
+    rocblas_int blocks = (n - 1) / 32 + 1;
+    dim3 grid(blocks, blocks, batch_count);
+    dim3 threads(32, 32);
```
tfalders, line 77: 32 should be replaced by BS2 (the 2D thread-block size defined in ideal_sizes.hpp)

clients/gtest/lauum_gtest.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // quick return
+    {0, 1},
+    // invalid
+    {-1, 1},
+    {10, 0},
+    // normal (valid) samples
+    {3, 3},
+    {10, 10},
+    {20, 20},
+    {30, 30}};
```
tfalders, line 29: lda should be larger than n in some cases

clients/gtest/lauum_gtest.cpp
```diff
@@ -0,0 +1,97 @@
...
+TEST_P(LAUUM, __float_complex)
+{
+    run_tests<rocblas_float_complex>();
+}
+
+TEST_P(LAUUM, __double_complex)
+{
+    run_tests<rocblas_double_complex>();
+}
+*/
```
tfalders, line 91: All functions in the API should be tested. We should either get the complex tests working, or remove the complex cases from the API until the tests are working.

library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -0,0 +1,109 @@
...
+    T one = 1;
+    T zero = 0;
+
+    rocblas_fill uploC = (uplo == rocblas_fill_upper) ? rocblas_fill_lower : rocblas_fill_upper;
+
+    // put the triangular factor of interest in work
+    ROCSOLVER_LAUNCH_KERNEL(set_zero<T>, grid, threads, 0, stream, n, n, work, shiftA, lda, strideA,
+                            uploC);
+    ROCSOLVER_LAUNCH_KERNEL(copy_mat<T>, grid, threads, 0, stream, n, n, A, shiftA, lda, strideA,
+                            work, shiftA, lda, strideA, no_mask{}, uplo);
```
tfalders, line 87: work should not be using the same values as A in this case. In particular, if `lda > n`, you should expect a segfault. For work, use shiftA -> 0, lda -> n, strideA -> n * n (or define ldw = n, strideW = n * n).

Also applies to trmm and kernel calls below.

library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    rocblas_side side;
+    if(uplo == rocblas_fill_upper)
+    {
+        side = rocblas_side_right;
+    }
+    else
+    {
+        side = rocblas_side_left;
+    }
```
tfalders, line 97: I'd prefer if this was condensed using the ( ? : ) notation.

tfalders (commented):
Not bad for a first attempt at adding a new function, though there are some issues to address, namely the failing complex tests and a potential segfault in the implementation.

Also, you need to add the new functions to the documentation. The changelog needs to be updated, as well as userguide_intro.rst and api_auxiliaryfunc.rst.

clients/gtest/lauum_gtest.cpp
qjojo: I changed the error tolerance as you suggested but the complex cases are still failing. I'll remove the complex versions from the public API and docs so I can get this in.

tfalders (approved):
LGTM! Be sure to wait for Juan's approval before merging, and do a squash-and-merge.

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -1069,6 +1069,45 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlasyf(rocblas_handle handle,
                                                  rocblas_int* info);
 //! @}
 
+/*! @{
+    \brief LAUUM computes the product of a triangular matrix A with its transpose.
+
+    \details
+    If uplo denotes upper, then \f$AA^T\f$ is computed, if uplo denotes lower then \f$A^TA\f$ is computed instead.
+    For complex types computes \f$AA^H\f$ or \f$A^HA\f$ depending on uplo.
```
cgmb, line 1077: Just to pick on the wording here a bit:

- Every other function uses the phrasing "If uplo indicates upper".
- "if uplo denotes lower then \f$A^TA\f$ is computed instead" is an independent phrase, so there's a comma splice here. Maybe just make it an independent sentence. It's also missing a comma before "then".
- The description for complex types is inconsistent in style compared to the description of the real functions. Given that we're not including the operations on complex types at the moment, I think the description of the operations on complex types can be dropped for now.

cgmb (commented):
I'm picking on the wording a bit, but it looks good.

cgmb (approved):

library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    // put the triangular factor of interest in work
+    ROCSOLVER_LAUNCH_KERNEL(set_zero<T>, grid, threads, 0, stream, n, n, work, 0, n, strideW, uploC);
+    ROCSOLVER_LAUNCH_KERNEL(copy_mat<T>, grid, threads, 0, stream, n, n, A, shiftA, lda, strideA,
+                            work, 0, n, strideW, no_mask{}, uplo);
+
+    rocblas_side side = (uplo == rocblas_fill_upper) ? rocblas_side_right : rocblas_side_left;
+
+    // work = work * A' or work = A' * work
+    rocblasCall_trmm<false, true, T>(handle, side, uplo, rocblas_operation_conjugate_transpose,
```
jzuniga-amd, line 89: Not the best use of TRMM. work is also triangular and being size n-by-n, conceptually, half of the operations of the whole product are multiplications by zero!

jzuniga-amd (commented):
Looks good for initial implementation, good job.  We may need to revisit it soon for optimizations, though. I was hoping for an implementation along the lines of a blocked algorithm (with gemms and symmetric rank updates)... the current implementation has too many data movements; it will probably not show the performance expected by the group requesting the API. We will see.

Just a few minor comments that require attention:

library/src/auxiliary/rocauxiliary_lauum.hpp
jzuniga-amd: we will revisit this in the optimization round

jzuniga-amd (approved):

qjojo:
CI failure is the known issue with managed malloc, I will manually merge after consulting the rest of the team.

PR #491, opened 2022-12-13
Remove host_vector and device_vector from clients

cgmb:
The `host_vector` is only used in `norm_error` (which I also consolidated). `device_vector` is entirely unused.


qjojo (approved):

tfalders (approved):
LGTM!

PR #492, opened 2022-12-13
Add asynchronous execution example

qjojo:
With the latest driver and rocBLAS changes our codebase now experimentally supports asynchronous execution via the hipGraph API. This adds an example (similar to the rocBLAS one) that shows how to use rocSOLVER functions in this way.

### DIFF
--- a/clients/samples/example_graph.c
+++ b/clients/samples/example_graph.c
@@ -1,0 +1,125 @@
+#include <hip/hip_runtime_api.h> // for hip functions
+#include <rocsolver/rocsolver.h> // for all the rocsolver C interfaces and type declarations
+#include <stdio.h>   // for printf
+#include <stdlib.h>  // for malloc
+
+// Example: Compute the QR Factorization of a matrix asynchronously on the GPU
+
+double *create_example_matrix(rocblas_int *M_out,
+                              rocblas_int *N_out,
+                              rocblas_int *lda_out) {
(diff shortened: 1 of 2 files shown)


tfalders (approved):

clients/samples/example_graph.c
```diff
@@ -0,0 +1,125 @@
+#include <hip/hip_runtime_api.h> // for hip functions
+#include <rocsolver/rocsolver.h> // for all the rocsolver C interfaces and type declarations
+#include <stdio.h>   // for printf
+#include <stdlib.h>  // for malloc
+
+// Example: Compute the QR Factorization of a matrix asynchronously on the GPU
```
cgmb, line 6: This is the comment from the basic example. Maybe adjust it to mention that this example is to show use with the hipGraph API.

cgmb (commented):
Do we not have to associate the stream with the handle? I'd have thought this would require a call to `rocblas_set_stream` somewhere.

clients/samples/example_graph.c
qjojo: You're right, it's my understanding that the stream has to be recorded on the rocblas stream but it can be executed on a different one (that's how it is in the rocBLAS sample anyway).

qjojo:
The CI failures on this are the known issues with the managed malloc test, this PR is just adding a sample so it can be merged safely.

jzuniga-amd (approved):

PR #494, opened 2022-12-14
Fix getf2_npvt/getrf_npvt singular value tests

cgmb:
The getf2_getrf_npvt_initData function uses an unconstrained template argument for the type of hinfo and host_strided_batched<T> has an implicit conversion to bool, which allows these two variables to be passed in the wrong order without triggering any type errors.

I noticed this error in the process of rewriting host_strided_batched<T> because I removed the implicit conversion to bool.


jzuniga-amd (approved):

tfalders (approved):
Good catch!

PR #496, opened 2022-12-14
Added geblttrf_npvt and geblttrs_npvt

tfalders:
This PR adds the block tridiagonal factorization and solve methods. The interleaved batched cases are currently not working so I've removed them for now.


clients/include/testing_geblttrf_npvt.hpp
```diff
@@ -0,0 +1,693 @@
...
+                            const rocblas_int ldc,
+                            const rocblas_int bc,
+                            Th& hA,
+                            Th& hB,
+                            Th& hC,
+                            const bool singular)
+{
+    if(CPU)
+    {
+        T tmp;
```
cgmb, line 143: This variable appears to be unused.

clients/include/testing_geblttrf_npvt.hpp
```diff
@@ -0,0 +1,693 @@
...
+        //                                     hB, hC, singular);
+
+        // // cpu-lapack performance (only if not in perf mode)
+        // *cpu_time_used = get_time_us_no_sync();
+        // for(rocblas_int b = 0; b < bc; ++b)
+        // {
+        //     cpu_getrf(m, n, hA[b], lda, hIpiv[b], hInfo[b]);
+        // }
+        // *cpu_time_used = get_time_us_no_sync() - *cpu_time_used;
+        *cpu_time_used = nan("");
```
cgmb, line 416: ?

clients/include/testing_geblttrs_npvt.hpp
```diff
@@ -0,0 +1,639 @@
...
+        //                                        dX, ldx, bc, hA, hB, hC, hX);
+
+        // // cpu-lapack performance (only if not in perf mode)
+        // *cpu_time_used = get_time_us_no_sync();
+        // for(rocblas_int b = 0; b < bc; ++b)
+        // {
+        //     cpu_getrs(trans, n, nrhs, hA[b], lda, hIpiv[b], hB[b], ldb);
+        // }
+        // *cpu_time_used = get_time_us_no_sync() - *cpu_time_used;
+        *cpu_time_used = nan("");
```
cgmb, line 357: ?

clients/include/testing_geblttrf_npvt.hpp
tfalders: These functions do not have a vanilla LAPACK equivalent, so a comparison with the cpu time doesn't really make sense. Declaring the cpu time to be NaN is also done for gesvdx, though in that case it's because LAPACK's gesvdx crashes and we can't get a cpu timing result.

cgmb: Could you add something to that effect as a comment in the code? e.g., "there's no direct cpu-lapack equivalent for performance comparison, but the lines below can be uncommented for a comparison with the more general cpu-lapack getrf" or something.

tfalders: Sure, I'll add a comment. The lines can't actually be uncommented though - not only do m, n, and hIpiv not exist, but hA is in a blocked format that getrf does not expect.

tfalders: Maybe I'll even remove those lines since they serve no purpose.

clients/include/testing_geblttrf_npvt.hpp
tfalders: I'll remove it before merging. No sense restarting CI for an unused variable removal.

qjojo (approved):

jzuniga-amd (commented):
I added a bunch of modifications (mainly to the documentation) in a PR to your branch. Please have a look.

jzuniga-amd (approved):
LGTM

PR #497, opened 2022-12-14
Disabled failing tests for FC

tfalders:
STEDC and dependent functions are now failing on my local machine. In order to ensure failing tests do not block the mainline merge, I've moved the failing `checkin_lapack` tests to `known_bug` and disabled the `daily_lapack` tests (moving both to `known_bug` creates a name conflict). Hopefully we can get a hotfix in to reenable these once they're no longer failing.


jzuniga-amd (approved):
LGTM. Thanks Troy

qjojo (approved):

PR #501, opened 2022-12-21
Modified rocBLAS_iamax

NaveenElumalaiAMD:
Modified level 1 rocBLAS function iamax as per changes in rocBLAS [PR:1580](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1580)

### DIFF
--- a/library/src/include/rocblas.hpp
+++ b/library/src/include/rocblas.hpp
@@ -43,7 +43,7 @@ rocblas_status rocblasCall_axpy(rocblas_handle handle,
 }
 
 // iamax
-template <bool ISBATCHED, typename T, typename S, typename U>
+template <typename T, typename S, typename U>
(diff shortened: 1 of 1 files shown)


tfalders (approved):
LGTM! Currently, there aren't any rocSOLVER functions calling `rocblasCall_iamax`, so we could even remove this method entirely.

jzuniga-amd (approved):

jzuniga-amd:
> LGTM! Currently, there aren't any rocSOLVER functions calling `rocblasCall_iamax`, so we could even remove this method entirely.

rocSOLVER has its own internal iamax method (which gives better performance at the moment), but better to leave the rocblas call wrapper in case we need it in the future.

cgmb (approved):

PR #502, opened 2022-12-21
Fix rare syevj bug

tfalders:
This PR should(?) address a rarely-occurring bug in syevj/heevj. It's a little difficult to be certain, because the bug only shows up once in every 5 or so test runs I do, but it hasn't shown up in any of my recent runs.


jzuniga-amd:
will open a single PR with all hotfixes as per instructions

PR #503, opened 2022-12-23
fix synchronization bug in stedc

jzuniga-amd:
This PR should fix the bugs in stedc and all its related functions.


jzuniga-amd:
will open a single PR with all hotfixes as per instructions

PR #507, opened 2022-12-27
Move internal headers to a subdirectory

cgmb:
The only rocsolver header intended for direct inclusion by users is <rocsolver/rocsolver.h>. The other rocsolver headers were not explicitly marked as internal, but that is what this change is fixing. All headers within internal/ are implementation details and may change without warning.


jzuniga-amd (commented):
If we will follow the structure of other libraries (like rocBLAS or even our own hipSOLVER), why still keep the `rocsolver` directory? 
Main header should be `rocsolver/library/include/rocsolver.h`, and the internal headers should be at `rocsolver/library/include/internal/`

cgmb:
I don't remember why it matters for the source tree, but they were moved into the rocsolver directory as a part of the build system reorganization. It was helpful to have the source tree layout match the install tree layout and that directory is the part before the slash in `#include <rocsolver/rocsolver.h>`.

cgmb:
Though... looking through other libraries, you're right. I thought that was done to other libraries as well, but it seems I was mistaken. @lawruble13, do you remember?

qjojo (approved):

tfalders (approved):

cgmb:
> If we will follow the structure of other libraries (like rocBLAS or even our own hipSOLVER), why still keep the `rocsolver` directory?
> Main header should be `rocsolver/library/include/rocsolver.h`, and the internal headers should be at `rocsolver/library/include/internal/`

Can we defer that until after `BUILD_FILE_REORG_BACKWARD_COMPATIBILITY` is removed? It's going to take some investigation to determine which method is correct, and the backwards-compatibility stuff more than doubles the complexity of that work.

Moving the headers around in our source tree doesn't affect the public API, so we can do that at any time. It won't have to be in ROCm 6.0.

jzuniga-amd (approved):

PR #509, opened 2023-01-06
workaround for hetrd

jzuniga-amd:
This PR adds a workaround to hetd2/hetrd to match culsolver output when the input matrix is not Hermitian. This is not a fix for a defect; there is really no correct tridiagonal form associated to a non-Hermitian input, but cusolver gets rid to the imaginary part in the diagonal elements, and for some reason, there is a request for rocsolver to do the same.

This addresses JIRA SWDEV-374304


cgmb (commented):
Ideally, we would have tests for any behaviours that are important to our users. Is it reasonably straightforward to add a check to our unit tests?

Given that this is mathematically unnecessary, it might also be a good idea to put the reason why this is being done as a comment. If you were to order various places where information might be recorded by the likelihood that it will still be accessible in twenty years, I think it would probably look something like:
1. code comments
2. commit messages
3. GitHub
4. Jira
5. Confluence

tfalders (approved):

cgmb:
If a test would be a pain then don't worry about it.

I'd be happy with just adding a brief comment to the implementation that basically says the same thing as your PR description.

jzuniga-amd:
This cannot be tested. We need to modify the generation of the input matrix so that it is not hermitian. Even then, our current implicit test will fail as there is no way to reconstruct the original non-hermitian A from a real tridiagonal form. So, we would need to write an entirely different unit test for this non-sense case.

cgmb (approved):

PR #510, opened 2023-01-11
Enable [cz]lauum

qjojo:
This changes lauum to use syrk/herk as the kernel to provide better performance and accuracy than trmm. Also enables support for complex numbers.


cgmb (dismissed):
You've partially updated the clients, but the tests still need to be updated to run these.

tfalders (dismissed):
CI is failing for the real upper triangular cases.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -52,12 +52,12 @@ template <typename T, typename U>
...
                                         const rocblas_fill uplo,
                                         const rocblas_int n,
-                                        U* A,
+                                        U A,
                                         const rocblas_int shiftA,
                                         const rocblas_int lda,
                                         const rocblas_stride strideA,
                                         const rocblas_int batch_count,
-                                        U* work)
+                                        U work)
```
tfalders, line 60: It's better for `work` to be a `T*` or a `void*`. Auxiliary functions don't have batch versions, but they can be called by batched functions behind the scenes. If `U` is `T* const*`, then we'll have to set up `work` as a batched array, which is a pain.

jzuniga-amd: Yes, this is important and needs to be changed.

clients/common/testing_lauum.cpp
```diff
@@ -6,4 +6,4 @@
 
 #define TESTING_LAUUM(...) template void testing_lauum<__VA_ARGS__>(Arguments&);
 
-INSTANTIATE(TESTING_LAUUM, FOREACH_REAL_TYPE, APPLY_STAMP)
+INSTANTIATE(TESTING_LAUUM, FOREACH_SCALAR_TYPE, APPLY_STAMP)
```
jzuniga-amd, line 9: You also need to put `lauum` in the map with no limited precision in 
`clients/include/rocsolver_dispatcher.hpp`. Otherwise you wont be able to run `lauum` with complex precision with the rocsolver_bench client.

jzuniga-amd (dismissed):

clients/common/testing_lauum.cpp
jzuniga-amd: complex tests are still failing everywhere. I will have a look.

jzuniga-amd:
the complex tests are still failing... I will have a look.

qjojo:
Juan determined that the CPU implementation of lauum is incorrect, so I changed the test cases to evaluate it directly instead. All tests should pass now.

EdDAzevedo:
There may be an implicit assumption that the matrix A is Hermitian where conceptually   conj(transpose(A)) == A, even though only the upper or lower triangular part is accessed. Thus there might be an implicit assumption that the diagonal entries of matrix are also  "real" numbers (imaginary part is exactly zero).   So another work around is to set the diagonal entries of the matrix to be real numbers (not complex numbers or to have the imaginary part of the diagonal entries to be exactly zero).   Just a thought.

clients/include/testing_lauum.hpp
```diff
@@ -71,7 +76,7 @@ void lauum_initData(const rocblas_handle handle,
     }
 }
 
-template <typename T, typename Td, typename Th>
+template <typename T, typename Td, typename Th, bool COMPLEX = rocblas_is_complex<T>>
```
jzuniga-amd, line 79: not really important, but the macro COMPLEX is not needed anymore.

library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -52,12 +52,12 @@ template <typename T, typename U>
...
                                         const rocblas_fill uplo,
                                         const rocblas_int n,
-                                        U* A,
+                                        U A,
                                         const rocblas_int shiftA,
                                         const rocblas_int lda,
                                         const rocblas_stride strideA,
                                         const rocblas_int batch_count,
-                                        U* work)
+                                        void* work)
```
jzuniga-amd, line 60: Just for consistency, in the .cpp file you better call `rocsolver_lauum_template` with `(T*)work`, and I would define `work` here as `T*` instead of `void` and casting every time I use it in the body of the function. 

In general, the way we use the typenames across the library is:
`T`: defines the base precision (float, double, float_complex or double_complex). Normally passed explicitly when calling the template functions.
`U`: could be either a pointer to `T` (`U = T*`) for the normal and strided_batched cases, or an array of pointers to `T` (`U = T* const []`) for the batched cases. `U` is normally inferred. 
S: some functions mix real and complex arrays. In this cases, `S` defines the base real precision (ie. `S` is float when `T` is float or float_complex, for example). `S` is normally inferred.

jzuniga-amd:
Also don't forget to update the documentation and changelog to reflect that LAUUM supports complex now.

jzuniga-amd:
And we should be more explicit in the documentation with the fact that the triangular factors in the computations are expected to come from a symmetric/hermitian matrix.

jzuniga-amd (changes requested):
Please look at PR in your branch for documentation updates.

jzuniga-amd (approved):
LGTM

cgmb (approved):

clients/common/lapack_host_reference.cpp
```diff
@@ -905,6 +905,21 @@ void zgelqf_(int* m,
...
+             int* lda,
+             rocblas_float_complex* B,
+             int* ldb);
+void zlacpy_(char* uplo,
+             int* n,
+             int* m,
+             rocblas_double_complex* A,
+             int* lda,
+             rocblas_double_complex* B,
+             int* ldb);
```
tfalders, line 921: Minor issue, but m should come before n, right?

clients/common/lapack_host_reference.cpp
```diff
@@ -2502,6 +2517,33 @@ void cpu_lacgv<rocblas_double_complex>(rocblas_int n, rocblas_double_complex* x,
...
+                                       rocblas_int n,
+                                       rocblas_int m,
+                                       rocblas_double_complex* A,
+                                       rocblas_int lda,
+                                       rocblas_double_complex* B,
+                                       rocblas_int ldb)
+{
+    char uploC = rocblas2char_fill(uplo);
+    zlacpy_(&uploC, &n, &m, A, &lda, B, &ldb);
+}
```
tfalders, line 2545: Same as above

clients/include/lapack_host_reference.hpp
```diff
@@ -339,6 +339,15 @@ void cpu_gelqf(rocblas_int m, rocblas_int n, T* A, rocblas_int lda, T* ipiv, T*
...
 void cpu_lacgv(rocblas_int n, T* x, rocblas_int incx);
 
+template <typename T>
+void cpu_lacpy(rocblas_fill uplo,
+               rocblas_int n,
+               rocblas_int m,
+               T* A,
+               rocblas_int lda,
+               T* B,
+               rocblas_int ldb);
```
tfalders, line 349: Same as above

tfalders (approved):
Looks good! Very minor issue to correct.

qjojo:
The CI failures are unrelated to the changes in this PR, I will force squash and merge.

PR #511, opened 2023-01-11
Remove deprecated functions and types

cgmb:
We'll be bumping the API version with ROCM 6.0, so we can now drop the deprecated functions and types.


library/CMakeLists.txt
```diff
@@ -8,16 +8,4 @@ set(rocsolver_SOVERSION 0.1)
...
-  include/rocsolver/rocsolver.h
-  include/rocsolver/internal/rocsolver-extra-types.h
-  include/rocsolver/internal/rocsolver-aliases.h
-  include/rocsolver/internal/rocsolver-functions.h
-  ${PROJECT_BINARY_DIR}/include/rocsolver/internal/rocsolver-version.h
-  ${PROJECT_BINARY_DIR}/include/rocsolver/internal/rocsolver-export.h
-)
-
-# For IDE integration
-source_group("Header Files\\Public" FILES ${rocsolver_headers_public})
```
tfalders, line 21: Out of curiosity, why are these being removed?

tfalders (approved):
Looks good! I quite enjoy seeing unnecessary code get the boot.

library/CMakeLists.txt
cgmb: That code exists because it improves the organization of the rocSOLVER sources when the project is opened in certain IDEs that have CMake integration. It doesn't actually have anything to do with building the project and I got tired of updating it when I don't think anyone working on rocSOLVER uses such an IDE.

jzuniga-amd (approved):

PR #513, opened 2023-01-12
Fix reserved identifiers in include guards

cgmb:
The include guards have been changed to the filename in uppercase letters, with all non-alphanumeric symbols replaced by underscore. This include guard pattern matches the guard that is used for the generated file `rocsolver-export.h`.

There are two reasons for this:
1. The C and C++ standards reserve all identifiers that begin with underscore followed by a capital letter [C99 §7.1.3] [C++11 §17.6.4.3.2].
2. In user-visible code, the rocSOLVER library should prevent name conflicts by only using identifiers that begin with `rocblas`, `ROCBLAS`, `rocsolver` or `ROCSOLVER`.

I ignored `rocsolver-aliases.h` to prevent conflicts with #511.

Ticket: SWDEV-343846

### DIFF
--- a/library/include/rocsolver/internal/rocsolver-extra-types.h
+++ b/library/include/rocsolver/internal/rocsolver-extra-types.h
@@ -1,9 +1,9 @@
 /* ************************************************************************
- * Copyright (c) 2019-2022 Advanced Micro Devices, Inc.
+ * Copyright (c) 2019-2023 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
-#ifndef ROCSOLVER_EXTRAS_H_
-#define ROCSOLVER_EXTRAS_H_
+#ifndef ROCSOLVER_EXTRA_TYPES_H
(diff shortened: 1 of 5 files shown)


library/include/rocsolver/internal/rocsolver-extra-types.h
```diff
@@ -121,4 +121,4 @@ typedef enum rocblas_srange_
     rocblas_srange_index = 263, /**< The \f$il\f$-th through \f$iu\f$-th singular values will be found.*/
 } rocblas_srange;
 
-#endif /* ROCSOLVER_EXTRAS_H_ */
+#endif /* ROCSOLVER_EXTRAS_H */
```
tfalders, line 124: `ROCSOLVER_EXTRA_TYPES_H`?

tfalders (approved):
LGTM!

jzuniga-amd (approved):

Issue #514, opened 2023-01-13
Change upper bound for fmt dependency on rocsolver spack package


aurianer:
Is the upper bound of the fmt dependency of the rocsolver spack package here for a specific reason or just because the other versions are not tested? https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/rocsolver/package.py#L125
I would need fmt@9 for my spack build-env, and it seems to work good when I remove this upper bound, I didn't test rocsolver exhaustively though.

cgmb:
The constraint on the rocsolver spack packages was added because fmt 9.0 and 9.1 had issues with hip and sycl [1]. There is a fix on the fmt development branch [2]. Spack has backported the fix to fmt 9.0 and 9.1 [3], so I believe that constraint could be removed from the rocsolver Spack package.

If you'd like to open a PR on spack/spack that removes the upper bound, I'll approve it.

[1] https://github.com/fmtlib/fmt/issues/3005
[2] https://github.com/fmtlib/fmt/commit/0b0f7cfbfcebd021c910078003d413354bd843e2
[3] https://github.com/spack/spack/pull/33733

cgmb:
Oh, I forgot that older versions of rocSOLVER do have a legitimate constraint. We added support for `fmt@8.1:` in `rocsolver@5.2`. We might also need to backport [the patch](https://github.com/ROCmSoftwarePlatform/rocSOLVER/commit/2bbfb8976f6e4d667499c77e41a6433850063e88) to entirely eliminate constraints on older versions of rocsolver.

PR #515, opened 2023-01-14
Fix use of fmt 9.0 and later in logging test

cgmb:
The implicitly defined formatter that rocsolver was using for `std::filesystem::path` has been removed from fmt 9.0 and later.

This change doesn't actually fix compatibility with the official fmt 9.0 or 9.1 releases, because those releases are still incompatible with HIP. However, it is sufficient for using rocsolver with fmt@9 from spack (because the fix for the HIP incompatibility has been backported into fmt@9 by the Spack package maintainer).


cgmb:
This partially addresses #514.

tfalders (approved):

jzuniga-amd (approved):

PR #508, opened 2023-01-05
Revert "Remove rpath/runpath from ROCm binaries and libraries (#483)"

cgmb:
SWDEV-310152 has been deferred to a later release. It would be worse to have only some of the libraries changed in ROCm 5.5 than to have them all configured one way or the other, so we will undo the changes made to rocSOLVER for this release.

This reverts commit f62bb7f76a0c671f8c6b7463f41831353eafffa5.

PR #516, opened 2023-01-19
Hotfix call to set_zero kernel in lauum

tfalders:
Testing on new hardware has revealed a bug in LAUUM where the part of the workspace matrix that is supposed to be zero is instead filled with uninitialized garbage values. This is because the `set_zero` kernel, confusingly, will zero out the part of the matrix that is NOT specified by `uplo` (the part specified by `uplo` will remain unchanged). This could be the culprit for why the complex versions of the function failed their tests.

I'm opening this PR as a hotfix for 5.5. Because of https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/508 and https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/510, I'm not sure if this will get merged back into develop, so I haven't altered the `set_zero` kernel. I suggest we do so on the develop branch for the sake of clarity and to avoid this kind of bug in the future.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_lauum.hpp
+++ b/library/src/auxiliary/rocauxiliary_lauum.hpp
@@ -69,6 +69,11 @@ rocblas_status rocsolver_lauum_template(rocblas_handle handle,
     hipStream_t stream;
     rocblas_get_stream(handle, &stream);
 
+    // everything must be executed with scalars on the host
+    rocblas_pointer_mode old_mode;
+    rocblas_get_pointer_mode(handle, &old_mode);
+    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+
     rocblas_int strideW = n * n;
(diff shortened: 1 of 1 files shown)


library/src/auxiliary/rocauxiliary_lauum.hpp
```diff
@@ -76,10 +76,8 @@ rocblas_status rocsolver_lauum_template(rocblas_handle handle,
     T one = 1;
     T zero = 0;
```
cgmb, line 77: Do we need to set `rocblas_pointer_mode_host` for calling TRMM with host scalars?

tfalders: Good point! I'll add that tomorrow.

tfalders: Oops, almost forgot lol

jzuniga-amd:
Closing in favor of #520

PR #517, opened 2023-01-20
Improve error reporting for zero-tolerance test failures

cgmb:
The rocsolver tests report certain error conditions (like unexpected info values) by adding a positive integer to the measured error and thereby guaranteeing that error exceeds tolerance. This can make it difficult to understand what is wrong when reviewing the logs for a failed set of tests (especially when there are multiple properties that could all increment the error).
    
The Google Test `EXPECT` functions will check a condition, but allow the test to continue. This means that an error can be recorded at the place where it occurs without losing data. The macros are defined to do nothing for rocsolver-bench, so this additional error output is only emitted in rocsolver-test.

### DIFF
--- a/clients/include/testing_gesvdx.hpp
+++ b/clients/include/testing_gesvdx.hpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (c) 2022 Advanced Micro Devices, Inc.
+ * Copyright (c) 2022-2023 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
 #pragma once
@@ -447,6 +447,7 @@ void gesvdx_getError(const rocblas_handle handle,
     double err = 0;
     for(rocblas_int b = 0; b < bc; ++b)
     {
+        EXPECT_EQ(hNsv[b][0], hNsvRes[b][0]) << "where b = " << b;
         if(hNsv[b][0] != hNsvRes[b][0])
             err++;
     }
--- a/clients/include/testing_gesvdx.hpp
+++ b/clients/include/testing_gesvdx.hpp
@@ -434,10 +434,12 @@ void gesvdx_getError(const rocblas_handle handle,
     //  meaning in gesvd_, however, We expect the used input matrices to always converge)
     /*for(rocblas_int b = 0; b < bc; ++b)
     {
+        EXPECT_EQ(hinfo[b][0], hinfoRes[b][0]) << "where b = " << b;
         if(hinfo[b][0] != hinfoRes[b][0])
             *max_err += 1;
         for(int j = 0; j < hNsv[b][0]; ++j)
         {
+            EXPECT_EQ(hifail[b][j], hifailRes[b][j]) << "where b = " << b << ", j = " << j;
             if(hifail[b][j] != hifailRes[b][j])
                 *max_err += 1;
         }
--- a/clients/include/testing_stein.hpp
+++ b/clients/include/testing_stein.hpp
@@ -274,7 +274,7 @@ void stein_getError(const rocblas_handle handle,
         err = 0;
         for(int j = 0; j < hInfo[0][0]; j++)
         {
-            EXPECT_EQ(hIfailRes[0][j], 0) << "j = " << j;
+            EXPECT_NE(hIfailRes[0][j], 0) << "j = " << j;
             if(hIfailRes[0][j] == 0)
                 err++;
         }
--- a/clients/include/testing_stein.hpp
+++ b/clients/include/testing_stein.hpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (c) 2020-2022 Advanced Micro Devices, Inc.
+ * Copyright (c) 2020-2023 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
 #pragma once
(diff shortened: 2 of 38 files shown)


jzuniga-amd (commented):
Looks good to me, I just wonder why not all functions that report `info` are getting similar assertions. For example, getf2/getfr, gels, getri, stedc, steqr, potf2/potrf, and many others also return some type of `info` parameter.

cgmb:
I found these checks by searching for `err += 1`, so I may have missed info checks that didn't use that pattern. I'll expand it to include those as well.

cgmb:
This is some example output in a case where I've forced a failure by setting all elements of hInfoRes to 125:
```
[ RUN      ] checkin_lapack/TRTRI.strided_batched__double_complex/17
/home/cgmb/rocSOLVER/clients/include/testing_trtri.hpp:193: Failure
  hInfo[b][0]
    Which is: 0
  hInfoRes[b][0]
    Which is: 125
where b = 0
/home/cgmb/rocSOLVER/clients/include/testing_trtri.hpp:193: Failure
  hInfo[b][0]
    Which is: 0
  hInfoRes[b][0]
    Which is: 125
where b = 1
/home/cgmb/rocSOLVER/clients/include/testing_trtri.hpp:193: Failure
  hInfo[b][0]
    Which is: 0
  hInfoRes[b][0]
    Which is: 125
where b = 2
/home/cgmb/rocSOLVER/clients/include/testing_trtri.hpp:442: Failure
Expected: ((max_error)) <= ((n)*get_epsilon<T>()), actual: 3 vs 2.22045e-14
```

cgmb:
Added checks for all functions that use `err++` or `err = 1`. I think this is once again ready for review, so please take another look when you have a moment.

cgmb:
For a more concrete example, these are the current Windows failures we've been seeing in PRs for the past while:
```
[ RUN      ] checkin_lapack/BDSVDX.__double/139
C:/workspace/clients/include\testing_bdsvdx.hpp(225): error: Expected equality of these values:
hNsv[0][0]
Which is: 1
hNsvRes[0][0]
Which is: 0
C:/workspace/clients/include\testing_bdsvdx.hpp(528): error: Expected: ((max_error)) <= ((n)*get_epsilon<T>()), actual: 1 vs 2.22045e-16
[  FAILED  ] checkin_lapack/BDSVDX.__double/139, where GetParam() = (L, { 1, 2, 1 }, { 1, 5, 15, 0, 0 }) (9 ms)
[ RUN      ] checkin_lapack/BDSVDX.__double/140
C:/workspace/clients/include\testing_bdsvdx.hpp(225): error: Expected equality of these values:
hNsv[0][0]
Which is: 1
hNsvRes[0][0]
Which is: 0
C:/workspace/clients/include\testing_bdsvdx.hpp(528): error: Expected: ((max_error)) <= ((n)*get_epsilon<T>()), actual: 1 vs 2.22045e-16
[  FAILED  ] checkin_lapack/BDSVDX.__double/140, where GetParam() = (L, { 1, 2, 1 }, { 1, 0, 15, 0, 0 }) (8 ms)
[ RUN      ] checkin_lapack/BDSVDX.__double/141
C:/workspace/clients/include\testing_bdsvdx.hpp(225): error: Expected equality of these values:
hNsv[0][0]
Which is: 0
hNsvRes[0][0]
Which is: 1
C:/workspace/clients/include\testing_bdsvdx.hpp(528): error: Expected: ((max_error)) <= ((n)*get_epsilon<T>()), actual: 2 vs 2.22045e-16
[  FAILED  ] checkin_lapack/BDSVDX.__double/141, where GetParam() = (L, { 1, 2, 1 }, { 1, 15, 20, 0, 0 }) (8 ms)
```

clients/include/testing_gesvdx.hpp
```diff
@@ -447,6 +447,7 @@ void gesvdx_getError(const rocblas_handle handle,
     double err = 0;
     for(rocblas_int b = 0; b < bc; ++b)
     {
+        EXPECT_EQ(hNsv[b][0], hNsvRes[b][0]) << "where b = " << b;
```
jzuniga-amd, line 450: The code is commented until we sort out the problem with reference lapack, but if you add the assertions for `ifail` and `info` as well, that would be fine.

jzuniga-amd (approved):
LGTM. Nice job.

cgmb:
@tfalders, @qjojo: could you please review these changes?

clients/include/testing_stein.hpp
```diff
@@ -272,6 +274,7 @@ void stein_getError(const rocblas_handle handle,
         err = 0;
         for(int j = 0; j < hInfo[0][0]; j++)
         {
+            EXPECT_EQ(hIfailRes[0][j], 0) << "j = " << j;
             if(hIfailRes[0][j] == 0)
                 err++;
```
tfalders, line 279: Here, we're expecting `hIfailRes[0][j]` to be non-zero. It should be `EXPECT_NE`.

tfalders (changes requested):
Looks good! Just one minor change is needed.

tfalders (approved):
LGTM. Thanks for adding this!

clients/include/testing_stein.hpp
cgmb: Thanks. I missed that these checks were backwards. It does concern me a little that the tests were still passing. That implies that this code isn't executed. In any case, I've fixed the hIfail checks.

tfalders: This code is only executed if the algorithm doesn't converge, which is a harder situation to create than introducing a singularity into a matrix. I don't think any of our tests right now try to force an algorithm to not converge.

PR #518, opened 2023-01-24
Add AOCL BLAS/LAPACK find modules

cgmb:
These are the FindBLAS and FindLAPACK from CMake 3.25.2, with a few modifications to support finding AOCL BLIS and libFLAME (including the multi-threaded variant, which the upstream FindBLAS is unable to find).

They are added to rocSOLVER in the `cmake/Modules` subfolder, so that one can choose to build rocsolver-test or rocsolver-bench with AOCL as the CPU reference library using the CMake arguments

    -DROCSOLVER_FIND_PACKAGE_LAPACK_CONFIG=ON \
    -DCMAKE_MODULE_PATH=<srcdir>/cmake/Modules \
    -DBLA_VENDOR=AOCL (or -DBLA_VENDOR=AOCL_mt)

These custom find modules are not automatically added to the `CMAKE_MODULE_PATH`. This ensures that they are entirely optional. One can imagine that users with future versions of CMake may still want to be able to use their built-in find modules, rather than being stuck using the version from 3.25.2 (even with our improvements).

### DIFF
--- a/cmake/Modules/FindBLAS.cmake
+++ b/cmake/Modules/FindBLAS.cmake
@@ -122,7 +122,6 @@ BLAS/LAPACK Vendors
   AMD Core Math Library
 
 ``AOCL``, ``AOCL_mt``
-  .. versionadded:: 3.26
   AMD Optimizing CPU Libraries
 
 ``Apple``, ``NAS``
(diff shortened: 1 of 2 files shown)


doctorcolinsmith:
The changes look fine, but it is not clear to me whether the failures are related.

jzuniga-amd (approved):
LGTM

PR #519, opened 2023-01-25
Replace host_batch_vector and host_strided_batch_vector

cgmb:
This is essentially a rewrite of the `host_batch_vector` and `host_strided_batch_vector` classes. I've deleted a number of unused options, enabled error checking on failed allocations by default, and added a number of assertions to catch errors.

The diff is probably not useful for this change. It's better to just review the final result. Or, at least, it will be when I'm done. This is still just a draft.


tfalders:
Closing for now

PR #523, opened 2023-02-08
Improved bdsqr performance

tfalders:
This PR adds parallelization to bdsqr and contains various other fixes and improvements.

Some important changes:
- Long-standing failures when calling rocsolver-bench on bdsqr have been resolved.
- The tests for bdsqr have been updated to be more like the tests for gesvd, where vectors needed for the implicit tests are computed in a complementary call to bdsqr.
- All bdsqr kernel names are now prefixed with bdsqr_ in order to avoid name conflicts.
- Calls to bdsqr from gesvd no longer need to use `workArr` in order to mix batched and strided-batched inputs.
- A newly-added kernel (`bdsqr_init`) will check for NaNs or Infs in the input and prevent further processing if they are found. It also does other pre-processing, mainly splitting the matrix into diagonal blocks that can be processed in parallel.
- Sorting the singular values/vectors, ensuring the singular values are positive, and finalizing `info` have been moved out of `bdsqr_kernel` into `bdsqr_sort`.
- `bdsqr_kernel` has been restructured and parallelized. Threads in x handle singular vector updates, thread groups in y handle split diagonal blocks, and thread groups in z handle batch instances.


qjojo (approved):

clients/include/testing_bdsqr.hpp
```diff
@@ -144,14 +145,14 @@ void bdsqr_initData(const rocblas_handle handle,
...
         {
-            for(rocblas_int i = 0; i < nv - 1; ++i)
+            for(rocblas_int i = 0; i < n - 1; ++i)
             {
                 E[i] = hE[0][i];
                 D[i] = hD[0][i];
             }
-            D[nv - 1] = hD[0][nv - 1];
+            D[n - 1] = hD[0][n - 1];
         }
```
cgmb, line 156: This feels a bit weird to me, but it's probably fine. Passing `test` just makes for an inconsistent abstraction, as it's telling the callee about the caller's context rather than telling the function what to do. The abstraction provided by the init function was pretty minimal anyway, so it's not really terribly important.

clients/include/testing_bdsqr.hpp
```diff
@@ -220,16 +225,33 @@ void bdsqr_getError(const rocblas_handle handle,
...
+    bdsqr_initData<true, false, T>(handle, uplo, n, nvRes, nuRes, nc, dD, dE, dV, ldv, dU, ldu, dC,
+                                   ldc, dInfo, hD, hE, hV, hU, hC, hInfo, D, E, true);
+
+    // execute computations:
+    // complementary execution to compute all singular vectors if needed
+    if(nvT > 0 || nuT > 0)
+    {
+        // send data to GPU
+        bdsqr_initData<false, true, T>(handle, uplo, n, nvT, nuT, 0, dD, dE, dV, ldv, dU, ldu, dC,
+                                       ldc, dInfo, hD, hE, hV, hU, hC, hInfo, D, E, false);
```
cgmb, line 241: If it were me, I would be introducing an enum or constant for this new flag rather than using `true` and `false`. Up to you, though.

cgmb (approved):
Minor comments. Nothing blocking.

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -261,221 +343,469 @@ ROCSOLVER_KERNEL void bdsqrKernel(const rocblas_int n,
...
+
+    // update output
+    if(found)
+    {
+        for(rocblas_int i = 0; i < n - 1; i++)
+        {
+            D[i] = nan("");
+            E[i] = nan("");
+        }
+        D[n - 1] = nan("");
```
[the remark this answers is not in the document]
tfalders: According to https://en.cppreference.com/w/cpp/numeric/math/NAN and https://en.cppreference.com/w/cpp/numeric/math/nan, the `NAN` macro is undefined if quiet NaNs are not supported, whereas `nan("")` will default to zero. I'm not sure we'll ever run into a situation where quiet NaN is not supported, but that might be why rocBLAS uses this function.

clients/include/testing_bdsqr.hpp
tfalders: This is another one of those things that I copied from elsewhere in the code and would prefer to change in a separate PR so that we can catch all instances.

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ## (Unreleased) rocSOLVER
 ### Added
 ### Optimized
+- Improved the performance of BDSQR and GESVD
```
jzuniga-amd, line 8: Please indicate that the the vectors must be computed to see the improvements

jzuniga-amd (approved):
Just a couple of things for the backlog:
The rocsolver-bench is still failing with GESVD for some combinations of parameters; it would be good to review the test code and apply some of the changes implemented here for BDSQR. Also, we should parallelize our interal LASR in general; we already have 2 or 3 parallel versions of the application of a series of rotations here and there... better to have a centralized optimized routine for the same functionality. 

But this PR looks good to me, good job!

PR #524, opened 2023-02-09
Revert "Move internal headers to a subdirectory (#507)"

cgmb:
Some libraries depend directly on rocsolver-version.h. Moving the header to internal/ subdirectory as was done in 4700f87f6472 would be a breaking change. While the rocsolver-version.h header was not intended for external use, it is simple and may be useful for some users. This change ~partially~ reverts #507, ~explicitly supporting the direct inclusion of <rocsolver/rocsolver-version.h> by library users~.


cgmb:
I'm also fine with entirely reverting #507 if that's preferred. My goal is just to make it clear what parts of the rocsolver API are supported parts of the public interface and which parts are not. If we're willing to support users that directly include `rocsolver-export.h`,`rocsolver-functions.h`, `rocsolver-extra-types.h`, `rocsolver-aliases.h`, etc., I can put those back too.

cgmb:
Actually, let's just make this simple. I've reverted #507 for this release. We can figure out which headers are internal details and which are for direct inclusion in a future release.

jzuniga-amd (approved):

cgmb:
Windows build is hanging forever and linux build failed with "No space left on device". I'm just going to merge.

PR #525, opened 2023-02-14
Updating syr2k/her2k block sizes from rocblas

daineAMD:
For https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1647

### DIFF
--- a/library/src/include/rocblas.hpp
+++ b/library/src/include/rocblas.hpp
@@ -1196,7 +1196,7 @@ rocblas_status rocblasCall_syr2k_her2k(rocblas_handle handle,
         : std::is_same<T, float>::value ? ROCBLAS_SSYR2K_NB
         : std::is_same<T, double>::value ? ROCBLAS_DSYR2K_NB
         : std::is_same<T, rocblas_float_complex>::value ? ROCBLAS_CSYR2K_NB
-                                        : ROCBLAS_ ZSYR2K_NB;
+                                        : ROCBLAS_ZSYR2K_NB;
 
     return rocblas_internal_syr2k_her2k_template<NB, BATCHED, TWOK, HERK>(
         handle, uplo, trans, n, k, cast2constType<T>(alpha), cast2constType<T>(A), offsetA, lda,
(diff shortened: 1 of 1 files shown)


library/src/include/rocblas.hpp
```diff
@@ -1192,7 +1194,9 @@ rocblas_status rocblasCall_syr2k_her2k(rocblas_handle handle,
     constexpr bool HERK = false;
     constexpr rocblas_int NB = BATCHED  ? ROCBLAS_SDSYR2K_BATCHED_NB
         : std::is_same<T, float>::value ? ROCBLAS_SSYR2K_NB
-                                        : ROCBLAS_DCZSYR2K_NB;
+        : std::is_same<T, double>::value ? ROCBLAS_DSYR2K_NB
+        : std::is_same<T, rocblas_float_complex>::value ? ROCBLAS_CSYR2K_NB
+                                        : ROCBLAS_ ZSYR2K_NB;
 
     return rocblas_internal_syr2k_her2k_template<NB, BATCHED, TWOK, HERK>(
```
cgmb, line 1201: I feel somewhat obligated to bring this up until we actually get to fixing it: there is no reason for `NB` to be part of the exported symbol name.

It would be possible to define an ABI that is stable over these changes by employing the [hourglass pattern](https://youtu.be/PVYdHDm0q6Y). All that's required to create a stable ABI is to define [an alias](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2729.htm#inspiration-existing-practice) for each explicit instantiation. We can then define a template function in rocSOLVER within rocblas.hpp that calls the appropriate alias for each type `T`.

e.g. some of the aliases for these functions might be:
```
rocblas_internal_ssyr2k_twok
rocblas_internal_dsyr2k_twok
rocblas_internal_ssyr2k_twok_batched
rocblas_internal_dsyr2k_twok_batched
```

Then there won't be any need to change rocSOLVER (or even rebuild it) when the block size changes in rocBLAS. It's not like rocSOLVER can call these functions with different block sizes anyway, so we don't lose anything by dropping it from the ABI.

If that was done consistently, there'd be no need to make any of these update PRs again, and we wouldn't have breakages due to mismatched rocBLAS and rocSOLVER versions.

daineAMD: Fair enough, I created [LWPMLSE-371](https://ontrack-internal.amd.com/browse/LWPMLSE-371) to make an example of what these changes would look like to discuss with our team.

cgmb: Thanks Daine. That's perfect.

cgmb (approved):

tfalders (approved):
Thanks @daineAMD!

daineAMD:
rocBLAS PR will probably go through tonight, will try to remember to check tomorrow so develop isn't broken for long.

PR #526, opened 2023-03-01
sparse LU and direct solver APIs (re-factorization)

jzuniga-amd:


library/include/rocsolver/internal/rocsolver-extra-types.h
```diff
@@ -121,4 +122,19 @@ typedef enum rocblas_srange_
...
+typedef struct rocsparse_rfinfo_
+{
+    rocsparse_handle sphandle;
+    rocsparse_mat_descr descrL;
+    rocsparse_mat_descr descrU;
+    rocsparse_mat_descr descrT;
+    rocsparse_mat_info infoL;
+    rocsparse_mat_info infoU;
+    rocsparse_mat_info infoT;
+} rocsparse_rfinfo;
```
cgmb, line 138: The size of the structure becomes part of ABI, so we may want to adjust this definition so we can more easily change the structure in the future. Though, admittedly, I'm not sure of the best way of doing that. One possibility is that we could hide the size by using an opaque pointer like we do with handle, but I think there are probably other possibilities as well.

EdDAzevedo: Should it be called rocsolver_rfinfo versus  rocsparse_rfinfo? I have a concern that there may be a confusion that this belongs to rocsparse library. Perhaps there can be  "rocsolver_rfinfo_t" pointer to the struct and exposed as "void *" to the user application?    

The cusolver interface assumes their handle hold other data related to the (batched) LU factors so if this is not contained in the rocsolver software, then the data of LU factor (and related logic code to allocate and deallocate data) has to be held in the hipsolver layer. I find using the header-only rocthrust library can greatly simplify the allocation/deallocation/copy  of host arrays and device arrays.

We may need to clarify what are the return status type and status codes. The status type and status codes may need to be future proof to accommodate the cusolverSP library as well.    One may consider taking the nvidia cusolver header file and change "cu" to "hip" and "cu" to "roc".

qjojo: > I think there are probably other possibilities as well.

If we don't want to make an opaque pointer type, the way Vulkan handles this is to add a void* pNext at the end of the structure that we can use to point to another structure with more data. If don't want a void* lurking around, we could add a template parameter for the pointer and then typedef it away for the public API. Personally, I think I prefer the opaque pointer.

cgmb: > Should it be called rocsolver_rfinfo versus rocsparse_rfinfo? I have a concern that there may be a confusion that this belongs to rocsparse library. Perhaps there can be "rocsolver_rfinfo_t" pointer to the struct and exposed as "void *" to the user application?

Agreed.

library/include/rocsolver/internal/rocsolver-functions.h
```diff
@@ -23725,6 +23725,640 @@ ROCSOLVER_EXPORT rocblas_status
...
+    \details
+    @param[out]
+    rfinfo      #rocsolver_rfinfo.\n
+                The pointer to the rfinfo struct to be initialized.
+    @param[in]
+    handle      rocblas_handle.\n
+    ********************************************************************/
+
+ROCSOLVER_EXPORT rocblas_status rocsolver_rfinfo_create(rocsolver_rfinfo* rfinfo,
+                                                        rocblas_handle handle);
```
tfalders, line 23747: It might be nice to have these be `rocsolver_create_rfinfo` and `rocsolver_destroy_rfinfo` to be consistent with `rocblas_create_handle` and `rocblas_destroy_handle`.

clients/include/rocsolver_test.hpp
```diff
@@ -13,6 +13,9 @@
 #include <fmt/ostream.h>
 #include <fmt/ranges.h>
 
+// Location of the sparse data directory for the re-factorization tests
+#define SPARSEDATA_DIR "/home/jcarzu/Development/rocSOLVER/sparse/clients/sparsedata"
```
tfalders, line 17: We will definitely need to change this to be general for other users and build layouts. Perhaps the best way is to copy the data to the executable clients folder as part of the build. I'm sure @cgmb would have a better idea of how to do this than I.

tfalders: Also, we should probably abort the test if a file is not found. Right now, the tests on my machine are giving memory access faults because they can't find the files.

clients/gtest/logging_gtest.cpp
```diff
@@ -3,7 +3,7 @@
  * ************************************************************************ */
 
 #include <cstdlib>
-#if __has_include(<filesystem>)
+#if __cpp_lib_filesystem
```
cgmb, line 6: This is not correct. The `__cpp_lib_filesystem` macro is defined within `<filesystem>`.

library/src/refact/rocsolver_rfinfo.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return rocblas_status_success;
+cleanup:
+    rocsparse_destroy_mat_info(impl->infoU);
+    rocsparse_destroy_mat_info(impl->infoL);
+    rocsparse_destroy_mat_descr(impl->descrT);
+    rocsparse_destroy_mat_descr(impl->descrU);
+    rocsparse_destroy_mat_descr(impl->descrL);
+    rocsparse_destroy_handle(impl->sphandle);
+    free(impl);
+    return result;
```
cgmb, line 96: The code that sets the result was lost when the macros were changed.

library/src/refact/rocsolver_rfinfo.cpp
```diff
@@ -0,0 +1,121 @@
...
+    GOTO_IF_ROCSPARSE_ERROR(rocsparse_create_mat_info(&impl->infoL), cleanup);
+    GOTO_IF_ROCSPARSE_ERROR(rocsparse_create_mat_info(&impl->infoU), cleanup);
+    GOTO_IF_ROCSPARSE_ERROR(rocsparse_create_mat_info(&impl->infoT), cleanup);
+
+    impl->solve_policy = rocsparse_solve_policy_auto;
+    impl->analysis_policy = rocsparse_analysis_policy_reuse;
+    *rfinfo = impl;
+    return rocblas_status_success;
+cleanup:
+    rocsparse_destroy_mat_info(impl->infoU);
```
cgmb, line 89: If we have a single cleanup, I suppose we should include `rocsparse_destroy_mat_info(impl->infoT);`. It will never be necessary to destroy it in the current code, but the single cleanup label doesn't force you to examine the previous check when you add a new check, so it may be forgotten when a new field is added.

clients/CMakeLists.txt
```diff
@@ -104,20 +110,46 @@ if(BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
...
   )
 
   prepend_path("${CMAKE_CURRENT_SOURCE_DIR}/" common_source_files common_source_paths)
   target_sources(clients-common INTERFACE ${common_source_paths})
 
+  # Copy and point to sparse test data
+  file(COPY
+    ${CMAKE_CURRENT_SOURCE_DIR}/sparsedata/
+    DESTINATION ${PROJECT_BINARY_DIR}/sparsedata/
+  )
```
cgmb, line 141: Why do we need to copy it into the build directory, anyway? Could we just use the data from the source tree? Or is this is necessary for Windows?

clients/CMakeLists.txt
```diff
@@ -104,20 +110,46 @@ if(BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
...
+  install(DIRECTORY
+    ${CMAKE_CURRENT_SOURCE_DIR}/sparsedata/
+    DESTINATION ${CMAKE_INSTALL_DATADIR}/rocsolver/test
```
cgmb, line 144: What do you think about making this `data/spmat_X_Y` rather than `sparsedata/mat_X_Y`? It would make the sparseness slightly more obvious when it's installed in `/opt/rocm/share/rocsolver/test/spmat_X_Y` without adding another subdirectory, and `data` is a [semi-standard name](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs#tld.data) for this sort of directory in the build tree.

cgmb (commented):
This is pretty much there. I have a few bikeshedding suggestions, but there's also a couple things that do need to be addressed as well. I'll directly fix the things that need to be fixed, and open a PR for the bits that are just suggestions.

clients/include/rocsolver_test.hpp
```diff
@@ -9,6 +9,14 @@
 #include <ostream>
 #include <stdexcept>
 
+#if __cpp_lib_filesystem
```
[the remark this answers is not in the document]
tfalders: I tried using `__has_include` in hipSOLVER and it failed to build on the NVIDIA machine I use for testing. I got this code from hipBLAS and figured I should update both rocSOLVER and hipSOLVER in case rocSOLVER ever runs into a similar issue.

clients/CMakeLists.txt
```diff
@@ -104,20 +110,46 @@ if(BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
...
+    COMPONENT tests
+  )
+
+  target_compile_definitions(clients-common INTERFACE
+    -DSPARSEDATA_DIR="${PROJECT_BINARY_DIR}/sparsedata/"
```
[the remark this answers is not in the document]
tfalders: That's a good point. Would you be able to put that together?

clients/CMakeLists.txt
tfalders: I figured it made more sense to copy it, so that the client always has the data on hand that it expects. If we use it from the source tree, then a previously built client might start failing if we checkout a different source branch.

clients/CMakeLists.txt
tfalders: I'd rather deal with that later

clients/include/rocsolver_test.hpp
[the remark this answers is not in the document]
cgmb: `__cpp_lib_filesystem` is defined in `<filesystem>` in C++17 and in `<version>` in C++20. I don't think you can depend on it being defined to do this check.

clients/CMakeLists.txt
cgmb: Sounds good.

qjojo (approved):

cgmb (approved):
I have mixed feelings about my own contributions to this, but you guys did a great job. There's still plenty that could be improved, but I don't see anything that must be.

tfalders (approved):

PR #531, opened 2023-03-14
Fix incorrect CMake version check

tfalders:
It looks like the last version bump accidentally changed a CMake version check as well. We should fix this for the upcoming release and possibly for 5.5 as well.


cgmb (approved):
Good catch. We could cherry-pick it into ROCm 5.5.1, but this is not important enough to go into ROCm 5.5.0. There's pretty much nobody still using CMake 3.21.0, 3.21.1 or 3.21.2.

jzuniga-amd (approved):

PR #536, opened 2023-04-03
Final rocblas interface changes

daineAMD:
Cleans up the rest of the rocBLAS internal interface. Ideally there will be no more changes needed to coordinate rocBLAS/rocSOLVER internal interface after this.

See https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1743 for rocBLAS changes. No CI for now until rocBLAS is in.

### DIFF
--- a/library/src/include/rocblas.hpp
+++ b/library/src/include/rocblas.hpp
@@ -261,12 +261,48 @@ rocblas_status rocblasCall_ger(rocblas_handle handle,
 
     if constexpr(CONJ)
         return rocblas_internal_gerc_template(
-            handle, m, n, alpha, stridea, cast2constType<T>(x), offsetx, incx, stridex,
-            cast2constType<T>(y), offsety, incy, stridey, A, offsetA, lda, strideA, batch_count);
+            handle, m, n, alpha, stridea, x, offsetx, incx, stridex,
+            y, offsety, incy, stridey, A, offsetA, lda, strideA, batch_count);
     else
         return rocblas_internal_ger_template(
-            handle, m, n, alpha, stridea, cast2constType<T>(x), offsetx, incx, stridex,
-            cast2constType<T>(y), offsety, incy, stridey, A, offsetA, lda, strideA, batch_count);
+            handle, m, n, alpha, stridea, x, offsetx, incx, stridex,
+            y, offsety, incy, stridey, A, offsetA, lda, strideA, batch_count);
+}
+
+// ger batched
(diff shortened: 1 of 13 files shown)


cgmb (commented):
~~Could you change `find_package(rocblas REQUIRED CONFIG PATHS ${ROCM_PATH})` to `find_package(rocblas 4.0 REQUIRED CONFIG PATHS ${ROCM_PATH})` in CMakeLists.txt to reflect the new version requirement? With the interface between the two libraries becoming stable, that CMake version check can now be more useful.~~

Actually, let's not add that yet, just in case the tentative rocBLAS version changes. We can make that sort of change later, when we have more certainty about the release.

library/src/include/rocblas.hpp
```diff
@@ -1492,74 +1586,39 @@ rocblas_status rocblasCall_symv_hemv(rocblas_handle handle,
...
+    else
+        return rocblas_internal_hemv_template(
+            handle, uplo, n, alpha, stridea, A, offsetA, lda,
+            strideA, x, offsetx, incx, stridex, beta, strideb, y,
+            offsety, incy, stridey, batch_count, work);
 }
 
-// hemv overload
-template <typename T, typename U, std::enable_if_t<rocblas_is_complex<T>, int> = 0>
+// symv/hemv overload - batched with strided y
```
tfalders, line 1606: I don't see a purely batched version of symv/hemv. Would you add that in?

tfalders (approved):
Looks good! You might need to run clang-format on rocblas.hpp though.

library/src/include/rocblas.hpp
daineAMD: It built and ran tests fine against my rocBLAS changes, so as far as I can tell it wouldn't be used. I can add it if you think there's a use case.

tfalders: It would be nice as a bit of future-proofing, in case we need it later and forget it's not in there yet.

daineAMD: Sure, just testing the build now.

cgmb (approved):
Thank you so much for this, Daine!

daineAMD:
I'll merge the rocBLAS PR today and the rocSOLVER one over the weekend, that's probably the least disruptive timing I'm guessing.

jzuniga-amd (approved):
LGTM. Good job Daine!

PR #549, opened 2023-05-18
Interleaved batched methods for geblttrf and geblttrs

tfalders:
This PR adds interleaved batched support to geblttrf and geblttrs (as well as getrs, getrf, getf2, and laswp behind the scenes). I've added custom kernels for gemm and ger, and adjusted the custom kernels for trsm, in order to get these working. I did not change the small-size getf2 kernels as I didn't want to affect their performance (the interleaved batched functions should not use these kernels except in the standard use case).


jzuniga-amd (commented):
Looks good to me. Good job!
My main concern is that this is a lapack-like function so, even if we interleave elements of the different matrices in the batch, I think we still need to ensure that we have column major format within a matrix. All my comments below revolve around this point:

jzuniga-amd:
Also, some tests of the logging functionality are failing. This test suite uses GETRF to generate logging outputs. Probably is not related, but as GETRF code changed, it would be good to verify if the logging test suite needs some updates.

tfalders:
> Also, some tests of the logging functionality are failing. This test suite uses GETRF to generate logging outputs. Probably is not related, but as GETRF code changed, it would be good to verify if the logging test suite needs some updates.

I added some arguments to the GETRF log. It might be due to that. I'll take a look

library/src/auxiliary/rocauxiliary_lasyf.hpp
```diff
@@ -17,6 +17,34 @@
...
+    for(int e = tid; e < m * n; e += MAX_THDS)
+    {
+        int i = e % m;
+        int j = e / m;
+        temp = 0;
+        for(int l = 0; l < k; l++)
+            temp += A[i + l * lda] * B[j + l * ldb];
+        C[i + j * ldc] = *alpha * temp + *beta * C[i + j * ldc];
+    }
+}
```
cgmb, line 46: I know this snippet is neither newly written nor intended to be truly optimized yet, but I think it is good for illustrating two principles:

1. For primitive types, it is not useful to manually hoist the declaration out of the loop. That is a trivial optimization that occurs even at -O1. See Compiler Explorer for a diff view of the compiler output after reducing the scope of temp (https://godbolt.org/z/34d37vPKd).
2. It is useful to hoist an indirect access through a pointer out of the loop. See Compiler Explorer for a diff view of the compiler output after manually hoisting the loading of alpha and beta out of the loop (https://godbolt.org/z/bYf64nPza). Note the number of loads from global memory (`global_load_dword`) dropped after the hoisting. I believe this is because `alpha`, `beta` and `C` are all pointers of the same type, and the compiler couldn't know whether `&C[i + j * ldc]` might be the address pointed to by `alpha` or `beta`. The compiler therefore had to reload `alpha` and `beta` on each iteration of the loop.

This is not terribly important given that this is obviously not an optimized GEMM kernel, but I wanted to make note of these general principles.

cgmb (approved):
LGTM.

jzuniga-amd (approved):
LGTM

tfalders:
Auto-merge enabled. I'll check back later tonight, and if it hasn't been merged yet, I'll force the merge.

PR #553, opened 2023-05-25
Fix failing Windows build

tfalders:
The Windows build for rocSOLVER appears to be failing because Windows is unable to do implicit conversions between std::string and filesystem::path.

### DIFF
--- a/clients/include/testing_csrrf_analysis.hpp
+++ b/clients/include/testing_csrrf_analysis.hpp
@@ -160,27 +160,27 @@ void csrrf_analysis_initData(rocblas_handle handle,
 
         // read-in A
         file = testcase / "ptrA";
-        read_matrix(file, 1, n + 1, hptrA.data(), 1);
+        read_matrix(file.string(), 1, n + 1, hptrA.data(), 1);
         file = testcase / "indA";
-        read_matrix(file, 1, nnzA, hindA.data(), 1);
+        read_matrix(file.string(), 1, nnzA, hindA.data(), 1);
         file = testcase / "valA";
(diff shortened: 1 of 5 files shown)


clients/include/testing_csrrf_analysis.hpp
```diff
@@ -403,9 +403,9 @@ void testing_csrrf_analysis(Arguments& argus)
     fs::path testcase;
     if(n > 0)
     {
-        testcase = get_sparse_data_dir() / fmt::format("mat_{}_{}", n, nnzM);
+        testcase = get_sparse_data_dir() / fmt::format("mat_{}_{}", n, nnzM).c_str();
```
cgmb, line 406: Use `fs::path(fmt::format("mat_{}_{}", n, nnzM))` to convert explicitly. Whenever I see `.c_str()` like this, I need to mentally check that the reference to the C-string doesn't exceed the lifetime of the string it came from. In theory, the conversion directly from `std::string` to `std::filepath` might also be slightly more efficient as you lose information about the length when passing a null-terminated `const char*`.

cgmb:
Hmm... Unrelated, but we will need to cherry-pick some docs changes into rocm-rel-5.6.

tfalders:
> Hmm... Unrelated, but we will need to cherry-pick some docs changes into rocm-rel-5.6.

Might as well get it done all at once, as long as it won't conflict once we merge back to develop. Feel free to open a PR on my fork.

tfalders:
Looks like I will also need to make the same changes in hipSOLVER, assuming this works.

jzuniga-amd (approved):

qjojo (approved):

PR #557, opened 2023-05-30
Check for test data relative to running executable

qjojo:
Adds a utility function to get the path of the running executable in clients.
Adds checks for test data relative to the running executable to fix test data not being found when installed from package.

### DIFF
--- a/clients/include/rocsolver_test.hpp
+++ b/clients/include/rocsolver_test.hpp
@@ -171,7 +171,7 @@ inline fs::path get_sparse_data_dir()
         return installed;
 
     fmt::print(
-        stderr, "Warning: default sparse data directories ({}, {}) not found, defaulting to current working directory.",
+        stderr, "Warning: default sparse data directories ({}, {}) not found, defaulting to current working directory.\n",
         exe_relative, installed);
 
     return fs::current_path();
--- a/clients/include/rocsolver_test.hpp
+++ b/clients/include/rocsolver_test.hpp
@@ -143,14 +143,7 @@ inline fs::path get_sparse_data_dir()
     fs::path p = fs::current_path();
     fs::path p_parent = p.parent_path();
     fs::path installed = p.root_directory() / "opt" / "rocm" / "share" / "rocsolver" / "test";
-    fs::path exe_relative = rocsolver_exepath() + "../" + "share/" + "rocsolver/" + "test/";
-    try
-    {
-        exe_relative = fs::canonical(exe_relative);
-    }
-    catch(const std::exception& ex)
-    {
-    }
+    fs::path exe_relative = fs::path(rocsolver_exepath()) / ".." / "share" / "rocsolver" / "test";
 
     // check relative to the current directory and relative to each parent
     while(p != p_parent)
--- a/clients/include/rocsolver_test.hpp
+++ b/clients/include/rocsolver_test.hpp
@@ -143,6 +143,14 @@ inline fs::path get_sparse_data_dir()
     fs::path p = fs::current_path();
     fs::path p_parent = p.parent_path();
     fs::path installed = p.root_directory() / "opt" / "rocm" / "share" / "rocsolver" / "test";
+    fs::path exe_relative = rocsolver_exepath() + "../" + "share/" + "rocsolver/" + "test/";
+    try
+    {
+        exe_relative = fs::canonical(exe_relative);
+    }
+    catch(const std::exception& ex)
+    {
+    }
 
     // check relative to the current directory and relative to each parent
     while(p != p_parent)
@@ -154,9 +162,17 @@ inline fs::path get_sparse_data_dir()
         p_parent = p.parent_path();
     }
 
+    // check relative to the running executable
+    if(fs::exists(exe_relative))
+        return exe_relative;
+
     // check relative to default install path
     if(fs::exists(installed))
         return installed;
+
+    fmt::print(
+        stderr, "Warning: default sparse data directories ({}, {}) not found, defaulting to current working directory.",
+        exe_relative, installed);
 
     return fs::current_path();
 }
--- a/clients/rocblascommon/clients_utility.cpp
+++ b/clients/rocblascommon/clients_utility.cpp
@@ -12,6 +12,20 @@
 #include "clients_utility.hpp"
 #include "rocblas_random.hpp"
 
+#ifdef _WIN32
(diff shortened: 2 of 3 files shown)


clients/rocblascommon/clients_utility.hpp
```diff
@@ -13,6 +13,21 @@
...
+namespace fs = std::filesystem;
+#else
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#endif
+
+#else
+#include <fcntl.h>
+#endif
+
```
cgmb: In the interests of minimizing build times, please default to putting things in implementation files, and only putting things in headers when they need to be there. (We often break that rule, but that's part of why our build times are so long.)

clients/rocblascommon/clients_utility.cpp
```diff
@@ -25,6 +25,46 @@ const std::thread::id main_thread_id = std::this_thread::get_id();
 // different seed but deterministically based on the thread id's hash function.
 thread_local rocblas_rng_t rocblas_rng = get_seed();
 
+// Return the path to the currently running executable
+std::string rocsolver_exepath()
+{
+#ifdef WIN32
```
cgmb: It's better to check for `_WIN32` as that macro is defined by the compiler and will always be defined when building on Windows, whereas `WIN32` is defined by the Windows SDK and is only defined after `windows.h` is included.

clients/rocblascommon/clients_utility.hpp
```diff
@@ -13,6 +13,21 @@
 
 #include "rocblas_test.hpp"
 
+#ifdef WIN32
+#define strcasecmp(A, B) _stricmp(A, B)
```
cgmb: This appears to be unused.

clients/include/rocsolver_test.hpp
```diff
@@ -143,6 +143,14 @@ inline fs::path get_sparse_data_dir()
     fs::path p = fs::current_path();
     fs::path p_parent = p.parent_path();
     fs::path installed = p.root_directory() / "opt" / "rocm" / "share" / "rocsolver" / "test";
+    fs::path exe_relative = rocsolver_exepath() + "../" + "share/" + "rocsolver/" + "test/";
```
cgmb: Using `+` with strings like this is a bit of an anti-pattern. You end up constructing and discarding a bunch of temporary strings as each pair is combined to create a new string.

When possible, use as few strings as possible and/or use `+=`. Could `"../" + "share/" + "rocsolver/" + "test/"` be replaced with `"../share/rocsolver/test/"`?

clients/include/rocsolver_test.hpp
```diff
@@ -143,6 +143,14 @@ inline fs::path get_sparse_data_dir()
...
+    try
+    {
+        exe_relative = fs::canonical(exe_relative);
+    }
+    catch(const std::exception& ex)
+    {
+    }
```
cgmb: The the generic `std::exception` and the empty catch block raises red flags to me, but I'm not entirely sure what we'd want to do about failure here anyway. Do we actually need to canonicalize this path? One way to have fewer errors to handle is to just do less stuff.

clients/include/rocsolver_test.hpp
tfalders: Not only that, this is an implicit conversion from string to path. It will probably break the Windows build. I'd prefer something like the previous line:
`fs::path exe_relative = fs::path(rocsolver_exepath()) / ".." / "share" / "rocsolver" / "test";`

qjojo:
Moved the `fs` include into the implementation file, removed unnecessary canonicalization, fixed preprocessor checks for windows.

clients/rocblascommon/clients_utility.cpp
```diff
@@ -12,6 +12,20 @@
...
+#ifdef _WIN32
+
+#ifdef __cpp_lib_filesystem
+#include <filesystem>
+namespace fs = std::filesystem;
+#else
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#endif
+
```
cgmb, line 27: [`GetModuleFileNameA`](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamea) is defined in `libloaderapi.h`, so you'll need to include that header on Windows.

clients/rocblascommon/clients_utility.cpp
```diff
@@ -12,6 +12,20 @@
 #include "clients_utility.hpp"
 #include "rocblas_random.hpp"
 
+#ifdef _WIN32
+
+#ifdef __cpp_lib_filesystem
```
cgmb: The [`__cpp_lib_filesystem`](https://en.cppreference.com/w/cpp/header/version) macro is not guaranteed to be defined until after either:
- `#include <version>`, which makes it rather pointless as the version header is newer than the filesystem header.
- `#include <filesystem>`, which is a bit of a Catch-22.

Any standard library header is allowed to include any other standard library header, so you will find that even if you do not include `#include <filesystem>` or `#include <version>` yourself, this macro will _probably_ be defined if you've included other standard library headers first. I believe that's why rocSPARSE can do this check 'wrong' and it doesn't usually cause problems.

Nevertheless, please use `#if __has_include(<filesystem>)` instead. You can find an example in [logging_gtest.cpp](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/rocm-5.5.1/clients/gtest/logging_gtest.cpp#L6-L12).

cgmb (commented):
You need to fix the Windows build. It's close, but you're missing an include.

clients/rocblascommon/clients_utility.cpp
cgmb: You might also (or maybe instead?) need `#include <windows.h>`.

tfalders: Yeah, the Windows CI doesn't seem to like `libloaderapi.h`. Switching it out for `windows.h` should hopefully resolve the build failure.

cgmb (approved):
The Windows build needs a minor change to fix, but aside from that this looks good. Thanks Jonah!

tfalders (approved):
LGTM, as long as the Windows build goes through

tfalders:
Looks like the build has passed and it's ready to be merged

PR #569, opened 2023-06-24
Specify minimum rocBLAS version

cgmb:
From testing, it seems that rocBLAS 3.1 is the current minimum version that you can use to build rocSOLVER. Note that the minimum version of rocBLAS required to build rocSOLVER may be completely different from the minimum required for the DEB/RPM packages. Everything related to the DEB/RPM packages has a bunch of hardcoded assumptions, and is therefore a lot more strict about library versions.

This change adds some restrictions that block users from doing things that don't work (building with rocblas < 3.1), and relaxes restrictions on things that should be fine (installing future versions of rocBLAS 4.x).


tfalders:
Are we confident there will be no breaking changes in 4.0? I have a feeling that 4.0 will remove `rocblas_is_user_managing_device_memory` and `rocblas_set_device_memory_size`, which were marked as deprecated but are used by hipSOLVER. Even if nothing breaks for rocSOLVER, I think it would be good for hipSOLVER and rocSOLVER to have the same version requirement for rocBLAS.

cgmb:
I'm not sure I agree that we should raise the rocSOLVER minimum requirements just because hipSOLVER has higher requirements, but I agree that we should think about both libraries before committing to anything for either. I've converted the [hipSOLVER PR](https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/178) to a draft while I rework it, as you are correct that we will need to remove our use of deprecated functionality to properly support rocBLAS 4.0.

TorreZuk:
> Are we confident there will be no breaking changes in 4.0? I have a feeling that 4.0 will remove `rocblas_is_user_managing_device_memory` and `rocblas_set_device_memory_size`, which were marked as deprecated but are used by hipSOLVER. Even if nothing breaks for rocSOLVER, I think it would be good for hipSOLVER and rocSOLVER to have the same version requirement for rocBLAS.

I agree with @tfalders we have breaking changes in 4.0 so don't know what this buys you over added support costs unless this is rigorously tested.  @amcamd is removing trmm functions and Ruban will be back next week and likely continue on those device_memory breaking changes.

cgmb:
> I agree with @tfalders we have breaking changes in 4.0 so don't know what this buys you over added support costs unless this is rigorously tested. @amcamd is removing trmm functions and Ruban will be back next week and likely continue on those device_memory breaking changes.

IMO, there should be no breakage to rocSOLVER or hipSOLVER when rocBLAS makes their changes. To provide a smooth transition, we should be updating our libraries to support building against either version of rocBLAS before the breaking changes are merged.

There are three reasons why we should retain support for building against older versions (where possible):

1. It makes it easier for contributors to build the develop branch of our libraries. It's not easy to build rocBLAS, so it's nice to support building against a version with prebuilt binaries available.
2. Many users have good reasons for sticking with older versions of ROCm, which is why we support things like installing multiple versions on the same system. Third-party libraries may therefore want to be able to support building against multiple versions of ROCm. If maintaining support for building against multiple versions of rocBLAS is difficult for rocSOLVER, that could be a sign that the changes will be difficult for other downstream libraries as well.
3. I'm going to have to do something like this anyway when rocSOLVER is packaged for distros. In a rolling release, the libraries don't all get updated atomically like with official ROCm releases. To have a smooth update process, it helps a lot to be able to support building against multiple versions through the transition. In that way, it's actually a lot like our CI.

cgmb:
As shown by #586, there are some tricky issues to consider when checking the version. Let's shelve this for now.

PR #571, opened 2023-06-28
[SWDEV-404616] Updated Copyright Headers

jibrealkhan:
Description: Updated the copyright notice by adding "All rights reserved" and add the BSD-2 license to all files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,13 @@
 # ########################################################################
-# Copyright (c) 2019-2023 Advanced Micro Devices, Inc.
+#   Copyright (C) 2019-2023 Advanced Micro Devices, Inc. All rights reserved.
+#  
+#   Redistribution and use in source and binary forms, with or without modification, 
+#   are permitted provided that the following conditions are met:
+#   1)Redistributions of source code must retain the above copyright notice, 
+#   this list of conditions and the following disclaimer.
+#   2)Redistributions in binary form must reproduce the above copyright notice, 
+#   this list of conditions and the following disclaimer in the documentation 
+#   and/or other materials provided with the distribution
 # ########################################################################
 
 cmake_minimum_required(VERSION 3.13)
--- a/clients/common/testing_csrrf_analysis.cpp
+++ b/clients/common/testing_csrrf_analysis.cpp
@@ -1,5 +1,13 @@
 /* ************************************************************************
- * Copyright (c) 2023 Advanced Micro Devices, Inc.
+ * Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, 
+ * are permitted provided that the following conditions are met:
+ * 1)Redistributions of source code must retain the above copyright notice, 
+ * this list of conditions and the following disclaimer.
+ * 2)Redistributions in binary form must reproduce the above copyright notice, 
+ * this list of conditions and the following disclaimer in the documentation 
+ * and/or other materials provided with the distribution.
  * ************************************************************************ */
 
 #include <testing_csrrf_analysis.hpp>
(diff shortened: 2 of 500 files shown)


jzuniga-amd (commented):
Shouldn't we include the new text in all the files?
For example in `library/scr/lapack/roclapack_gebd2.hpp` after line 7, etc...

qjojo (approved):

clients/common/testing_csrrf_analysis.cpp
```diff
@@ -1,5 +1,13 @@
 /* ************************************************************************
- * Copyright (c) 2023 Advanced Micro Devices, Inc.
+ * Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
```
tfalders, line 2: Minor issue, but the dates on these csrrf files seem to have been changed.

CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
...
+#   Copyright (C) 2019-2023 Advanced Micro Devices, Inc. All rights reserved.
+#  
+#   Redistribution and use in source and binary forms, with or without modification, 
+#   are permitted provided that the following conditions are met:
+#   1)Redistributions of source code must retain the above copyright notice, 
+#   this list of conditions and the following disclaimer.
+#   2)Redistributions in binary form must reproduce the above copyright notice, 
+#   this list of conditions and the following disclaimer in the documentation 
+#   and/or other materials provided with the distribution
 # ########################################################################
```
cgmb, line 11: There's a few issues here. In order from most important to least important:

1. Clause (1) states that `Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.` but no disclaimer follows. You've left out an important part of the license.
2. The lack of whitespace in and around the numbered clauses make them difficult to read.
3. It bugs me when the comment blocks at the top and bottom are too small for the containing text. One thing that would help would be to continue using just a single space after the `#`.

To address these issues for shell, Python and CMake files, I suggest using this template:
https://github.com/cgmb/rocSOLVER/blob/6fda8116bafbe67fe05ee0cceb92709737d244cb/CMakeLists.txt#L1-L26

For C and C++ files, I suggest using this template:
https://github.com/cgmb/rocSOLVER/blob/6fda8116bafbe67fe05ee0cceb92709737d244cb/library/include/rocsolver/rocsolver.h#L1-L26

I've opened https://github.com/jibrealkhan/rocSOLVER/pull/1 with these suggested templates. Since you've been quite consistent in these headers, it should be pretty straight-forward to write a simple find-and-replace script to apply this change (or one like it) to update to all files.

library/include/rocsolver/rocsolver.h
```diff
@@ -1,5 +1,13 @@
 /* ************************************************************************
- * Copyright (c) 2019-2023 Advanced Micro Devices, Inc.
+ * Copyright (C) 2019-2023 Advanced Micro Devices, Inc. All right reserved.
```
cgmb, line 2: This should be `All rights reserved.`

cgmb (changes requested):
Sorry for taking so long to review this.

Please note that you also need to update the template in `scripts/reformat-files`.

If you want to see what I see when I review a library's copyright statements, install scan-copyrights with `sudo apt install cme libconfig-model-dpkg-perl` and then run `scan-copyrights` from the root directory of the project.

clients/common/testing_csrrf_analysis.cpp
cgmb: In general, I don't think any existing copyright dates should be changed by this update. The guidelines are that we should update the date whenever we make a substantial modification. A change to the copyright block itself probably doesn't count. That said, it's not a big deal either way.

cgmb:
With regards to the JIRA ticket number, I would put it in the body text of your PR rather than in the title. There are two reasons for that:

1. rocSOLVER is an open-source project and most development is done in public. However, that number refers to a bug tracker that most users and external contributors do not have access to. It doesn't deserve such a prominent position when it's only useful to some people.
2. If it's in the body text of the PR, it will be automatically turned into a hyperlink. For people who have access to our bug tracker, it's actually more convenient to have it in the extended description.

So, I would suggest just putting it at the end of your PR description, like:

Ticket: SWDEV-404616

PR #587, opened 2023-07-24
add scripts to generate sparse data for tests

jzuniga-amd:
This PR adds the matlab/octave scripts used to generate the sparse matrices for the tests of the refactorization functionality. 
Also, it adds new sparse matrices to test the upcoming Cholesky refactorization functions.


scripts/spdata/cleanmat.m
```diff
@@ -0,0 +1,51 @@
...
+	(This script is for internal use only. It is not part of
+	rocSOLVER library interface and could change or be removed
+	without any notice)
+-------------------------------------------------------------
+%}
+
+
+n = size(B,1);
+A=full(B);
+
```
EdDAzevedo, line 30: Perhaps the code may consider using the "find" function, say   [ivec, jvec, avec] = find(B);   to find all the non-zero entries, then do something  like  ind = find(  (ivec != jvec) ) to find all non-diagonal entries.    If you wish to remove the "largest magnitude" entries first, then one can also sort first by something like    
[dummy, iperm] = sort( -abs(avec));   ivec = ivec(iperm); jvec = jvec(iperm); avec = avec(iperm);

One can then do something like    the following to set the entries to be "removed" to be very small.
mremove = toclean;  m = min(mremove, length(ind));  avec( ind(1:m) ) = realmin;  
then reconstruct the sparse matrix    A = sparse( ivec, jvec, avec, n,n);
Note matlab (or octave) may internally remove  entries that are exactly zero  and so may affect the sparsity pattern and "nnz". If it is acceptable for matlab (or octave) to modify  the sparsity pattern and nnz, then you can set "avec(ind(1:m)) = 0"

scripts/spdata/csr2full.m
```diff
@@ -0,0 +1,38 @@
...
+	without any notice)
+-----------------------------------------------------------
+%}
+
+
+A=zeros(m,m);
+
+for i = 1:m,
+	k = ptr(i+1) - ptr(i);
+	for j=1:k,
```
EdDAzevedo, line 32: May consider using  vector notation instead of for loop,      j = 1:k; A(i, ind(ptr(i)+j)+1) = val( ptr(i) + j);

scripts/spdata/fixmat.m
```diff
@@ -0,0 +1,50 @@
...
+	rocSOLVER library interface and could change or be removed
+	without any notice)
+-------------------------------------------------------------
+%}
+
+
+n = size(B,1);
+A=full(B);
+
+for j=1:n,
```
EdDAzevedo, line 32: You might consider something like  "ind_zero_diag = find( diag(B,0) == 0);" to find all columns that has a zero diagonal.  To modify the diagonal entries, one can try something like "B = B + sparse( 1:n,1:n,  -diag(B,0) + diagval * ones(n,1),  n,n);"  Then process the necessary columns related to zero diagonal entries using the vector "ind_zero_diag".  
Just a suggestion.

scripts/spdata/full2csr.m
```diff
@@ -0,0 +1,44 @@
...
+  	(This script is for internal use only. It is not part of
+	rocSOLVER library interface and could change or be removed
+	without any notice)
+-----------------------------------------------------------
+%}
+
+n = size(A,1);
+
+ptr = 0;
+ind = [];
```
EdDAzevedo, line 29: Note the "nnz()" function also works on dense matrices so it is possible to preallocate the correct amount of storage.  The "find()" function can also work on a row of a dense matrix to find the non-zero entries in that row. 
Say something like the following

nnzA = nnz(A);  ptr = zeros(n+1,1); ind = zeros( nnzA,1); val = zeros(nnzA,1);
ipfree = 1;
for irow=1:n,
   [ii,jj,aa] = find(A(irow,:));  nz = length(aa);
   ptr(irow) = ipfree;  
   istart = ipfree; iend = ipfree + (nz-1);
   ind( istart:iend ) = jj; 
   val( istart:iend ) = aa;
   ipfree = ipfree + nz;
end;
ptr(n+1) = nnz + 1;

% adjust starting index to be zero
ptr = ptr - 1;
ind = ind - 1;

clients/sparsedata/posmat_250_300/valT
```diff
@@ -0,0 +1 @@
...11388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 0.001989009072323235 15.811388175737225 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 0.024416997530140771 15.811369447654799 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 0.00047819463810443803 15.811388293610714 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 0.015644067604499635 0.054527894089619194 15.811286537531187 15.811388300841896 15.811388300841896 0.0098570651355969208 15.811385228317818 15.811388300841896 15.811388300841896 0.019450522259298587 15.811376337219473 15.811388300841896 15.811388300841896 0.054375229894693705 15.811294802588872 15.811388300841896 15.811388300841896 15.811388300841896 15.811388300841896 0.039593158842017258 15.811338728323195 15.811388300841896 0.019166789845428089 15.811376683709963 15.811388300841896 15.811388300841896 0.00037894219127470371 15.811388296300954
```
cgmb: Could we clean up the whitespace on these data files?

    find clients/sparsedata -type f -exec sed -i -e 's/\s*$//' -e '$a\' {} \;

scripts/spdata/printall.m
```diff
@@ -0,0 +1,118 @@
...
+isok = print_x('X_30',X);
+if (!isok),
+  	error(sprintf('Error writting files for matrix X with 30 rhs'));
+end;
+isok = print_x('B_30',B);
+if (!isok),
+  	error(sprintf('Error writting files for matrix B with 30 rhs'));
+end;
+
+end
```
cgmb: We should add `\.m` to the list of extensions in `scripts/reformat-files`.

The command I suggested for the matrix files can also fixup these.

    find scripts/spdata -type f -exec sed -i -e 's/\s*$//' -e '$a\' {} \;

clients/sparsedata/posmat_250_300/valT
tfalders: Pursuant to that, maybe there are some changes to the scripts that can be made to ensure they don't have these issues in the first place?

jzuniga-amd: whitespaces removed on current data files, and scripts corrected for future use

scripts/spdata/printall.m
jzuniga-amd: .m added, and .m scripts reformated accordingly

scripts/spdata/cleanmat.m
jzuniga-amd: I'm sure there are many ways we can improve/optimize these scripts, but given they are only for internal use, I will leave them as is for now to focus on other commitments for the release cycle. We can open another PR with modifs any time we want later.

cgmb (approved):

tfalders (approved):
Looks good to me!

PR #593, opened 2023-08-08
Divide & conquer + Jacobi (part 1)

jzuniga-amd:
Generalizes stedc to use different solvers
Adapts syevj kernels to be used from stedc


library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -289,23 +303,22 @@ ROCSOLVER_KERNEL void syevj_small_kernel(const rocblas_esort esort,
...
-            info[bid] = 0;
+            n_sweeps[0] = sweeps;
+            info[0] = 0;
         }
         else
         {
-            n_sweeps[bid] = max_sweeps;
-            info[bid] = 1;
+            n_sweeps[0] = max_sweeps;
+            info[0] = 1;
```
tfalders: Not too important, but I think I would prefer pointer dereferences rather than array accesses here. E.g., `*residual = sqrt(local_res)` instead of `residual[0] = sqrt(local_res)`.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -767,16 +784,32 @@ ROCSOLVER_KERNEL void __launch_bounds__(BDIM) stedc_kernel(const rocblas_int n,
...
         // determine ideal number of sub-blocks
-        levs = stedc_num_levels(bs);
+        levs = stedc_num_levels(bs, solver_mode);
         blks = 1 << levs;
 
-        // if split block is too small, solve it with steqr
+        // if split block is too small
+        // solver_mode = 1 solve the split blocks using classic QR iteration
+        // solver_mode = 2 solve the split blocks using Jacobi
+        // solver_mode = 3 solve the split blocks using bisection and inverse iteration
```
tfalders: It might make things easier to use `#define` and `#undef` to give names to these numbers.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -767,16 +784,32 @@ ROCSOLVER_KERNEL void __launch_bounds__(BDIM) stedc_kernel(const rocblas_int n,
...
-            if(id == 0)
+            if(solver_mode == 3)
+            {
+                // TODO
+            }
+
+            else if(solver_mode == 2)
+            {
+                // TODO
+            }
```
tfalders: Could we use a switch statement here instead?

tfalders (approved):
Looks good to me! I just have some minor comments.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -9,14 +9,18 @@
...
 #include "rocauxiliary_sterf.hpp"
 #include "rocblas.hpp"
 #include "rocsolver/rocsolver.h"
 
-#define BDIM 512 // Number of threads per thread-block used in main stedc kernel
+#define STEDC_BDIM 512 // Number of threads per thread-block used in main stedc kernel
 #define MAXITERS 50 // Max number of iterations for root finding method
+#define CLASSICQR 1 // used to specify classic QR iteration solver (default case)
+#define JACOBI 2 // used to specify Jacobi iteration solver
+#define BISECTION 3 // used to specify bisection and inverse iteration solver
```
cgmb, line 23: I'd consider using an enum for this sort of thing. Then the compiler can help with things like warnings when switches don't cover all possible values.

cgmb (approved):
I had a minor suggestion, but nothing that necessarily needs to be changed.

library/src/auxiliary/rocauxiliary_stedc.hpp
jzuniga-amd: will merge as is and maybe change it if needed in the following PR (part2)

PR #594, opened 2023-08-10
366827 - Disable file reorg backward compatibility support by default

arvindcheru:
Changes proposed
- Backward compatibility support flag set to false by default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ option(WERROR "Treat warnings as errors" OFF)
 option(BUILD_COMPRESSED_DBG "Enable compressed debug symbols" ON)
 
 cmake_dependent_option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY
-  "Build with file/folder reorg backward compatibility enabled" ON "NOT WIN32" OFF)
+  "Build with file/folder reorg backward compatibility enabled" OFF "NOT WIN32" OFF)
 if(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)
   rocm_wrap_header_dir(
     ${CMAKE_SOURCE_DIR}/library/include/rocsolver
--- a/install.sh
+++ b/install.sh
@@ -327,7 +327,7 @@ optimal=true
 cleanup=false
 build_sanitizer=false
 build_codecoverage=false
-build_freorg_bkwdcomp=true
+build_freorg_bkwdcomp=false
 build_with_sparse=true
 unset architecture
 unset rocblas_dir
--- a/install.sh
+++ b/install.sh
@@ -325,7 +325,6 @@ optimal=true
 cleanup=false
 build_sanitizer=false
 build_codecoverage=false
-build_freorg_bkwdcomp=false
 build_with_sparse=true
 unset architecture
 unset rocblas_dir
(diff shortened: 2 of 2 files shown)


cgmb (commented):
Please change the default for users that are building directly with CMake. e.g.
```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ option(WERROR "Treat warnings as errors" OFF)
 option(BUILD_COMPRESSED_DBG "Enable compressed debug symbols" ON)
 
 cmake_dependent_option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY
-  "Build with file/folder reorg backward compatibility enabled" ON "NOT WIN32" OFF)
+  "Build with file/folder reorg backward compatibility enabled" OFF "NOT WIN32" OFF)
 if(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)
   rocm_wrap_header_dir(
     ${CMAKE_SOURCE_DIR}/library/include/rocsolver
```

IMO, it would be nice if our install.sh file set no options whatsoever to eliminate the duplication between the CMake and the install script. That's a bigger job than you really need to handle in this PR, though. Just ensure you also update the cmake dependent option for this change.

arvindcheru:
> Please change the default for users that are building directly with CMake. e.g.
> 
> ```diff
> --- a/CMakeLists.txt
> +++ b/CMakeLists.txt
> @@ -115,7 +115,7 @@ option(WERROR "Treat warnings as errors" OFF)
>  option(BUILD_COMPRESSED_DBG "Enable compressed debug symbols" ON)
>  
>  cmake_dependent_option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY
> -  "Build with file/folder reorg backward compatibility enabled" ON "NOT WIN32" OFF)
> +  "Build with file/folder reorg backward compatibility enabled" OFF "NOT WIN32" OFF)
>  if(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)
>    rocm_wrap_header_dir(
>      ${CMAKE_SOURCE_DIR}/library/include/rocsolver
> ```
> 
> IMO, it would be nice if our install.sh file set no options whatsoever to eliminate the duplication between the CMake and the install script. That's a bigger job than you really need to handle in this PR, though. Just ensure you also update the cmake dependent option for this change.

Thanks for the comments, Kindly review updated version

cgmb (commented):
Removing the `--rm-legacy-include-dir` option makes sense to me, but if you're doing that, please remove the rest of the code that implements that flag from `install.sh`. There's no reason for `build_freorg_bkwdcomp` to exist at all anymore.

cgmb (approved):

samjwu (approved):
The test stage failures do not appear to be caused by this PR

Same error observed in separate PR:
http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin%2FrocSOLVER/detail/PR-597/2/pipeline

cgmb:
@jzuniga-amd, it seems something has changed in size again. 
```
/var/jenkins_home/workspace/main_precheckin_rocSOLVER_PR-594/ipFk1GtG8/rocsolver/clients/gtest/memory_model_gtest.cpp:93: Failure
  size
    Which is: 87053184
  size1
    Which is: 53498752
```

samjwu:
Merging this; will monitor future builds for errors

PR #597, opened 2023-08-23
Added sparse Cholesky re-factorization

tfalders:
This PR adds sparse Cholesky re-factorization to rocSOLVER.

I've made the following changes to @EdDAzevedo's branch:
- Renamed `rocsolver_rfinfo_mode_general` and `rocsolver_rfinfo_mode_symmetric` to `rocsolver_rfinfo_mode_lu` and `rocsolver_rfinfo_mode_cholesky`. This should make it easier to add additional modes for these same matrix types (e.g. QR factorization for general matrices).
- Implemented `rocsolver_set_rfinfo_mode`. CSRRF_ANALYSIS and CSRRF_SOLVE will now change behaviour depending on the mode that is set. Also, the mode cannot be changed after CSRRF_ANALYSIS is called.
- Separated code from `rocrefact_csrrf_refact.hpp` into `rocrefact_csrrf_refactlu.hpp` and `rocrefact_csrrf_refactchol.hpp`. Both CSRRF_REFACTLU and CSRRF_REFACTCHOL will fail if the mode does not match and analysis has not been performed.
- Moved `rf_search.hpp` to `refact_helpers.hpp`. I also removed some extraneous files.
- Added tests for CSRRF_ANALYSIS and CSRRF_SOLVE in Cholesky mode. Because of failures in the new CSRRF_SOLVE tests, I reverted the changes to the sparse test data. This caused failures in the CSRRF_REFACTCHOL tests, which I resolved by rewriting `rf_add_QAQ_kernel` (using `rf_add_PAQ_kernel` as the base).


EdDAzevedo (approved):

clients/gtest/csrrf_analysis_gtest.cpp
```diff
@@ -10,7 +10,10 @@ using ::testing::Values;
 using ::testing::ValuesIn;
 using namespace std;
 
-typedef std::tuple<int, int> csrrf_analysis_tuple;
+typedef std::tuple<int, int, printable_char> csrrf_analysis_tuple;
+
+// if mode = 0, then the factorization is LU
+// if mode = 1, then the factorization is Cholesky
```
jzuniga-amd: Just to avoid confusions, through the code you are using mode=1 for LU and mode=2 for Chol

clients/include/testing_csrrf_analysis.hpp
```diff
@@ -338,8 +352,12 @@ void testing_csrrf_analysis(Arguments& argus)
     rocblas_int n = argus.get<rocblas_int>("n");
     rocblas_int nnzM = argus.get<rocblas_int>("nnzM");
     rocblas_int nnzT = argus.get<rocblas_int>("nnzT");
+    char modeC = argus.get<char>("mode", '1');
```
jzuniga-amd: An option to manually pass the mode must be added to the benchmark client (client.cpp) as well

clients/gtest/csrrf_refactchol_gtest.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+// case when n = 0 and nnz = 60 also execute the bad arguments test
+// (null handle, null pointers and invalid values)
+
+// for checkin_lapack tests
+const vector<int> n_range = {
+    // quick return
+    0,
+    // invalid
+    -1,
```
jzuniga-amd, line 24: an invalid case where the analysis is done in LU mode, before the cholesky refactorization, must be systematically tested (see other related comments)

clients/include/testing_csrrf_refactchol.hpp
```diff
@@ -0,0 +1,513 @@
...
+                                             dindT, dvalT, dpivQ, hptrA, hindA, hvalA, hptrT, hindT,
+                                             hvalT, hpivQ, testcase);
+
+    // execute computations
+    // GPU lapack
+    CHECK_ROCBLAS_ERROR(rocsolver_csrrf_analysis(
+        handle, n, 0, nnzA, dptrA.data(), dindA.data(), dvalA.data(), nnzT, dptrT.data(),
+        dindT.data(), dvalT.data(), dpivQ.data(), dpivQ.data(), (T*)nullptr, n, rfinfo));
+
+    CHECK_ROCBLAS_ERROR(rocsolver_csrrf_refactchol(handle, n, nnzA, dptrA.data(), dindA.data(),
```
jzuniga-amd, line 208: We need to test that refactchol returns an error if the analysis was done in mode LU or if the mode is switched to LU before the call.
I suggest we codify this case(s) somehow in the gtest file so that  we don't repeat it(them) with every test case. (see other related comments)

-- A similar test logic should be included in refactlu

clients/include/testing_csrrf_refactchol.hpp
```diff
@@ -0,0 +1,513 @@
...
+{
+    // get arguments
+    rocblas_local_handle handle;
+    rocsolver_local_rfinfo rfinfo(handle);
+    rocblas_int n = argus.get<rocblas_int>("n");
+    rocblas_int nnzA = argus.get<rocblas_int>("nnzA");
+    rocblas_int nnzT = argus.get<rocblas_int>("nnzT");
+    rocblas_int hot_calls = argus.iters;
+
+    CHECK_ROCBLAS_ERROR(rocsolver_set_rfinfo_mode(rfinfo, rocsolver_rfinfo_mode_cholesky));
```
jzuniga-amd, line 309: we need to test return or internal error if mode= lu
(see other related comments)

clients/include/testing_csrrf_solve.hpp
```diff
@@ -285,7 +297,7 @@ void csrrf_solve_getPerfData(rocblas_handle handle,
     {
         csrrf_solve_initData<false, true, T>(handle, n, nrhs, nnzT, dptrT, dindT, dvalT, dpivP,
                                              dpivQ, dB, ldb, hptrT, hindT, hvalT, hpivP, hpivQ, hB,
-                                             hX, testcase, false);
+                                             hX, testcase, mode, false);
```
jzuniga-amd, line 300: here we also need to test invalid combinations (analysis in LU mode with solve in Chol mode, etc)
See other related comments.

library/src/refact/rocrefact_csrrf_refactchol.hpp
```diff
@@ -0,0 +1,213 @@
...
+                                                   void* work)
+{
+    ROCSOLVER_ENTER("csrrf_refactchol", "n:", n, "nnzA:", nnzA, "nnzT:", nnzT);
+
+    // quick return
+    if(n == 0)
+        return rocblas_status_success;
+
+    // check state of rfinfo
+    if(!rfinfo->analyzed || rfinfo->mode != rocsolver_rfinfo_mode_cholesky)
```
jzuniga-amd: This condition is not strong enough. We need to guarantee that the analysis was done in cholesky mode, and that the call to refactchol is also in cholesky mode, i.e.:

(mode in analysis  +  mode in refactchol   =   result)
none (not done)  +  LU  =  error
none (not done)  +  CHOL =  error
LU + LU = error
LU + CHOL = error    
CHOL + LU = error
CHOL + CHOL = OK    

Similar truth tables must be implemented for refactlu and solve functions.
(See also related comments in the test infrastructure. This must be systematically tested if the user is going to receive an internal error from the API)

library/src/refact/rocrefact_csrrf_refactlu.hpp
```diff
@@ -215,6 +161,10 @@ rocblas_status rocsolver_csrrf_refactlu_template(rocblas_handle handle,
     if(n == 0)
         return rocblas_status_success;
 
+    // check state of rfinfo
+    if(!rfinfo->analyzed || rfinfo->mode != rocsolver_rfinfo_mode_lu)
+        return rocblas_status_internal_error;
```
jzuniga-amd, line 165: see related comments

library/src/refact/rocrefact_csrrf_solve.hpp
```diff
@@ -227,6 +288,10 @@ rocblas_status rocsolver_csrrf_solve_template(rocblas_handle handle,
     if(n == 0 || nrhs == 0)
         return rocblas_status_success;
 
+    // check state of rfinfo
+    if(!rfinfo->analyzed)
+        return rocblas_status_internal_error;
```
jzuniga-amd, line 294: see related comments

library/src/refact/rocsolver_rfinfo.cpp
```diff
@@ -127,3 +138,32 @@ extern "C" rocblas_status rocsolver_destroy_rfinfo(rocsolver_rfinfo rfinfo)
...
+{
+#ifdef HAVE_ROCSPARSE
+    if(!rfinfo)
+        return rocblas_status_invalid_pointer;
+
+    if(mode != rocsolver_rfinfo_mode_lu && mode != rocsolver_rfinfo_mode_cholesky)
+        return rocblas_status_invalid_value;
+
+    if(rfinfo->analyzed)
+        return rocblas_status_internal_error;
```
jzuniga-amd: If I understand this logic correctly, once an analysis is done (and thus `analyzed` set to true), it is not possible any more to switch modes unless rfinfo is destroyed and a new one is created. 
I suggest a better behavior would be to allow the mode to change, but `analysis` should record in which mode was used for the last time, and `refactlu`, `refactchol` and `solve` must verify the corresponding truth tables (see the other related comments).

For this, I suggest `rfinfo` not only contains the bolean `analyzed` which indicates if the analysis has been done, but also an extra field, let's say `last_mode` (or something similar) which is set to `mode` after the analysis. 
If `rfinfo->mode` and `rfinfo->last_mode` does not coincide and `rfinfo_analyzed=true`, then `solve`, `refactlu` and `refactchol` should return error.

scripts/spdata/test100.m
```diff
@@ -0,0 +1,17 @@
+
+% ********************************************************************
+% Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+% ********************************************************************
+dir='/home/edazeved/MERGE/rocSOLVER/build/debug/clients/sparsedata/posmat_100_300';
```
jzuniga-amd: These scripts are not very useful unless `dir` is passed as parameter...
Alternatively, we could remove the scripts and only leave `regen_prob.m`

jzuniga-amd (commented):
It looks really good, but there are a few things that we need to consider:

1. I open a PR to your branch with some revisions from my original draft that were not carried into this PR. Please have a look (it's mainly documentation stuff, but it dictates how the APIs should behave)
2. According on these revisions, there are some things that may need modification. Please look at the following comments in the code.
3. Also, (not expressed in the comments but) we need to:
      a. implement the get function for the rfinfo mode
      b. relax the conditions for invalid pointers to accept pivp = null when mode = cholesky

library/src/refact/rocsolver_rfinfo.cpp
EdDAzevedo: I think the mode (cholesky or LU) is part of the rfinfo handle. So once the mode is set say it is cholesky mode then perhaps we can assume it stays the same. Thus  the call to analysis must be analysis for Choleksy and  calling refactlu is an error. Similarly a call to solve must be using L L' automatically. I wonder whether we need "last_mode". 

 Perhaps this assumption that once the mode in rfinfo is set it cannot be changed may simplify the logic.

clients/gtest/csrrf_analysis_gtest.cpp
EdDAzevedo: Perhaps mode can be  enum  type  or  maybe  simply mode == 'C' for Cholesky and mode == 'L' for LU (instead of 0 and 1)?

library/src/refact/rocrefact_csrrf_refactchol.hpp
tfalders: The condition should be strong enough if the mode cannot be changed after analysis. If we allow the mode to be changed after analysis, then yes, I will need to beef up the condition.

jzuniga-amd: Right, this is supposing that the setter `rfinfo_set_mode` can set the mode of a living rfinfo object anytime...
This is what the documentation suggests... otherwise, we need to clearly specify that `rfinfo` will be fixed to LU or CHOL after the first analysis.

library/src/refact/rocsolver_rfinfo.cpp
jzuniga-amd: Yes, what you describe is the current implementation and it works fine. What bothers me is the fact that `rfinfo` is not reusable anymore (which I think was the goal from the beginning)...

If a user executes a LU refactorization and, later on, the application needs a Cholesky one,  it will be necessary to destroy the handle, create a new one, and then "set" it to be cholesky, rather than simply change the mode of the existing handle.

EdDAzevedo: IMHO I think it is acceptable for the user application to have two rfinfo handles (one for Cholesky and one for LU). The extra memory cost on the host should not be a problem and there may not be significant savings in processing time in sharing a single handle.  One may expect the most costly part is the numerical factorization.   Just my $0.02

clients/include/testing_csrrf_workflow.hpp
```diff
@@ -0,0 +1,464 @@
...
+#include "clientcommon.hpp"
+#include "lapack_host_reference.hpp"
+#include "norm.hpp"
+#include "rocsolver.hpp"
+#include "rocsolver_arguments.hpp"
+#include "rocsolver_test.hpp"
+
+/*
+ * ===========================================================================
+ *    testing_csrrf_workflow is a modified version of testing_csrrf_workflow
```
jzuniga-amd: do you mean modified version of testing_csrrf_analysis?

(I was thinking on somehow verify the workflows within the test codes of the functions, but I guess having a separate test is cleaner and easier to debug... good idea)

jzuniga-amd (approved):
LGTM.
I would remove the `test.m` scripts for now as they work only with a local directory.

clients/include/testing_csrrf_workflow.hpp
tfalders: Right, nice catch! I forgot to change it back after I did a Replace All.

tfalders:
CI failures appear to be unrelated, so I'll force the merge

cgmb:
> CI failures appear to be unrelated, so I'll force the merge

Narrator voice: "They were not unrelated."

cgmb:
```
C:/jenkins/workspace/rocmmathlib_rocsolver_pr-600/clients/include\testing_csrrf_refactchol.hpp:141:14: error: no viable overloaded '='
        file = testcase / "ptrA";
        ~~~~ ^ ~~~~~~~~~~~~~~~~~
```

tfalders:
> ```
> C:/jenkins/workspace/rocmmathlib_rocsolver_pr-600/clients/include\testing_csrrf_refactchol.hpp:141:14: error: no viable overloaded '='
>         file = testcase / "ptrA";
>         ~~~~ ^ ~~~~~~~~~~~~~~~~~
> ```

Looks like the refactchol files were based on an old version of the refactlu files, from before the Windows fixes. *Sigh*

PR #600, opened 2023-08-31
Resolve test failures

tfalders:
In order to unblock the release, I'm disabling the failing memory model test and adding stream synchronizations as a temporary workaround to the intermittent failures on gfx940.

### DIFF
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -101,7 +101,7 @@ set(others_test_source
   # unified memory model
   managed_malloc_gtest.cpp
   # rocblas memory model
-  # memory_model_gtest.cpp (TODO: Resolve test failures)
+  memory_model_gtest.cpp
   # rocsolver logging
   logging_gtest.cpp
   # helpers
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -101,7 +101,7 @@ set(others_test_source
   # unified memory model
   managed_malloc_gtest.cpp
   # rocblas memory model
-  memory_model_gtest.cpp
+  # memory_model_gtest.cpp (TODO: Resolve test failures)
   # rocsolver logging
   logging_gtest.cpp
   # helpers
--- a/library/src/specialized/roclapack_trsm_specialized_kernels.hpp
+++ b/library/src/specialized/roclapack_trsm_specialized_kernels.hpp
@@ -825,6 +825,15 @@ void rocsolver_trsm_lower(rocblas_handle handle,
         return;
     }
 
+    // TODO: Temporary workaround for gfx940 synchronization issue with rocBLAS
+    int device;
+    hipGetDevice(&device);
+    hipDeviceProp_t deviceProperties;
+    hipGetDeviceProperties(&deviceProperties, device);
+    std::string deviceFullString(deviceProperties.gcnArchName);
+    std::string deviceString = deviceFullString.substr(0, deviceFullString.find(":"));
+    bool do_sync = (deviceString.find("gfx940") != std::string::npos);
+
     // ****** MAIN LOOP ***********
     if(isleft)
     {
@@ -853,6 +862,9 @@ void rocsolver_trsm_lower(rocblas_handle handle,
                 offA = idx2D(j, j, inca, lda);
                 offB = idx2D(j, 0, incb, ldb);
                 FORWARD_SUBSTITUTIONS;
+
+                if(do_sync)
+                    hipStreamSynchronize(stream);
 
                 // update right hand sides
                 rocsolver_gemm<BATCHED, STRIDED, T>(
@@ -894,6 +906,9 @@ void rocsolver_trsm_lower(rocblas_handle handle,
                 offA = idx2D(m - nextpiv, m - nextpiv, inca, lda);
                 offB = idx2D(m - nextpiv, 0, incb, ldb);
                 BACKWARD_SUBSTITUTIONS;
+
+                if(do_sync)
+                    hipStreamSynchronize(stream);
 
                 // update right hand sides
                 rocsolver_gemm<BATCHED, STRIDED, T>(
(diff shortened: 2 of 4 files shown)


clients/gtest/CMakeLists.txt
```diff
@@ -101,7 +101,7 @@ set(others_test_source
   # unified memory model
   managed_malloc_gtest.cpp
   # rocblas memory model
-  memory_model_gtest.cpp
+  # memory_model_gtest.cpp (TODO: Resolve test failures)
```
jzuniga-amd: mmm what are these test failures?

cgmb: [There is a GTest feature for temporarily disabling tests](http://google.github.io/googletest/advanced.html#temporarily-disabling-tests). Please prefix the test name as `DISABLED_`. This will ensure they are not run unless `--gtest_also_run_disabled_tests` is passed. That is better because:
1. We can then still run the tests without having to rebuild the library from source.
2. The CI will be ensuring that `memory_model_gtest.cpp` can still compile.

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
```diff
@@ -825,6 +825,15 @@ void rocsolver_trsm_lower(rocblas_handle handle,
...
     }
 
+    // TODO: Temporary workaround for gfx940 synchronization issue with rocBLAS
+    int device;
+    hipGetDevice(&device);
+    hipDeviceProp_t deviceProperties;
+    hipGetDeviceProperties(&deviceProperties, device);
+    std::string deviceFullString(deviceProperties.gcnArchName);
+    std::string deviceString = deviceFullString.substr(0, deviceFullString.find(":"));
+    bool do_sync = (deviceString.find("gfx940") != std::string::npos);
```
jzuniga-amd: It looks good, but I wonder about the performance implications of the getdeviceproperties API, probably not important, but out of curiosity:
Why do we have to do this at runtime?
Would it be possible to do this at compile time with a macro if we know we are generating code for gfx940?

clients/gtest/CMakeLists.txt
tfalders: I think rocBLAS might have changed the amount of device memory that is allocated by default. This is breaking some assumptions in the memory model tests.

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
TorreZuk: @jzuniga-amd host side code is compiled once for all GPU so must be runtime decision.  Only kernel launch selects GPU ISA at runtime and kernels can have #if defines for arch as compiled for each GPU arch.  We cache the arch info in the rocBLAS handle and use if for such runtime decisions....

cgmb: If it were done at compile time, the synchronization would apply to all GPU architectures, since the same host code is shared by all users. You need to do it dynamically if you want to ensure it only applies to a particular GPU. It's easy to do things in the device code on a per-architecture basis by checking for the `__GFX940__` macro, but in the host code, the dynamic check is the best way.

tfalders: @TorreZuk Is there a way we can get that cached info for use by rocSOLVER?

tfalders: @jzuniga-amd If the performance of getdeviceproperties is a concern, should we instead do the synchronization unconditionally?

clients/gtest/CMakeLists.txt
cgmb: This isn't the first time this has happened. I would suggest we disable the test until we can rework it to be more robust against false positives.

qjojo (approved):
Looks good. IMO the extra checks look like they'll get along well with host branch prediction so performance impact will probably be almost nothing (aside from any work done for the actual synchronization).

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
TorreZuk: > @TorreZuk Is there a way we can get that cached info for use by rocSOLVER?

Sorry I think you use rocblas_handle as an opaque struct, we would have to expose it via new exported method.
If you have something similar that is done once you could do it there, just suggesting the pattern we use to avoid repeatedly querying this information.  Can consider adding it as new function but won't be a quick fix for you, and we need rocblas review to expose internal info.

cgmb: > @jzuniga-amd If the performance of getdeviceproperties is a concern, should we instead do the synchronization unconditionally?

There may be performance implications to that choice as well, but those would be device-side concerns. The performance on the device is more critical than the performance on the host, so I think we should stick to the dynamic query.

tfalders: It sounds like the cleanest option is probably to do the query in our trsm_mem function, and pass the `do_sync` bool to our trsm calls. This would unfortunately require changes to all functions that call trsm. Or, I could temporarily repurpose the `optim_mem` bool, which we always set to true anyway. It's not an elegant solution, but it would save time and effort.

clients/gtest/CMakeLists.txt
TorreZuk: @rkamd can advise on rocblas changes

cgmb (approved):

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
cgmb: IMO, the way it is now seems simple enough and we can improve upon it later. Maybe in the next rocSOLVER technical discussion, we can talk about design changes to make this sort of thing nicer. We'll have to do a similar query as well, if we want to split our specialized kernels out into separate files.

clients/gtest/CMakeLists.txt
rkamd: AFAIK, there is no change in the default size for temporary workspace. Do you see a failure in validating default 32MB size ?

cgmb:
```
[ RUN      ] checkin_misc_MEMORY_MODEL.rocblas_managed
../../../../clients/gtest/memory_model_gtest.cpp:93: Failure
  size
    Which is: 87053184
  size1
    Which is: 53498752
../../../../clients/gtest/memory_model_gtest.cpp:110: Failure
  size
    Which is: 87053184
  size1
    Which is: 53498752
../../../../clients/gtest/memory_model_gtest.cpp:119: Failure
  size
    Which is: 87053184
  size1
    Which is: 53498752
[  FAILED  ] checkin_misc_MEMORY_MODEL.rocblas_managed (644 ms)
```

PR #603, opened 2023-09-05
Implement test for GETRF with large matrices.

jibrealkhan:
Added tests for large matrices for GETRF API.
[SWDEV-409347]


tfalders (approved):

cgmb (commented):
What's the difference between this change vs. just adding larger test sizes to the existing GETRF tests? I see that you're using rocBLAS for the check. Is it just the same check, but using the GPU to check the results rather than the CPU?

tfalders:
The existing GETRF test compares the results of rocsolver_getrf with the CPU LAPACK result, which is really slow at these sizes. This new test is an implicit test where we use GETRF and GETRS to solve `Ax = b` for `x`, then use GEMM to calculate `Ax` and compare the result with `b`. So, it's technically testing GETRS (and GEMM) as well, but it's a lot faster than the existing test thanks to the bulk of the computation resting on the GPU. And when I say a lot faster, we're talking ~20s vs ~1hr.

jzuniga-amd (approved):
LGTM
Just wondering if the new failures detected are in the gemms inside GETRF or the gemm in the implicit test. In any case, we just investigate and work with tensile/rocblas folks to resolve

cgmb:
> The existing GETRF test compares the results of rocsolver_getrf with the CPU LAPACK result, which is really slow at these sizes. This new test is an implicit test where we use GETRF and GETRS to solve Ax = b for x, then use GEMM to calculate Ax and compare the result with b. So, it's technically testing GETRS (and GEMM) as well, but it's a lot faster than the existing test thanks to the bulk of the computation resting on the GPU. And when I say a lot faster, we're talking ~20s vs ~1hr.

Could we include something explaining that in the comments? The change to the type of test is the entire thing that justifies it being a whole separate file. It's rather important and should be highlighted.

cgmb (approved):

tfalders:
The new large size test has been put under a new "weekly_lapack" test suite. I suspect that means it won't be run as part of the precheckin or extended CI tests and we will need to create a new CI job. Perhaps @cgmb @eidenyoshida @yoichiyoshida can advise?

cgmb:
It will need a Jenkins configuration for weekly, [like rocBLAS](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/5c7549987d8413a4c33e79eec26bd38660ec46ca/.jenkins/weekly.groovy).

PR #615, opened 2023-09-29
Fix debug build for CSRRF tests

tfalders:


jzuniga-amd (approved):

clients/include/testing_csrrf_analysis.hpp
```diff
@@ -450,8 +450,10 @@ void testing_csrrf_analysis(Arguments& argus)
             matname = fmt::format("posmat_{}_{}", n, nnzM);
 
         testcase = get_sparse_data_dir() / fs::path(matname);
-        fs::path file = testcase / "ptrT";
-        read_last(file.string(), &nnzT);
+        fs::path fileA = testcase / "ptrA";
+        read_last(fileA.string(), &nnzM);
+        fs::path fileT = testcase / "ptrT";
+        read_last(fileT.string(), &nnzT);
```
cgmb, line 456: I don't get it. Why was this only a problem in the debug build?

tfalders: Think of `nnzT` as the storage capacity for `indT` and `valT`. The actual number of non-zero values stored in `indT` and `valT` is given by the last element of `ptrT`, but `nnzT` can be larger than this without creating any problems in the algorithm.

The real issue is that `read_matrix` didn't fail even though it was reading past the end of the file. I guess we didn't realize that `assert` only applies to the debug build. Perhaps I should replace the assertion with an exception instead.

cgmb (approved):
When you go to squash-and-merge, please edit the commit message to include the information from [your comment](https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/615/files#r1343087520).

cgmb:
You will need to apply clang-format before merging.

PR #619, opened 2023-10-04
Check HIP errors in test teardown

cgmb:
In ROCm 6.0, the behaviour of the HIP API is expected to change such that successful calls do not reset the last error status. By checking the HIP error status in the teardown of the test, we can ensure that there were no unchecked errors within the test itself. This will also ensure that the error status is reset for the next test, thereby preventing a problem in one test to cause another test to fail.

This check is of limited value until the behaviour of HIP changes, but it is harmless to check anyway.


tfalders (approved):
Huh. Somehow I missed this PR getting opened. Nice job!

jzuniga-amd (approved):
LGTM

PR #621, opened 2023-10-11
add explicit string conversion for read_path to fix a build error

qjojo:
Adds some explicit conversions for calls the read_last function to fix a master specific build failure. This is the same solution we used for develop.


tfalders:
From what I understand, develop should be promoted to master. I'm not sure why it hasn't happened for a long time, but I don't think we should be adding commits to master that are not in develop.

TorreZuk:
develop->staging-> ...??? ..  ->master   @lawruble13  maybe can advise.

qjojo:
Closing in favor of automatic promotion, I've reached out to some people to see about getting that working for rocSOLVER again.

PR #624, opened 2023-10-18
Fast-forward master to include commits up to 6.0 FC

qjojo:
This brings master up to where it would be had the automatic promotion for 6.0 worked properly. This is strictly a fast-forward and does not introduce any new commits (it works locally with `--ff-only`). All the commits in this pull request have gone through CQE successfully. 

@lawruble13 advised that this is the proper way to bring master to where it should be, and won't cause any problems since the role of the master branch is no longer to act as the source of truth for what is in mainline.


cgmb (changes requested):
I don't think it's possible to fast-forward merge via a PR. *Merge* will create an extra commit and *squash and merge* will change the commits. *Rebase* almost seems like it could be used for a fast-forward merge, but I discovered when I merged #596 that it performs the rebase even if the base is already the target branch commit and this changes the commit date/author metadata.

You will need to push directly to the official rocSOLVER repo, so add it to your remotes:

    git remote add official git@github.com:ROCmSoftwarePlatform/rocSOLVER.git

Then, do a fast-forward merge to master locally and push:

    git checkout master
    git merge --ff release/rocm-rel-6.0
    git push official master
    
If you wish to live life on the edge, you can push directly to master:

    git push official release/rocm-rel-6.0:master

PR #625, opened 2023-10-25
Relax array length requirements for gesvdx

tfalders:
When `srange == rocblas_srange_index`, the value of `nsv` is known exactly. We can leverage this to reduce the amount of memory required by `S`, `U`, and `V`, and align the memory requirements with what is required by cuSOLVER's gesvda.


jzuniga-amd (approved):
LGTM

cgmb (approved):

tfalders:
It looks like the docs build status is the only thing preventing the merge of this PR. The docs are passing on CI, but for some reason the status isn't updating. If there are no objections, I'll force the merge of this PR.

EDIT: Oh, never mind. It looks like it merged successfully!

PR #627, opened 2023-10-30
Fix singular vector normalization in bdsvdx

tfalders:


qjojo (approved):

library/src/auxiliary/rocauxiliary_bdsvdx.hpp
```diff
@@ -92,29 +100,41 @@ ROCSOLVER_KERNEL void bdsvdx_reorder_vect(const rocblas_fill uplo,
...
+        // compute normalization factors for U and V
+        iamax<MAX_THDS, W>(tid, n, work, 2, sval1, sidx);
+        nrm2<MAX_THDS, W>(tid, n, work, 2, sval2);
+        __syncthreads();
+        iamax<MAX_THDS, W>(tid, n, work + 1, 2, sval1 + 1, sidx + 1);
+        nrm2<MAX_THDS, W>(tid, n, work + 1, 2, sval2 + 1);
+        __syncthreads();
+
+        scl1 = (work[sidx[0] - 1] >= 0 ? W(1) / sval2[0] : W(-1) / sval2[0]);
+        scl2 = (work[sidx[1] - 1] >= 0 ? W(1) / sval2[1] : W(-1) / sval2[1]);
```
jzuniga-amd: As discussed, if the vectors coming from the composed tridiagonal form are orthonormal (which I suppose is the case as you are adding the test to bdsvdx and it is passing?), then the scaling factor should always be sqrt(2). 

(I am not sure, however, what would be faster to compute, the norm of the vector or sqrt(2) ??)

jzuniga-amd (approved):
LGTM

library/src/auxiliary/rocauxiliary_bdsvdx.hpp
cgmb: I don't think I understand the question. `sqrt(2)` is a constant. There's no need for a run-time computation at all.

jzuniga-amd: mmm, sorry for the confusion, it was rather a rhetorical question based on my discussion with Troy. I was doubting if it would be easier to simply divide by the norm of the vector than to ensure (not "compute") that the norm is 1/sqrt(2); something that should be done in the `stein`  function...

jzuniga-amd (approved):
So, I wonder now if the increase in the tolerance of the tests is due to the normalization done by multiplying by sqrt(2) vs dividing by the actual norm??

In any case, we also need to include the code that tests if the eigenvectors form an orthonormal basis in the `stein` routine

tfalders:
> So, I wonder now if the increase in the tolerance of the tests is due to the normalization done by multiplying by sqrt(2) vs dividing by the actual norm??
> 
> In any case, we also need to include the code that tests if the eigenvectors form an orthonormal basis in the `stein` routine

There weren't many cases failing with the smaller test tolerance. Just the 20x20 case with float complex precision.

PR #631, opened 2023-11-14
splitLU optimization

EdDAzevedo:
New code to improve the performance of rocsolver_dcsrrf_splitlu() by using the header-only rocthrust library to compute in-place inclusive prefix scan sum.
(SWDEV-431135)


CMakeLists.txt
```diff
@@ -197,6 +197,8 @@ if(BUILD_WITH_SPARSE)
   list(APPEND rocsolver_pkgdeps "rocsparse >= 2.2")
 endif()
 
+find_package(rocprim REQUIRED CONFIG PATHS ${ROCM_PATH})
```
tfalders, line 198: Since this is currently only required by splitlu, I'd suggest only requiring rocPRIM when BUILD_WITH_SPARSE is true. It's probably also a good idea to specify a minimum version. Perhaps @cgmb can advise.

library/src/CMakeLists.txt
```diff
@@ -297,6 +297,7 @@ target_link_libraries(rocsolver
     roc::rocblas
   PRIVATE
     $<BUILD_INTERFACE:rocsolver-common> # https://gitlab.kitware.com/cmake/cmake/-/issues/15415
+    roc::rocprim
```
tfalders, line 312: If we only require rocPRIM when we BUILD_WITH_SPARSE, then this will need to be moved down into the BUILD_WITH_SPARSE section.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -229,10 +413,90 @@ rocblas_status rocsolver_csrrf_splitlu_template(rocblas_handle handle,
...
+        // -----------------
+        // copy into L and U
+        // -----------------
+        ROCSOLVER_LAUNCH_KERNEL(rf_splitLU_copy_kernel<T>, dim3(nblocks), dim3(nthreads), 0, stream,
+                                n, nnzT, ptrT, indT, valT, Lp, indL, valL, Up, indU, valU);
+    }
 
     return rocblas_status_success;
 }
+#undef HIP_CHECK
```
tfalders: This can be removed

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -148,7 +319,19 @@ void rocsolver_csrrf_splitlu_getMemorySize(const rocblas_int n,
...
+
+    // ------------------------------------------
+    // query amount of temporary storage required
+    // ------------------------------------------
+    size_t rocprim_size_bytes = 0;
+    void* temp_ptr = nullptr;
+
+    const hipError_t istat = rocprim::inclusive_scan(temp_ptr, rocprim_size_bytes, LUp, LUp, n,
+                                                     rocprim::plus<rocblas_int>());
+    assert(istat == hipSuccess);
```
tfalders: assert only works in the debug build. I would instead suggest returning a status code, which the _impl method will need to check.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -97,47 +253,62 @@ ROCSOLVER_KERNEL void rf_splitLU_kernel(const rocblas_int n,
...
         Li[j] = irow;
-        Lx[j] = T(1);
+        Lx[j] = static_cast<T>(1);
     };
 }
 
 template <typename T>
 void rocsolver_csrrf_splitlu_getMemorySize(const rocblas_int n,
                                            const rocblas_int nnzT,
+                                           rocblas_int* LUp,
```
tfalders: For consistency, please rename to ptrT.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -229,10 +413,90 @@ rocblas_status rocsolver_csrrf_splitlu_template(rocblas_handle handle,
         return rocblas_status_success;
     }
 
-    rocblas_int const nthreads = BS1;
-    rocblas_int const nblocks = 1;
-    ROCSOLVER_LAUNCH_KERNEL(rf_splitLU_kernel<T>, dim3(nblocks), dim3(nthreads), 0, stream, n, nnzT,
-                            ptrT, indT, valT, ptrL, indL, valL, ptrU, indU, valU, work);
+    const bool use_alg1 = (n <= 64);
+    if(use_alg1)
```
tfalders: This is the kind of situation where we would add a named constant to ideal_sizes.hpp and use that instead of 64, e.g. n <= SPLITLU_SWITCH_SIZE. use_alg1 is also unnecessary and can be moved directly into the if statement.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -229,10 +413,90 @@ rocblas_status rocsolver_csrrf_splitlu_template(rocblas_handle handle,
...
+        ROCSOLVER_LAUNCH_KERNEL(rf_splitLU_kernel<T>, dim3(nblocks), dim3(nthreads), 0, stream, n,
+                                nnzT, ptrT, indT, valT, ptrL, indL, valL, ptrU, indU, valU, work);
+    }
+    else
+    {
+        rocblas_int const nthreads = BS1;
+        rocblas_int const nblocks = max(1, min(1024, (n - 1) / nthreads + 1));
+
+        rocblas_int* const Lp = ptrL;
+        rocblas_int* const Up = ptrU;
```
tfalders, line 561: This is unnecessary. Please use ptrL and ptrU directly.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -229,10 +413,90 @@ rocblas_status rocsolver_csrrf_splitlu_template(rocblas_handle handle,
...
+            size_t storage_size_bytes = max(storage_size_bytes_Lp, storage_size_bytes_Up);
+            assert(size_work >= storage_size_bytes);
+
+            if(size_work < storage_size_bytes)
+            {
+                return (rocblas_status_internal_error);
+            };
+
+            temp_ptr = (void*)work;
+            storage_size_bytes = size_work;
```
tfalders: This looks like it belongs in getMemorySize, rather than the template method

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -229,10 +413,90 @@ rocblas_status rocsolver_csrrf_splitlu_template(rocblas_handle handle,
...
+        // ------------------------
+        // set Lp[0] = 0, Up[0] = 0
+        // ------------------------
+        {
+            const rocblas_int ival = static_cast<rocblas_int>(0);
+
+            HIP_CHECK(hipMemcpyAsync(Lp, &ival, sizeof(rocblas_int), hipMemcpyHostToDevice, stream));
+
+            HIP_CHECK(hipMemcpyAsync(Up, &ival, sizeof(rocblas_int), hipMemcpyHostToDevice, stream));
+        }
```
tfalders, line 610: I would suggest using our reset_batch_info kernel in lib_device_helpers.hpp to do this instead of hipMemcpyAsync. Reason being that reset_batch_info can work with batched and strided batched arrays, and we might want to add batched versions of splitlu at some point.

Although, this opens up a question: Does rocPRIM have batch functionality?

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -45,48 +182,67 @@ ROCSOLVER_KERNEL void rf_splitLU_kernel(const rocblas_int n,
...
     // 1st pass to determine number of non-zeros per row
     // and set up Lp and Up
     // -------------------------------------------------
-    rocblas_int i, j;
-    rocblas_int irow, icol, istart, iend;
-    __shared__ rocblas_int nnzL;
-    __shared__ rocblas_int nnzU;
 
-    for(irow = tid; irow < n; irow += hipBlockDim_x)
+    for(auto irow = wid; irow < n; irow += nwaves)
```
tfalders: I don't understand what you're trying to accomplish here. What exactly is waveSize supposed to represent? Wouldn't it make more sense to launch a 2D kernel directly rather than convert indices within a 1D kernel?

Plus, I'm pretty sure our test data never meets the conditions in cal_wave_size, meaning waveSize is always 1 and we aren't testing whether this code works properly.

tfalders (changes requested):
Here are some preliminary comments. I haven't fully reviewed the kernel code yet because I would like have some concerns about the indexing scheme. Once we've settled those concerns, I'll continue my review.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: #undef HIP_CHECK is removed

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: rocsolver_csrrf_splitlu_getMemorySize returns a rocblas_status as well

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: renamed LUp to ptrT

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: added SPLITLU_SWITCH_SIZE in ideal_sizes.hpp

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I think using intermediate variables as const expression should have no or insignificant impact on code quality. The compiler can (should) inline the const expression if the compiler chooses.

I think the use of intermediate variables is for the benefit of the developer.   For example,  it may be necessary to be compatible with cuda library to use say function argument variables as csrRowPtrA, csrColIndA, csrValA but the convention with another library, say,  SuiteSparse uses Ap, Ai, Ax for conceptually the similar arrays. So it may be convenient for one developer to use intermediate variables  as something like   "auto const Ap =csrRowPtrA;" when working with  SuiteSparse Library or in writing more compact code.     

Conversely,  a developer not familiar with the naming convention of SuiteSparse might prefer a more meaningful variable name and prefer to use  "auto const csrRowPtrA = Ap;" instead.

I see no harm in using intermediate variables for the benefit of the developer.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: This is extra check for robustness.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I think at some point if we wish to add batch capability, then all  relevant code (and code interface) should be modified to have batch capability. A mixed hybrid code with partially implemented batch capability might not be the best approach.  Rocsolver will also need to wait for batch capability in rocsparse.

If the batch interface assumes all sparse matrices in the batch list have the same sparsity pattern but may have different numerical values, then we do not need batch capability in rocPRIM.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: The performance of sparse matrix algorithms is commonly related to data access and not limited by flops. Ideally, we would like to have stride-1 coaslesced access to the arrays in compressed sparse row format. Thus one may wish to assign at least a  full warp to process one row. However, if the average number of non-zeros per row is small say about 15 nonzeros, then using all 64 threads in a warp may not be efficient since most of the threads will have no work. One may consider assigning say 8 threads to process one row and process 8 rows concurrently by the warp.  I use a conceptual virtual "wave" of threads to process one row. 

In using a 2D array of threads in a block, a programmer dealing with Fortran data structure might want threads in the conceptual "x" dimension to be contigous threads in the same warp to have coaslesced data access. Similarly, a programmer dealing with C array data structure might want threads in the conceptual "y" dimension to be contiguous threads in the same warp.  By using a 1D decomposition, there is no ambiguity and clearly that all threads  in a conceptual virtual "wave" are contiguous threads in the same warp.  There is also more flexibility in decoupling of the launch configuration and kernel code. By using a 1D decomposition, the kernel code will still work correctly independent of the launch configuration.

If the test data never meets the conditions in cal_wave_size() that the average number of non-zeros per row in our test matrices is always a small number, then I think we should add more tests. For example, we might consider storing a fully dense matrix or wide banded matrix in CSR format to test the code (not just splitlu) works properly.

CMakeLists.txt
EdDAzevedo: I'll leave cmake issues to Cory the expert.

library/src/CMakeLists.txt
EdDAzevedo: I'll leave cmake issues with Cory the expert.

library/src/refact/rocrefact_csrrf_splitlu.hpp
tfalders: The main advantage of using a 2D kernel in this case would be that you'd only have to do these computations once, rather than in each kernel. This would also reduce code redundancy, which could also be accomplished by moving some of the code into cal_wave_size.

There's an additional possibility that the kernel could be launched with nthreads not evenly divisible by waveSize, e.g. if a user altered the value of BS1. I believe this would cause some elements to be missed by the kernel. This shouldn't happen with an explicitly 2D kernel.

It would be good to have the explanation for what a wave is included as comments in the code.

EdDAzevedo: The number of threads assigned to process each row is based on a heuristic related to the average number of nonzeros per row and this number of nonzeros is assumed to be computed by the ptrT array on GPU. Thus one would need to copy this value back to host to set the 2D launch configuration.  I'll add more comments about the conceptual virtual "wave" of threads assigned to process each row in the routine cal_wave_size(). 

I don't think whether 1D or 2D will make a big difference since a significant amount of  the time is spent in the rocPRIM for computing prefix sum. 

The code assumes warpSize is a power of 2 and the number of threads per thread block is also a multiple of warpSize. I'll add an assertion to check this so that the correctness of the code can be checked during internal rocsolver test. I'll add  SPLITLU_BS1 variable  so that it will be a multiple of warpSize.

It would be interesting to see if  BS1 is modified to say BS1=127, or BS2 = 13, whether the rocsolver library will still work correctly.

CMakeLists.txt
cgmb: Sounds good to me.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,66 +27,239 @@
...
+        // -------------------
+        for(auto k = kstart + lid; (k < kend); k += waveSize)
+        {
+            const auto icol = Mi[k];
+            if(icol == irow)
+            {
+                const auto kdiag = k;
+                nzUarray[irow] = kend - kdiag;
+                nzLarray[irow] = (kdiag - kstart);
+                nzLarray[irow] += 1; // add 1 for unit diagonal
```
tfalders: This is two writes to global memory. It'll be faster if you combine them, i.e. `nzLarray[irow] = (kdiag - kstart)`

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,66 +27,239 @@
...
+
+    SETUP_THREADS(lid, wid, waveSize, nwaves);
 
-    rocblas_int* const nzLp = work;
-    rocblas_int* const nzUp = work + n;
+    auto const tid = hipThreadIdx_x + hipThreadIdx_y * hipBlockDim_x
+        + hipThreadIdx_z * (hipBlockDim_x * hipBlockDim_y);
+    auto const nthreads = (hipBlockDim_x * hipBlockDim_y) * hipBlockDim_z;
+
+    rocblas_int* const diagpos = work;
```
tfalders: You may wish to consider using shared memory instead of global memory for this. That should make it a little faster.

tfalders (commented):
I have just some minor comments on the remainder of the code. There is one comment from my previous review that I'd like to reopen, but other than that I think it's almost ready.

library/src/refact/rocrefact_csrrf_splitlu.hpp
tfalders: I'd still like to see this moved to getMemorySize. Reason being that it doesn't make sense for the function to fail when the cause of the failure can be easily addressed on our side. If size_work is taken to be `max({rocprim_size_bytes, storage_size_bytes_Lp, storage_size_bytes_Up, size_work_LU})`, then we can ensure that `size_work >= storage_size_bytes` and therefore this check is no longer needed and the function won't fail.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I'll add 1 in line 116.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: Added code to check whether the array required can fit in LDS shared memory, otherwise use global memory.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: the code to check storage for rocprim is removed

library/src/include/ideal_sizes.hpp
```diff
@@ -291,3 +291,11 @@
...
 #define TRTRI_BATCH_INTERVALS 32, 245, 1009
 #define TRTRI_BATCH_BLKSIZES 0, 16, 32, 0
+
+/************************** splitlu ***************************************
+*******************************************************************************/
+/*! \brief Determines the size at which rocSOLVER switches from
+    the kernel using a single thread block or using faster algorithm using rocPRIM */
+
+#define SPLITLU_SWITCH_SIZE 64
+#define SPLITLU_BS1 256
```
tfalders: Now that the kernel has been made 2D, I think this can be removed and BS1 can be used instead

library/src/refact/rocrefact_csrrf_splitlu.cpp
```diff
@@ -66,7 +66,14 @@ rocblas_status rocsolver_csrrf_splitlu_impl(rocblas_handle handle,
...
 
-    rocsolver_csrrf_splitlu_getMemorySize<T>(n, nnzT, &size_work);
+    {
+        rocblas_status const istat
+            = rocsolver_csrrf_splitlu_getMemorySize<T>(n, nnzT, ptrT, &size_work);
+        if(istat != rocblas_status_success)
+        {
+            return (istat);
+        }
+    }
```
tfalders: You can simplify this code by using the ROCBLAS_CHECK macro, i.e. `ROCBLAS_CHECK(rocsolver_csrrf_splitlu_getMemorySize<T>(n, nnzT, ptrT, &size_work));`

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,128 +27,336 @@
...
+#ifndef SPLITLU_SWITCH_SIZE
+#define SPLITLU_SWITCH_SIZE 64
+#endif
+
+#ifndef SETUP_THREADS
+#define SETUP_THREADS(lid, wid, waveSize, nwaves)             \
+    {                                                         \
+        waveSize = hipBlockDim_x;                             \
+        nwaves = hipBlockDim_y * hipGridDim_x;                \
+        wid = hipThreadIdx_y + hipBlockIdx_x * hipBlockDim_y; \
```
tfalders, line 47: I'm confused why it's hipBlockDim_y you're using here instead of hipBlockDim_x. How does this work?

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,128 +27,336 @@
...
+    // ------------------------------------------
+    size_t rocprim_size_bytes = 0;
+    void* temp_ptr = nullptr;
+
+    const hipError_t istat = rocprim::inclusive_scan(temp_ptr, rocprim_size_bytes, ptrT, ptrT, n,
+                                                     rocprim::plus<rocblas_int>());
+    if(istat != hipSuccess)
+    {
+        return (rocblas_status_internal_error);
+    }
```
tfalders: You can simplify this using the HIP_CHECK macro

tfalders (approved):
A couple more minor comments, but it's in a state now where I feel comfortable approving.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: The launch configuration is dim3(nblocks,1,1), dim3(nx,ny,1),   where nx is the number of threads in a conceptual "wave",  and there are ny waves per thread block. 

Thus there are a total of (nblocks * ny ) waves.  The "wid"  identification number of the wave is   threadIdx_y +   blockIdx_x * ny,   where ny is BlockDim_y.

library/src/refact/rocrefact_csrrf_splitlu.cpp
EdDAzevedo: I have changed the code to use ROCBLAS_CHECK macro.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I have changed the code to use HIP_CHECK macro.

library/src/include/ideal_sizes.hpp
EdDAzevedo: SPLITLU_BS1 is removed from ideal_sizes.hpp,  refact/rocrefact_csrrf_splitlu.hpp modified to use BS1

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,128 +27,429 @@
...
 #include "rocblas.hpp"
 #include "rocsolver/rocsolver.h"
 #include "rocsparse.hpp"
 
+#ifndef SPLITLU_SWITCH_SIZE
+#define SPLITLU_SWITCH_SIZE 64
+#endif
+
+#ifndef SETUP_THREADS
+#define SETUP_THREADS(lid, wid, waveSize, nwaves)             \
```
cgmb: Could we perhaps call this `SPLITLU_SETUP_THREADS` to reflect that it is specific to this function?

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,128 +27,429 @@
...
+        nnzLrow += 1; // add one for unit diagonal
+
+        {
+            bool const isvalid = (nnzUrow == nzUarray[irow]) && (nnzLrow == nzLarray[irow])
+                && (nnzTrow == (nnzLrow - 1) + nnzUrow);
+
+            if(!isvalid)
+            {
+                printf("irow=%d,nnzTrow=%d,nnzLrow=%d,nzLarray[irow]=%d,nnzUrow=%d,nzUarray=%d\n",
+                       irow, nnzTrow, nnzLrow, nzLarray[irow], nnzUrow, nzUarray[irow]);
```
cgmb, line 127: I think this printf should probably be guarded behind `#ifndef NDEBUG` so it's only in debug builds. However, if you really want to keep it in release builds distributed to end users, it needs to have a message that can be understood by those users.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,128 +27,429 @@
...
+
     if(tid == 0)
     {
-        nnzL = 0;
-        nnzU = 0;
+        rocblas_int nnzL = 0;
+        rocblas_int nnzU = 0;
 
-        for(i = 0; i < n; i++)
+        for(auto irow = 0 * n; irow < n; irow++)
```
cgmb: `auto irow = 0 * n;` is a rather awkward construction. I realize that the multiplication will be optimized away, but this is really just `rocblas_int irow = 0;`. I think that would be more clear. If you really want to use type deduction for the index, I would go with `using index_type = decltype(n); index_type irow = 0;`.

library/src/refact/rocrefact_csrrf_splitlu.hpp
```diff
@@ -27,128 +27,429 @@
...
+    // ------------------------------------------
+    size_t rocprim_size_bytes = 0;
+    void* temp_ptr = nullptr;
+
+    HIP_CHECK(rocprim::inclusive_scan(temp_ptr, rocprim_size_bytes, ptrT, ptrT, n,
+                                      rocprim::plus<rocblas_int>()));
+
+    *size_work = (rocprim_size_bytes + size_work_LU);
+
+    return (rocblas_status_success);
```
cgmb: It's not a big deal, but ideally all team members would write code in a similar enough style to be indistinguishable. The parentheses around return values are unnecessary and don't match the rest of the project, so I'd recommend omitting them.

That said, these have no actual impact on the end result and I don't think it will make the code any less clear for other contributors. This is just a nitpick.

cgmb (commented):
I had a few minor comments. The most important is related to the printf, as that has user-visible impact. The rest are merely small suggestions.

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I'll use the form "return i + j * ld  " instead of "return( i + j * ld )"

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I'll declare the type to use "for( rocblas_int irow = 0".

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I'll guard this using "NDEBUG".

library/src/refact/rocrefact_csrrf_splitlu.hpp
EdDAzevedo: I'll change this to SPLITLU_SETUP_THREADS.

cgmb (approved):

cgmb:
There was a logical conflict in the changes to the CI configuration needed for this PR and #677. It added getRocPRIM only for the static case, but this PR needs it for the dynamic library build as well. You'll need to back out the `getRocPRIM`-related changes from #677 in `.jenkins/common.groovy` to resolve the conflict.

jzuniga-amd (approved):
LGTM. 
Probably better to merge #689 first.

tfalders:
There were some new files added: TAGS, testing_csrrf_splitlu.hpp.dev, and testing_csrrf_splitlu.hpp.org. Are these necessary, or can they be removed?

tfalders:
~I'll do the merge once the tests pass again. I'll merge my getf2 PR first. I don't expect any merge conflicts, so you should be fine to leave this PR as it is.~ EDIT: Actually, scratch that. Apparently my PR has a merge conflict that needs to be resolved.

PR #634, opened 2023-11-22
Load rocsparse using dlopen

cgmb:
This is more or less the same strategy that I expect we'll use for cholmod / suitesparse in hipSOLVER.


cgmb:
If you have any questions, I'll be monitoring this PR. However, I won't be able to address any comments. Maintainers are allowed to edit this pull request, so feel free to push fixes to my branch.

tfalders (approved):
Nice job! I've only skimmed over all the different method definitions/instantiations, but everything else looks good.

By the way, what will happen if a method fails to load but a user tries to call it anyway? We'll produce an error code for rfinfo being null?

CMakeLists.txt
```diff
@@ -127,7 +127,7 @@ endif()
 
 option(BUILD_LIBRARY "Build rocSOLVER library" ON)
 option_opposite(BUILD_LIBRARY SKIP_LIBRARY)
-option(BUILD_WITH_SPARSE "Build rocSOLVER sparse re-factorization and solvers" ON)
+option(BUILD_WITH_SPARSE "Build with rocsparse available at build time" OFF)
```
tfalders: Almost forgot: the option to remove sparse functionality in the install script is `--no-sparse`. I don't think we have an equivalent to enable it. If the default is now going to be to build without sparse functionality, we should add a new option to the install script to enable it.

tfalders: We should probably also update the changelog.

tfalders: I imagine that this option needs to be enabled when we build rocSOLVER statically. Will that be handled by the cmake code? I didn't notice anything that would force that.

library/src/refact/rocsolver_rfinfo.cpp
```diff
@@ -56,15 +62,162 @@
...
     if(!handle)
         return rocblas_status_invalid_handle;
 
     if(!rfinfo)
         return rocblas_status_invalid_pointer;
 
+#ifndef HAVE_ROCSPARSE
+    if(!rocsolver::try_load_rocsparse())
+        return rocblas_status_not_implemented;
+#endif
```
tfalders: I think you might have to move this to the beginning of the function, before the handle check. The sparse tests will skip if this function returns rocblas_status_not_implemented, but that won't happen since the sparse tests are passing nullptr for both arguments.

CMakeLists.txt
tfalders: Now that dlopen is an optional dependency, we should probably document that as well.

cgmb: I've updated the changelog, docs for the rocsparse dependency, and defaults for the CMake option. The dlopen functionality is provided as a part of libc, so I don't think we need to worry about it as a dependency.

cgmb: I suppose we still need `install.sh --sparse`.

tfalders: The option I added to hipSOLVER is `--embed-sparse`. If you'd like something different, I'm open to changing it, just want to make sure rocSOLVER and hipSOLVER use the same name.

cgmb:
> what will happen if a method fails to load but a user tries to call it anyway? We'll produce an error code for rfinfo being null?

The current behaviour is that we'll return `rocblas_status_not_implemented` from `rocsolver_create_rfinfo`. There will be no way to create the rfinfo struct and it will remain untouched. I suppose some users could be confused by a failure if they do not initialize the rfinfo handle to null and does not check the return code of `rocsolver_create_rfinfo` before calling other sparse functions.

qjojo (approved):
Looks good!

PR #637, opened 2023-11-28
Fix refactchol

EdDAzevedo:
Remove argument "beta" from  rf_add_QAQ_kernel and use hipMemsetAsync() to set array "valT" to zero.


library/src/specialized/roclapack_trsm_specialized_kernels.hpp
```diff
@@ -29,6 +29,14 @@
 
 #include "rocsolver_run_specialized_kernels.hpp"
 
+#ifndef HIP_CHECK
+#define HIP_CHECK(fcn)               \
+    {                                \
+        hipError_t istat = (fcn);    \
+        assert(istat == hipSuccess); \
+    }
+#endif
```
tfalders: As I have been recently reminded, assert only works in the debug build. We'll probably have to add a rocblas_status to these TRSM methods and return an internal error if the status is not a success, which is a bit beyond the scope of this PR. I'd recommend moving the changes to this file into a new PR where we can iron out what needs to be done.

tfalders (changes requested):
Nice job! Thanks again for finding the cause of the bug.

I've opened a PR on your local fork to adjust the code style for consistency. Please look it over when you have some time.

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
jzuniga-amd: In fact, the workaround in TRSM should not be needed anymore (with the new hardware/firmware stack). So all this code could be removed completely. I am running the entire test suite on the node to confirm this is the case.

EdDAzevedo: Perhaps there should be a consistent standard HIP_CHECK() macro in a standard header file  for all  of the rocsolver source code?

jzuniga-amd: I ran the entire rocsolver test suite without this TRSM workaround and everything is passing. This code was only temporal and is not needed anymore. It can be safely removed.

tfalders: Even on 940?

jzuniga-amd: right!! 940 and 941 should stay in the condition.

tfalders: I have opened a new PR against @EdDAzevedo's fork to address these issues. Please have a look when you're able.

tfalders (approved):

jzuniga-amd (approved):
LGTM

PR #651, opened 2024-01-12
Improve performance of sorting eigenvalues and eigenvectors using shell sort

EdDAzevedo:
Improve performance of sorting eigenvalues and eigenvectors in "stedc_sort"  kernel using shell sort (instead of selection sort).  The shell sort algorithm is described on wikipedia. 

The shell sort algorithm works on a single thread block so that all threads can be synchronized using __syncthreads().  A single thread is first used to sort the eigenvalues and to generate the permutation vector.   Then repeated swap operations on columns of "C" matrix  containing eigenvector (using all threads in the thread block)  are used to implement the action of permutation vector. 

Note  integer temporary storage of length (n) (for each thread block) will be needed to hold the permutation vector.


library/src/auxiliary/rocauxiliary_bdsvdx.hpp
```diff
@@ -345,9 +345,45 @@ rocblas_status rocsolver_bdsvdx_template(rocblas_handle handle,
...
+            assert(map_array != nullptr);
+            if(map_array == nullptr)
+            {
+                return (rocblas_status_internal_error);
+            };
+        };
         ROCSOLVER_LAUNCH_KERNEL(syevx_sort_eigs<T>, dim3(1, batch_count, 1), dim3(BS1, 1, 1), 0,
                                 stream, ntgk, nsv, Stmp, ntgk, Z, shiftZ, ldz, strideZ, ifail,
-                                strideF, info);
+                                strideF, info, map_array);
```
tfalders: This is an unacceptable way to declare map_array. A better (and easier) way to do it is to pick a single workspace array that will be reused for map_array, ideally one that already has enough space such as isplit or iblock, and stick with it.

To help with readability, the name of the workspace array you pick should be changed, e.g. to isplit_map or iblock_map, across the entire file. Changing it in the associated cpp files is recommended but not strictly necessary. Changing it also in functions that call bdsvdx is a nice-to-have, but I won't complain if it's not done.

library/src/lapack/roclapack_syevx_heevx.hpp
```diff
@@ -341,8 +413,46 @@ rocblas_status rocsolver_syevx_heevx_template(rocblas_handle handle,
         // sort eigenvalues and eigenvectors
         dim3 grid(1, batch_count, 1);
         dim3 threads(BS1, 1, 1);
+
+        rocblas_int* map_array = nullptr;
```
tfalders: Same comment as for bdsvdx. Both iblock and isplit are good candidates to reuse for map_array.

library/src/lapack/roclapack_syevdx_heevdx_inplace.hpp
```diff
@@ -283,8 +283,47 @@ rocblas_status rocsolver_syevdx_heevdx_inplace_template(rocblas_handle handle,
             (T*)work3, (T**)nsplit_workArr);
 
         // sort eigenvalues and eigenvectors
-        ROCSOLVER_LAUNCH_KERNEL(syevx_sort_eigs<T>, grid1, threads1, 0, stream, n, d_nev, W,
-                                strideW, A, shiftA, lda, strideA, (rocblas_int*)nullptr, 0, info);
+        rocblas_int* map_array = nullptr;
```
tfalders: Same comment as for bdsvdx. Both iblock and isplit are good candidates to reuse for map_array.

library/src/lapack/roclapack_syevx_heevx.hpp
```diff
@@ -39,6 +39,89 @@
...
+                                          rocblas_int ldz,
+                                          rocblas_int* ifail)
+{
+    auto const tid = hipThreadIdx_x + hipThreadIdx_y * hipBlockDim_x
+        + hipThreadIdx_z * (hipBlockDim_x * hipBlockDim_y);
+    auto const nthreads = (hipBlockDim_x * hipBlockDim_y) * hipBlockDim_z;
+
+    auto const k_start = tid;
+    auto const k_inc = nthreads;
+    ;
```
tfalders: This semicolon can be removed

library/src/lapack/roclapack_syevx_heevx.hpp
```diff
@@ -39,6 +39,89 @@
...
+    {
+        // ------------------------------------------------------
+        // double check map[] is restored to identity permutation
+        // ------------------------------------------------------
+        __syncthreads();
+        for(auto k = k_start; k < nev; k += k_inc)
+        {
+            assert(map[k] == k);
+        };
+        __syncthreads();
```
tfalders, line 122: Assert only works in the debug build. In the release build, this loop will waste time doing nothing. It should either be placed in a preprocessor macro or removed (or changed to do something in release mode but I'm not sure how to go about that).

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2092,8 +2100,19 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
...
+            // -------------------------------------
+            // reuse array T* tempvect of size n * n
+            // -------------------------------------
+            rocblas_int* work = (rocblas_int*)tempvect;
+
+            auto constexpr max_blocks = 64 * 1000;
+            auto const nblocks = max(1, min(max_blocks, batch_count));
+
+            ROCSOLVER_LAUNCH_KERNEL((stedc_sort<T>), dim3(1, 1, nblocks), dim3(BS1), 0, stream, n,
+                                    D + shiftD, strideD, C, shiftC, ldc, strideC, batch_count, work);
```
tfalders: If I'm reading this correctly, this kernel is always going to be launched with 64000 blocks, which is way too many in the vast majority of cases. I know we've talked about the possibility of batch_count exceeding the maximum number of thread blocks, but this is not a suitable remedy. I would prefer that we stuck with launching batch_count blocks, at least until we've devised a suitable pattern that can be used across the library.

library/src/include/lib_device_helpers.hpp
```diff
@@ -773,3 +784,490 @@ ROCSOLVER_KERNEL void check_singularity(const rocblas_int n,
...
+    // -----------------------------------------------
+    if((n <= 0) || (a == nullptr))
+    {
+        return;
+    };
+
+    // ------------------------
+    // Shell sort (from wikipedia)
+    // Sort an array a[0...(n-1)].
+    // ------------------------
```
tfalders: I'm not sure if deriving code from Wikipedia is okay for us to do or not, and to what extent (if any) we need to differentiate our code from theirs.
It also looks like the comments are verbatim the same as on Wikipedia, which will probably need to be removed or rephrased.
Perhaps @doctorcolinsmith can advise?

library/src/include/lib_device_helpers.hpp
```diff
@@ -773,3 +784,490 @@ ROCSOLVER_KERNEL void check_singularity(const rocblas_int n,
...
+    bool const is_root_thread
+        = (hipThreadIdx_x == 0) && (hipThreadIdx_y == 0) && (hipThreadIdx_z == 0);
+
+    int constexpr ngaps = 8;
+    int const gaps[ngaps] = {701, 301, 132, 57, 23, 10, 4, 1}; // # Ciura gap sequence
+
+    auto const k_start = hipThreadIdx_x + hipThreadIdx_y * hipBlockDim_x
+        + hipThreadIdx_z * (hipBlockDim_x * hipBlockDim_y);
+
+    auto const k_inc = (hipBlockDim_x * hipBlockDim_y) * hipBlockDim_z;
```
tfalders, line 810: It's a little strange for me to see the x, y, and z dimensions being referenced even though the function seems to be 1D. I'm guessing it's to allow the function to be used by 1D, 2D, and 3D kernels. Usually we don't use these built-in HIP functions in our device functions; instead we would pass k_start and k_inc as parameters. Not a big deal, but I thought I'd mention it.

library/src/include/lib_device_helpers.hpp
```diff
@@ -773,3 +784,490 @@ ROCSOLVER_KERNEL void check_singularity(const rocblas_int n,
...
+        {
+            assert(D[k] >= D[k + 1]);
+        };
+    };
+    __syncthreads();
+#endif
+}
+
+template <typename S, typename I>
+__device__ static void shell_sort_descending(const I n, S* a, I* map = nullptr)
```
tfalders: A minor point, but it would be nice to have associated methods grouped together in the file. An example ordering of methods in the file could be:
- shell_sort_ascending
- shell_sort_descending
- shell_sort
- selection_sort_ascending
- selection_sort_descending
- selection_sort
- permute_vector

tfalders (changes requested):
Looks pretty good, though I have a number of comments to be addressed.

library/src/auxiliary/rocauxiliary_bdsvdx.hpp
EdDAzevedo: The original method is robust with defensive programming in the sense that it will always work and does not require detailed knowledge of other parts of implementation. If say work1_iwork[]  has sufficient storage, then it will always use work1_iwork[]. It also tries to use storage arrays named "work" and try not use named arrays, in case those arrays might contain information that needs to be preserved across different kernel launches.   It is also future proof in the sense in case in future  if the code reduces the storage for isplit[] to less than n integers for each batch item,  the original method will still work since it will try to use another temporary array with sufficient storage.  A general name like "map_array" has the advantage that if in future isplit[] has less than n integers and we need to use say iblock[] instead, then we'll need to modify the code to rename isplit_map[] to iblock_map[].  The original method also has insignificant impact to performance. 

Ideally, one can call the stream-oriented pool memory allocator to allocate and deallocate as needed. This can greatly simplify the code and improve robustness and maintainability.
--------------------
I'll modify the code to always use isplit[] and rename map_array[] to isplit_map[].

library/src/lapack/roclapack_syevx_heevx.hpp
EdDAzevedo: I'll add #ifdef NDEBUG  to perform the extra check only during testing and not in production code.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: The launch configuration will use the min(  max_blocks, batch_count), so if batch_count is say 47, then it will use only 47 blocks. If the batch_count is say  1,000,000,  it will use only max_blocks of 64000 thread blocks. Internally the kernel will use a loop to process 64000 batch items concurrently in each iteration.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: I'll remove the comments.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: Yes, it is defensive programming for robustness that this library routine will still work even with different launch configuration of 1D, or 2D or 3D grid of threads in a thread block. The kernel may be launched with say a 3D grid of threads in a thread block, but this library routine will still correctly perform the sorting.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: The code is rearranged.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: Another possible advantage in restricting the launch configuration  is if batch_count is large, we need scratch storage that scale as  O( min( batch_count,max_blocks))  and not O( batch_count ) since we know there will be at most O(max_blocks ) active concurrent computations.    

If the machine has only about say 128  Compute Units (CU), then maybe launch configuration of say max_blocks=512  blocks are already sufficient to keep the machine fully occupied. Using a smaller number of max_blocks may reduce the amount of scratch space required to scaling as O(512) instead of O(batch_count).

library/src/lapack/roclapack_syevx_heevx.hpp
EdDAzevedo: I have removed that extra semicolon.

library/src/lapack/roclapack_syevdx_heevdx_inplace.hpp
EdDAzevedo: I'll rename map_array[] to isplit_map[] and assign  isplit[] to isplit_map[].

library/src/auxiliary/rocauxiliary_stedc.hpp
tfalders: Ah, somehow I missed the min in that call. That makes a lot more sense now.

This does bring up some different comments though, namely:
1. There are other kernels in the function that are not using this pattern. If batch_count exceeds the maximum number of blocks, then they will cause the function to fail before it ever reaches this kernel.
2. 64000 is what we call a "magic number". It would be better if it were defined as a constant in ideal_sizes.hpp, or better yet, queried from the device properties.

Whenever we change a pattern like this, I prefer to do it all at once across the entire library in a separate PR. So, while I still recommend sticking with launching batch_count blocks, this is with the understanding that we will later adopt a pattern similar to this across the entire library.

EdDAzevedo: OK. I'll remote max_blocks.  

I think the main advantage is in reducing the amount of scratch space required to scaling as O(max_blocks) instead of scaling as O(batch_count), especially when the value of batch_count is large.  In the case of sorting, we need O(n) temporary storage to hold the permutation vector. By restricting the launch configuration to only max_blocks thread blocks, it can reduce the storage requirement from "n * batch_count" to only "n * max_blocks".

According to this web page, in cuda the y and z grid dimensions have a limit of 65535. So it may be  possible that a hip program using hipsolver library with batch_count larger than this limit may work on AMD gpu but may fail on cuda.

https://en.wikipedia.org/wiki/CUDA#Version_features_and_specifications

tfalders:
The changes to map_array are an improvement, though not quite what I had in mind. I've opened a PR on your fork with my suggested changes.

tfalders (approved):
Looks good! A 2nd review will also be needed, and it would nice to have a go-ahead from @doctorcolinsmith, but after that we can squash-and-merge and get this into the library.

doctorcolinsmith:
@tfalders No concerns from me.

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -680,6 +680,105 @@ ROCSOLVER_KERNEL void bdsqr_init(const rocblas_int n,
 }
 
 /** BDSQR_SORT sorts the singular values and vectors by selection sort if applicable. **/
```
jzuniga-amd: this comment is no longer valid and not in the right place

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -680,6 +680,105 @@ ROCSOLVER_KERNEL void bdsqr_init(const rocblas_int n,
...
+                                          const I ldu,
+                                          const I nc,
+                                          T* C,
+                                          const I ldc,
+                                          I* map)
+{
+    // --------------------------------------------
+    // perform swaps to implement permutation vector
+    // the permutation vector will be restored to the
+    // identity permutation 0,1,2,...
```
jzuniga-amd: You can add the description of the function right before its declaration as for all the other functions in the file. Just to keep consistency.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -51,12 +53,14 @@ typedef enum rocsolver_stedc_mode_
 template <rocsolver_stedc_mode MODE>
 __host__ __device__ inline rocblas_int stedc_num_levels(const rocblas_int n);
 
-/***************** Device auxiliary functions *****************************************/
+/***************** Device auxiliary functions
+ * *****************************************/
```
jzuniga-amd: there is a misplaced end of line here

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -663,14 +686,15 @@ __host__ __device__ inline rocblas_int stedc_num_levels<rocsolver_stedc_mode_qr>
     return levels;
 }
 
-/*************** Main kernels *********************************************************/
+/*************** Main kernels
+ * *********************************************************/
```
jzuniga-amd, line 689: same here

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1928,10 +1966,14 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
...
         local_gemm<BATCHED, STRIDED, T>(handle, n, C, shiftC, ldc, strideC, tempvect, tempgemm,
-                                        (S*)work_stack, 0, ldt, strideT, batch_count, workArr);
-
-        ROCSOLVER_LAUNCH_KERNEL((stedc_sort<T>), dim3(batch_count), dim3(1), 0, stream, n,
-                                D + shiftD, strideD, C, shiftC, ldc, strideC);
+                                        static_cast<S*>(work_stack), 0, ldt, strideT, batch_count,
+                                        workArr);
+
+        // finally sort eigenvalues and eigenvectors
+        auto const nblocks = max(1, batch_count);
```
jzuniga-amd: this `max` takes some unnecessary extra time. batch_count cannot be less than 1 at this point (invalid values should be detected with the check_argument helper of the batched functions that may call stedc)

library/src/auxiliary/rocauxiliary_stedcj.hpp
```diff
@@ -35,20 +37,22 @@
 
 #define MAXSWEEPS 20 // Max number of sweeps for Jacobi solver (when used)
 
-/***************** Device auxiliary functions *****************************************/
+/***************** Device auxiliary functions
+ * *****************************************/
```
jzuniga-amd: extra end line again

library/src/auxiliary/rocauxiliary_stedcj.hpp
```diff
@@ -273,7 +278,8 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
     }
 }
 
-/******************* Host functions ********************************************/
+/******************* Host functions
+ * ********************************************/
```
jzuniga-amd: idem

jzuniga-amd (commented):
Just a few very minor comments below

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -754,46 +866,57 @@ ROCSOLVER_KERNEL void bdsqr_sort(const rocblas_int n,
...
-            {
-                m = j;
-                p = D[j];
-            }
-        }
+        shell_sort_descending(n, D, map);
         __syncthreads();
-
-        if(m != i)
+        bdsqr_permute_swap(n, nv, V, ldv, nu, U, ldu, nc, C, ldc, map);
```
jzuniga-amd: You need to call the `permute_swap` function only `if(nu || nv || nc)`.
You can save some cycles this way because `permute_swap` will still restore map into the identity even if there are no vectors to permutate

library/src/auxiliary/rocauxiliary_bdsqr.hpp
EdDAzevedo: I have updated the position of the comment and mention shell sort.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: I have joined the two lines. The line split may be due to clang-format trying to reformat code and long comments to be less than 80 columns.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: I have removed the max.

library/src/auxiliary/rocauxiliary_stedcj.hpp
EdDAzevedo: I have joined the two lines.

library/src/auxiliary/rocauxiliary_stedcj.hpp
EdDAzevedo: I have joined the two lines.

library/src/auxiliary/rocauxiliary_bdsqr.hpp
EdDAzevedo: add test for "bool const need_swap = (nu > 0) || (nv > 0) || (nc > 0);" before calling permute_swap().

library/src/auxiliary/rocauxiliary_bdsqr.hpp
EdDAzevedo: I have moved the description of the function to right before its declaration.

jzuniga-amd (approved):
LGTM. Don't forget to merge Troy's changes to stedcj.
Also, please add an item in the Changelog under the **Optimized** title. 
"Improved performance of sorting algorithms used in different eiegnsolvers" or something like this.

library/src/include/lib_device_helpers.hpp
```diff
@@ -773,3 +784,452 @@ ROCSOLVER_KERNEL void check_singularity(const rocblas_int n,
...
+                    {
+                        map[j] = itemp;
+                    };
+                };
+            };
+        };
+    };
+    __syncthreads();
+#ifdef NDEBUG
+#else
```
cgmb, line 859: `#ifndef`?

library/src/include/lib_device_helpers.hpp
```diff
@@ -773,3 +784,452 @@ ROCSOLVER_KERNEL void check_singularity(const rocblas_int n,
...
+    //
+    // Note: performs in a thread block
+    // --------------------------------------------
+    {
+        bool const has_work = (n >= 1) && (C != nullptr) && (ldc >= 1) && (map != nullptr);
+        if(!has_work)
+        {
+            return;
+        };
+    };
```
cgmb: It's not important enough to be worth going back and changing, but in the future please match the style of the surrounding code and omit these semi-colons. The ideal would be for all contributors to have a consistent style.

library/src/lapack/roclapack_syevx_heevx.hpp
```diff
@@ -50,61 +135,43 @@ ROCSOLVER_KERNEL void __launch_bounds__(BS1) syevx_sort_eigs(const rocblas_int n
...
-        for(jj = j + 1; jj < nev; jj++)
-        {
-            if(W[jj] < tmp1)
-            {
-                i = jj;
-                tmp1 = W[jj];
-            }
-        }
-        __syncthreads();
+    assert(nev <= n);
```
cgmb, line 160: IMO, this may be good place for `ROCSOLVER_ASSUME_X(condition, message)`. It resolves to [`__builtin_assume`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-assume) in release builds and `assert` in debug builds.

Although, I think I might have broke the ability to use it in device code. So, maybe this isn't a suggestion for right now, but for next time...

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -108,6 +111,11 @@ __device__ void seq_eval(const rocblas_int type,
         gout = k;
         hout = k + 1;
     }
+    else
+    {
+        // unexpected value for type, something is wrong
+        assert(false);
+    }
```
cgmb, line 118: There's a `ROCSOLVER_UNREACHABLE_X(message)` macro for this. It resolves to [`__builtin_unreachable`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-unreachable) in release mode and `assert(false)` in debug mode...

Although, I may broken it in device code by trying to be too clever with displaying the error message... so maybe that's a suggestion for once I fix that.

cgmb (approved):
I have only superficial comments.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: I was trying to avoid double negative  "if not defined not debug"

library/src/lapack/roclapack_syevx_heevx.hpp
EdDAzevedo: In the url https://clang.llvm.org/docs/LanguageExtensions.html#builtin-assume, it seems to me "__builtin_assume" is intended for clang compiler optimization and perhaps not intended for the purpose of checking correctness. There is mention "If the condition is violated during execution, the behavior is undefined."

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: It seems to me "__builtin_unreachable" is intended for clang compiler optimization. It may not be intended for checking for coding error.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: I have removed the semi-colons.

randyh62 (commented):
Looks good to me.

PR #652, opened 2024-01-13
Add cmake ninja compile and link jobs

trixirt:
Building may fail or thrash when memory is exhausted. Allow the build to fine tune the compile and link jobs with these cmake options

  -DROCSOLVER_PARALLEL_COMPILE_JOBS=<num>
  -DROCSOLVER_PARALLEL_LINK_JOBS=<num>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,32 @@ if(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)
   )
 endif()
 
+#
+# Seperate linking jobs from compiling
+# Too many concurrent linking jobs can break the build
+# Copied from LLVM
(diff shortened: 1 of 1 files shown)


cgmb:
Interesting. Thanks for the suggestion, Tom. We're also exploring options to reduce the library size, which may reduce the linker memory usage.

cgmb:
I'm glad that this patch worked well for you, but I'm not comfortable copying code from LLVM without also copying across the applicable LLVM license information. On the other hand, this functionality does not seem substantial enough to be worth adding a new license to rocSOLVER (even if only at build-time). I also wonder if this functionality might be better moved into CMake itself.

In any case, the underlying problem being addressed in this PR seems like a minor issue that will disappear if we can reduce the size of librocsolver.so. However, if you feel that this patch is worth reconsidering, please reopen the PR and we can further discuss the matter.

PR #654, opened 2024-01-17
Added stedcj

tfalders:
Adds stedcj to the internal API


clients/common/testing_stedcj.cpp
```diff
@@ -0,0 +1,32 @@
+/* **************************************************************************
+ * Copyright (C) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
```
jzuniga-amd: This is a new file, maybe just put Copyright (C) 2024

clients/include/testing_stedcj.hpp
```diff
@@ -0,0 +1,580 @@
+/* **************************************************************************
+ * Copyright (C) 2020-2024 Advanced Micro Devices, Inc. All rights reserved.
```
jzuniga-amd: 2024

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -42,9 +41,16 @@
 #define STEDC_BDIM 512 // Number of threads per thread-block used in main stedc kernels
 #define MAXITERS 50 // Max number of iterations for root finding method
 #define MAXSWEEPS 20 // Max number of sweeps for Jacobi solver (when used)
```
jzuniga-amd: MAXSWEEPS probably not needed here anymore ?

library/src/auxiliary/rocauxiliary_stedcj.cpp
```diff
@@ -0,0 +1,165 @@
+/* **************************************************************************
+ * Copyright (C) 2021-2024 Advanced Micro Devices, Inc. All rights reserved.
```
jzuniga-amd: 2024

library/src/auxiliary/rocauxiliary_stedcj.cpp
```diff
@@ -0,0 +1,165 @@
...
+                                     const rocblas_int ldc,
+                                     rocblas_int* info)
+{
+    ROCSOLVER_ENTER_TOP("stedcj", "--evect", evect, "-n", n, "--ldc", ldc);
+
+    if(!handle)
+        return rocblas_status_invalid_handle;
+
+    // argument checking
+    rocblas_status st = rocsolver_stedcj_argCheck(handle, evect, n, D, E, C, ldc, info);
```
jzuniga-amd: maybe can reuse `stedc_argCheck` ?
signatures are identical

library/src/auxiliary/rocauxiliary_stedcj.hpp
```diff
@@ -0,0 +1,479 @@
+/************************************************************************
+ * Derived from the BSD3-licensed
+ * LAPACK routine (version 3.7.0) --
+ *     Univ. of Tennessee, Univ. of California Berkeley,
+ *     Univ. of Colorado Denver and NAG Ltd..
+ *     December 2016
+ * Copyright (C) 2021-2024 Advanced Micro Devices, Inc. All rights reserved.
```
jzuniga-amd: Year 2024. 
Also maybe we can remove the LAPACK notice. This routine does not exist in LAPACK.

library/src/auxiliary/rocauxiliary_stedcj.hpp
```diff
@@ -0,0 +1,479 @@
...
+    *size_splits = sizeof(rocblas_int) * (5 * n + 2) * batch_count;
+
+    // size for temporary diagonal and rank-1 modif vector
+    *size_tmpz = sizeof(S) * (2 * n) * batch_count;
+}
+
+//--------------------------------------------------------------------------------------//
+/** This helper check argument correctness for stedcj API **/
+template <typename T, typename S>
+rocblas_status rocsolver_stedcj_argCheck(rocblas_handle handle,
```
jzuniga-amd: This helper is the same as in `stedc`. Maybe we can reduce repetition and use just that one.

jzuniga-amd (changes requested):
Looks good. Cleaner and more organized than having all combined in a single file. 
(We can probably re-use some code on the clients side, but we can leave it for later; I want to do a thorough revision of all the clients code anyways).
Just a few minor comments below. 
Also, as we are leaving the test code, please add it to the bench client as well (`rocsolver_dispatcher.hpp`, etc). It would be useful to call `stedcj` from the bench client to do direct comparisons with `stedc` or `stedcx` for example.

jzuniga-amd (approved):
LGTM

qjojo (approved):

clients/include/testing_stedcj.hpp
```diff
@@ -0,0 +1,580 @@
...
+                        double* cpu_time_used,
+                        const rocblas_int hot_calls,
+                        const int profile,
+                        const bool profile_kernels,
+                        const bool perf)
+{
+    constexpr bool COMPLEX = rocblas_is_complex<T>;
+    using S = decltype(std::real(T{}));
+
+    int lgn = floor(log(n - 1) / log(2)) + 1;
```
cgmb, line 375: We should really make a helper function for this operation, as this has shown up a couple times now.

cgmb (approved):
Trivial comment. LGTM.

clients/include/testing_stedcj.hpp
EdDAzevedo: I think there is a log2() function in <cmath>. You can also ask AI to generate a function to compute  the next power of 2 using bit operations.

PR #655, opened 2024-01-24
Add codeowners for documentation

samjwu:
Relates to https://github.com/ROCm/ROCm/issues/2832
Adding documentation team as codeowners for documentation.
Goal is to make this group of code owners reviewers when merging documentation changes. The header directory is related to documentation and thus changes to that are also flagged for documentation review.

**Please note that this requires the team to be added as a role with Write permissions.**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,7 @@
 * @jzuniga-amd @tfalders @cgmb @qjojo
+# Documentation files
+docs/* @ROCm/rocm-documentation
+*.md @ROCm/rocm-documentation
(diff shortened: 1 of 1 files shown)


cgmb (approved):

cgmb (dismissed):
Could you remove executable permissions from the code owner file? I didn't notice that at first, but you've changed the file permissions.

samjwu:
> Could you remove executable permissions from the code owner file? I didn't notice that at first, but you've changed the file permissions.

Good catch; I have restored file permissions to its original settings.

cgmb (approved):

PR #657, opened 2024-01-25
Reduce rocblas minimum version to 4.1

cgmb:
The rocSOLVER library doesn't use any functions from rocBLAS 4.2, and the rocBLAS library now promises interface stability for the functions that rocSOLVER uses (since 4.0). The minimum version will only need to be increased when rocSOLVER is changed to use a rocBLAS function that was added after rocBLAS 4.1.

I had rocblas 4.1 installed when I built the latest rocSOLVER, and I discovered that I couldn't install the package because of the version check.


tfalders:
Perhaps @qjojo can comment more, but my understanding was that there are some functions that will need to be added to rocBLAS as part of our efforts to add 64-bit support, and these would be expected in the next release.

tfalders (approved):
Per today's discussion, the code as-is only requires 4.1, and we can update to 4.2 once the 64-bit APIs are added.

jzuniga-amd (approved):

PR #658, opened 2024-01-26
Drop gfx803 and gfx900 from default build targets

cgmb:
The rocBLAS library has dropped these targets from its default build targets. While users can still build for those targets by specifying them explicitly, there's little point enabling them in the official binary releases when they are not enabled in rocSOLVER dependencies.
    
(cherry picked from commit 2c64a42b390696fe5047360cbdfa3f59e5cfdbc8)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Full documentation for rocSOLVER is available at [rocsolver.readthedocs.io](http
 ### Changed
 - Relaxed array length requirements for GESVDX with `rocblas_srange_index`.
 
+### Removed
+- Removed gfx803 and gfx900 from default build targets.
+
 ### Fixed
 - Corrected singular vector normalization in BDSVDX and GESVDX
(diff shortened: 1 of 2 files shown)


jzuniga-amd (approved):

tfalders (approved):

cgmb:
Hmmm, there's a failure on gfx908. It is clearly unrelated, but it does need further investigation.
```
[ RUN      ] checkin_lapack/SYEVX.strided_batched__float/37
<...>/rocsolver/clients/include/testing_syevx_heevx.hpp:345: Failure
  hNev[b][0]
    Which is: 50
  hNevRes[b][0]
    Which is: 0
where b = 2

<...>/rocsolver/clients/include/testing_syevx_heevx.hpp:757: Failure
Expected: ((max_error)) <= ((2 * n)*get_epsilon<T>()), actual: 1 vs 1.1920928955078125e-05

[  FAILED  ] checkin_lapack/SYEVX.strided_batched__float/37, where GetParam() = ({ 50, 60, 50, -15, -5, 20, 30 }, { V, A, U }) (9 ms)
```

PR #665, opened 2024-02-06
optimization of Cholesky factorization

EdDAzevedo:
This pull request is related to optimizing the performance of Cholesky factorization by introducing a new implementation of "potf2_lds" to load a small N by N matrix (N <= POTRF_BLOCKSIZE). The small matrix is loaded into LDS shared memory, and then  Cholesky factorization is performed with a single kernel launch. Note that each batch item is factored in one thread block.


tfalders:
I've taken the quickest of looks at the PR, and opened a PR on your fork to move the new kernel to the specialized kernels folder. I've also noticed some intermittent test failures when running tests on my local machine. My guess is that it's a synchronization issue, but I haven't looked close enough to know where you'd need a syncthreads.

tfalders:
I've done a full review and opened a PR on your fork with my suggested changes. After it is merged in, I will approve this PR.

tfalders (approved):
LGTM!

tfalders:
It looks like my changes introduced a memory access fault into SYGV. I'll take a look at it after the weekend.

tfalders:
I've opened a new PR on your fork that should fix the memory access faults my changes introduced.

library/src/lapack/roclapack_potf2.hpp
```diff
@@ -117,6 +118,13 @@ void rocsolver_potf2_getMemorySize(const rocblas_int n,
     // size of scalars (constants)
     *size_scalars = sizeof(T) * 3;
 
+    if(n <= POTRF_BLOCKSIZE(T))
```
jzuniga-amd: Strictly speaking, this condition should use another constant (POTF2_MAX_SMALL_SIZE or something like this) in case future tuning reveals that the size at which the small specialized kernel stops being effective is different from the block size in POTRF.

library/src/lapack/roclapack_potf2.hpp
```diff
@@ -196,77 +204,88 @@ rocblas_status rocsolver_potf2_template(rocblas_handle handle,
...
-                                    shiftA, strideA, idx2D(j, j, lda), j, pivots, info);
+        // ----------------------
+        // use specialized kernel
+        // ----------------------
+        potf2_run_small<T>(handle, uplo, n, A, shiftA, lda, strideA, info, batch_count);
+    }
+    else
+    {
+        // (TODO: When the matrix is detected to be non positive definite, we need
+        // to
```
jzuniga-amd: remove the new line ??

library/src/lapack/roclapack_potrf.hpp
```diff
@@ -74,8 +116,9 @@ void rocsolver_potrf_getMemorySize(const rocblas_int n,
         return;
     }
 
-    rocblas_int jb = POTRF_BLOCKSIZE;
-    if(n <= POTRF_POTF2_SWITCHSIZE)
+    // rocblas_int nb = POTRF_BLOCKSIZE(T);
```
jzuniga-amd: is this comment intended?

library/src/lapack/roclapack_potrf.hpp
```diff
@@ -156,8 +204,9 @@ rocblas_status rocsolver_potrf_template(rocblas_handle handle,
 
     // if the matrix is small, use the unblocked (BLAS-levelII) variant of the
     // algorithm
-    rocblas_int nb = POTRF_BLOCKSIZE;
-    if(n <= POTRF_POTF2_SWITCHSIZE)
+    // rocblas_int nb = POTRF_BLOCKSIZE(T);
```
jzuniga-amd: same here ?

library/src/lapack/roclapack_potrf.hpp
```diff
@@ -37,6 +37,48 @@
...
+    {
+        auto const lds_size = get_lds_size();
+        bool const is_ok = ((sizeof(T) * (nb_max * (nb_max + 1)) / 2) <= lds_size);
+        assert(is_ok);
+    }
+#endif
+
+    rocblas_int const npass = iceil(n, nb_max);
+    rocblas_int const nb = iceil(n, npass);
+    return (nb);
```
jzuniga-amd: I did not see any performance improvements when using this variable `nb` vs the fixed `nb_max`. More tuning will be required in the future to determine the ideal block size as a function of `n`.  

My main problem with this approach is that `nb` could be very different from `nb_max`. If `nb_max` is determined to optimize some criteria (in this case the use of LDS, for example), then you may want to factorize as many blocks as possible of that size...

Take for example `n = 518` and `nb_max = 128`; this gives `nb = 104`. If using `nb_max`, the algorithm will factorize four optimal blocks of 128 columns and one sub-optimal block with only 6 columns. On the other hand, using `nb`, it will factorize five sub-optimal blocks, four with 104 columns and one with 102. 
(Something to think about).

jzuniga-amd (approved):
LGTM. 
If no more tuning will be tried for now, this is ready to merge. 

A few minor comments below. Also please add an entry to the changelog; something like "improved performance of Cholesky factorization" or something like this.

library/src/lapack/roclapack_potf2.hpp
EdDAzevedo: I'll keep POTRF_BLOCKSIZE(T) for now  to make the code simpler and consider POTF2_MAX_SMALL_SIZE in future in case this is a performance issue.

library/src/lapack/roclapack_potf2.hpp
EdDAzevedo: I'll remove the new line.

library/src/lapack/roclapack_potrf.hpp
EdDAzevedo: I'll change the code back to using a fixed block size. If the last panel is size k, then the amount of work required is only factorization of a small k by k matrix.  This should have only a small overall impact.

library/src/lapack/roclapack_potrf.hpp
EdDAzevedo: I'll change the code to use a fixed block size.

library/src/lapack/roclapack_potrf.hpp
EdDAzevedo: I'll change the code to use a fixed block size.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Changed
 - Renamed install script arguments of the form *_dir to *-path. Arguments of the form *_dir remain functional for
   backwards compatibility.
+- Improved performance of Cholesky factorization.
```
jzuniga-amd: It should go under ###Optimized

EdDAzevedo: I'll make the change to go under ###Optimized.

PR #666, opened 2024-02-07
Add ifndef guards to macros

cgmb:
This makes it easier to set the ideal sizes at build time. For example, users could set `export CXXFLAGS="-DGEQxF_BLOCKSIZE=128"` in their environment before calling CMake.

In the future, it might be nice to display information about compiler options set when building rocsolver in the rocsolver-bench output.

### DIFF
--- a/library/src/include/ideal_sizes.hpp
+++ b/library/src/include/ideal_sizes.hpp
@@ -301,8 +301,9 @@
 #define GETF2_SSKER_MAX_N 64 //always <= wavefront and <= GETF2_SPKER_MAX_N
 #endif
 #ifndef GETF2_OPTIM_NGRP
-#define GETF2_OPTIM_NGRP #endif
-16, 15, 8, 8, 8, 8, 8, 8, 6, 6, 4, 4, 4, 4, 4, 4, 3, 3, 3, 3, 3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2
+#define GETF2_OPTIM_NGRP \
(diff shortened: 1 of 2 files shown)


library/src/include/ideal_sizes.hpp
```diff
@@ -226,68 +272,161 @@
...
 #define GETF2_SSKER_MAX_M 512 //always <= 512 and <= GETF2_SPKER_MAX_M
+#endif
+#ifndef GETF2_SSKER_MAX_N
 #define GETF2_SSKER_MAX_N 64 //always <= wavefront and <= GETF2_SPKER_MAX_N
-#define GETF2_OPTIM_NGRP \
-    16, 15, 8, 8, 8, 8, 8, 8, 6, 6, 4, 4, 4, 4, 4, 4, 3, 3, 3, 3, 3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2
+#endif
+#ifndef GETF2_OPTIM_NGRP
+#define GETF2_OPTIM_NGRP #endif
+16, 15, 8, 8, 8, 8, 8, 8, 6, 6, 4, 4, 4, 4, 4, 4, 3, 3, 3, 3, 3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2
```
tfalders: This looks iffy. The #endif appears before the actual definition.

tfalders (approved):
Looks good! (Aside from one definition)

library/src/include/ideal_sizes.hpp
cgmb: Thanks for catching that.

cgmb: I'd used
```
sed -i 's@#define \(\w\+\) \(.*\)@#ifndef \1\n#define \1 \2\n#endif@' ./library/src/include/ideal_sizes.hpp
```
to make the changes, but that pattern didn't account for multi-line macros.

jzuniga-amd (approved):

PR #667, opened 2024-02-07
cmake: add fmt/rocsparse as package deps if building static lib

evetsso:
When a static library is built, no linking is done.  (Basically static libraries are just archives of object files.)

So if `libfoo.a` depends on `libbar.a`, CMake just has to remember that when you link `libfoo.a`, it also needs to pull in `libbar.a`.  The problem is that CMake only knows about `libbar.a` if a `find_package(bar)` is done.  This PR fixes our generated `foo-config.cmake` file to find this dependency, when we're building a static library.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -321,11 +321,11 @@ if(BUILD_WITH_SPARSE)
 endif()
 
 if(ROCSOLVER_EMBED_FMT)
-  target_link_libraries(rocsolver PRIVATE fmt::fmt-header-only)
+  target_link_libraries(rocsolver PRIVATE $<BUILD_INTERFACE:fmt::fmt-header-only>)
 else()
+  list(APPEND static_depends PACKAGE fmt)
   target_link_libraries(rocsolver PRIVATE fmt::fmt)
 endif()
-list(APPEND static_depends PACKAGE fmt)
 
 # In ROCm 4.0 and earlier, the default maximum threads per block is 256
 target_compile_options(rocsolver PRIVATE --gpu-max-threads-per-block=1024)
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -309,7 +309,7 @@ target_link_libraries(rocsolver
     $<$<PLATFORM_ID:Linux>:--unwindlib=libgcc>
 )
 
-set(static_depends)
+set(static_depends PACKAGE rocblas)
 
 if(BUILD_WITH_SPARSE)
   target_link_libraries(rocsolver PRIVATE roc::rocsparse)
@@ -421,7 +421,6 @@ rocm_export_targets(
   TARGETS roc::rocsolver
   DEPENDS
     PACKAGE hip
-    PACKAGE rocblas
   STATIC_DEPENDS
     ${static_depends}
   NAMESPACE roc::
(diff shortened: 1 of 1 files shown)


cgmb (approved):

tfalders (approved):

library/src/CMakeLists.txt
```diff
@@ -322,6 +325,7 @@ if(ROCSOLVER_EMBED_FMT)
 else()
   target_link_libraries(rocsolver PRIVATE fmt::fmt)
 endif()
+list(APPEND static_depends PACKAGE fmt)
```
TorreZuk: I would have thought if ROCSOLVER_EMBED_FMT is true there is no need to later link against fmt it is already internalized?

library/src/CMakeLists.txt
```diff
@@ -418,6 +422,8 @@ rocm_export_targets(
   DEPENDS
     PACKAGE hip
     PACKAGE rocblas
+  STATIC_DEPENDS
+    ${static_depends}
```
TorreZuk: Also expected rocblas in static_depends, and why isn't rocsparse in DEPENDS section?

TorreZuk (commented):
Assume this is tested outside CI ?

library/src/CMakeLists.txt
evetsso: It's not internalized "enough".  CMake will put `fmt::fmt-header-only` into the `INTERFACE_LINK_LIBRARIES` of the generated rocsolver-targets.cmake file in that situation, so downstream users of rocSOLVER will hit errors like:

```
CMake Error at /prefix/lib/cmake/rocsolver/rocsolver-targets.cmake:61 (set_target_properties):
  The link interface of target "roc::rocsolver" contains:

    fmt::fmt-header-only

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

unless they happen to have done their own `find_package(fmt)`.

library/src/CMakeLists.txt
evetsso: DEPENDS is stronger than STATIC_DEPENDS, so it covers both cases for rocblas to only be in DEPENDS.

I can add rocsparse to DEPENDS, but it looks like that should be conditional on BUILD_WITH_SPARSE?

TorreZuk: Sounds good on DEPENDS handling both.   Maybe list new for ${shared_depends} , agree should be conditional on BUILD_WITH

library/src/CMakeLists.txt
cgmb: It's possible to hide it with `$<BUILD_INTERFACE:fmt::fmt-header-only>`. We do that for the rocsolver-common target already.

evetsso: @cgmb Can you point to an example?  In rocFFT we're using object libraries internally and I didn't want those exported.  So instead of doing `target_link_libraries` I'm setting the LINK_LIBRARIES on the target directly.  I think that would work here but I don't think that's what you had in mind.

cgmb: https://github.com/ROCm/rocSOLVER/blob/2fe99038667e51a9d8700d31b6fcdd1afb848aad/library/src/CMakeLists.txt#L306

cgmb (approved):

evetsso:
> Assume this is tested outside CI ?

I've run the rocSOLVER tests manually on an all-static math libs build and they pass.  Obviously it would be better if math-ci can run them automatically but we might want to merge PRs like this just so that everything builds to begin with.

PR #668, opened 2024-02-08
Add rocsolver team as code owners for docs

cgmb:
When a file matches multiple patterns in the CODEOWNERS file, it is assigned the code owners listed in the last matching pattern (not the union of all code owners in all matching patterns). This means that the additions to CODEOWNERS in 9da4ca8be567adfebafafa3caa2af9c27163a9fd implicitly removed the rocsolver team members from documentation files.


jzuniga-amd (approved):
Good catch!

tfalders (approved):

PR #670, opened 2024-02-09
Rename install script arguments

tfalders:
This PR renames the following arguments in install.sh:
1. rocblas_dir -> rocblas-path
2. rocsolver_dir -> rocsolver-path
3. rocsparse_dir -> rocsparse-path

This is to bring more consistency between rocSOLVER's install script with hipBLAS's and hipSOLVER's. I've also used this naming convention for https://github.com/ROCm/rocSPARSE-internal/pull/729.

The existing arguments names can still be used, but they aren't documented in the help message.

We might also want to apply the same changes to lib_dir, build_dir, and install_dir. Let me know if you think that's a good idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Added
 ### Optimized
 ### Changed
+- Renamed install script arguments of the form *_dir to *-path. Arguments of the form *_dir remain functional for
+  backwards compatibility.
+
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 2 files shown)


cgmb (approved):
Note that you can also export `rocblas_DIR`, `rocsolver_DIR`, and `rocsparse_DIR` environment variables before calling ./install.sh to do the same thing as the `--<lib>-path` flags.

jzuniga-amd (approved):

jzuniga-amd:
Is this worth mentioning in the CHANGELOG ?

tfalders:
> Is this worth mentioning in the CHANGELOG ?

Sure, I can add that in

PR #672, opened 2024-02-14
Switch from using hipcc to using Clang compiler

lawruble13:
Summary of proposed changes:
-  Change the default compiler from hipcc to amdclang
  - As hipcc is an increasingly thin wrapper around clang, we want to reduce complexity and switch to just use clang as the default.
- Users may still manually specify the compilers using the `CXX` and `CC` environment variables.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -1,5 +1,5 @@
 # ##########################################################################
-# Copyright (C) 2021-2023 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2021-2024 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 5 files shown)


cgmb (commented):
Please also update `docs/userguide/install.rst`.

randyh62 (commented):
Looks good to me

install.sh
```diff
@@ -480,8 +480,6 @@ case "${ID}" in
   ;;
 esac
 
-export CXX="hipcc"
-export CC="clang"
 export FC="gfortran"
```
cgmb: I'm not sure about 20e9de6937a810995a9bdab0e893b50f9d968f9a. We set these variables in `install.sh` to ensure that we use the same compiler when building the dependencies with `./install.sh -d` as when building the library itself. That helped minimize weird differences between platforms, particularly when we supported platforms where the system compiler was very old (like on CentOS 6).

common/CMakeLists.txt
```diff
@@ -54,6 +54,15 @@ if(WIN32)
   )
 endif()
 
+# hipcc adds these options automatically. These options might have an effect
+# on performance, so add them when building with clang directly, just in case.
+if(CMAKE_CXX_COMPILER MATCHES ".*clang++.*")
```
jmachado-amd: Hey Cory, I think you mean to write `.*clang\\+\\+.*` as each `\\` should be parsed as a single `\` in quoted arguments (https://cmake.org/cmake/help/latest/command/string.html#regex-specification).

cgmb: Thanks! I started with checking that it wasn't hipcc and I was so careful to match [rocRAND's tried and true checks](https://github.com/ROCm/rocRAND/blob/rocm-6.0.2/cmake/VerifyCompiler.cmake#L45). Then I switched to checking it was clang++ and forgot to do the same. :(

cgmb: I'll fix that and apply these flags onto the rocsolver library directly instead of rocsolver-common, as there is no need for these flags on the clients.

cgmb (approved):

jzuniga-amd (approved):
LGTM

PR #683, opened 2024-03-04
use std::max(), std::min(), instead of max(), min()  to avoid 32bit overflow

EdDAzevedo:
The original max(t1,t2) function computes as 32bit signed integer even if t1, t2 are of type size_t unsigned 64 bit.  

This can lead to problems related to  32bit overflow such as getri_strided_batched for n=260, batch_count=2500, type rocblas_double_complex. 

This update creates an overloaded function that performs std::max(t1,t2) instead of max(t1,t2) to use 64bit arithmetic when t1, t2 are of type size_t.

There is an overloaded funtion for std::min(t1,t2) to replace min(t1,t2) as well.

### DIFF
--- a/common/include/rocblas_utility.hpp
+++ b/common/include/rocblas_utility.hpp
@@ -466,3 +466,15 @@ catch(...)
 }
 
 #undef ROCSOLVER_ROCBLAS_HAS_F8_DATATYPES
+
+template <typename T>
+static T max(T x, T y)
+{
+    return (std::max<T>(x, y));
+}
+
+template <typename T>
+static T min(T x, T y)
+{
+    return (std::min<T>(x, y));
+}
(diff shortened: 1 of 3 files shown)


common/include/rocblas_utility.hpp
```diff
@@ -466,3 +466,15 @@ catch(...)
 }
 
 #undef ROCSOLVER_ROCBLAS_HAS_F8_DATATYPES
+
+template <typename T>
+static T max(T x, T y)
+{
+    return (std::max<T>(x, y));
+}
```
TorreZuk: You may as well "static inline" if writing these, but maybe just switch all to using std::min and std::max ?

EdDAzevedo: I'll add the inline hint.  This is intended to make minimal changes to existing code.

EdDAzevedo:
This will be resolved by Jibreal using namespace in rocsolver codes.

PR #684, opened 2024-03-05
Update links

LisaDelaney:


docs/userguide/memory.rst
```diff
@@ -9,7 +9,7 @@ In contrast to LAPACK, however, pointers to the workspace are not explicitly pas
 functions as arguments; instead, they are managed behind-the-scenes using a configurable device memory
 model.
 
-rocSOLVER makes use of and is integrated with `the rocBLAS memory model`_. Workspace memory, and the
+rocSOLVER makes use of and is integrated with `the rocBLAS memory model`. Workspace memory, and the
```
cgmb: This change seems to have removed the link. If that was the intent, there's still pieces of it leftover at the bottom of the file that should also be removed:

```
.. _the rocBLAS memory model: https://rocblas.readthedocs.io/en/latest/API_Reference_Guide.html#device-memory-allocation-in-rocblas
```

There's also a similar link to the `Device Memory Allocation Functions` in rocBLAS that should probably be updated or removed later on in the file.

yhuiYH: Right, should this link be retained and pointing here instead? https://rocm.docs.amd.com/projects/rocBLAS/en/latest/API_Reference_Guide.html#device-memory-allocation-in-rocblas

yhuiYH: And the other referenced link at the bottom of this page should be pointing here: https://rocm.docs.amd.com/projects/rocBLAS/en/latest/API_Reference_Guide.html#device-memory-allocation-functions

yhuiYH (approved):

PR #686, opened 2024-03-08
Added APIs to the rocsolver namespace

jibrealkhan:
In this PR I have created a Marco definition for the rocsolver namespace and added APIs to the rocsolver namespace.


clients/include/clientcommon.hpp
```diff
@@ -38,3 +38,4 @@
 //#include "rocblascommon/device_strided_batch_vector.hpp"
 //#include "rocblascommon/host_batch_vector.hpp"
 //#include "rocblascommon/host_strided_batch_vector.hpp"
+using namespace rocsolver;
```
cgmb: Please ensure that all source files end with a newline.

common/include/rocblas_utility.hpp
```diff
@@ -32,11 +32,21 @@
...
-
 #include <hip/hip_runtime.h>
 #include <rocblas/rocblas.h>
 #include <rocsolver/rocsolver.h>
 
+// Macro for Namespace
+#ifndef ROCSOLVER_BEGIN_NAMESPACE
+#define ROCSOLVER_BEGIN_NAMESPACE    \
+    namespace rocsolver {            \
+    inline namespace v3_26_0{
```
cgmb: This is good, but it could be improved by using the macros from rocsolver-version for these numbers.

cgmb (changes requested):
Thanks Jibreal. I think there are a couple tweaks required, but it generally looks good.

common/include/common_host_helpers.hpp
```diff
@@ -60,6 +60,8 @@
     } while(0)
 
 /* =============================================================================================== */
+
+ROCSOLVER_BEGIN_NAMESPACE
```
cgmb: Oh. I suppose that this header defines functions both within the library and the clients. The rocsolver namespace (especially with the version number) should only be used for functions defined within the library. The `ROCSOLVER_BEGIN_NAMESPACE` should probably be guarded within `#ifdef ROCSOLVER_LIBRARY` to ensure that it's only applied when using this header in library code.

cgmb: That will make the `using namespace rocsolver;` in `clientscommon.hpp` unnecessary, as neither the rocsolver public API nor any client functions would be in the rocsolver namespace.

common/include/rocblas_utility.hpp
```diff
@@ -32,11 +32,39 @@
...
+#if ROCSOLVER_VERSION_PATCH < 10
+#define ROCSOLVER_VERSION_PATCH_PADDED ROCSOLVER_CONCAT2(0, ROCSOLVER_VERSION_PATCH)
+#else
+#define ROCSOLVER_VERSION_PATCH_PADDED ROCSOLVER_VERSION_PATCH
+#endif
+
+#ifndef ROCSOLVER_BEGIN_NAMESPACE
+#define ROCSOLVER_BEGIN_NAMESPACE    \
+    namespace rocsolver {            \
+    inline namespace ROCSOLVER_CONCAT4(v, ROCSOLVER_VERSION_MAJOR, ROCSOLVER_VERSION_MINOR_PADDED, ROCSOLVER_VERSION_PATCH_PADDED) {
```
tfalders: Minor question, but will this have the periods between the major/minor/patch numbers? I.e. Will it be 3.26.abcde vs 326abcde?

tfalders (approved):
There are some changes I'd like to make to the newlines for better visual consistency, but that's entirely cosmetic and not needed ATM.

common/include/rocblas_utility.hpp
cgmb: No. It needs to be a valid C identifier, so it must start with a letter and can only include alphanumeric characters or underscore. The current inline namespace name is `v32600`, where 3 is the major version, 26 is the minor version and 00 is the patch version.

cgmb: At our current pace, we could reach minor version 100 in ~12 years if we don't make any breaking API changes. However, this is an internal implementation detail, so we can always change the pattern in the future if it becomes a problem. For now, I've tried to keep it compact so it can be easily ignored when browsing through big lists of symbol names.

jzuniga-amd (commented):
LGTM but there are some files with conflicts (maybe just whitespace or stuff like this?)

cgmb:
> LGTM but there are some files with conflicts (maybe just whitespace or stuff like this?)

There are more issues than whitespace, unfortunately. _All_ library functions should be going into the namespace, except:
1. Functions that are purposely inserted into another library's namespace (e.g., fmt print specializations).
2. The rocsolver C API functions.

No client functions should go into the namespace. The namespace is just for the library, so there should be no changes to the clients.

That means I need to extend this pattern to the new functions added in 6.2 (which is trivial), but as I've been applying my suggestions above, it has revealed that there are functions not being included in the namespace that should be. I'll have to adjust the BEGIN/END for a number of files.

tfalders:
We should probably retarget this to release-staging, along with the hipSOLVER PR, and merge back to develop afterwards.

cgmb (approved):
We can do some cleanup later, but it is now functionally complete.

PR #689, opened 2024-03-18
Add random sparse matrix generator to tests

jzuniga-amd:
1. adds initializer methods that generate a random sparse matrix of a chosen form (fill and diagonal type) of a given size and given number of non-zeros
2. adds the methods to `data_initializer.hpp`. In the future, the idea is to have in this file different initializer methods that can be reused by the different `_initData()` functions in our test codes.
3. Use the new initializers to generate test cases for `csrrf_sumlu` and `csrrf_splitlu`. This allows us to have a wide range of test cases and benchmark data; the fixed test cases in the `sparsedata` files are no longer required for these 2 APIs.  
4. Finally, this PR proposes a file reorganization and renaming. Many files are touched, but the changes are minor (files are either just renamed/moved and/or the path to included headers modified accordingly).


clients/include/testing_csrrf_splitlu.hpp
```diff
@@ -152,50 +154,15 @@ void csrrf_splitlu_initData(rocblas_handle handle,
...
                             Uh& hindU,
-                            Th& hvalU,
-                            const fs::path testcase,
-                            bool test = true)
+                            Th& hvalU)
 {
     bool mat_zero = (nnzT == 0);
 
-    // if not matrix zero, read data from files
+    // if not matrix zero, generate input data
```
EdDAzevedo, line 163: Suggestion: perhaps add comments to explain the different use cases of variable nnzT. One may think nnzT = (nnzL - n) + nnzU since matrix  T = (L - eye) + U.

clients/include/testing_csrrf_splitlu.hpp
```diff
@@ -235,12 +204,11 @@ void csrrf_splitlu_getError(rocblas_handle handle,
...
-                            const fs::path testcase)
+                            double* max_err)
 {
     // input data initialization
-    csrrf_splitlu_initData<true, true, T>(handle, n, nnzT, dptrT, dindT, dvalT, hptrT, hindT, hvalT,
-                                          hptrL, hindL, hvalL, hptrU, hindU, hvalU, testcase);
+    csrrf_splitlu_initData<true, true, T>(handle, n, nnzT, nnzL, nnzU, dptrT, dindT, dvalT, hptrT,
+                                          hindT, hvalT, hptrL, hindL, hvalL, hptrU, hindU, hvalU);
 
     // execute computations
```
EdDAzevedo, line 226: Suggestion: If L and U have been generated, perhaps also consider testing other edge cases, say  with T = L   (strictly lower triangular),  and T = U (strictly upper triangular),   T = diagonal matrix ,  T = identity matrix,  and T is fully dense. 

 I wonder whether T  is matrix with all zeros with nnzT = 0 would cause a  problem since (nnzT == 0) may have a special meaning.

clients/include/testing_csrrf_splitlu.hpp
```diff
@@ -395,11 +406,17 @@ void testing_csrrf_splitlu(Arguments& argus)
...
     {
-        // arbitrarily assign ~55% of nnzT to U, ~45% to L
-        nnzU = (nnzT * 55) / 100;
+        if(nnzT > n * n)
+            nnzT = n * n;
+
+        // assign a random number of nonzeros to L and U
+        rocblas_seedrand();
+        rocblas_int high, low;
+        low = max(0, nnzT - n * (n - 1) / 2);
```
EdDAzevedo: Suggestion: Maybe use std::max, std::min to be safe.

clients/rocblascommon/rocblas_init.hpp
```diff
@@ -318,3 +318,212 @@ void rocblas_copy_matrix(const T* A,
...
+            // update ptrA
+            for(rocblas_int j = i + 1; j <= n; ++j)
+                ptrA[j] += nn;
+        }
+    }
+
+    /////////////////////////////
+    // Generates indA and valA //
+    /////////////////////////////
+
```
EdDAzevedo, line 399: Suggestion: It may be simpler to generate a random sparse matrix in the coordinate (COO) format (where we explicilty store (i,j, A(i,j)), then convert from COO to CSR format.  

If we need say a M by N sparse matrix,  there is a simple mapping between coordinate (i,j) -> i + j * M to a value in 0..( M * N - 1 ) and given a value 0 <= ij < ( M * N ),  we can recover (i,j) as i=mod(ij,M), j = (ij-i)/M.  Thus one can generate  nnzA  random numbers in range [0, M*N), then generate the coordinates (i,j).  This may generate the exact number of non-zeros without duplicates.

If we need a symmetric random matrix, I think there is also a mapping between the compact storage of   N * (N + 1)/2  non-zero entries to "(i,j)" coordinate. Maybe something like    (i,j) -> (i + (j*( j + 1))/2     and the inverse operation.

EdDAzevedo (approved):

clients/include/testing_csrrf_splitlu.hpp
jzuniga-amd: I added a comment. I don't see what use cases nnzT has other than the number of non-zeros in T. If nnzT=0, then L=I and U=0, so the factorized matrix is the matrix zero.

clients/include/testing_csrrf_splitlu.hpp
jzuniga-amd: nnzT=0 has no special meaning. T=0 is in the test suite as well.

clients/include/testing_csrrf_splitlu.hpp
jzuniga-amd: changed

clients/rocblascommon/rocblas_init.hpp
jzuniga-amd: I won't have time to look into this, but I will leave a TODO comment in case we want to revisit it.

cgmb:
I've opened a PR with my suggestions relating to the directory reorg. My goal with the suggestions is to make it more obvious which sources are shared by rocsolver-test and rocsolver-bench. https://github.com/jzuniga-amd/rocSOLVER/pull/12

cgmb (approved):

clients/common/misc/data_initializer.hpp
```diff
@@ -318,3 +318,217 @@ void rocblas_copy_matrix(const T* A,
...
+                for(rocblas_int j = i; j < n; ++j)
+                    ops[j - i] = j;
+            }
+
+            // if no random diag, then make sure the position in diagonal is non-zero
+            in = 0;
+            if(!randomdiag)
+            {
+                z[0] = i;
+                op = (fill == 2) ? 0 : i;
```
tfalders: Isn't `fill == 2` comparing an enum against and int? It would be preferable to have an enum against enum comparison.

tfalders (approved):
Looks good! Just one minor comment.

clients/common/misc/data_initializer.hpp
jzuniga-amd: right, good catch. I forgot to replace this one. Done.

PR #691, opened 2024-03-21
fix argument checking for invalid pointers

jzuniga-amd:
The documentation of several functions indicates that some of their argument arrays are size n-1. Within the argument-checking methods, however, we return invalid pointer status if n > 0 and the pointer is null. 

This PR fixes this discrepancy and addresses the error in [SWDEV-445494](https://ontrack-internal.amd.com/browse/SWDEV-445494) 

Also changed min and max to std::min and std::max everywhere in the library code

### DIFF
--- a/clients/include/testing_stedc.hpp
+++ b/clients/include/testing_stedc.hpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2020-2023 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2020-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -77,8 +77,8 @@ void testing_stedc_bad_arg()
     // safe arguments
     rocblas_local_handle handle;
     rocblas_evect evect = rocblas_evect_original;
-    rocblas_int n = 1;
-    rocblas_int ldc = 1;
+    rocblas_int n = 2;
+    rocblas_int ldc = 2;
 
     // memory allocations
     device_strided_batch_vector<S> dD(1, 1, 1, 1);
--- a/library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
+++ b/library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
@@ -112,7 +112,7 @@ rocblas_status rocsolver_ormtr_argCheck(rocblas_handle handle,
         return rocblas_status_continue;
 
     // 3. invalid pointers
-    if((nq > 0 && !A) || (nq > 1 && !ipiv) || (m * n && !C))
+    if((nq > 0 && !A) || (nq > 1 && !ipiv) || (m && n && !C))
         return rocblas_status_invalid_pointer;
 
     return rocblas_status_continue;
--- a/library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
+++ b/library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
@@ -4,7 +4,7 @@
  *     Univ. of Tennessee, Univ. of California Berkeley,
  *     Univ. of Colorado Denver and NAG Ltd..
  *     December 2016
- * Copyright (C) 2019-2022 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2019-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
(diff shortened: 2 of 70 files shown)


clients/include/testing_stedc.hpp
```diff
@@ -77,8 +77,8 @@ void testing_stedc_bad_arg()
     // safe arguments
     rocblas_local_handle handle;
     rocblas_evect evect = rocblas_evect_original;
-    rocblas_int n = 1;
-    rocblas_int ldc = 1;
+    rocblas_int n = 2;
+    rocblas_int ldc = 2;
```
EdDAzevedo, line 81: It is not clear to me why the test case n=1 needs to be changed. Perhaps the test need to check n=1 and also n=2 as well?

library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
```diff
@@ -112,7 +112,7 @@ rocblas_status rocsolver_ormtr_argCheck(rocblas_handle handle,
         return rocblas_status_continue;
 
     // 3. invalid pointers
-    if((nq > 0 && !A) || (nq > 0 && !ipiv) || (m * n && !C))
+    if((nq > 0 && !A) || (nq > 1 && !ipiv) || (m * n && !C))
```
EdDAzevedo: Minor suggestion:  Perhaps test (m == 0) || (n == 0) instead of testing ( (m*n) == 0) to avoid possible overflow or  the product happen to get to 0 by modulo arithmetic. 

 Fortunately, (2^31 - 1) for signed 32bit integer  is also a prime number.  However, say (2^15 - 1)  for signed 16bit integer is not a prime number,  and (2^63 - 1) for signed 64bit is not a prime number.

EdDAzevedo (approved):

library/src/lapack/roclapack_gebd2.hpp
```diff
@@ -91,7 +91,8 @@ rocblas_status rocsolver_gebd2_gebrd_argCheck(rocblas_handle handle,
         return rocblas_status_continue;
 
     // 3. invalid pointers
-    if((m * n && !A) || (m * n && !D) || (m * n && !E) || (m * n && !tauq) || (m * n && !taup))
+    if((m * n && !A) || (m * n && !D) || (min(m, n) > 1 && !E) || (m * n && !tauq)
```
tfalders: Please use std::min instead of min

tfalders (approved):
LGTM! Just one minor comment

library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
EdDAzevedo: I still feel uncomfortable with using "(m * n) != 0" as short-cut  for "(m != 0) && (n != 0)". 

For example, let m = n = 2^16 = 65536, then "m * n " can be computed to return  0 due to overflow.

If we have 64bit interface and variables m, n are of type int64_t,  then let   m =  n = 2^32 = 4294967296,  then  "m * n" can return 0 due to overflow.

library/src/lapack/roclapack_gebd2.hpp
EdDAzevedo: Perhaps consider  std::min(m,n) instead of (m * n)  in the tests or more explicitly ( (m != 0) && (n != 0)) ? 

Otherwise, m = n = 2^16  can yield "m * n" to be 0 due to overflow.

Perhaps introduce intermediate variables something like

bool const has_work = (m >= 1) && (n >= 1);
if( (has_work) &&
     ( (A == nullptr) || (D == nullptr) || (tauq == nullptr) || (taup == nullptr))  ){ 
          return( rocblas_status_invalid_pointer); 
          }
    
    if ((std::min(m,n) > 1) && (E == nullptr)) { return(rocblas_status_invalid_pointer); }

clients/include/testing_stedc.hpp
jzuniga-amd: This is not to test case n = 1.  It is to test bad arguments (invalid pointers, invalid values, etc.) These sizes are any "safe" size used to create dummy arrays for the test.

library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
jzuniga-amd: I changed these conditions to (m && n) everywhere

library/src/lapack/roclapack_gebd2.hpp
jzuniga-amd: now using std::min and std::max everywhere

library/src/auxiliary/rocauxiliary_ormtr_unmtr.hpp
tfalders: I wonder if we should start checking for overflows in the argCheck methods, e.g. `m * n > 0`. That would probably be a separate PR though, and touch many files.

cgmb: Signed integer overflow is undefined behaviour. You can't use `m * n > 0` to check for signed overflow, but I do generally agree we should be checking some sort of maximum bounds in argCheck.

cgmb: For overflow checks in rocSOLVER, I would use the [builtin clang extensions](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html).

cgmb (approved):

PR #693, opened 2024-03-26
Added larf_64

tfalders:
This PR adds 64-bit APIs for LARF. I intend to follow the same general structure for future 64-bit APIs, so let me know what you think as soon as you're able.

One of the important questions we need to address with the 64-bit APIs is how to test them. For now, I'm using the same test cases as for the 32-bit API. We can probably stick with this approach as long as we ensure there's no downcasting happening within LARF or future 64-bit APIs, e.g. by reenabling some of the warnings we've disabled and flagging them as failures on CI.

### DIFF
--- a/clients/include/testing_larf.hpp
+++ b/clients/include/testing_larf.hpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2020-2022 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2020-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -35,7 +35,7 @@
 #include "rocsolver_arguments.hpp"
 #include "rocsolver_test.hpp"
 
-template <typename T>
+template <bool API64, typename T>
 void larf_checkBadArgs(const rocblas_handle handle,
                        const rocblas_side side,
                        const rocblas_int m,
@@ -47,31 +47,31 @@ void larf_checkBadArgs(const rocblas_handle handle,
                        const rocblas_int lda)
 {
     // handle
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(nullptr, side, m, n, dx, inc, dt, dA, lda),
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, nullptr, side, m, n, dx, inc, dt, dA, lda),
                           rocblas_status_invalid_handle);
 
     // values
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(handle, rocblas_side_both, m, n, dx, inc, dt, dA, lda),
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, handle, rocblas_side_both, m, n, dx, inc, dt, dA, lda),
                           rocblas_status_invalid_value);
 
     // pointers
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(handle, side, m, n, (T) nullptr, inc, dt, dA, lda),
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, handle, side, m, n, (T) nullptr, inc, dt, dA, lda),
                           rocblas_status_invalid_pointer);
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(handle, side, m, n, dx, inc, (T) nullptr, dA, lda),
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, handle, side, m, n, dx, inc, (T) nullptr, dA, lda),
                           rocblas_status_invalid_pointer);
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(handle, side, m, n, dx, inc, dt, (T) nullptr, lda),
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, handle, side, m, n, dx, inc, dt, (T) nullptr, lda),
                           rocblas_status_invalid_pointer);
 
     // quick return with invalid pointers
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(handle, rocblas_side_left, 0, n, (T) nullptr, inc,
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, handle, rocblas_side_left, 0, n, (T) nullptr, inc,
                                          (T) nullptr, (T) nullptr, lda),
                           rocblas_status_success);
-    EXPECT_ROCBLAS_STATUS(rocsolver_larf(handle, rocblas_side_right, m, 0, (T) nullptr, inc,
+    EXPECT_ROCBLAS_STATUS(rocsolver_larf(API64, handle, rocblas_side_right, m, 0, (T) nullptr, inc,
                                          (T) nullptr, (T) nullptr, lda),
                           rocblas_status_success);
 }
 
-template <typename T>
+template <bool API64, typename T>
 void testing_larf_bad_arg()
 {
     // safe arguments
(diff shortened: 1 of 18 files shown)


library/src/auxiliary/rocauxiliary_larf.hpp
```diff
@@ -156,9 +156,9 @@ rocblas_status rocsolver_larf_template(rocblas_handle handle,
     //      ZERO ENTRIES ****
 
     // compute the matrix vector product  (W=-A'*X or W=-A*X)
-    rocblasCall_gemv<T>(handle, trans, m, n, cast2constType<T>(scalars), 0, A, shiftA, lda, stridea,
-                        x, shiftx, incx, stridex, cast2constType<T>(scalars + 1), 0, Abyx, 0, 1,
-                        order, batch_count, workArr);
+    rocblasCall_gemv<T, rocblas_int>(
```
EdDAzevedo: I wonder whether the compiler can deduce the "I" type from the types of variables in the argument list,  so that the original code with "rocblasCall_gemv\<T \>"  (instead of "rocblasCall_gemv\<T,rocblas_int\>") will still work without changes?  

If so, this might save some work.

library/src/include/lib_host_helpers.hpp
```diff
@@ -41,12 +41,15 @@
  * ===========================================================================
  */
 
-inline size_t idx2D(const size_t i, const size_t j, const size_t lda)
+inline rocblas_stride idx2D(const rocblas_stride i, const rocblas_stride j, const rocblas_stride lda)
```
EdDAzevedo: I wonder whether a templated version or lambda expression would be more flexible that will work for 32bit interface and 64bit interface. Say something like

template<typename I>
inline int64_t idx2D(const I i, const I j, const I lda) { 
                                   return   i + j *static_cast<int64_t>(lda); };

library/src/include/lib_device_helpers.hpp
```diff
@@ -389,10 +389,10 @@ ROCSOLVER_KERNEL void reset_batch_info(U info, const rocblas_stride stride, cons
         inf[idx] = T(val);
 }
 
-template <typename T>
-ROCSOLVER_KERNEL void get_array(T** out, T* in, rocblas_stride stride, rocblas_int batch)
+template <typename T, typename I>
+ROCSOLVER_KERNEL void get_array(T** out, T* in, rocblas_stride stride, I batch)
 {
-    int b = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+    I b = static_cast<I>(hipBlockIdx_x) * hipBlockDim_x + hipThreadIdx_x;
```
EdDAzevedo: Will all similar routines consistently also have the extra "typename I" template argument?

clients/include/testing_larf.hpp
```diff
@@ -240,13 +241,13 @@ void larf_getPerfData(const rocblas_handle handle,
...
         start = get_time_us_sync(stream);
-        rocsolver_larf(handle, side, m, n, dx.data(), inc, dt.data(), dA.data(), lda);
+        rocsolver_larf(API64, handle, side, m, n, dx.data(), inc, dt.data(), dA.data(), lda);
         *gpu_time_used += get_time_us_sync(stream) - start;
     }
     *gpu_time_used /= hot_calls;
 }
 
-template <typename T>
+template <bool API64, typename T>
```
EdDAzevedo, line 250: I wonder whether we can test with (M very large, N = 2 or N = 1), and (M = 2 or M = 1, N very large) where "very large" exceed 2^31 - 1. Perhaps this can still fit in device memory?  The test code might need to query the amount of device memory available and skip some tests if the data will not fit.

TorreZuk (commented):
If you haven't already reviewed rocBLAS we have the ILP64 for both Level 1 and Level 2 functions in develop branch and feel free to ask questions.   Our test framework we went with single instantiation using 64bit args for both cases and runtime dispatch as we already did FORTRAN that way, but yes testing actual sizes over 32bits is a work in progress due to practical memory limits.  Skinny matrices would be expected.

library/src/auxiliary/rocauxiliary_larf.hpp
tfalders: One would think. I was getting some compiler errors before implementing this change, though I think they were coming from latrd and labrd; larf and the others weren't a problem. It deserves a closer look at some point; my concern right now is just getting it to work. If you think it would be worthwhile, I can revert this change to larf and the others.

library/src/include/lib_host_helpers.hpp
tfalders: Not a bad idea. I'll put it in.

library/src/include/lib_device_helpers.hpp
tfalders: That's the idea, and what should become standard for 64-bit functions going forward unless we decide on a different convention.

clients/include/testing_larf.hpp
tfalders: That could be something to look into, yes, especially if rocBLAS goes a similar route. We will still have to consider how to handle methods that work on square matrices, though, since the 64-bit APIs will probably be extended to such functions at some point in the future.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: Another  nit picking comment or very minor suggestion:  If hipBlockDim_x has limit of 2^10, and hipGridDim_x has uint32_t limit of 2^32 - 1, then the max value for variable "b" is less than 2^42, which is still less than limit for 64bit integer, say batch_count = 2^43.       Fortunately, we don't expect to see batch_count > 2^42 in the near future.

One suggestion is to code using the for-loop of the form below so that it will work no matter the limits of the launch configuration

I const b_start = static_cast\<I\>(hipBlockIdx_x) * hipBlockDim_x  + hipThreadIdx_x;
I const b_inc = static_cast\<I\>(hipGridDim_x) * hipBlockDim_x;

for( I b=b_start; b < batch; b += b_inc) { .... }

library/src/auxiliary/rocauxiliary_larf.hpp
tfalders: So, it seems that larf does require this change after all. I get compiler errors when removing `I` from either ger or gemv.

library/src/include/lib_device_helpers.hpp
TorreZuk: The product is limited https://rocm.docs.amd.com/projects/HIP/en/latest/user_guide/faq.html#what-is-maximum-limit-of-kernel-launching-parameter

tfalders: Oh, interesting! I wasn't aware of that. Thanks for the heads-up!

library/src/auxiliary/rocauxiliary_larf.hpp
EdDAzevedo: I wonder perhaps the compiler has issue with the cast2constType\<T\>. 

Maybe just using  another temporary variable can get around this cast2constType\<T\>? 

Maybe something like
{
T const * const alpha = scalars;
T const * const beta = scalars + 1;
... rocblasCall_gemv( handle, ..., alpha, ...., beta ... ); 
}


The compiler may even be able to deduce the "T" type by just using "rocblasCall_gemv(",  even without the "T" template argument.  



Just a suggestion.

library/src/include/lib_host_helpers.hpp
tfalders: I think I might end up reverting this change. It works with larf, but now that I'm working on trsm it's having difficulty deducing the template argument.

library/src/include/lib_device_helpers.hpp
EdDAzevedo: I think it is more robust to use the "for( I b=b_start; b < batch; b += b_inc) { ... }" form since there is a max limit of 2^32 total threads but   batch_count > 2^32 is not impossible in near future hardware, say 1x1 matrices of float type with batch_count = 2^32 + 1. Otherwise, the code may produce incorrect results using the "if (b < batch)" way of coding since it relies on  max value of variable "b" to be at least "batch-1".

library/src/include/lib_host_helpers.hpp
EdDAzevedo: You might consider more than 1 integer template argument if that is the problem, say
template\<typename T, typename I, typename J \>
inline rocblas_stride idx2D( const I i, const I j, const J lda) { return ... }

Another possibility is a lambda expression (that will be automatically inlined by the compiler) so that the compiler can automatically match the types. Maybe something like the following (where variables i, j, lda can each be a different type). Note if the lambda expression is in the body of a device function, then it will automatically work inside the device function as well.

auto idx2D = [](auto i, auto j, auto lda) -> rocblas_stride { return .... };

tfalders:
Closing for now since larf is not required for the next release.

PR #695, opened 2024-03-28
Added getrs_64

tfalders:
This PR adds 64-bit versions of GETRS to the rocSOLVER API.

I've disabled CI for now since I'm getting an undefined symbol error from rocBLAS when running the tests. Will reenable once it's resolved.


library/src/include/rocsolver_run_specialized_kernels.hpp
```diff
@@ -39,107 +39,109 @@
  */
 
 // trsm
-template <bool BATCHED, bool STRIDED, typename T>
+template <bool BATCHED, bool STRIDED, typename T, typename I>
```
EdDAzevedo, line 42: Question: Is the convention the type "I" is right after "T" or is type "I" is consistently the last template argument, or there is no convention?

tfalders: The convention I'm using is to place `I` directly after `T`

library/src/include/lib_host_helpers.hpp
```diff
@@ -75,9 +75,10 @@ T* const* cast2constPointer(T** array)
     return array;
 }
 
-inline rocblas_int get_index(rocblas_int* intervals, rocblas_int max, rocblas_int dim)
+template <typename I>
+inline I get_index(I* intervals, I max, I dim)
```
cgmb: Templates are implicitly inline, so that can be dropped.

clients/rocblascommon/device_strided_batch_vector.hpp
```diff
@@ -194,6 +194,30 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
...
+    //! @brief Returns pointer.
+    //! @param batch_index The batch index.
+    //! @return A mutable pointer to the batch_index'th vector.
+    //!
+    T* operator[](int64_t batch_index)
+    {
+        return (this->m_stride >= 0)
+            ? this->m_data + batch_index * this->m_stride
+            : this->m_data + (batch_index + 1 - this->m_batch_count) * this->m_stride;
+    }
```
jmachado-amd: In regard to all the `[device, host]*vector` classes, I am afraid that `m_n`, `m_inc` and `m_batch_count` are also supposed to become `int64_t`.

clients/rocblascommon/device_strided_batch_vector.hpp
```diff
@@ -194,6 +194,30 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
...
+
+    //!
+    //! @brief Returns non-mutable pointer.
+    //! @param batch_index The batch index.
+    //! @return A non-mutable mutable pointer to the batch_index'th vector.
+    //!
+    const T* operator[](int64_t batch_index) const
+    {
```
jmachado-amd: I know this is minor, and unrelated to your PR, but it is a bit hard to stop looking when the code and the comments make conflicting claims: these methods clearly return non-const pointers to const pointees.

clients/rocblascommon/device_strided_batch_vector.hpp
```diff
@@ -194,6 +194,30 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
...
+        return (this->m_stride >= 0)
+            ? this->m_data + batch_index * this->m_stride
+            : this->m_data + (batch_index + 1 - this->m_batch_count) * this->m_stride;
+    }
+
```
jmachado-amd: Another minor point that applies to pieces of code similar to this one, to consider for the future. Since overflow of signed integers is undefined behavior, a compiler could (and likely would) perform the 32-bit version of this computation with 64-bit registers, effectively masking potential overflows (which would appear as accesses to out of bound elements of the vector). Now that the computation has been promoted to 64-bits, overflows here can lead to subtle bugs in the (distant) future.

library/src/auxiliary/rocauxiliary_laswp.hpp
```diff
@@ -14,31 +14,31 @@
...
+                                   const I k2,
+                                   const I* ipivA,
+                                   const rocblas_stride shiftP,
+                                   I incp,
                                    const rocblas_stride strideP)
 {
-    int id = hipBlockIdx_y;
-    int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+    I id = hipBlockIdx_y;
+    I tid = hipBlockIdx_x * static_cast<I>(hipBlockDim_x) + hipThreadIdx_x;
```
jmachado-amd, line 33: I like the idea of promoting `hipBlockDim_x` to avoid issues with 32-bit modular arithmetic. Please don't listen to anyone saying that the maximum number of threads per block is way below int-max. :)

library/src/include/rocblas.hpp
```diff
@@ -1895,33 +1897,48 @@ void rocblasCall_trsm_mem(rocblas_side side,
         rocblas_workmode parameter **/
 
     // can't infer batched based on input params
-    if constexpr(BATCHED)
-        rocblas_internal_trsm_batched_workspace_size<T>(side, transA, m, n, batch_count, 0, x_temp,
-                                                        x_temp_arr, invA, invA_arr, &no_opt_size);
+    if constexpr(std::is_same<I, rocblas_int>::value)
+    {
```
jmachado-amd, line 1901: I am a little uncomfortable with this use of `std::is_same<>` to determine which API to use (32-bit versus 64-bit). On one hand this method is low risk, as type deduction on the pass-by-value `I` variables will discard most qualifiers and likely do what you want. However, a piece of code for which, say, `I = unsigned int` will call the 64-bit API, which is unlikely to be correct and might lead to hard to debug performance regressions.

For the time being, I would suggest inverting the tests if possible (i.e., testing whether `I` is `int64_t` and falling back on the 32-bit API for everything else) -- improper use of the 32-bit API should be simple to catch in a unit test.

If everyone agrees, for the future, I'd suggest using SFINAE to make sure that methods like `rocblasCall_trsm_mem` can't be called when `I` is neither `int64_t` or `rocblas_int`. Something in the lines of the [`is_any_of` concept (in the example)](https://en.cppreference.com/w/cpp/concepts/same_as) should work:

```cpp
template<typename I, typename ...>
struct is_any_of : std::false_type {};

/// `is_any_of<I, T, Ts>::value` is defined if type `I` is any of the valid types `T`, `Ts...`
template<typename I, typename T, typename... Ts>
struct is_any_of<I, T, Ts...> : std::bool_constant<std::is_same<I, T>::value || is_any_of<I, Ts...>::value> {};

template <bool BATCHED, typename T, typename I,
    typename = typename std::enable_if<is_any_of<std::decay_t<I>, // Defined if `I` is a valid type
        int64_t, rocblas_int                                      // List of valid types goes here
        >::value>::type>
void rocblasCall_trsm_mem(rocblas_side side, // (...)
```

library/src/lapack/roclapack_getri.hpp
```diff
@@ -239,7 +239,8 @@ void rocsolver_getri_getMemorySize(const rocblas_int n,
     t1 = n * blk * sizeof(T) * batch_count;
 
     // requirements for calling TRSM
-    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_right, rocblas_operation_none, n, blk + 1,
+    rocblas_int nn = (n % 128 != 0) ? n : n + 1;
+    rocblasCall_trsm_mem<BATCHED, T>(rocblas_side_right, rocblas_operation_none, nn, blk + 1, 1, 1,
```
jmachado-amd, line 243: Not exactly part of this PR but somewhat important, I found this comment elsewhere on the necessity of making sure `n` is not a multiple of `128`:

```cpp
        // (Note: rocblas TRSM workspace size is less than expected when the number of rows is multiple of 128.
        //  For this reason, when trying to set up a workspace that fits all the TRSM calls for m <= blk,
        //  blk cannot be multiple of 128.)
```

If this is still the case, then we should consider the possibility of modifying `rocblasCall_trsm_mem` to work around this issue and prevent ourselves from forgetting to do it again while making it easier to revert the workaround globally when rocblas fix this issue.

clients/include/testing_getrs.hpp
```diff
@@ -35,20 +35,20 @@
...
+                        Id dIpiv,
                         const rocblas_stride stP,
-                        T dB,
-                        const rocblas_int ldb,
+                        Td dB,
+                        const I ldb,
                         const rocblas_stride stB,
-                        const rocblas_int bc)
+                        const I bc)
 {
```
jmachado-amd: I understand this is just following the original design and it is not very significant as it is just a test, but I wonder why `Td` and `Id` are being used instead of pointers/references to `T` and `I`. The side effect I see is that pointers `dA`, `dB` and `dIpiv` passed by value will lose their modifiers (such as "constness"), but I don't see how this would be useful here.

library/src/include/lib_host_helpers.hpp
```diff
@@ -41,12 +41,12 @@
  * ===========================================================================
  */
 
-inline size_t idx2D(const size_t i, const size_t j, const size_t lda)
+inline int64_t idx2D(const int64_t i, const int64_t j, const int64_t lda)
 {
     return j * lda + i;
 }
```
jmachado-amd, line 48: This comment applies to the overloading set of `idx2D`. Small functions that are used everywhere can be a huge source of pain. Although no numerical code depending on level-2 or -3 methods should come close to using variables that would cause overflow here in the foreseeable future, the potential use of this method in other places might lead to hard to debug overflow or narrowing bugs. I would suggest reverting the types back to `size_t` unless you are certain that new code requires the returned value to be signed.

jmachado-amd (commented):
Hello Troy, that is lot of work! The code reads well, the only major point I have is that I can't compile this branch in my workstation, and I am not sure whether CI has run since you turned it on again. If you can confirm that this branch compiles as is, I'll happily approve it.

Given the time constraints, I understand that there is a necessity for moving forward with this PR and that further changes to the code might not be feasible now. That said, the main comments I have are not about design changes you introduced in this PR, but about the abundance of code that might lead to signed integer overflow and modular arithmetic issues when used with large integers -- it is natural to overlook these issues with methods that perform level-2 or level-3 operations on the grounds that their significant time and memory complexity would make such issues unlikely, but maybe it is time to consider paying more attention to them now.

Also, I would suggest making sure that default calls to BLAS routines are made targeting their 32-bit versions (please have a look at my comments on file `rocblas.hpp` for details).

clients/rocblascommon/device_batch_vector.hpp
```diff
@@ -189,6 +189,26 @@ class device_batch_vector : private d_vector<T, PAD, U>
...
+    {
+        return this->m_data[batch_index];
+    }
+
+    //!
+    //! @brief Constant random access.
+    //! @param batch_index The batch index.
+    //! @return Constant pointer to the array on device.
+    //!
+    const T* operator[](int64_t batch_index) const
```
cgmb: We could probably drop the 32-bit overloads for these, but it's fine either way.

cgmb (approved):

clients/rocblascommon/device_strided_batch_vector.hpp
cgmb: I'm not a fan of these vectors in general and would like to entirely replace them. I have a draft PR where I started that, but it's incomplete...

tfalders:
> Hello Troy, that is lot of work! The code reads well, the only major point I have is that I can't compile this branch in my workstation, and I am not sure whether CI has run since you turned it on again. If you can confirm that this branch compiles as is, I'll happily approve it.

I'm resolving the merge conflicts now and will push the fix along with addressing the review comments. It does build for me and I've run the tests successfully (for the most part, I think my card really is starting to fail).

clients/include/testing_getrs.hpp
tfalders: `Td` can be either `T*` (non-batched and strided-batched case) or `T* const*` (batched case), so I believe templating is necessary for that at least. `Id` is pretty much always `I*` (only exception I'm aware of is geqrf_ptr_batched), but I think we use the same pattern as `Td` for consistency.

library/src/lapack/roclapack_getri.hpp
tfalders: Yes, perhaps moving it into rocblasCall_trsm_mem would be a good move. We do the same thing in rocsolver_trsm_mem (which calls rocblasCall_trsm_mem) and most trsm calls have been changed to go through rocsolver_trsm, but getri and trtri haven't been moved over yet due to some missing features in rocsolver_trsm.

library/src/include/rocblas.hpp
tfalders: The library convention has been to use `rocblas_int` instead of `int` or `unsigned int`, so I thought it should be safe, but if you think this will be safer then I will change it over.

library/src/lapack/roclapack_getri.hpp
tfalders: Oh, I think I remember now why we didn't put it in rocblasCall_trsm_mem. Sometimes we only need to call trsm once, and in that case we can use the smaller amount of memory when the size is a multiple of 128. rocsolver_trsm_mem accounts for this with its `inblocked` parameter.

I think I'll keep it as it is for now, but we'll want to move getri and trtri (and all future functions) over to use rocsolver_trsm instead of rocblasCall_trsm at the earliest convenience.

EdDAzevedo: I have created a ticket SWDEV-448096 for rocblas 
rocblas_internal_trsm_workspace_maxsize() to return upper bound of workspace size.
Once that is available, we can call "maxsize()" then we don't need to check for these special internal magic values.

clients/rocblascommon/device_batch_vector.hpp
tfalders: Combined with @jmachado-amd's comments above, I think it makes sense to transition the containers from 32-bit to 64-bit.

clients/rocblascommon/device_strided_batch_vector.hpp
tfalders: Combined with @cgmb's comment below, I think it makes sense to transition the containers fully from 32-bit to 64-bit.

clients/rocblascommon/device_strided_batch_vector.hpp
tfalders: I've replaced the comments in this file with comments from a different container file

clients/rocblascommon/device_strided_batch_vector.hpp
tfalders: Combined with @cgmb's comment below, I think it makes sense to transition the containers fully from 32-bit to 64-bit.

library/src/include/lib_host_helpers.hpp
tfalders: Done

library/src/include/lib_host_helpers.hpp
tfalders: The reason why I changed it was because, by convention, these parameters will all be either `rocblas_int` (aka `int32_t`) or `int64_t`. Typically, the return value is added to a `rocblas_stride` value (which is an alias of `int64_t`) and fed into a `rocblas_stride` argument. So, as far as I can tell, the inputs are all signed, and the output will be treated as signed as well.

jmachado-amd (approved):

PR #697, opened 2024-03-28
Added getf2_64

tfalders:
This PR adds 64-bit versions of GETF2 to the rocSOLVER API.

Some code is duplicated from https://github.com/ROCm/rocSOLVER/pull/695 so that I could open it as a separate PR.

Although I've added the template parameter `I` to the small-size GETF2 kernel, I've only instantiated it for `rocblas_int`. I did change the two shift parameters from `rocblas_int` to `rocblas_stride`, however. I don't think this should affect performance, but we should definitely check.

Unfortunately, the idea we discussed on using the 32-bit small-size kernel in the 64-bit API won't work. This is because the `ipiv` and `info` arrays have different types in the 32-bit and 64-bit APIs. For now, the small-size kernel won't be used by the 64-bit API, meaning we can expect a sizeable performance hit for GETF2_64.


library/src/include/lib_host_helpers.hpp
```diff
@@ -41,12 +41,15 @@
  * ===========================================================================
  */
 
-inline size_t idx2D(const size_t i, const size_t j, const size_t lda)
+inline rocblas_stride idx2D(const rocblas_stride i, const rocblas_stride j, const rocblas_stride lda)
```
EdDAzevedo: Minor comment: do we want to consistently use int64_t or consistently use rocblas_stride?

library/src/specialized/roclapack_getf2_specialized_kernels.hpp
```diff
@@ -210,56 +210,56 @@ ROCSOLVER_KERNEL void __launch_bounds__(GETF2_SSKER_MAX_M)
...
+    const I ty = hipThreadIdx_y;
+    const I id = hipBlockIdx_z;
+    const I bdx = hipBlockDim_x;
+    const I bdy = hipBlockDim_y;
 
     // batch instance
     T* A = load_ptr_batch<T>(AA, id, shiftA, strideA);
-    rocblas_int *ipiv, *permut;
-    ipiv = load_ptr_batch<rocblas_int>(ipivA, id, shiftP, strideP);
+    I *ipiv, *permut;
```
EdDAzevedo: Minor suggestion: it might be safe to initialize the variable at the time of declaration, so that the variable has a known value. For example, something like

I * const ipiv = loadPtr_batch<I>(....);
I * const  permut = (permut_idx != nullptr) ? permut_idx + id * stridePI : nullptr;

library/src/include/lib_host_helpers.hpp
tfalders: My approach to this has been to remain consistent with what rocBLAS is doing in their 64-bit API. From what I can understand, rocblas_stride is used for parameters that are int64 in both the 32-bit and 64-bit API (so far limited to just the shift and stride parameters), and int64_t is used for the rest.

I suppose I should change the parameters you've highlighted to int64_t. That would be more consistent.

jmachado-amd:
Similarly to #695, I can't compile this PR in my workstation, and it looks like the merge conflicts are preventing CI from running.

jmachado-amd (approved):
The code reads well and everything looks good. I understand that there are opportunities for fine tuning of the small size kernels, which is definitely best postponed.

cgmb (approved):
LGTM

PR #698, opened 2024-04-03
Improve accuracy of Jacobi eigensolver

jmachado-amd:
This commit adds a hotfix to tackle accuracy issues in the Jacobi eigensolver that could happen when off diagonal entries of a matrix are too small, but still relevant for precision.

This is not meant to be a definitive solution; I've chosen the new threshold rather conservatively to balance the necessity of improving the accuracy of the current eigensolver while leaving some extra space to make sure subsequent computations would not suffer from lack of precision or underflow. This choice is, most likely, too conservative, but more tests (in particular including matrices with small norm) are required to guarantee that no spurious bugs would appear if a more aggressive thresholding strategy were used.

### DIFF
--- a/library/src/lapack/roclapack_syevj_heevj.hpp
+++ b/library/src/lapack/roclapack_syevj_heevj.hpp
@@ -166,6 +166,7 @@ __device__ void run_syevj(const rocblas_int dimx,
         local_diag += std::real(sines_diag[i]);
     }
     S tolerance = (local_res + local_diag) * abstol * abstol;
+    S small_num = get_safemin<S>()/eps;
 
     // execute sweeps
     rocblas_int count = (half_n - 1) / dimx + 1;
@@ -187,7 +188,7 @@ __device__ void run_syevj(const rocblas_int dimx,
                     aij = Acpy[i + j * n];
                     mag = std::abs(aij);
 
-                    if(mag < eps)
+                    if(mag*mag < small_num)
                     {
                         c = 1;
                         s1 = 0;
@@ -651,6 +652,8 @@ ROCSOLVER_KERNEL void syevj_diag_kernel(const rocblas_int n,
         sh_bottom[tix] = x2;
     }
 
+    S small_num = get_safemin<S>()/eps;
+
     // for each off-diagonal element (indexed using top/bottom pairs), calculate the Jacobi rotation and apply it to A
     i = x1;
     j = x2;
@@ -662,7 +665,7 @@ ROCSOLVER_KERNEL void syevj_diag_kernel(const rocblas_int n,
             mag = std::abs(aij);
 
             // calculate rotation J
-            if(mag < eps)
+            if(mag*mag < small_num)
             {
                 c = 1;
                 s1 = 0;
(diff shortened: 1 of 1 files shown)


library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -166,6 +166,7 @@ __device__ void run_syevj(const rocblas_int dimx,
         local_diag += std::real(sines_diag[i]);
     }
     S tolerance = (local_res + local_diag) * abstol * abstol;
+    S small_num = get_safemin<S>()/eps;
```
EdDAzevedo: I see std::norm(aij) used. Is there a danger about numerical overflow, say aij = {1.5e+154, 1.5e+154}, then std::abs( aij ) is representable, but  std::norm( aij ) may cause overflow, or say aij = {1.0e-200, 1.0e-200}, std::abs(aij) is non-zero, but std::norm(aij) is zero.

What if we do not make the test "if (mag < eps)" but test "if (mag == 0)" then try to compute c, s1 otherwise? There is the step "s1 = s * aij / mag", maybe it should be "s1 = (aij/mag) * s", where (aij/mag) should have magnitude of around 1?

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -187,7 +188,7 @@ __device__ void run_syevj(const rocblas_int dimx,
                     aij = Acpy[i + j * n];
                     mag = std::abs(aij);
 
-                    if(mag < eps)
+                    if(mag*mag < small_num)
```
EdDAzevedo: I have a concern about "mag * mag" causing overflow. Would testing  "if (mag < sqrt_small_num)" be feasible?

library/src/lapack/roclapack_syevj_heevj.hpp
jmachado-amd: Those are good points! In the end I committed the smallest change that would prevent the original bug from happening, while guaranteeing that no subsequent computation would suffer.

Scaling the matrix (if its inf norm is either too big or too small) before doing the actual eigen decomposition would mitigate the majority of the underflow and overflow issues, however the tests currently don't include matrices with large or tiny inf norms. Thus, I thought that it would not be advisable to do significant changes to the code at this point in time and potentially break cases that might not be very accurate but still work. I plan to revisit this once the new tests are ready.

library/src/lapack/roclapack_syevj_heevj.hpp
jmachado-amd: As I understand, if `mag * mag` overflows to +Inf or if `mag` is an NaN the computation proceeds to the `else` case in the same way the test "if (mag < sqrt_small_num)" would.

tfalders (approved):
LGTM!

library/src/lapack/roclapack_syevj_heevj.hpp
EdDAzevedo: My point is "if (mag < sqrt_small_num)" is conceptually the same as "if (mag * mag < small_num)" but testing with "sqrt_small_num" will avoid unnecessary overflow or underflow when computing "mag * mag" may cause overflow or underflow.

jzuniga-amd (approved):
LGTM

PR #701, opened 2024-04-16
Added getrf_64

tfalders:


library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -71,7 +73,7 @@ ROCSOLVER_KERNEL void gemm_noconj_kernel(const rocblas_int m,
     T temp = 0;
     if(i < m && j < n)
     {
-        for(rocblas_int l = 0; l < k; l++)
+        for(I l = 0; l < k; l++)
             temp += A[i * inca + l * lda] * B[l * incb + j * ldb];
```
EdDAzevedo: Perhaps these should be computed using 64bit arithmetic, even for 32bit rocblas_int? 
Minor suggestion: avoid using "l" (ell) as variable since it may look like "1" (one).  
Perhaps consider something like
for(I el=0; el < k; el++) {
    temp += A[  static_cast<rocblas_stride>(i) * inca + static_cast<rocblas_stride>(el) * lda ]   ...

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -178,7 +180,11 @@ rocblas_status rocsolver_gemm(rocblas_handle handle,
     if(m == 0 || n == 0 || k == 0 || batch_count == 0)
         return rocblas_status_success;
 
-    if(inca == 1 && incb == 1 && incc == 1)
+    bool is_inc1 = (inca == 1 && incb == 1 && incc == 1);
+    bool is_32bit
+        = (!std::is_same<I, int64_t>::value
+           || (lda * k < INT_MAX && ldb * n < INT_MAX && ldc * n < INT_MAX && batch_count < INT_MAX));
```
EdDAzevedo, line 186: Question: should expressions  like "lda * k < INT_MAX " or "ldb * n" or "ldc * n" be computed using 64bit arithmetic to avoid possible 32 bit integer overflow? Otherwise, the expression might always be less than INT_MAX, even if the product is larger than INT_MAX, due to 32 bit  overflow.

Maybe it is harmless and more robust to use something like   
(   (static_cast<int64_t>(lda) * k) < INT_MAX   )

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -178,7 +180,11 @@ rocblas_status rocsolver_gemm(rocblas_handle handle,
     if(m == 0 || n == 0 || k == 0 || batch_count == 0)
         return rocblas_status_success;
 
-    if(inca == 1 && incb == 1 && incc == 1)
+    bool is_inc1 = (inca == 1 && incb == 1 && incc == 1);
```
EdDAzevedo: Very minor nit picking suggestion: it may be a good practice to add "const" qualifier whenever possible. It helps the compiler analysis and makes the code more robust by avoiding unintended modification.

Maybe something like
const     bool is_inc1 = (inca == 1) && (incb == 1) && (incc == 1);

EdDAzevedo (approved):

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
tfalders: The expression `!std::is_same<I, int64_t>::value` should return true whenever `I` is not 64-bit and short-circuit the boolean evaluation. Therefore, "lda * k", "ldb * n", and "ldc * n" will only ever be evaluated using 64-bit arithmetic.

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
tfalders: Sounds good. I'll add it in.

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
EdDAzevedo: A very minor nit picking comment: would testing  directly
std::is_same<I,int32_t>      be more clear?

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
tfalders: gemm is only one of the functions called by getrf (and usually we'll call rocBLAS rather than use this internal kernel), so I don't expect changing this to improve correctness in a meaningful way. We'd have to convert all index calculations to 64-bit across all nested functions in order to see a benefit from this, which I consider outside the scope of this PR.

As for the variable "l", I'm not really sure what else to call it since i, j, k, m, and n are all taken. "idx", perhaps?

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
tfalders: That's what I had originally, but @jmachado-amd suggested that it might be possible for `I` to be a different 32-bit type that is not `int_32_t`, such as `unsigned int`, and that this might be a safer approach.

EdDAzevedo: Just a minor suggestion.    If it is possible to encounter say int32_t and uint32_t, or int64_t and uint64_t, then perhaps  consider break up the bool conditions into simpler steps. 

Maybe something like 



bool const is_32bit_int = (std::is_same<I,int32_t>::value() ||  std::is_same<I, uint32_t>::value());
bool const is_size_fit_32bit  = ( int64_t(lda) * k < INT_MAX)  &&  
                                                     (int64_t(ldb) * n < INT_MAX) && 
                                                     (int64_t(ldc) * n < INT_MAX); 
bool const use_32bit_interface = is_32bit_int ||    is_size_fit_32bit;

--------------
A separate issue: would the 32bit interface still work for cases   lda * k > INT_MAX,    ldb * n > INT_MAX,    ldc * n > INT_MAX?  

One may think as long as the internal index calculations use 64bit arithmetic the 32bit interface should still work correctly.      

Perhaps the test is         

bool const is_size_fit_32bit =  (lda < INT_MAX) && (ldb < INT_MAX) && (ldc < INT_MAX) && (k < INT_MAX) && (n < INT_MAX)?

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -73,8 +73,8 @@ ROCSOLVER_KERNEL void gemm_noconj_kernel(const I m,
     T temp = 0;
     if(i < m && j < n)
     {
-        for(I l = 0; l < k; l++)
-            temp += A[i * inca + l * lda] * B[l * incb + j * ldb];
+        for(I idx = 0; idx < k; idx++)
+            temp += A[i * inca + idx * lda] * B[idx * incb + j * ldb];
```
EdDAzevedo, line 77: Perhaps consider using 64bit arithmetic for index calculations, even for 32bit index type.

Maybe something like the following.  It should be harmless if I is already type int64_t.

temp += A[ int64_t(i) * inca + int64_t(idx) * lda ] * B[ int64_t(idx) * incb + int64_t(j) * ldb ];

EdDAzevedo (approved):

cgmb (approved):
Thanks Troy. This looks like a good start, and something that we can continue to improve going forward. LGTM.

jzuniga-amd (approved):
LGTM

tfalders:
For some reason, CI status isn't being reported to github. It looks like all the tests have passed, so if there are no objections, I'll force the merge.

cgmb:
The tests are passing, so feel free to force the merge.

http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin%2FrocSOLVER/detail/PR-701/3/pipeline

PR #711, opened 2024-05-03
Update soversion

cgmb:
We haven't been updating the soversion, but we should be doing so each release. The minor version gets bumped whenever we add a new symbol to the public ABI. The major version gets bumped when we make an incompatible change to the ABI. It's basically like the normal version number but it's specific to the ABI. The patch number is set by rocm-cmake during the official builds, so we leave it unset.


jzuniga-amd (approved):

jzuniga-amd:
oh! This is targeting 6.2 release branch. Does this mean that so version in develop should become 0.3?

cgmb:
> oh! This is targeting 6.2 release branch. Does this mean that so version in develop should become 0.3?

Yes.

tfalders (approved):

PR #714, opened 2024-05-07
Fix stein initial eigenvectors' choices

jmachado-amd:
This PR fixes a bug in `roclapack_stein` where eigenvectors would fail to converge because the initial values used in stein would be orthogonal to the real eigenvectors (to numerical precision).  It improves stein's initial choice by sampling the components of the starting vector from a pseudorandom number generator.  The generator is stateless, previous iterates are kept as local variables in the caller.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_stein.hpp
+++ b/library/src/auxiliary/rocauxiliary_stein.hpp
@@ -35,9 +35,9 @@
 #include <type_traits>
 
 #include "lapack_device_functions.hpp"
-#include "rocsolver_prng.hpp"
 #include "rocblas.hpp"
 #include "rocsolver/rocsolver.h"
+#include "rocsolver_prng.hpp"
 
 ROCSOLVER_BEGIN_NAMESPACE
 
@@ -126,7 +126,7 @@ __device__ void run_stein(const int tid,
             ifail[i] = 0;
 
     auto normalize = [](rocblas_int value, rocblas_int max) -> S {
-        return static_cast<S>(static_cast<double>(value)/max);
+        return static_cast<S>(static_cast<double>(value) / max);
     };
     // iterate over submatrix blocks
     for(rocblas_int nblk = 0; nblk < iblock[nev - 1]; nblk++)
@@ -180,10 +180,11 @@ __device__ void run_stein(const int tid,
                 // deterministic seed, varies with thread ids and with each new computed eigenvalue
                 seed = tid + (j + nblk * nev) * MAX_THDS + 1;
 
-                // rocSOLVER's prng can be initialized with one or two seeds (better results are achieved if 
+                // rocSOLVER's prng can be initialized with one or two seeds (better results are achieved if
                 // two, independent, seeds are used).  The two seeds used here are not independent, they just spread
                 // the first pseudo-random number each thread computes more evenly.
-                rocsolver_int_prng<rocblas_int> prng(seed, rocsolver_int_prng<rocblas_int>::max() + 1 - seed);
+                rocsolver_int_prng<rocblas_int> prng(
+                    seed, rocsolver_int_prng<rocblas_int>::max() + 1 - seed);
 
                 for(i = tid; i < blksize; i += MAX_THDS)
                 {
--- a/library/src/auxiliary/rocauxiliary_stein.hpp
+++ b/library/src/auxiliary/rocauxiliary_stein.hpp
@@ -32,6 +32,8 @@
 
 #pragma once
 
+#include <type_traits>
+
 #include "lapack_device_functions.hpp"
 #include "rocblas.hpp"
 #include "rocsolver/rocsolver.h"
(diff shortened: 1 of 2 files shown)


jmachado-amd:
I made this PR to start the conversation. If everyone agrees that the proposed solution is adequate, we might start discussing whether this can stay as is or if it would be better to move the pseudorandom number generator elsewhere (and to add it to the library tests).

Of course, the code style here is a bit different, please let me know of anything you think should be changed to better reflect rocsolver's codding style.

library/src/auxiliary/rocauxiliary_stein.hpp
```diff
@@ -88,6 +90,190 @@ __device__ void stein_reorthogonalize(rocblas_int i,
     }
 }
 
+namespace detail {
```
EdDAzevedo: If a way to generate random numbers is needed, perhaps we can use system software (say drand48())  or reuse existing software and don't need to bring in another random number generator? 

 Perhaps we don't need the highest quality of random numbers?

jmachado-amd: That is a good point which I forgot to clarify, the new code generates pseudorandom numbers in the GPU without further dependencies. If there is a known source of good random numbers for `__device__` methods or if it is the plan to add a true random number generator as a dependency to rocSOLVER I would be more than happy to update the code to use that.

library/src/auxiliary/rocauxiliary_stein.hpp
```diff
@@ -88,6 +90,199 @@ __device__ void stein_reorthogonalize(rocblas_int i,
...
+    //
+    //! \return Number (Y - Z)/(2^31 - 1) cast to type S.
+    //!
+    template <typename S,
+              typename K = I,
+              typename = typename std::enable_if<std::is_same<std::decay_t<K>, I>::value>::type>
+    __device__ __host__ static auto uniform(K&& Y, K&& Z = 0) -> S
+    {
+        constexpr double range = static_cast<double>(m_m31_mXY);
+        I Xnext = next_X(Y, Z);
```
tfalders: My understanding of C++ was that a method couldn't be used until after it had been declared, but this seems to refute that. Am I missing something?

tfalders (changes requested):
Thanks for taking care of this, @jmachado-amd! Looks good! As discussed previously, I'd like to see the pseudorandom generator moved into a separate file. I'll approve after that.

Also, I had a quick question since it seems like I'm behind on some C++ developments.

library/src/auxiliary/rocauxiliary_stein.hpp
```diff
@@ -168,9 +364,14 @@ __device__ void run_stein(const int tid,
...
                 // initialize starting eigenvector
-                // TODO: how to make it random?
+                Ytid = Ztid = tid + (j + nblk * nev) * MAX_THDS + 1;
+                prng::range_check_Y(Ytid);
+                prng::range_check_Z(Ztid);
                 for(i = tid; i < blksize; i += MAX_THDS)
-                    work[i] = (i == j - j1 ? S(1) : S(-1) / (blksize - 1));
+                {
+                    // work[i] lies in the interval [-1, 1)
+                    work[i] = 2 * (prng::uniform<S>(Ytid, Ztid) - 1.0 / 2);
```
EdDAzevedo: Is there an internal state (say the "seed") in this random number generator? Is it taking "Ytid" and "Ztid" as the initial "seed".  

Would prng::uniform(Ytid,Ztid)   give the same random value for fixed unchanging value of Ytid, and Ztid? 

A concern is whether multiple entries in "work[i]" might get the identical random number value for fixed tid  (and so fixed Ytid and Ztid values).  

Perhaps one need something like 
"Ytid = next_Y(Ytid); Ztid = next_Z(Ztid)"  
or perhaps Ytid, and Ztid are generated in an expression that involve variable "i" as well, such as 

Ytid = Ztid = i + (j + nblk * nev) * MAX_THDS + 1;    // inside the for(i = tid) loop

to update the state to advance to another random number?    

Just checking.

library/src/auxiliary/rocauxiliary_stein.hpp
cgmb: I'm not going to pretend I know much about C++ parsing, but my understanding is that the whole class is parsed before any inline member function bodies are considered. You can imagine that these inline function bodies are just a convenient shorthand that is equivalent to putting the implementation immediately following the class.

jmachado-amd: I understand that both of you are correct, in the sense that a function scope is the part of the translation unit that follows its declaration, but names declared inside a class are also visible inside all of its function bodies -- effectively as if the inline function bodies were defined immediately after the class declaration.

I think that this is clarified in Stroustrup's "The Design and Evolution of C++", but I don't have that book anymore to check or give pointers. For concreteness (and a little bit of terseness), class scope appears in item 3.3.7(1) of the C++11 standard ([N34242](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3242.pdf)), which I quote below:

> The potential scope of a name declared in a class consists not only of the declarative region following the name’s point of declaration, but also of all function bodies, brace-or-equal-initializers of non-static data members, and default arguments in that class (including such things in nested classes).

For C++17, see N4659:6.3.7 (1).

library/src/auxiliary/rocauxiliary_stein.hpp
jmachado-amd: The generator has no internal state, the variables `Ytid` and `Ztid` are not only the seeds but are also being updated at each call to `uniform` with the new computed values of their respective sequences.  What is happening is that all of those methods (`next_X`, `uniform`, etc.) modify the variables passed to them, thus the code:
```.cpp
 work[i] = 2 * (prng::uniform<S>(Ytid, Ztid) - 1.0 / 2);
```
is both returning a value (that is stored in `work[i]`) and updating the values of `Ytid` and `Ztid` (with `next_Y(Ytid)` and `next_Z(Ztid)`). The values stored in `work` won't repeat before the main sequence `Xn` runs over its entire period.

cgmb: As mentioned in the other thread, `- 1.0 / 2` will implicitly convert the computation to double. I would recommend `- S(0.5)`. We should perhaps consider taking the time to turn on more warnings as I believe there is a flag we could use to catch these sorts of implicit conversions.

It's probably not a big deal in this case, but it pays to keep these things tidy even where it doesn't matter, so that mistakes stand out in the places where it does.

EdDAzevedo: Is the meaning of "&&" double ampersand in  the argument list in something like "uniform(K&& Y, K&& Z = 0)" mean X and Y need to be reference to "rvalue" and so cannot be modified?  

So can   something like "uniform(Ytid, Ztid)" also modify variables Ytid and Ztid?

Perhaps the interface can be declared as "uniform(K& Y, K& Z=0)" ?

jmachado-amd: Agreed, will fix.

jmachado-amd: Ok, so this is a piece of C++ that can be confusing. Quoting [Scott Meyers](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers), the `&&` in a type declaration _sometimes_ means rvalue reference, but sometimes it means _either rvalue reference or lvalue reference_. In the latter case the `&&` is called a _[forwarding reference](https://en.cppreference.com/w/cpp/language/reference)_.

A somewhat simplified way to distinguish both cases is checking whether the declaration involves any type deduction. If there is no type deduction, then the `&&` do mean rvalue reference. But in the templatized method `uniform(K&& Y, K&& Z = 0)`, `Y` and `Z` have type `K&&` for some deduced type `K`, which implies that the two `&&`s are forward references and can mean either `K&&` or `K&` according to how the method `uniform` is called. The reason it is being used this way is exactly to allow calling those methods with lvalues or rvalues, as a declaration such as `uniform(K& Y, K& Z = 0)` would be ill-formed, the non-const lvalue reference `Z` cannot bind to the rvalue `0` in the declaration.

I do recommend you have a look at Scott Meyers [essay on the topic](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers), that was really useful for me. Just notice that he uses the name "universal reference" to denote "forwarding reference", I understand that, at the time he wrote the essay, there was no consensus on how those should be called. Anyway, I will think about a way to make the code clearer in this regard.

cgmb: If that's the only reason it supports taking rvalues, perhaps it could instead have a function that takes two arguments and an overload that takes one? IDK. I don't have particularly strong opinions on this.

EdDAzevedo: Hi Julio, Thanks for the link to the informative  article. I wasn't aware of the universal value meaning of "rT&&".  Would it then be too lax in accepting rvalue reference as well such as "uniform(Ytid + 1)" or "uniform(13,0)"?  If the "Z = 0" is not needed, perhaps just using "K&" is sufficient?   Just a thought.

jmachado-amd: Perhaps the code is trying to do too much. The struct provides 3 linear congruential sequences, two of which are independent (Yn, Zn) and have a period of order ~2^31, and the third (Xn) that uses the previous ones to yield a period of order ~2^56. The suggestion is to always use Xn, but there might be situations in which we want to use Yn and Zn directly, thus the attempt to provide an interface that would allow using all of them.

In the particular case of the `uniform` method, using `uniform(Y, Z)` will yield values distributed as Xn, using `uniform(Y)` (or `uniform(Y, 0)`) will yield values distributed as Yn, and using `uniform(0, Z)` will yield values distributed as Zn. Thus, to remove the forwarding references while keeping the same functionality one would have to add one overload to each case (and revisit the rest of the API too, the lvalue references are there to handle the updating logic of each sequence internally).

It looks like adding a simple wrapper to handle the internal state might make everything more clear.

gsm-mgv: Hi!

I randomly came across this PR  while looking at rocm related things.

I had a few comments/questions; I've included them all here to reduce clutter. Apologies if I am off the mark here, or intruding where I should not be.

Regarding STL and modularity:
- Is there any reason why the STL provided prng cannot be used? See: https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution
- Assuming the standalone implementation is required (ie. no STL), is it worth simplifying the implementation and making it more concrete/explicit to the stein impl? As it currently stands, it is heavily templated, which provides for a level of genericity, however, it is only being utilised specifically for this purpose (if I'm not mistaken). This could potentially be extracted and placed in a common header for use in other parts of the code, if required.
- The interface is a bit abstract; You could potentially consolidate most of the functionality into a class (or struct) and construct using the typical parameters for a distribution. For example, a gaussian distribution can be parameterised by the mean/stddev. In this instance, a uniform dist could be parameterised on the bounds of output. ie. [lower, upper). The dist object could subsequently be drawn from using a sample() method (or something to that effect) (note that this does not solve the problem of requiring an underlying prng).

Regarding code specifics:
- Why are integers preferred, instead of unsigned integers, especially when performing modulo arithmetic? (more precautions are required)
- For both of the range checks, why is a modulo operator not sufficient? (this may be related to the previous point)
- Regarding the use of &&, it's unclear if it provides much benefit in this instance. In almost all cases where it is being used, the value is immediately cast to an integer. I'm speculating, but it's likely it would be both more performant and easier to digest the code if the arguments were by value. By value would be strictly more performant than by ref, at least for 64 bit machines (assuming the size of the integer is < size of pointer, which is likely the case). It should be on par for the majority of cases, otherwise.

Anyway, just a few thoughts. Apologies again if this is not the place (or I'm plain wrong).

EdDAzevedo (approved):

library/src/include/rocsolver_prng.hpp
```diff
@@ -0,0 +1,276 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ * *************************************************************************/
+
+#pragma once
+
+#include <type_traits>
+
+ROCSOLVER_BEGIN_NAMESPACE
```
tfalders, line 34: You might want to include rocblas_utility.hpp, where this macro is defined.

tfalders (approved):
LGTM! Just one comment

library/src/include/rocsolver_prng.hpp
jmachado-amd: Yes, I did trim too much. Good catch as always, thanks!

PR #717, opened 2024-05-09
Multikernel option for bdsqr optimization

tfalders:
As previously discussed, I have been experimenting with optimizing BDSQR by using multiple kernel launches, with device synchronizations to determine the iterative loop's stopping condition.

Broadly speaking, I have made the following changes:
- `bdsqr_t2bQRstep` and `bdsqr_b2tQRstep` have been combined into a single device function. This was intended to allow different split blocks to be processed in the same thread group, but it wasn't performant. Still, I've kept this change.
- `bdsqr_kernel` has been divided into multiple different kernels, with the iterative loop moved to the CPU.
  - `bdsqr_compute` determines the shift for the current split block and applies the QR step to D and E. If nv, nu, and nc are less than the specified switch size, it will also update the singular vectors.
  - `bdsqr_rotate` updates the singular vectors using multiple thread groups.
  - `bdsqr_update_endpoints` updates the endpoints of each split block, and spawns new split blocks if zeroes are found in the middle of a split block.
- `bdsqr_chk_completed` is a new kernel that will determine if the stopping criterion has been met for each batch instance.
- The `work` and `splits_map` arrays have been expanded to accommodate new information, and a new `completed` array has been added to hold information about problem status. See the comment block on line 316 for an explanation of how data is stored in these arrays.


library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -71,772 +71,925 @@ __device__ T bdsqr_estimate(const rocblas_int n, T* D, T* E, int t2b, T tol, int
...
+            __syncthreads();
+
+            // -----------
+            // swap arrays
+            // -----------
+
+            auto const j_start = tid;
+            auto const j_inc = nthreads;
+
+            __syncthreads();
```
qjojo: Is this __syncthreads() needed? All that happens between it and the last one is creating two local variables from kernel arguments.

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -125,10 +122,15 @@ __device__ void bdsqr_QRstep(const rocblas_int tid,
             D[dk + dir] = c * D[dk + dir];
 
             // save rotations to update singular vectors
-            if((t2b && nv) || (b2t && (nu || nc)))
+            if(t2b && nv)
             {
-                rots[rk] = c;
-                rots[rk + 1] = s;
+                rots[ek] = c;
```
EdDAzevedo, line 127: Minor suggestion: perhaps a macro or lambda can hide the design decision on how the cosine and sines values are stored?  This may centralize the design decision to allow the storage format to be easily  change. 

Maybe something like

auto cosine = [=](auto ek) -> S& {  return( rots[ek] ); };
auto sine = [=](auto ek) -> S& { return( rots[ek + n]); };

then use as something like 

cosine(ek) = c;  
sine(ek) = c;

Just a suggestion.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_bdsqr.hpp
tfalders: I think you might be right. I can take a look at removing this one.

library/src/auxiliary/rocauxiliary_bdsqr.hpp
tfalders: That does seem like it could be helpful. I'll consider it when we revisit this code to integrate your LASR device function.

EdDAzevedo (approved):

jzuniga-amd (approved):
LGTM

PR #721, opened 2024-05-18
Partial eigenvalue decomposition with divide & conquer

jzuniga-amd:
- Modifies STEDC and STEBZ to accommodate divide & conquer for partial decompositions
-  Adds initial optimizations to merge phase of divide & conquer
- Adds new family of APIs divide & conquer + bisection (SYEVDX/HEEVDX, SYGVDX/HEGVDX)


clients/include/testing_stedcj.hpp
```diff
@@ -265,48 +242,44 @@ void stedcj_getError(const rocblas_handle handle,
...
-                    hA[0][i + j * lda] = hE[0][i];
-                else
-                    hA[0][i + j * lda] = 0;
-            }
+            if(i == j)
+                hA[0][i + j * lda] = hD[0][i];
+            else if(i + 1 == j)
+                hA[0][i + j * lda] = hE[0][i];
+            else
+                hA[0][i + j * lda] = 0;
```
EdDAzevedo: just double checking: will  the lower triangular part of hA need to be set to zero?

clients/include/testing_stedcj.hpp
```diff
@@ -325,33 +298,28 @@ void stedcj_getError(const rocblas_handle handle,
...
-                              hC[0] + j * ldc, 1);
-            }
+        // both eigenvalues and eigenvectors needed; need to implicitly test
+        // eigenvectors due to non-uniqueness of eigenvectors under scaling
 
-            // error is ||hC - hCRes|| / ||hC||
-            // using frobenius norm
-            *max_errv = norm_error('F', n, n, ldc, hCRes[0], hC[0]);
+        // multiply A with each of the n eigenvectors and divide by corresponding
+        // eigenvalues
```
EdDAzevedo: just double checking: are the eigenvalues guaranteed to be away from zero and not cause overflow by division?

do we need to check whether the eigen vectors are ortho-normal?

Perhaps check something like

norm( A - V * diag(eig) * V' ) <= tol * norm(A) 
and 
norm( V'*V - eye ) <= tol ?

docs/reference/intro.rst
```diff
@@ -205,6 +205,10 @@ LAPACK-like functions
     :ref:`rocsolver_sygvdj <sygvdj>`, x, x, ,
     :ref:`rocsolver_heevdj <heevdj>`, , , x, x
     :ref:`rocsolver_hegvdj <hegvdj>`, , , x, x
+    :ref:`rocsolver_syevdj <syevdx>`, x, x, ,
+    :ref:`rocsolver_sygvdj <sygvdx>`, x, x, ,
+    :ref:`rocsolver_heevdj <heevdx>`, , , x, x
+    :ref:`rocsolver_hegvdj <hegvdx>`, , , x, x
```
tfalders: Mismatch on the function names

library/src/CMakeLists.txt
```diff
@@ -29,156 +29,195 @@
...
   lapack/roclapack_gelqf_strided_batched.cpp
-  # Problem and matrix reductions (diagonalizations)
+  #- left column compression
+  lapack/roclapack_gerq2.cpp
+  lapack/roclapack_gerq2_batched.cpp
+  lapack/roclapack_gerq2_strided_batched.cpp
+  lapack/roclapack_gerqf.cpp
+  lapack/roclapack_gerqf_batched.cpp
+  lapack/roclapack_gerqf_strided_batched.cpp
+  #### Problem and matrix reductions (diagonalizations) ####
```
tfalders: Minor issue, but it would be nice if these headers all used consistent capitalization

library/src/include/ideal_sizes.hpp
```diff
@@ -295,6 +295,14 @@
 #define SYEVDJ_MIN_DC_SIZE 16
 #endif
 
+/****************************** syevdx ******************************************
+*******************************************************************************/
+/*! \brief Determines the minimum size required for the Bisection divide and conquer to be used.
+
+    \details If the size of the block is smaller than SYEVDX_MIN_DC_SIZE,
+    the eigenvectors are computed with the normal inverse iteration algorithm. */
+#define SYEVDX_MIN_DC_SIZE 16
```
cgmb, line 305: Please guard with `#ifndef SYEVDX_MIN_DC_SIZE`, for consistency.

library/src/auxiliary/rocauxiliary_stedcj.cpp
```diff
@@ -119,52 +116,48 @@ ROCSOLVER_END_NAMESPACE
 extern "C" {
 
 ROCSOLVER_EXPORT rocblas_status rocsolver_sstedcj(rocblas_handle handle,
-                                                  const rocblas_evect evect,
```
cgmb: I would agree with @tfalders that we should avoid changing the function signature. It's a private function, but it's easier to reason about library compatibility if we _never_ change the ABI for any visible symbols.

clients/gtest/auxiliary/stedcx_gtest.cpp
```diff
@@ -0,0 +1,158 @@
...
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<vector<int>, vector<int>> stedcx_tuple;
+
+// each size_range vector is a {n, ldc}
+
+// each ops_range vector is a {rng, vl, vu, il, iu}
+// if rng = 0, then find all eigenvalues
+// if rng = 1, then find eigenavlues in (vl, vu]
```
tfalders: Spelling on eigenvalues

library/src/lapack/roclapack_syevdx_heevdx.hpp
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    const rocblas_stride stride = n;
+
+    // reduce A to tridiagonal form
+    rocsolver_sytrd_hetrd_template<BATCHED, T>(handle, uplo, n, A, shiftA, lda, strideA, D, stride,
+                                               E, stride, tau, stride, batch_count, scalars,
+                                               (T*)work1, (T*)work2, (T*)work3, (T**)nsplit_workArr);
+
+    if(evect != rocblas_evect_original || n < SYEVDX_MIN_DC_SIZE)
```
tfalders, line 255: Might it make more sense to call syevx directly, rather than recreating the same functionality?

tfalders (commented):
Some preliminary comments, in addition to those from today's meeting (i.e. adding evect to stedcj and stedcx).

I think you might need to apply clang format as well.

library/src/include/ideal_sizes.hpp
tfalders: Seconded

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -0,0 +1,898 @@
...
+            rocsolver_bench_inform(inform_invalid_args);
+
+        return;
+    }
+
+    // determine sizes
+    size_t size_A = size_t(lda) * n;
+    size_t size_B = size_t(ldb) * n;
+    size_t size_W = size_t(n);
+    size_t size_Z = size_t(ldz) * n;
```
tfalders, line 665: It occurred to me that maybe we shouldn't allocate Z if evect is none. Not necessary for this PR, since we'll probably want to do this for all the eigensolvers.

clients/common/auxiliary/testing_stedcj.hpp
```diff
@@ -208,18 +212,37 @@ void stedcj_initData(const rocblas_handle handle,
             if(E == 1)
                 hD[0][n - 1] = 1;
         }
+
+        // initialize C to the identity matrix
+        if(evect == rocblas_evect_original)
+        {
+            for(rocblas_int i = 0; i < n; i++)
+            {
+                for(rocblas_int j = 0; j < n; j++)
```
EdDAzevedo: very minor nit picking suggestion: make "for i" to be the inner loop to get efficient stride-1 access to memory.

clients/common/auxiliary/testing_stedcj.hpp
```diff
@@ -242,44 +265,48 @@ void stedcj_getError(const rocblas_handle handle,
 
     int lgn = floor(log(n - 1) / log(2)) + 1;
```
EdDAzevedo, line 266: do we need to check that (n-1) > 0 before taking the log(n-1)?   just checking.

library/src/auxiliary/rocauxiliary_stebz.hpp
```diff
@@ -68,9 +68,12 @@ __device__ void increasing_order(const rocblas_int nn, T* X, rocblas_int* Y)
         {
             X[s2] = X[s1];
             X[s1] = v;
-            bv = Y[s2];
-            Y[s2] = Y[s1];
-            Y[s1] = bv;
+            if(Y)
```
EdDAzevedo, line 73: Minor suggestion: if this is taking a significant amount of time, perhaps consider calling shell_sort_ascending() or shell_sort() in lib_device_helpers.hpp

clients/include/testing_stedcj.hpp
jzuniga-amd: Not necessary. It's been used in functions with parameter uplo = upper_triangular

clients/include/testing_stedcj.hpp
jzuniga-amd: yes, eigenvalues are away from zero for the input matrices that we generate now. 
(We can think on other cases when doing the main refactoring of the unit tests down the road).

EdDAzevedo (approved):
I am not familiar with all the details but on the surface it looks ok to me.

library/src/include/ideal_sizes.hpp
jzuniga-amd: done

library/src/auxiliary/rocauxiliary_stedcj.cpp
jzuniga-amd: Corrected. stedcj (and now stedcx) have same signature as before with parameter evect

docs/reference/intro.rst
jzuniga-amd: corrected

library/src/CMakeLists.txt
jzuniga-amd: changed

library/src/auxiliary/rocauxiliary_stedcx.hpp
```diff
@@ -0,0 +1,610 @@
...
+        *size_workArr = 0;
+        *size_splits = 0;
+        *size_tmpz = 0;
+        return;
+    }
+
+    size_t s1, s2;
+
+    // requirements for solver of small independent blocks
+    rocsolver_steqr_getMemorySize<T, S>(evect, n, batch_count, size_work_steqr);
```
tfalders, line 371: As far as I can tell, STEDCX does not call STEQR, but work_steqr is still used. Could we set size_work_steqr directly and remove this call, or is there a reason for having it?

clients/gtest/auxiliary/stedcx_gtest.cpp
jzuniga-amd: corrected

library/src/lapack/roclapack_syevdj_heevdj.hpp
```diff
@@ -108,8 +108,8 @@ void rocsolver_syevdj_heevdj_getMemorySize(const rocblas_evect evect,
                                                     &unused);
 
     // extra requirements for computing eigenvalues and vectors (stedcj)
-    rocsolver_stedcj_getMemorySize<BATCHED, T, S>(rocblas_evect_tridiagonal, n, batch_count, &w12,
-                                                  &w22, &w32, size_work4, size_workSplits, &unused);
+    rocsolver_stedcj_getMemorySize<BATCHED, T, S>(evect, n, batch_count, &w12, &w22, &w32,
+                                                  size_work4, size_workSplits, &unused);
```
tfalders: It makes sense to pass rocblas_evect_tridiagonal instead of evect, since the code will only ever pass rocblas_evect_tridiagonal to stedcj.

library/src/lapack/roclapack_syevdx_heevdx.hpp
```diff
@@ -0,0 +1,312 @@
...
+        *size_isplit = sizeof(rocblas_int) * n * batch_count;
+
+        // size of array for temporary split off block sizes
+        *size_nsplit_workArr = std::max(*size_nsplit_workArr, sizeof(rocblas_int) * batch_count);
+    }
+    else
+    {
+        // extra requirements for computing eigenvalues and vectors (stedcx)
+        rocsolver_stedcx_getMemorySize<BATCHED, T, S>(
+            evect, n, batch_count, &a3, &b3, &c3, size_work4, size_work5, size_work6_ifail, &unused);
```
tfalders: Same comment as for syevdj, with passing rocblas_evect_tridiagonal instead of evect.

library/src/auxiliary/rocauxiliary_stedcx.hpp
```diff
@@ -0,0 +1,610 @@
...
+
+    // 1. invalid/non-supported values
+    if(range != rocblas_erange_all && range != rocblas_erange_value && range != rocblas_erange_index)
+        return rocblas_status_invalid_value;
+    if(evect != rocblas_evect_none && evect != rocblas_evect_tridiagonal
+       && evect != rocblas_evect_original)
+        return rocblas_status_invalid_value;
+
+    // 2. invalid size
+    if(n < 0 || ldc < n)
```
tfalders: Might want to future proof by only checking values for C when evect != none.

library/src/lapack/roclapack_syevdx_heevdx.hpp
jzuniga-amd: I had it like this before, but as ifail is not a parameter for syevdx, I needed to combine it with work6 to re-use some workspace. I'm going to leave it this way for now. We can change it once we organize better the whole workspace allocation process.

library/src/auxiliary/rocauxiliary_stedcx.hpp
```diff
@@ -0,0 +1,610 @@
...
+    ROCSOLVER_LAUNCH_KERNEL(reset_info, gridReset, threads, 0, stream, info, batch_count, 0);
+
+    // quick return
+    if(n == 1)
+    {
+        if(evect != rocblas_evect_none)
+            ROCSOLVER_LAUNCH_KERNEL(reset_batch_info<T>, dim3(1, batch_count), dim3(1, 1), 0,
+                                    stream, C, strideC, n, 1);
+        ROCSOLVER_LAUNCH_KERNEL(stedcx_case1_kernel, dim3(batch_count), dim3(1), 0, stream, erange,
+                                vl, vu, D, strideD, nev, W, strideW);
```
tfalders, line 504: If range is I, but vl >= vu, the case1 kernel will return no eigenvalues even if they are valid.

clients/common/auxiliary/testing_stedcj.hpp
jzuniga-amd: changed

library/src/auxiliary/rocauxiliary_stedcx.hpp
```diff
@@ -0,0 +1,610 @@
...
+    }
+    else
+    {
+        nev[bid] = 1;
+        W[0] = d;
+    }
+}
+
+//--------------------------------------------------------------------------------------//
+/** STEDCX_SPLIT_KERNEL implements the solver phase of the DC algorithm to
```
tfalders: to...?

library/src/auxiliary/rocauxiliary_stedcx.hpp
```diff
@@ -0,0 +1,610 @@
...
+    __shared__ S sval[STEBZ_SPLIT_THDS];
+    __shared__ rocblas_int sidx[STEBZ_SPLIT_THDS];
+
+    run_stebz_splitting<STEBZ_SPLIT_THDS>(tid, range, n, vl, vu, il, iu, D, E, nsplit, W, splits,
+                                          tmpIS, pivmin, Esqr, bounds, inter, ninter, sval, sidx,
+                                          eps, ssfmin, compact);
+}
+
+//--------------------------------------------------------------------------------------//
+/** STEDCX_SYNTHESIS_KERNEL implements the solver phase of the DC algorithm to
```
tfalders: to...?

clients/common/auxiliary/testing_stedcj.hpp
jzuniga-amd: yes! If n <= 1, the workspace size should be 1.
Corrected.

library/src/auxiliary/rocauxiliary_stedcx.hpp
```diff
@@ -0,0 +1,610 @@
...
+    for(rocblas_int k = 0; k < maxlevs; ++k)
+    {
+        /** TODO: at the last level, kernels in steps b, c, and d could skip computations of
+            eigen values and vectors that are out of the desired range. Whether this could be
+            exploited somehow to improve performance must be explored in the future. For now,
+            as all values and vectors are computed concurrently (by different threads), skiping
+            the computation of some of them does not seem to make much difference. **/
+
+        // a. prepare secular equations
+        ROCSOLVER_LAUNCH_KERNEL((stedc_mergePrepare_kernel<rocsolver_stedc_mode_qr, S>),
```
tfalders: Shouldn't these kernels be using rocsolver_stedc_mode_bisection instead of rocsolver_stedc_mode_qr?

library/src/auxiliary/rocauxiliary_stedcj.hpp
```diff
@@ -403,35 +403,64 @@ rocblas_status rocsolver_stedcj_template(rocblas_handle handle,
...
+    rocblas_int numgrps3 = ((n - 1) / maxblks + 1) * maxblks;
+
+    // launch merge for level k
+    /** TODO: using max number of levels for now. Kernels return immediately when passing
+        the actual number of levels in the split block. We should explore if synchronizing
+        to copy back the actual number of levels makes any difference **/
+    for(rocblas_int k = 0; k < maxlevs; ++k)
+    {
+        // a. prepare secular equations
+        ROCSOLVER_LAUNCH_KERNEL((stedc_mergePrepare_kernel<rocsolver_stedc_mode_qr, S>),
```
tfalders: Shouldn't these kernels be using rocsolver_stedc_mode_jacobi instead of rocsolver_stedc_mode_qr?

tfalders (approved):
I've only skimmed the stedc kernels, but it looks good to me!

Looks like you might need to merge the release branch.

library/src/auxiliary/rocauxiliary_stebz.hpp
jzuniga-amd: The time is not significant. I tried to use the shell sort device functions but could not. I think the grid configurations are not quite compatible. leave it as it is for now; needs more investigation later.

library/src/auxiliary/rocauxiliary_stedcx.hpp
jzuniga-amd: For the solve phase, STEDCX is still using stedc_solve_kernel which uses STEQR. 
(The partial decomposition only happens at the merge phase, but we still need to compute all the eigenvalues of the sub-blocks so I am using the normal solver).

library/src/lapack/roclapack_syevdj_heevdj.hpp
jzuniga-amd: I forgot to change it back. 
Corrected.

library/src/lapack/roclapack_syevdx_heevdx.hpp
jzuniga-amd: done

library/src/auxiliary/rocauxiliary_stedcx.hpp
jzuniga-amd: changed

library/src/auxiliary/rocauxiliary_stedcx.hpp
jzuniga-amd: I don't see why...
If range = index, vl and vu are ignored.

library/src/auxiliary/rocauxiliary_stedcx.hpp
jzuniga-amd: mmm I forgot to complete the comments.
Done

library/src/auxiliary/rocauxiliary_stedcx.hpp
jzuniga-amd: completed

library/src/auxiliary/rocauxiliary_stedcx.hpp
jzuniga-amd: oh good catch!!
(right now it does not matter as we are using the same number of blocks/levels for all the methods, but it will do after the tuning)
Changed.

library/src/auxiliary/rocauxiliary_stedcj.hpp
jzuniga-amd: changed

library/src/auxiliary/rocauxiliary_stedcx.hpp
tfalders: Oh, I see! My mistake.

clients/common/auxiliary/testing_stedcj.hpp
EdDAzevedo: Note there is lapack   xLASET(uplo, m, n, alpha, beta, A, lda )  to set off-diagonal entries to alpha and diagonal to beta.  

It can be used to initialize a matrix to the  identity matrix using alpha = 0, beta = 1.

Just a thought.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1514,61 +1832,202 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_merge_kernel(const roc
...
                 }
             }
-            __syncthreads();
             /* ----------------------------------------------------------------- */
+        }
+    }
+}
+
+//--------------------------------------------------------------------------------------//
+/** STEDC_MERGEVECTORS_KERNEL updates vectors after merges are done. A matrix in the batch
```
jmachado-amd: Very minor: the kernel name in the comment does not match its actual name (`stedc_mergeUpdate_kernel`).

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1347,13 +1311,195 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_merge_kernel(const roc
...
+                             S* vecsA,
+                             rocblas_int* splitsA,
+                             const S eps,
+                             const S ssfmin,
+                             const S ssfmax)
+{
+    // threads and groups indices
+    /* --------------------------------------------------- */
+    // batch instance id
+    rocblas_int bid = hipBlockIdx_z;
```
jmachado-amd, line 1350: I know this can't be addressed now, but it would be nice if we could have more than 65k matrices in a batch.

jmachado-amd (approved):
Looks good, thanks Juan!

library/src/lapack/roclapack_syevdx_heevdx.hpp
```diff
@@ -0,0 +1,313 @@
...
+        rocsolver_stedcx_template<BATCHED, STRIDED, T>(
+            handle, rocblas_evect_tridiagonal, erange, n, vl, vu, il, iu, D, stride, E, stride, nev,
+            W, strideW, Z, shiftZ, ldz, strideZ, info, batch_count, (S*)work1, (S*)work2, (S*)work3,
+            (S*)work4, (S*)work5, work6_ifail, (S**)nsplit_workArr);
+
+        rocblas_int h_nev = (erange == rocblas_erange_index ? iu - il + 1 : n);
+        rocsolver_ormtr_unmtr_template<BATCHED, STRIDED>(
+            handle, rocblas_side_left, uplo, rocblas_operation_none, n, h_nev, A, shiftA, lda,
+            strideA, tau, n, Z, shiftZ, ldz, strideZ, batch_count, scalars, (T*)work1, (T*)work2,
+            (T*)work3, (T**)nsplit_workArr);
```
tfalders, line 307: Should this be encased in an `if(evect != rocblas_evect_none)` block? It seems to me that if Z is null and n >= SYEVDX_MIN_DC_SIZE then this could segfault.

Similar comment for syevdj.

library/src/auxiliary/rocauxiliary_stedc.hpp
jzuniga-amd: I will correct it

library/src/auxiliary/rocauxiliary_stedc.hpp
jzuniga-amd: Yes, we need to coordinate with rocblas to see if we take some approach together.

library/src/lapack/roclapack_syevdx_heevdx.hpp
jzuniga-amd: It is! If evect != evect_original, then the function does not use the D&C approach. (see line 255).

jzuniga-amd:
All tests are passing. I will correct the minor error in the comment above and then merge right away.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: IMHO, I  think this also requires launching with at most 

"max_batch = 64 * 1024 - 1; batch_nblocks = std::min( max_batch, batch_count)"  

thread blocks in z-dimension and the kernel code slightly modified that instead of just using  
"bid = hipBlockIdx_z", we can use a for loop with something like

auto const bid_start = hipBlockIdx_z; 
auto const bid_inc = hipGridDim_z;
for(auto bid = bid_start; bid < batch_count; bid += bid_inc) {    // body of kernel ... }

We might also reduce the amount of temporary scratch space required that is related to 
"min( batch_count, max_batch)". Perhaps using max_batch as low as  the number of  physical Compute Units (CU) (which is about O(100)) on the GPU device  is sufficient to keep the GPU device fully busy.

Just a suggestion.

PR #726, opened 2024-05-27
Fix stebz pivots lower bound

jmachado-amd:
This PR fixes a bug in which `stebz` can fail when the lower bound for pivots is too small.

For example, the following input matrix:
```.cpp
  double A[9] = {
      1.000000105,  0.000000005, 0.0,
      0.000000005,  1.000000105, 0.0,
      0.0, 0.0,     1.000000111
  };
```
will cause an underflow in line:
https://github.com/ROCm/rocSOLVER/blob/9e10562c3c226b11a04b9bf6a6afc01029353d14/library/src/auxiliary/rocauxiliary_stebz.hpp#L305
which yields `pmin = 0`.

For this input matrix, `stebz` crashes with a segmentation fault because many variables are not initialized correctly, for instance, see `lc_maxite`:
https://github.com/ROCm/rocSOLVER/blob/9e10562c3c226b11a04b9bf6a6afc01029353d14/library/src/auxiliary/rocauxiliary_stebz.hpp#L552

### DIFF
--- a/library/src/auxiliary/rocauxiliary_stein.hpp
+++ b/library/src/auxiliary/rocauxiliary_stein.hpp
@@ -32,6 +32,8 @@
 
 #pragma once
 
+#include <type_traits>
+
 #include "lapack_device_functions.hpp"
 #include "rocblas.hpp"
 #include "rocsolver/rocsolver.h"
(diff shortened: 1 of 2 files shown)


library/src/auxiliary/rocauxiliary_stein.hpp
```diff
@@ -168,9 +364,14 @@ __device__ void run_stein(const int tid,
...
                 // initialize starting eigenvector
-                // TODO: how to make it random?
+                Ytid = Ztid = tid + (j + nblk * nev) * MAX_THDS + 1;
+                prng::range_check_Y(Ytid);
+                prng::range_check_Z(Ztid);
                 for(i = tid; i < blksize; i += MAX_THDS)
-                    work[i] = (i == j - j1 ? S(1) : S(-1) / (blksize - 1));
+                {
+                    // work[i] lies in the interval [-1, 1)
+                    work[i] = 2 * (prng::uniform<S>(Ytid, Ztid) - 1.0 / 2);
```
cgmb, line 373: There will be an implicit promotion to double when `S = float`. What do you think about replacing `- 1.0 / 2` with something like `- S(0.5)`?

tfalders:
Is the change to STEIN intended to be included in this PR? I'd be happy to approve the change to STEBZ, but for STEIN I'd like to see the pseudorandom number generator moved to a separate file.

library/src/auxiliary/rocauxiliary_stein.hpp
jmachado-amd: Ouch, that file is not supposed to be part of this PR, it looks like I based it on the wrong commit. Thanks for the suggestion anyway!

PR #730, opened 2024-05-29
Allow building without the 64bit functions

cgmb:
This change enables rocSOLVER to be built using rocBLAS 4.1 (with reduced functionality). The rocSOLVER interfaces that rely on newer rocBLAS functionality will return the not implemented status.

It's a pain when the minimum version of rocBLAS required to build rocSOLVER is greater than the most recently released ROCm version, so this makes it possible to use an older version.


library/src/lapack/roclapack_getrf_batched.cpp
```diff
@@ -293,9 +309,13 @@ rocblas_status rocsolver_sgetrf_npvt_batched_64(rocblas_handle handle,
                                                 int64_t* info,
                                                 const int64_t batch_count)
 {
+#ifdef HAVE_ROCBLAS_64
```
EdDAzevedo, line 312: This is just a minor  general suggestion related to coding style.

Just as assigning ipiv = nullptr and passing ipiv instead of directly passing nullptr can improve the readability of the code, may I also suggest in general adding other variables such as say  
const rocblas_stride strideP = 0; const bool pivot = false; 
and pass these in as well.  
It is a little more work but this may make the intention of the code more clear.
Similarly, even for template arguments, one can consider using something like
constexpr bool BATCHED = false; constexpr bool STRIDED = false;
rocsolver_getrf_getMemorySize<BATCHED,STRIDED,T>( ... )  
Just a suggestion.

EdDAzevedo (approved):

jzuniga-amd (approved):
LGTM 
This is good idea. Thanks Cory!

jmachado-amd (approved):
This is great, thanks for doing it Cory!

library/src/include/rocblas.hpp
```diff
@@ -37,6 +37,211 @@
 #include "rocblas/internal/rocblas_device_malloc.hpp"
 #include "rocsolver_logger.hpp"
 
+#ifndef HAVE_ROCBLAS_64
+#if ROCBLAS_VERSION_MAJOR > 4 || (ROCBLAS_VERSION_MAJOR == 4 && ROCBLAS_VERSION_MINOR >= 2)
+#define HAVE_ROCBLAS_64 1
+#endif
+#endif
+
+// These function tempaltes help to provide compatibility with older versions
```
tfalders: Spelling on templates

library/src/include/rocblas.hpp
```diff
@@ -37,6 +37,211 @@
 #include "rocblas/internal/rocblas_device_malloc.hpp"
 #include "rocsolver_logger.hpp"
 
+#ifndef HAVE_ROCBLAS_64
+#if ROCBLAS_VERSION_MAJOR > 4 || (ROCBLAS_VERSION_MAJOR == 4 && ROCBLAS_VERSION_MINOR >= 2)
```
tfalders, line 41: I'm wondering a bit about future-proofing, since rocBLAS 4.2 does not have support for 64-bit gemms. How do we plan to handle this once they're added? Will `HAVE_ROCBLAS_64` be updated so that a newer version of rocBLAS is required for the 64-bit APIs, or will we add a new macro to handle the new version specifically? Should we maybe rename the macro to `HAVE_ROCBLAS_4_2` in that case?

tfalders (commented):
Looks good! Just a couple comments

library/src/include/rocblas.hpp
TorreZuk: rocblas 4.3 will have the gemm _64 templates

library/src/lapack/roclapack_getrf_batched.cpp
cgmb: > Similarly, even for template arguments, one can consider using something like
constexpr bool BATCHED = false; constexpr bool STRIDED = false;
rocsolver_getrf_getMemorySize<BATCHED,STRIDED,T>( ... )

I would prefer we define an internal enum and a few helper functions. e.g.,
```
enum rocsolver_fntype {
  rocsolver_fntype_single,
  rocsolver_fntype_batched,
  rocsolver_fntype_strided_batched,
};
constexpr bool is_batched(rocsolver_fntype fntype) {
  return fntype == rocsolver_fntype_batched || fntype == rocsolver_fntype_stided_batched;
}
constexpr bool is_strided(rocsolver_fntype fntype) {
  return fntype == rocsolver_fntype_stided_batched;
}
```
We can then use calls like `rocsolver_getrf_getMemorySize<rocsolver_fntype_single,T>( ... )` rather than `rocsolver_getrf_getMemorySize<false,false,T>( ... )`. When interfacing with rocBLAS, we would use the constexpr functions to convert, like with
`rocblas_abcxyz_getMemorySize<is_batched(fntype),is_strided(fntype),T>( ... )`.

EdDAzevedo: I agree having the 3 enum values may make the code more clear. Currently there are 4 possibilities with BATCHED=true,false, and STRIDED=true,false so there might be some ambiguity.

library/src/include/rocblas.hpp
cgmb: IDK. We could go either way. Let's cross that bridge when we come to it.

cgmb: Let's assume that we'll just bump the required version. If the person implementing it finds a better way, we can consider it then.

PR #736, opened 2024-06-06
cmake: add C as a project language, as clients use it

evetsso:
Similar in spirit to https://github.com/ROCm/rocPRIM/pull/567, as this fixes a CMake problem where LLVM can't find Terminfo.  But rocSOLVER really **does** use C in its client programs and toolchain-linux.cmake is justified in setting both CMAKE_CXX_COMPILER and CMAKE_C_COMPILER.

So instead, even though clients might not be built, it's easiest to just add C as a project language at the top level.


cgmb (commented):
The C language is enabled in `clients/CMakeLists.txt`. What is the problem this is solving?

evetsso:
> The C language is enabled in `clients/CMakeLists.txt`. What is the problem this is solving?

In release-staging/rocm-rel-6.2, a static library build (using a static library build of HIP) will produce this CMake error:

```
CMake Error at /usr/local/lib64/python3.6/site-packages/cmake/data/share/cmake-3.25/Modules/Internal/CheckSourceCompiles.cmake:44 (message):
  check_source_compiles: C: needs to be enabled before use.
Call Stack (most recent call first):
  /usr/local/lib64/python3.6/site-packages/cmake/data/share/cmake-3.25/Modules/CheckCSourceCompiles.cmake:76 (cmake_check_source_compiles)
  /opt/rocm/llvm/lib/cmake/llvm/FindTerminfo.cmake:32 (check_c_source_compiles)
  /opt/rocm/llvm/lib/cmake/llvm/LLVMConfig.cmake:174 (find_package)
  /opt/rocm/llvm/lib/cmake/clang/ClangConfig.cmake:10 (find_package)
  /usr/local/lib64/python3.6/site-packages/cmake/data/share/cmake-3.25/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /opt/rocm/lib/cmake/amd_comgr/amd_comgr-config.cmake:8 (find_dependency)
  /usr/local/lib64/python3.6/site-packages/cmake/data/share/cmake-3.25/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /opt/rocm/lib/cmake/hip/hip-config-amd.cmake:99 (find_dependency)
  /opt/rocm/lib/cmake/hip/hip-config.cmake:149 (include)
  CMakeLists.txt:185 (find_package)
```

The relevant blob of FindTerminfo.cmake looks like:
```cmake
  if(DEFINED CMAKE_C_COMPILER)
    include(CheckCSourceCompiles)
    check_c_source_compiles("${Terminfo_LINKABLE_SRC}" Terminfo_LINKABLE)
  else()
    include(CheckCXXSourceCompiles)
    check_cxx_source_compiles("${Terminfo_LINKABLE_SRC}" Terminfo_LINKABLE)
  endif()
```

So basically it's using CMAKE_C_COMPILER as a crude proxy to know whether it can build a test program as C or CXX.  Building the test program using either languages requires it to be enabled at the top-level CMake project.

The problem is that the build is setting both CMAKE_C_COMPILER and CMAKE_CXX_COMPILER, even though the rocSOLVER library itself would only use CXX.  Incidentally, upstream LLVM seems to have removed this whole FindTerminfo concept entirely, but I don't think we can count on that to be pulled in to ROCm 6.2.

Failing that, I can think of a few obvious ways to address this:
* Only set CMAKE_C_COMPILER if the project really will use C.  Currently it's toolchain-linux.cmake that sets it, and it doesn't know whether clients will be built or not.  I could also change that to be conditional on clients being built but that seems a bit fragile.
* Just add C as a language to the top-level project (what this PR does).

I don't mind looking at other solutions too, if anyone can think of any.

cgmb (approved):
This doesn't feel like an elegant solution, but it's a harmless workaround. Enabling both `C` and `CXX` is extremely common in C++ projects using CMake because it's the default behaviour if no language is specified.

TorreZuk (approved):
For what it is worth LGTM

jmachado-amd (approved):

PR #738, opened 2024-06-12
Hotfix: Update static rocprim dependency with correct name

lawruble13:
The rocPRIM package for static builds had to be renamed to `rocprim-static-dev[el]`, in order for the dependencies to be correct. As such, the names here need to be updated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,8 +203,8 @@ if(BUILD_WITH_SPARSE)
 endif()
 
 find_package(rocprim REQUIRED CONFIG PATHS ${ROCM_PATH})
-rocm_package_add_rpm_dependencies(STATIC_DEPENDS "rocprim-devel")
-rocm_package_add_deb_dependencies(STATIC_DEPENDS "rocprim-dev")
+rocm_package_add_rpm_dependencies(STATIC_DEPENDS "rocprim-static-devel")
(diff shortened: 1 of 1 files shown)


jmachado-amd (approved):

tfalders (approved):

cgmb (approved):
This was probably already explained to me, but what's the difference between rocprim-static-dev and rocprim-dev? Isn't it header-only?

lawruble13:
> This was probably already explained to me, but what's the difference between rocprim-static-dev and rocprim-dev? Isn't it header-only?

The dependencies, `rocprim-dev` depends on the shared HIP runtime, `rocprim-static-dev` depends on the static runtime. It's certainly not ideal to had, so if you have alternative suggestions for fixing that I am all ears.

cgmb:
> The dependencies, `rocprim-dev` depends on the shared HIP runtime, `rocprim-static-dev` depends on the static runtime. It's certainly not ideal to had, so if you have alternative suggestions for fixing that I am all ears.

In Fedora or Debian packaging, they would probably use an `or` so that the dependency is satisfied by either the static or dynamic runtime devel package. In rpm that's the actual word `or`. In deb, that's the symbol `|`.

cgmb:
That said, I have not deeply studied this matter, so it's up to you and Freddy to evaluate whether that's applicable and what path forward you want to take.

PR #740, opened 2024-06-13
Add CMake HIP language support

cgmb:
Add `USE_HIPCXX` option to build using CMake's support for the HIP language. This option requires CMake 3.21.3 or newer.

The rocsolver clients shouldn't need to be built as HIP, but they require `rocblas_is_complex<T>`, which is only provided by the rocblas headers when compiling as HIP. In the future, this could be corrected.

Note that `set_source_files_properties` only affects the targets declared in the same CMake file as the call to set the properties. As such, it is difficult to use together with `target_sources`. All uses of latter have therefore been removed. The project declaration in the clients directory has also been removed so that `PROJECT_SOURCE_DIR` can always be used to refer to the root directory.


cgmb:
This should not change anything about the rocsolver library unless users pass `-DUSE_HIPCXX=ON` when building it. The approach here is the same as used for rocALUTION in ROCm 6.0:
https://github.com/ROCm/rocALUTION/commit/5bd014fa8c71ed743876eef5ae4171a1cfdc715c

This was my test setup in an Ubuntu 24.04 docker container:
```bash
apt-get -y update
apt-get -y upgrade
apt-get -y install gnupg2 wget sudo
sudo mkdir --parents --mode=0755 /etc/apt/keyrings
wget https://repo.radeon.com/rocm/rocm.gpg.key -O - | gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg > /dev/null
echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/6.4 noble main" | sudo tee --append /etc/apt/sources.list.d/rocm.list
echo -e 'Package: *\nPin: release o=repo.radeon.com\nPin-Priority: 600' | sudo tee /etc/apt/preferences.d/rocm-pin-600
apt-get -y update
apt-get -y install rocm-dev rocblas-dev rocsparse-dev rocprim-dev build-essential cmake git libfmt-dev gfortran libopenblas-dev libgtest-dev
git clone https://github.com/cgmb/rocSOLVER.git -b use-cmake-hiplang
cd rocSOLVER
HIPCXX=/opt/rocm/llvm/bin/amdclang++ cmake -S. -Bbuild \
  -DCMAKE_HIP_ARCHITECTURES=gfx906 \
  -DBUILD_TESTING=ON \
  -DROCSOLVER_FIND_PACKAGE_LAPACK_CONFIG=OFF \
  -DUSE_HIPCXX=ON
make -j16 -C build
```

Note that with `-DUSE_HIPCXX=ON`, the default build target will be the user's GPU, not the full list of all architectures that rocSOLVER supports. I think that's the default we eventually want to move to, so it will be nice to be able to have that behaviour accessible via this non-default build configuration.

cgmb:
@haampie, this is pretty much what I'm going to suggest for the various rocm libs. At some point, we can then make it the default. I'd also like to collect the flags used for AMD's official ROCm releases (e.g., CMAKE_INSTALL_PATH, CMAKE_INSTALL_RPATH, CMAKE_HIP_ARCHITECTURES, etc.), and move them into a centralized location that defines the flags for the builds of the packages that AMD provides.

clients/gtest/CMakeLists.txt
```diff
@@ -168,26 +188,30 @@ if(WIN32)
...
     COMMAND ${CMAKE_COMMAND}
-    ARGS -E copy_directory $ENV{rocblas_DIR}/bin/rocblas/library ${PROJECT_BINARY_DIR}/staging/library
+    ARGS -E copy_directory $ENV{rocblas_DIR}/bin/rocblas/library $<TARGET_FILE_DIR:rocsolver-test>/library
   )
 endif()
 
 target_link_libraries(rocsolver-test PRIVATE
   $<IF:$<TARGET_EXISTS:GTest::gtest>,GTest::gtest,GTest::GTest>
-  hip::device
+  GTest::GTest
```
tfalders: Potential conflict with line 202 (see #931)

tfalders (approved):
LGTM!

AGonzales-amd (approved):

tfalders:
Test failure is unrelated. I'm forcing the merge.
EDIT: Oh, wait, the potential conflict is still there. I'll hold off for now.

clients/gtest/CMakeLists.txt
cgmb: Thanks. Updated.

tfalders:
Forcing the merge (for real this time)

PR #741, opened 2024-06-14
Remove loop unrolls from getri and trtri

tfalders:
This PR removes the loop unrolls from the small-size getri and trtri kernels, reducing the library size and allowing them to be built even if the -n flag is passed to the install script. My tests on gfx90a show a minimal impact on performance, though I wasn't able to run the batch tests due to a hang.

### DIFF
--- a/library/src/specialized/roclapack_getri_specialized_kernels.hpp
+++ b/library/src/specialized/roclapack_getri_specialized_kernels.hpp
@@ -42,9 +42,10 @@ ROCSOLVER_BEGIN_NAMESPACE
     the library size.
 *************************************************************/
 
-template <rocblas_int DIM, typename T, typename U>
+template <typename T, typename U>
 ROCSOLVER_KERNEL void __launch_bounds__(TRTRI_MAX_COLS)
-    getri_kernel_small(U AA,
+    getri_kernel_small(const rocblas_int n,
+                       U AA,
                        const rocblas_int shiftA,
                        const rocblas_int lda,
                        const rocblas_stride strideA,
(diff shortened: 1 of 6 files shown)


library/src/specialized/roclapack_getri_specialized_kernels.hpp
```diff
@@ -68,19 +69,18 @@ ROCSOLVER_KERNEL void __launch_bounds__(TRTRI_MAX_COLS)
...
+    __shared__ T common[TRTRI_MAX_COLS];
     T temp;
     rocblas_int jp;
 
     // read corresponding row from global memory in local array
-    T rA[DIM];
-#pragma unroll
-    for(int j = 0; j < DIM; ++j)
+    T rA[TRTRI_MAX_COLS];
+    for(int j = 0; j < n; ++j)
```
EdDAzevedo, line 78: Minor comment:
If there are sufficient registers, perhaps it is still possible to use limited unrolling,  such as say,  "#pragma unroll 4"?

Should there be guards for computing "i + j * lda" to avoid 32-bit overflow?
Perhaps something like say  

bool const is_large_n = 
      (int64_t(n)*int64_t(n)) != int64_t(n*n);  // 32-bit integer overflow?
if (is_large_n) {
    for(int j=0; j < n; ++j) { rA[j] = A[i + int64_t(j)*lda]; }
    }
else {
   for(int j=0; j < n; ++j) { rA[j] = A[i + j*lda];} // use 32-bit arithmetic
   }

Just a thought.

tfalders:
Unfortunately, even with these changes the debug build is still exceeding the maximum binary size. I might look into removing the loop unrolls from getf2_npvt as well.

tfalders:
Closing until new update can be tested

PR #742, opened 2024-06-17
Hybrid cpu host + gpu device execution of bdsqr() 

EdDAzevedo:
Hybrid cpu host + gpu device execution of bdsqr() in an attempt to speed up SVD calculations.

bdsqr_host()  is heavily influenced by lapack routine dbdsqr()  where the "D" and "E" arrays are reduced on the CPU. The rotations are then copied to GPU to update the "V", "U", "C" arrays. 

A special case is if no rotations are needed for "V", "U", "C" arrays, then the lapack version of bdsqr() is called directly.


tfalders:
Hi @EdDAzevedo, I've opened 2 PRs against your local fork to update this PR. One is just a straightforward merge of develop into your branch, the second contains the infrastructure code I've added to integrate hybrid algorithms into the library.

library/include/rocsolver/rocsolver-extra-types.h
```diff
@@ -164,4 +164,21 @@ typedef enum rocsolver_rfinfo_mode_
     = 272, /**< To work with Cholesky factorization (for symmetric positive definite sparse matrices). */
 } rocsolver_rfinfo_mode;
 
+/*! \brief Used by specific functions to specify the algorithm mode.
+ ********************************************************************************/
+typedef enum rocsolver_alg_mode_
+{
+    rocsolver_alg_mode_gpu
+    = 281, /**< Computations are all performed on the GPU. This is the default mode. */
```
EdDAzevedo: Minor comment: perhaps double check whether these are unique values (say after LASR api)

library/src/common/rocsolver_handle.cpp
```diff
@@ -0,0 +1,132 @@
...
+            return rocblas_status_internal_error;
+    }
+
+    switch(func)
+    {
+    case rocsolver_function_gesvd:
+    case rocsolver_function_bdsqr:
+        if(mode == rocsolver_alg_mode_gpu || mode == rocsolver_alg_mode_hybrid)
+        {
+            handle_data->bdsqr_mode = mode;
```
EdDAzevedo, line 66: Minor comment: perhaps verify there are reasonable default values in handle_data. Just a thought.

tfalders: Done

library/include/rocsolver/rocsolver-extra-types.h
tfalders: Done

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -3831,6 +3870,10 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zunmtr(rocblas_handle handle,
     In order to carry out calculations, this method may synchronize the stream contained within the
     rocblas_handle.
 
+    \note
+    A hybrid (CPU+GPU) approach is available for BDSQR. Use \ref rocsolver_set_alg_mode "SET_ALG_MODE"
+    to enable it.
```
jzuniga-amd, line 3875: It would be really useful if we add a sample code in `clients/samples` that illustrates the algorithm selection functionality.  We can then add a comment  like 

"(see an example in the `example_algorithm_selection.cpp` file in `clients/sample` directory)"

here and in the cases below

jzuniga-amd (commented):
@EdDAzevedo @tfalders, I opened a PR in your branch with more corrections/suggestions. Please have a look. 
https://github.com/EdDAzevedo/rocSOLVER/pull/15

library/src/include/lapack_host_functions.hpp
```diff
@@ -0,0 +1,596 @@
...
+ *    and BLAS functionality.
+ * ===========================================================================
+ */
+
+static void call_lamch(char& cmach, double& eps)
+{
+    eps = ((cmach == 'E') || (cmach == 'e')) ? std::numeric_limits<double>::epsilon()
+        : ((cmach == 'S') || (cmach == 's')) ? std::numeric_limits<double>::min()
+        : ((cmach == 'B') || (cmach == 'b')) ? FLT_RADIX
+                                             : std::numeric_limits<double>::min();
```
jmachado-amd: Why not use a `switch`?

library/src/include/lapack_host_functions.hpp
```diff
@@ -0,0 +1,596 @@
...
+    // if g == 0, then cs == 1, sn == 0
+    // if f == 0, then cs = 0, sn is chosen so that r is real
+    // ------------------------------------------------------
+
+    auto dble = [](auto z) { return (static_cast<double>(real_part(z))); };
+    auto dimag = [](auto z) { return (static_cast<double>(imag_part(z))); };
+    auto disnan = [](auto x) -> bool { return (isnan(x)); };
+    auto dcmplx = [](auto x, auto y) -> T {
+        bool constexpr is_complex_type
+            = !(std::is_same<T, float>::value || std::is_same<T, double>::value);
```
jmachado-amd: I see that it is easier to check if `T` is not a floating point here, but this design choice is problematic. Let's leave it as it is for now, but address the issue at the library level by creating a concept like struct to simplify checking for complex numbers in the future.

library/src/include/lapack_host_functions.hpp
```diff
@@ -0,0 +1,596 @@
...
+    {
+        auto const ip = i * incx;
+        if(is_zero)
+        {
+            x[ip] = 0;
+        }
+        else
+        {
+            x[ip] *= a;
+        }
```
jmachado-amd: I don't like this design choice: If `x_in` contains `Inf`s or `NaN`s they will be destroyed. It is very annoying to see a numerical method fail for no clear reason, only to realize that a part of the code was hiding a bug by overwriting `NaN`s that should have been propagated.

library/src/include/lib_host_helpers.hpp
```diff
@@ -92,6 +92,74 @@ I get_index(I* intervals, I max, I dim)
...
 
+template <typename T, typename I>
+static void call_swap(I& n, T& x_in, I& incx, T& y_in, I& incy)
+{
+    T* const x = &(x_in);
+    T* const y = &(y_in);
+    for(I i = 0; i < n; i++)
+    {
+        I const ix = i * incx;
+        I const iy = i * incy;
```
jmachado-amd: If the analogous gpu code casts `incx` and `incy` to `int64_t`, then the host code should do it as well. Besides, on the cpu, this will have zero impact on efficiency.

jzuniga-amd (approved):
LGTM. I have no more comments/suggestions for now. 
If someone can add `example_algorithm_selection.cpp` to `clients/samples/`, that would be really useful for users. (Otherwise, it can be added later).

library/src/auxiliary/rocauxiliary_bdsqr_hybrid.hpp
```diff
@@ -0,0 +1,1668 @@
...
+        rocsolver_lasr_template<T, S>(handle, side, pivot, direct, m, n, dc, 0, ds, 0, &A, 0, lda,
+                                      0, I(1));
+        CHECK_HIP(hipStreamSynchronize(stream));
+    };
+
+    auto indx2f = [](auto i, auto j, auto ld) -> int64_t {
+        assert((1 <= i) && (i <= ld));
+        assert((1 <= j));
+        return ((i - 1) + (j - 1) * int64_t(ld));
+    };
```
jmachado-amd: This type of code can generate unnecessary compiler warnings due to type mismatch. In the future, it would be better to make sure that all variables have compatible types and cast the literal `1` as needed; or just adapt rocSOLVER's `idx2D` (found in [`lib_host_helpers.hpp`](https://github.com/ROCm/rocSOLVER/blob/b29339365edc98a18493e684bdd3f5c1df65d870/library/src/include/lib_host_helpers.hpp#L46)).

library/src/include/lapack_host_functions.hpp
```diff
@@ -0,0 +1,596 @@
...
+            auto const w = std::max(xabs, yabs);
+            auto const z = std::min(xabs, yabs);
+            if(z == zero)
+            {
+                ddlapy2 = w;
+            }
+            else
+            {
+                ddlapy2 = w * std::sqrt(one + square(z / w));
+            }
```
jmachado-amd, line 203: This is a general comment about patterns that are heavily used in the library, and we will want to address at
some point in time. There is much numerical code that was designed to target `X87` instructions and its high accuracy registers (e.g., lapack) and I think that this piece is a good example.

If this were compiled to use high accuracy registers, then it would provide good accuracy in general and exceptional accuracy for `float`s. With `SSE` instructions however, unnecessary underflow can still occur, and it is better to use a POSIX compliant version of [`hypot`](https://pubs.opengroup.org/onlinepubs/9799919799/). Generally speaking, a literal translation of lapack into c++ does not yield the same accuracy.

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -3831,6 +3870,10 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zunmtr(rocblas_handle handle,
     In order to carry out calculations, this method may synchronize the stream contained within the
     rocblas_handle.
 
+    \note
+    A hybrid (CPU+GPU) approach is available for BDSQR. Use \ref rocsolver_set_alg_mode
+    to enable it.
```
jmachado-amd: I would add to this note that the hybrid algorithm is primarily targeted to be used in homogeneous architectures.

library/src/include/lapack_device_functions.hpp
```diff
@@ -1111,4 +1111,81 @@ ROCSOLVER_KERNEL void axpy_kernel(const rocblas_int n,
...
+    I const i_start = hipThreadIdx_x + hipBlockIdx_x * hipBlockDim_x;
+    I const i_inc = hipBlockDim_x * hipGridDim_x;
+
+    S const zero = 0;
+    bool const is_da_zero = (da == zero);
+    if(incx == 1)
+    {
+        for(I i = i_start; i < n; i += i_inc)
+        {
+            x[i] = (is_da_zero) ? zero : da * x[i];
```
jmachado-amd: Similarly to the equivalent host function, if `x` contains `Inf`s or `NaN`s they will be destroyed (see comment on `call_scal` for context).

library/src/include/lapack_device_functions.hpp
```diff
@@ -1111,4 +1111,81 @@ ROCSOLVER_KERNEL void axpy_kernel(const rocblas_int n,
...
+    else
+    {
+        // ---------------------------
+        // code for non-unit increments
+        // ---------------------------
+
+        for(I i = i_start; i < n; i += i_inc)
+        {
+            auto const ix = 0 + i * static_cast<int64_t>(incx);
+            x[ix] = (is_da_zero) ? zero : da * x[ix];
```
jmachado-amd: See previous comment.

library/src/include/rocsolver_handle.hpp
```diff
@@ -0,0 +1,44 @@
...
+ROCSOLVER_BEGIN_NAMESPACE
+
+struct rocsolver_handle_data_
+{
+    rocblas_int checksum;
+
+    rocsolver_alg_mode bdsqr_mode = rocsolver_alg_mode_gpu;
+};
+
+typedef struct rocsolver_handle_data_* rocsolver_handle_data;
```
jmachado-amd, line 42: Every `typedef` is a missed `using` opportunity.

library/src/auxiliary/rocauxiliary_bdsqr_hybrid.hpp
```diff
@@ -0,0 +1,1668 @@
...
+    S const eight = 8;
+    S const meight = -one / eight;
+    I const maxitr = 6;
+    I ione = 1;
+
+    bool const lower = (uplo == 'L') || (uplo == 'l');
+    bool const upper = (uplo == 'U') || (uplo == 'u');
+
+    //rotate is true if any singular vectors desired, false otherwise
+    bool const rotate = (ncvt > 0) || (nru > 0) || (ncc > 0);
```
jmachado-amd, line 227: For the future: `dqds` is better when eigenvectors are not needed.

jmachado-amd (approved):
The code contains some design choices that will have to be revisited in the future; but, apart from that, it reads well and delivers important features.

I am approving the PR as is, but please allow a few more days for everyone to have a look at it before merging. If there is sufficient time to update it, I would point out the following subset of the comments (as potential improvements): (a) the type mismatch issues in lambdas such as `idnx2f`; (b) the missing casts to `int64_t` on `call_swap`; and (c) reviewing the suppression of `NaN`s on all versions of `scal`. The other comments are intended for our future reference.

We must also revisit our API in the future to make sure that hybrid methods can communicate to end users how much host memory they need to allocate.

library/src/include/lapack_host_functions.hpp
EdDAzevedo: One concern is the possibility by the user  of using alpha == 0 in the call to scaling routine  to set an array to zero, even if the array was containing uninitialized or unknown values or Infs or Nans.

library/src/include/lapack_host_functions.hpp
EdDAzevedo: I think in general whether a switch/case (implemented as a jump table or nested if-then-else) or the above form (conditional moves) is faster depends on the different cases, the compiler and the hardware. I think visually, it kind of looks like a switch/case  style.   

On the GPU, my guess is in general, a conditional move might be better in other to avoid thread divergence  caused by the branch and there might be a hardware conditional move instruction. 

In the specific case above in testing the same function argument variable, there should not be thread divergence.  

Just a thought.

jmachado-amd: For context, this code runs on the cpu and it is called just a handful of times; efficiency is not a concern.

library/src/include/rocsolver_handle.hpp
tfalders: I was using the same pattern as rocblas_handle and rocsolver_rfinfo, but I suppose it's not necessary for types that are not exposed to the user

PR #747, opened 2024-06-19
Add 64-bit versions of lacgv, larf, and larfg

qjojo:
These APIs are required for the 64-bit version of geqrf and geqr2


clients/common/misc/rocsolver.hpp
```diff
@@ -904,16 +904,28 @@ inline rocblas_status rocblas_gemm(bool STRIDED,
 /*****************************************************/
 
 /******************** LACGV ********************/
-inline rocblas_status
-    rocsolver_lacgv(rocblas_handle handle, rocblas_int n, rocblas_float_complex* x, rocblas_int incx)
+inline rocblas_status rocsolver_lacgv(bool API64,
+                                      rocblas_handle handle,
+                                      rocblas_int n,
+                                      rocblas_float_complex* x,
+                                      rocblas_int incx)
```
EdDAzevedo: I wonder whether there are other ways to implement this. It seems to me it is not passing in int64_t and so is not truly implementing 64bit interface.   Just a thought.

clients/common/auxiliary/testing_larf.hpp
```diff
@@ -137,7 +137,7 @@ void larf_initData(const rocblas_handle handle,
     }
 }
 
-template <typename T, typename Td, typename Th>
+template <bool API64, typename T, typename Td, typename Th>
```
EdDAzevedo: Is there a test for large n, say lacgv with n = 2^31 + 1, just to check the 64bit interface works for large n as expected?

library/src/auxiliary/rocauxiliary_lacgv.hpp
```diff
@@ -37,38 +37,37 @@
...
-    int j = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;
-    int b = hipBlockIdx_z;
+    I i = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+    I j = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;
+    I b = hipBlockIdx_z;
 
     T* Ap = load_ptr_batch<T>(A, b, shifta, stridea);
 
     if(i < m && j < n)
         Ap[i + j * lda] = conj(Ap[i + j * lda]);
```
EdDAzevedo: Perhaps use something like  " i + j * int64_t(lda)" to avoid potential 32bit overflow?

library/src/auxiliary/rocauxiliary_lacgv.hpp
```diff
@@ -109,12 +108,12 @@ rocblas_status rocsolver_lacgv_template(rocblas_handle handle,
     rocblas_get_stream(handle, &stream);
 
     // handle negative increments
-    rocblas_int offset = incx < 0 ? shiftx - (n - 1) * incx : shiftx;
+    rocblas_stride offset = incx < 0 ? shiftx - (n - 1) * incx : shiftx;
 
     // conjugate x
-    rocblas_int blocks = (n - 1) / 64 + 1;
+    I blocks = (n - 1) / 64 + 1;
     ROCSOLVER_LAUNCH_KERNEL(conj_in_place<T>, dim3(1, blocks, batch_count), dim3(1, 64, 1), 0,
```
EdDAzevedo, line 116: Perhaps use something like "BS1" or  something like "constexpr int LARCGV_NTHREADS = 64"? Is there a concern about max number of threads, or max number of blocks to be less than 64 * 1024?

 If  in 64bit interface n = 2^32 + 1, would it  still work?    Just checking.

qjojo:
Added explicit tests for sizes of UINT32_MAX + 1, addressed feedback, rearranged the templates to use the integer type itself instead of flags.

clients/common/auxiliary/testing_lacgv.hpp
```diff
@@ -68,12 +68,8 @@ void testing_lacgv_bad_arg()
     lacgv_checkBadArgs(handle, n, dA.data(), inc);
 }
 
-template <bool CPU, bool GPU, typename T, typename Td, typename Th>
-void lacgv_initData(const rocblas_handle handle,
-                    const rocblas_int n,
-                    Td& dA,
-                    const rocblas_int inc,
-                    Th& hA)
+template <bool CPU, bool GPU, typename T, typename Td, typename Th, typename I>
```
tfalders: Minor issue, but I've been trying to consistently have `I` come after `T`. The benefit of this is that we can specify `T` and `I`, and let any remaining template arguments be automatically interpreted.

The other test files seem to be good. It's only lacgv that is inconsistent with this.

docs/reference/auxiliary.rst
```diff
@@ -57,6 +57,10 @@ rocsolver_<type>lacgv()
 .. doxygenfunction:: rocsolver_zlacgv
    :outline:
 .. doxygenfunction:: rocsolver_clacgv
+   :outline:
+.. doxygenfunction:: rocsolver_zlacgv_64
+   :outline:
+.. doxygenfunction:: rocsolver_clacgv_64
```
tfalders: For consistency, please put the 64-bit functions above the 32-bit functions.

clients/common/misc/client_util.hpp
```diff
@@ -53,6 +53,9 @@
    wants to add. By calling the next function twice, it can double the number of times the stamp
    is instantiated. By appending different values in each call to the next function, it can
    instantiate the stamp with multiple different values. */
+#define FOREACH_BIT_VARIANT(STAMP, F, ...) \
+    F(STAMP, ##__VA_ARGS__, false)         \
+    F(STAMP, ##__VA_ARGS__, true)
```
tfalders: This does not appear to be used? Perhaps it can be removed.

clients/common/misc/clients_utility.hpp
```diff
@@ -53,6 +53,10 @@ static constexpr char LIMITED_MEMORY_STRING_GTEST[]
 // Return the path to the client executable, for finding test matrices on filesystem
 std::string rocsolver_exepath();
 
+// Evaluates to true if an integer type is 64 bits wide
+template <typename I>
+static constexpr bool rocsolver_is_64bit = sizeof(I) == 8;
```
tfalders: Same here. It does not appear to be used.

library/src/auxiliary/rocauxiliary_larfg.hpp
tfalders, line 1: The set_taubeta kernel should also have the changes applied to it. shifta should be type rocblas_stride and variable b should be type I.

clients/gtest/auxiliary/lacgv_gtest.cpp
```diff
@@ -54,21 +55,25 @@ const vector<vector<int>> range = {
     {30, 3},
     {30, -3}};
 
+const vector<vector<int64_t>> range_64 = {{UINT32_MAX + 1, 1}};
```
tfalders: This seems like a rather large size to test. I'm not sure we want to test that large. Perhaps just using the same numbers as for the 32-bit API is okay?

clients/gtest/auxiliary/lacgv_gtest.cpp
```diff
@@ -100,6 +112,20 @@ TEST_P(LACGV, __double_complex)
...
+    run_tests<rocblas_double_complex>();
+}
+
 INSTANTIATE_TEST_SUITE_P(daily_lapack, LACGV, ValuesIn(large_range));
 
 INSTANTIATE_TEST_SUITE_P(checkin_lapack, LACGV, ValuesIn(range));
+
+INSTANTIATE_TEST_SUITE_P(daily_lapack, LACGV_64, ValuesIn(range_64));
+
+INSTANTIATE_TEST_SUITE_P(checkin_lapack, LACGV_64, ValuesIn(range_64));
```
tfalders: Since you're passing in the same values, daily_lapack and checkin_lapack will run the same test. These should be differentiated (or one of them removed).

clients/gtest/auxiliary/larf_gtest.cpp
```diff
@@ -69,30 +70,35 @@ const vector<vector<int>> matrix_size_range = {
     {20, 15, 20},
     {35, 35, 50}};
 
+const vector<vector<int64_t>> matrix_size_range_64 = {{UINT32_MAX + 1, 1, 1}};
+const vector<vector<int64_t>> incx_range_64 = {{1, 1}};
```
tfalders: Same comment as above

clients/gtest/auxiliary/larfg_gtest.cpp
```diff
@@ -70,22 +71,38 @@ const vector<int> large_n_size_range = {
...
+    0,
+    // normal (valid) samples
+    1,
+    5,
+    8,
+    10,
+};
+
+// for checkin_lapack tests
+const vector<int64_t> n_size_range_64 = {UINT32_MAX + 1};
```
tfalders: Same comment as above

tfalders (commented):
Looks good! I have some minor comments to be addressed, and then I'll approve.

Note to self: I need to remember to update the gtest files for getrf since you've added the `I` parameter to these gtest files whereas I did not add them to getrf.

clients/gtest/auxiliary/lacgv_gtest.cpp
TorreZuk: UINT32_MAX + 1 is likely 0.  Even if you reduce the size one long standing thing I don't like about your solver tests is that they don't report the specifc test problem sizes like rocblas does.   Any bad problem size can't be noticed by inspection.  Maybe consider a ticket for that.  Back to large sizes:  the sign bit of a 32bit int we sometimes use in rocblas as the minimally large size that will be a bad value if calling the 32bit APIs.  We have a stress category which includes the 64bit sizes as they can be minutes for a single test.

EdDAzevedo: For these cases of generating or applying the Householder block reflectors  H = (eye -  Y * T * Y') or unblocked H = (eye - y * tau * y'), even if the "y" vector is very long, I suspect it should not be too expensive in taking O(N) flops and O(N) data access so it may still be acceptable to test with vector length  N larger than  (2^31 -1).    
Just a thought.

clients/common/auxiliary/testing_larf.hpp
qjojo: That test takes an extremely long time to execute, I've run it locally and it works but in terms of adding it to test coverage I don't think it's appropriate for daily or checkin tests. We should think about adding a long run stress test category like rocblas has in the future.

tfalders (approved):

cgmb (approved):

qjojo:
The CI failures on this exist on all open PRs (sygvdx errors) and are not related to the changes in this PR so I will squash and merge.

PR #748, opened 2024-06-19
Use correct fortran package name for mariner OS

tfalders:
This will need to be cherry-picked into 6.2 once it is merged.


qjojo (approved):

CMakeLists.txt
```diff
@@ -226,6 +226,8 @@ if(BUILD_CLIENTS_TESTS OR BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_SAMPLES)
     endif()
   elseif(CLIENTS_OS STREQUAL "ubuntu" AND CLIENTS_OS_VERSION VERSION_GREATER_EQUAL "20.04")
     set(GFORTRAN_DEB "libgfortran5")
+  elseif(CLIENTS_OS STREQUAL "mariner")
+    set(GFORTRAN_RPM "gfortran")
```
jmachado-amd, line 230: Why is it gfortran and not flang? Be careful that the gcc suite tends to default to `-fPIC` and `-fPIE` while clang doesn't.

tfalders: I've reproduced what other libraries have done, see e.g. https://github.com/ROCm/rocBLAS-internal/pull/2557 and https://github.com/ROCm/rocSPARSE/pull/397

jmachado-amd (approved):

CMakeLists.txt
jmachado-amd: Ok, but please keep an eye on CI for potential linking issues.

jzuniga-amd (approved):

PR #752, opened 2024-06-25
Update changelog to mention the `stebz` synchronization defect

jmachado-amd:


tfalders (approved):
Please take out "that manifests on Windows" since it likely affects Linux as well. Please also add SYGVX/HEGVX, BDSVDX, and GESVDX to the list of affected functions.

EdDAzevedo (approved):

EdDAzevedo (approved):

PR #755, opened 2024-07-03
Fix memory allocation issues in Jacobi solvers

jmachado-amd:
The Jacobi solvers where trying to allocate more memory than they needed, and this led to allocation failures with clients that manage their own memory (hipSOLVER in particular).

### DIFF
--- a/library/src/lapack/roclapack_syevj_heevj.cpp
+++ b/library/src/lapack/roclapack_syevj_heevj.cpp
@@ -80,7 +80,7 @@ rocblas_status rocsolver_syevj_heevj_impl(rocblas_handle handle,
     // memory workspace allocation
     void *Acpy, *J, *norms, *top, *bottom, *completed;
     rocblas_device_malloc mem(handle, size_Acpy, size_J, size_norms, size_top, size_bottom,
-                              size_completed, size_norms);
+                              size_completed);
 
     if(!mem)
         return rocblas_status_memory_error;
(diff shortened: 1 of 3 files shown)


library/src/lapack/roclapack_syevj_heevj.cpp
```diff
@@ -80,7 +80,7 @@ rocblas_status rocsolver_syevj_heevj_impl(rocblas_handle handle,
     // memory workspace allocation
     void *Acpy, *J, *norms, *top, *bottom, *completed;
     rocblas_device_malloc mem(handle, size_Acpy, size_J, size_norms, size_top, size_bottom,
-                              size_completed, size_norms);
+                              size_completed);
```
cgmb, line 83: I forget why size_norms was unnecessary here. Could you (briefly) remind me? It would be good to have the explanation in the PR for posterity anyway.

tfalders (approved):

library/src/lapack/roclapack_syevj_heevj.cpp
tfalders: size_norms is already included earlier in the size allocations. This one is redundant and erroneous. Not sure how I didn't notice it before.

EdDAzevedo: Is "size_norms = sizeof(S) * batch_count"? 
If so, one may think "size_norms" is small compared to say "size_Acpy = sizeof(T) * n * n * batch_count;"
----------------------
[Aside] 
 If  the GPU has say 100 Compute Units (CU), then a launch configuration for processing say 2*100 batch items at a time (inside a for loop) should be sufficient to keep the GPU busy. Then the amount of temporary storage may be significantly reduced to something like the following

static int get_num_cu(int deviceId = 0) {
  int ival = 0;
  auto const attr = hipDeviceAttributeMultiprocessorCount;
  HIP_CHECK(hipDeviceGetAttribute(&ival, attr, deviceId));
  return (ival);
}


auto const max_blocks = 2 * get_num_cu();
auto const nblocks = std::min( max_blocks, batch_count );

size_t const size_Acpy = sizeof(T) * n * n * nblocks;
...
ROCSOLVER_LAUNCH_KERNL( dim3( 1,1,  nblocks) ...., batch_count )

EdDAzevedo (approved):

PR #759, opened 2024-07-08
Added small-size larfg and larf kernels

tfalders:


library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -38,88 +38,86 @@
...
-
-    T* a = load_ptr_batch<T>(alpha, b, shifta, stride);
-    T* t = tau + b * strideP;
-
-    if(norms[b] > 0)
+    if(norms[0] > 0)
     {
-        T n = sqrt(norms[b] + a[0] * a[0]);
-        n = a[0] >= 0 ? -n : n;
+        T n = sqrt(norms[0] + alpha[0] * alpha[0]);
```
EdDAzevedo, line 46: Minor suggestion: consider using  "m,n,i,j,k" to name  integer variables (fortran convention).

library/src/specialized/rocauxiliary_larfg_specialized_kernels.hpp
```diff
@@ -48,13 +50,40 @@ ROCSOLVER_KERNEL void __launch_bounds__(LARFG_SSKER_THREADS)
...
+    I tid = hipThreadIdx_x;
+
+    // select batch instance
+    T* a = load_ptr_batch<T>(alpha, bid, shiftA, strideA);
+    T* x = load_ptr_batch<T>(xx, bid, shiftX, strideX);
+    T* tau = load_ptr_batch<T>(tauA, bid, 0, strideP);
+
+    // shared variables
+    __shared__ T sval[LARFG_SSKER_THREADS];
+    __shared__ T sh_x[LARFG_SSKER_MAX_N];
```
EdDAzevedo, line 70: I wonder whether there can be an extra "for" loop to handle the larger cases (larger than LARFG_SSKER_MAX_N).    Just a suggestion.

library/src/include/ideal_sizes.hpp
```diff
@@ -37,15 +37,15 @@
...
+#define LARF_SSKER_THREADS 256 // must be 64, 128, 256, 512, or 1024
 #endif
 
 #ifndef LARF_SSKER_BLOCKS
 #define LARF_SSKER_BLOCKS 64
 #endif
 
 #ifndef LARF_SSKER_MAX_DIM
-#define LARF_SSKER_MAX_DIM 512
+#define LARF_SSKER_MAX_DIM 2048
```
EdDAzevedo: I wonder whether this can be dependent on the type T or sizeof(T) so that it may be larger for float type sizeof(T) == 4 or sizeof(T) == 8, say something like
#define LARF_SSKER_MAX_DIM(T)  \
      ( (sizeof(T) == 16) ? 2048 : \
         (sizeof(T) == 4) ? 8192 : \
         (sizeof(T) == 8) ? 4096 :2048)

EdDAzevedo (approved):

qjojo (approved):

library/src/auxiliary/rocauxiliary_larfg.hpp
cgmb: Is this `sqrt` going to be double-precision? I'm not sure there's anything that guarantees that `float sqrt(float)` is available in the global namespace.

EdDAzevedo: Perhaps add something like the following,

auto sqrt  = [](auto x) { return( std::sqrt( x ) ); };

to get the code to work.
Just a suggestion.

tfalders: Can std::sqrt be accessed by a device function? I figured the sqrt was coming from HIP.

library/src/specialized/rocauxiliary_larfg_specialized_kernels.hpp
tfalders: For cases larger than LARFG_SSKER_MAX_N I think I would read from x directly, and remove the shared sh_x array. I can take a look at it.

library/src/auxiliary/rocauxiliary_larfg.hpp
cgmb: [The `sqrt` does call a float overload](https://cuda.godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIAruiakl9ZATwDKjdAGFUtEywZ7HAGTwNMAHLuAEaYxCAAzKQADqgKhHYMLm4eerHxtgK%2B/kEsoeFRVpg2iUIETMQEye6eXJaY1pkMZRUE2YEhYZGW5ZXVqXUKvW1%2BHXldEQCUlqgmxMjsHEPEJjYA1AyoAPpoDABuYfECWwQApADsAEIXACKnEdcaAIIEmCzRBq/3TsgGCgprAAq92way2WywezwC3BG22uwOxCODBOazwWEEeCoAE8TtjopgIMCAEwANlOZLWe0mawulzWxEwBDmDCp93pt1Ozy5T3BwFoqGCYlhe1Q6IZJhRSgIJyYJlC5WoAqYBAAVGtyiZSGsqMq1XDiCwFNrdagVeqxNEEExJjy6Ty1o60VQIJtDQpTgBWS4aL03WkRUEaW3PJ20q4OsNOuVEDYB/0KACOlVdJCNXp9ftpxPplutGd9nv9FtoVqYBb9toekajaIxthxeIJrqrjyetcdrPu/rz5e9hf9IO7aw0AYAYmsALSskAbdl20MdgD0S7WClEtD8wB1TBsJBANajbvT/azw64ADpR6uIL2K0WpxtWwv28vV5qD4va5r74OIv7XUfO9TyLGlVwYeduS/KMVzWBUmE/V9a2AzMH2HCDq2gzloIaJRD3taCw2PD0QL/f0uEgpCox/Uj4xHSiw2wp5bg4aZaE4T1eE8bheFQTgnDkG52ycClrhzNdZnmTBswiHhSAITRWOmABrEBPQ0fROEkLjFNIPiOF4BQQA0hSOC0aY4FgGBEBQVB3joMJyEoNB7PocIEDwaJfkMYBJ2IVBkBYSdSQ0LgNGJPg6FeJFKGCXTgj8CpsU4OSEuYYhsQAeWCbRilMuSXLYQRMoYWhkrM3gsGCExgCcMRaCMnjSCwFgfPECrmrwRkSgORqtFITBVGKExXhS3g/FediOs3YJiCSlwsF0ghiDwFgxtIRFgjiTAbjeNqt0U6ZdSYYAFAANTwTAAHdMoJbi5P4QQRDEdgpBkQRFBUdQOt0OoDCMFAzAsGajMgaZUGiJpGt4xEVqwUGIGmIoSnsCBHH6WoBoYdB2lyfI0jiBIBAxgmMkSXHOnCQYGjy0phhJ6nGjp1oKfGKmelaBmOcqVn8a4JHJIWCQ2I4nSOv0tZvKMBkAqC0kL0vYk1ggXBCBIGT%2Bd4UzzJUtSNKm7TSDW9TSG4/r9MM4z5MO0hLJs2YCGiEanIgFzogc4gAlYRYpe3fzAuChWL2JXhMHwIg4b0R7hA3V7pGjz61F037SCuubojGkWOE403dP0zKRqdgg1lQKhJf%2Bv3ZcDxXlZcVywg1yYtcO6YEEwJgsHCRHNI4Q3jY0s3eM4S2TJb0hVJNqaIjF83h%2BtirJizkPc/FuftaUjbDkSEBJCAA%3D%3D%3D), though I'm unclear on what guarantees that.

In [HIP Language Extensions](https://rocm.docs.amd.com/projects/HIP/en/docs-6.1.2/reference/kernel_language.html), the documented functions are `double sqrt(double)` and `float sqrtf(float)`. Maybe we're getting `std::sqrt` because it is declared in the global namespace due to C++11 §17.6.1:

> It is unspecified whether these names are first declared within the global namespace scope and are then injected into namespace std by explicit using-declarations (7.3.3).

In any case, I suppose this is fine for now.

cgmb (approved):

tfalders:
CI failures appear to be the result of ongoing issues in STEBZ. I will force the merge.

cgmb:
> CI failures appear to be the result of ongoing issues in STEBZ. I will force the merge.

@tfalders, I'm not so sure. The tests were consistently passing prior to this change and are consistently failing now that it has been merged. As shown by the CI logs, failures occur on gfx1101, but not on gfx908 or gfx90a, and presumably not on your gfx900 workstation. I suspect there is something wrong in this code that affects RDNA GPUs, while leaving Vega and CDNA GPUs unaffected. Maybe wave32 vs wave64?

Passing:
rocSOLVER version 3.27.0.bf215fb (with rocBLAS 4.3.0.66ab453c)

Failing:
rocSOLVER version 3.27.0.1300e80 (with rocBLAS 4.3.0.66ab453c)

I think we should back this change out until the problem can be determined.

PR #760, opened 2024-07-08
Use std::min and std::max

cgmb:
The clang compiler provides integer min and max functions when building for the HIP language. This may result in 64-bit integers being implicitly converted to 32-bit. To avoid this, we will always use std::min and std::max as those functions are strict about matching arguments.

### DIFF
--- a/clients/common/auxiliary/testing_orgbr_ungbr.hpp
+++ b/clients/common/auxiliary/testing_orgbr_ungbr.hpp
@@ -106,7 +106,7 @@ void orgbr_ungbr_initData(const rocblas_handle handle,
     if(CPU)
     {
         using S = decltype(std::real(T{}));
-        size_t s = max(hIpiv.n(), 2);
+        size_t s = std::max(hIpiv.n(), int64_t(2));
         std::vector<S> E(s - 1);
         std::vector<S> D(s);
         std::vector<T> P(s);
@@ -168,7 +168,7 @@ void orgbr_ungbr_getError(const rocblas_handle handle,
                           Th& hIpiv,
                           double* max_err)
 {
-    size_t size_W = max(max(m, n), k);
+    size_t size_W = std::max(std::max(m, n), k);
     std::vector<T> hW(size_W);
 
     // initialize data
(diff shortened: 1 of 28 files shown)


clients/common/auxiliary/testing_orgbr_ungbr.hpp
```diff
@@ -106,7 +106,7 @@ void orgbr_ungbr_initData(const rocblas_handle handle,
     if(CPU)
     {
         using S = decltype(std::real(T{}));
-        size_t s = max(hIpiv.n(), 2);
+        size_t s = std::max(hIpiv.n(), int64_t(2));
```
EdDAzevedo, line 109: Is the promotion "int64_t(2)" necessary or the compiler can automatically perform the promotion? If explicit promotion is necessary, should it be "size_t(2)" since it will be assigned to "size_t s" ?

cgmb: The cast is required as `2` is `int` and `hIpiv::n()` returns `int64_t` (a.k.a. `long int` or `long long int`). `std::max` needs both input values to have the same type, so the alternative would be to cast both `hIpiv.n()` and `2` to `size_t`.

EdDAzevedo (approved):

tfalders (approved):
I'm not a fan of all these casts, but you can't argue with necessity. LGTM!

cgmb:
Just thinking about the use of `std::max` with floating point values... what is the expected behaviour of rocSOLVER with NaN values in the input matrices?

EdDAzevedo:
> Just thinking about the use of `std::max` with floating point values... what is the expected behaviour of rocSOLVER with NaN values in the input matrices?

I think the behaviour is not well defined. 

Let   double x = nan("");  and double y = -1.0 then

(x > y) is false
(x < y) is false
std::max(x,y) is nan
std::max(y,x) is y

cgmb:
It is undefined behaviour to pass values to `std::min` or `std::max` that do not obey a strict weak ordering. That much I'm clear on. The question is what the behaviour of rocSOLVER should be when there are NaN values in its inputs.

I think, perhaps, we can use `std::fmin` and `std::fmax` for user-provided floating point values in most cases. For `find_max_tridiag`, the result seems reasonable (returning the largest non-NaN value). It's less clear for LAGTF, but I think it might still be reasonable for that case. Those are the only two places where it's being used with floats.

cgmb:
@EdDAzevedo, as you mentioned offline, we'll need to take another look at our use of `abs`. We can do that separately, I think.

EdDAzevedo (approved):

tfalders (approved):

PR #762, opened 2024-07-09
Fix jacobi cycle-sync

jmachado-amd:
This PR adds a work-around to fix the convergence (and related performance) defects in the Jacobi eigensolver that 
 steamed from improper synchronization. The root cause of the synchronization defect will be addressed elsewhere.

### DIFF
--- a/library/src/lapack/roclapack_syevj_heevj.hpp
+++ b/library/src/lapack/roclapack_syevj_heevj.hpp
@@ -1161,40 +1161,30 @@ ROCSOLVER_KERNEL void
 {
     rocblas_int n = half_blocks - 1;
 
-    auto f = [n = n](auto i) -> auto
+    auto cycle = [n = n](auto i) -> auto
     {
         using I = decltype(i);
         i = (i - 1) % (2 * n + 1) + 1;
         I j{};
 
-        if((1 <= i) && (i <= n))
+        if (i % 2 == 0)
         {
-            j = 2 * i;
+            j = i + 2;
+            if (j > 2*n)
+            {
+                j = 2*n + 1;
+            }
         }
         else
         {
-            j = 2 * (2 * n + 1 - i) + 1;
+            j = i - 2;
+            if (j < 1)
+            {
+                j = 2;
+            }
         }
 
         return j;
-    };
-
-    auto g = [n = n](auto j) -> auto
-    {
-        using I = decltype(j);
-        j = (j - 1) % (2 * n + 1) + 1;
-        I i{};
-
-        if(j % 2 == 0)
-        {
-            i = j / 2;
-        }
-        else
-        {
-            i = 2 * n + 1 - (j - 1) / 2;
-        }
-
-        return i;
     };
 
     rocblas_int tidx = hipThreadIdx_x;
@@ -1203,18 +1193,14 @@ ROCSOLVER_KERNEL void
 
     if(tidx == 0)
     {
-        k = g(top[1]) + 1;
-        bottom[0] = f(2 * n + k);
+        bottom[0] = cycle(bottom[0]);
     }
-    __syncthreads();
-
-    if(tidx > 0)
+    else
     {
         for(rocblas_int l = tidx; l < half_blocks; l += dimx)
         {
-            k = g(top[tidx]) - (tidx - 1) + 1;
-            top[tidx] = f(tidx - 1 + k);
-            bottom[tidx] = f(2 * n - tidx + k);
+            top[l] = cycle(top[l]);
+            bottom[l] = cycle(bottom[l]);
         }
     }
 }
--- a/library/src/lapack/roclapack_syevj_heevj.hpp
+++ b/library/src/lapack/roclapack_syevj_heevj.hpp
@@ -1103,6 +1103,60 @@ ROCSOLVER_KERNEL void syevj_offd_rotate(const bool skip_block,
     so only one thread group is needed.) **/
 template <typename T>
 ROCSOLVER_KERNEL void
+    syevj_cycle_pairs_old(const rocblas_int half_blocks, rocblas_int* top, rocblas_int* bottom)
+{
+    rocblas_int tix = hipThreadIdx_x;
+    rocblas_int i, j, k;
+
+    if(half_blocks <= hipBlockDim_x && tix < half_blocks)
+    {
+        if(tix == 0)
+            i = 0;
+        else if(tix == 1)
+            i = bottom[0];
+        else if(tix > 1)
+            i = top[tix - 1];
+
+        if(tix == half_blocks - 1)
+            j = top[half_blocks - 1];
+        else
+            j = bottom[tix + 1];
+        __syncthreads();
+
+        top[tix] = i;
+        bottom[tix] = j;
+    }
+    else
+    {
+        // shared memory
+        extern __shared__ double lmem[];
+        rocblas_int* sh_top = reinterpret_cast<rocblas_int*>(lmem);
+        rocblas_int* sh_bottom = reinterpret_cast<rocblas_int*>(sh_top + half_blocks);
+
+        for(k = tix; k < half_blocks; k += hipBlockDim_x)
+        {
+            sh_top[k] = top[k];
+            sh_bottom[k] = bottom[k];
+        }
+        __syncthreads();
+
+        for(k = tix; k < half_blocks; k += hipBlockDim_x)
+        {
+            if(k == 1)
+                top[k] = sh_bottom[0];
+            else if(k > 1)
+                top[k] = sh_top[k - 1];
+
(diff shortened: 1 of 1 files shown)


library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1105,55 +1106,123 @@ template <typename T>
...
-        else if(tix > 1)
-            i = top[tix - 1];
+    auto f = [n = n](auto i) -> auto {
+        using I = decltype(i);
+        i = (i - 1) % (2 * n + 1) + 1;
+        I j{};
 
-        if(tix == half_blocks - 1)
-            j = top[half_blocks - 1];
+        if ((1 <= i) && (i <= n))
```
EdDAzevedo: Minor suggestion:  you might also consider something like

auto const j = ((1 <= i)  && (i <= n)) ? 2*i : 2 * (2 * n + 1 - i) + 1;

to avoid  "using I = decltype(i)"

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1105,55 +1106,123 @@ template <typename T>
...
+    if (tidx > 0)
+    {
+        /* k = g(top[tidx]) - (tidx - 1) + 1; */
+        k = top[0];
+        top[tidx] = f(tidx - 1 + k);
+        bottom[tidx] = f(2 * n - tidx + k);
     }
+    /* printf("Output pair %d: [%d, %d]\n", tidx, top[tidx], bottom[tidx]); */
 }
+/* template <typename T> */
```
EdDAzevedo: Minor comment: this may be just a matter of taste but I use something like
"#if (0)  .... #endif"   to comment out large sections of code.   Just a suggestion.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -872,8 +872,9 @@ ROCSOLVER_KERNEL void syevj_offd_kernel(const rocblas_int blocks,
     if(completed[bid + 1])
         return;
 
-    rocblas_int i = top[hipBlockIdx_x];
+    rocblas_int i = hipBlockIdx_x == 0 ? 0 : top[hipBlockIdx_x];
```
EdDAzevedo: Minor suggestion: perhaps the code can pre-generate the tournament schedule of cycle pairs in an array or table to copy to device memory. 

Maybe a table  say   schedule(0:1, 0:(npairs-1), 0:(num_rounds-1)), where npairs = n/2, num_rounds = n - 1 so that    
player0  = schedule(0, ipair, iround) 
plays against   
player1 = schedule(1, ipair, iround)  
as the "ipair" in the "iround" in the tournament.

One can also  write  (cpu) code to check the schedule.  For example, each player never plays himself and each player plays against each other player only once and in separate rounds. This may add robustness to the correctness of the tournament schedule.   Otherwise, it may not be  immediately obvious just by examining the C++ code whether the tournament schedule generated is correct.    Just a thought.

EdDAzevedo (approved):

tfalders:
The math seems rather complicated. I was imagining something simpler. For example, check the value of top[tid] or bottom[tid]. If it's even, decrement by two, and if the result is <= 0 then change it to be 1. If it's odd, increment by two, and if the result is >= half_blocks then change it to be half_blocks-2.

EDIT: The zero should, of course, stay where it is. This approach shouldn't need any synchronization between threads.

jmachado-amd:
> The math seems rather complicated. I was imagining something simpler. For example, check the value of top[tid] or bottom[tid]. If it's even, decrement by two, and if the result is <= 0 then change it to be 1. If it's odd, increment by two, and if the result is >= half_blocks then change it to be half_blocks-2.
> 
> EDIT: The zero should, of course, stay where it is. This approach shouldn't need any synchronization between threads.

Your suggestion works, of course, but it accomplishes something that is different from what I was trying to do. The original idea was to completely remove the `top` and `bottom` arrays and just use a single integer to keep track of the current step while keeping the logic of the off-diagonal kernels largely unchanged. The way I am doing it, given a current step, all threads would be able to compute their pair in a constant number of predicated instructions without having to know the previous pair they used (if any), while what you suggest depends on the knowledge of the previous pair and keeps using the `top` and `bottom` arrays. If you try to model your approach in a similar way as I did, you'll find that to remove the need to iterate `O(half_blocks)` times you'll get something very close to what the `f` and `g` functions are accomplishing already.

That said, I did not have the time to complete what I was trying to do, and one can argue that what I was doing is largely irrelevant anyway, because even after I remove the `top` and `bottom` arrays we would still need to keep the `syevj_cycle_pairs` kernel (whose launch accounts for most of the time spent to cycle the pairs for most matrix sizes anyway). Thus, I'll just implement your solution and defer any further change to a refactoring of Jacobi. :)

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1103,44 +1103,88 @@ ROCSOLVER_KERNEL void syevj_offd_rotate(const bool skip_block,
...
+    rocblas_int n = half_blocks - 1;
 
-    auto g = [n = n](auto j) -> auto
+    auto cycle = [n = n](auto i) -> auto
     {
-        using I = decltype(j);
-        j = (j - 1) % (2 * n + 1) + 1;
-        I i{};
+        using I = decltype(i);
+        i = (i - 1) % (2 * n + 1) + 1;
```
EdDAzevedo, line 1113: Can variable i become negative?  If so, perhaps  add (2*n+1) to use something like 

i = (  (i - 1) + (2*n+1)) % (2*n+1)  + 1

to avoid mod arithmetic on negative value?

EdDAzevedo (approved):
One suggestion is to just leave the top/bottom/cycle code there but unused and read the independent pairs from a table. The code to generate the table on cpu host (and code to check the table) are available.  It will just require code to allocate and copy the table to device and minor change to the kernel code to read the table,  something like
p = table[ 2*iwave + iround * m ]; q = table[ 2*iwave + 1 + iround * m ]; // pair (p,q)

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1155,6 +1155,56 @@ ROCSOLVER_KERNEL void
...
+        bottom[0] = cycle(bottom[0]);
+    }
+    else
+    {
+        for(rocblas_int l = tidx; l < half_blocks; l += dimx)
+        {
+            top[l] = cycle(top[l]);
+            bottom[l] = cycle(bottom[l]);
+        }
+    }
```
tfalders: Since tidx 0 is not participating in the loop, I think the loop will actually miss some entries. The following should be more robust:

```
if(tidx == 0)
    bottom[0] = cycle(bottom[0])
for(rocblas_int l = tidx + 1; l < half_blocks; l += dimx)
    [ ... ]
```

tfalders (approved):
Looks good! I just have one comment

library/src/lapack/roclapack_syevj_heevj.hpp
tfalders: It looks like you also need to run clang-format

jmachado-amd: Good catch!

library/src/lapack/roclapack_syevj_heevj.hpp
jmachado-amd: The idea is that the inputs of the `cycle` lambda are thread ids (in particular in the range `{1, 2, ..., 2*n+1}`), we might defend ourselves from it being called with integers larger than `2*n` but negative numbers are not expected -- so I didn't include the extra `+ 2*n  + 1`. Now that I think of it, the code has become so simple that it seems completely harmless to remove that line.

EDIT: The input of the `cycle` lambda are the contents of the `top` and `bottom` vectors now, not thread ids. But again, those are indices and should not be negative at any time.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1189,19 +1135,15 @@ ROCSOLVER_KERNEL void
...
     rocblas_int tidx = hipThreadIdx_x;
     rocblas_int dimx = hipBlockDim_x;
-    rocblas_int k{};
 
     if(tidx == 0)
     {
         bottom[0] = cycle(bottom[0]);
     }
-    else
+    for(rocblas_int l = tidx + 1; l < half_blocks; l += dimx)
```
EdDAzevedo, line 1143: Minor stylistic suggestion: Personally, I try to avoid using "l" (lower case L) as a variable since it can look like "1" (one) depending on the font. You might consider using  say "ell" instead of "l".  Just a suggestion.

PR #765, opened 2024-07-17
Use rocBLAS gemm_64

tfalders:


library/src/include/rocblas.hpp
```diff
@@ -240,6 +242,54 @@ rocblas_status rocblas_internal_trsm_batched_template_64(rocblas_handle,
                                                          rocblas_stride offset_invA = 0,
                                                          rocblas_stride stride_invA = 0)
     = delete;
+
+// gemm
+template <typename T>
+rocblas_status rocblas_internal_gemm_template_64(rocblas_handle,
```
cgmb, line 249: The preprocessor guards on these declarations need to be based on the version they were added in, rather than just on `HAVE_ROCBLAS_64`. We don't want to redeclare functions like `rocblas_internal_gerc_batched_template_64` if using rocblas 4.2.

tfalders: My understanding from https://github.com/ROCm/rocSOLVER/pull/730#discussion_r1637246614 was that we would continue to use `HAVE_ROCBLAS_64` and bump the required version number since the addition of gemm_64 was only one version off from the other 64-bit functions

cgmb: The linked comment is referring to requiring rocblas 4.3 to build with the 64-bit API, but this about when the 64-bit API is not available. We need to prevent the (non-GEMM) 64-bit rocblas functions from being defined twice when using rocblas 4.2. The deleted functions should only be declared if they're not declared by rocblas.

cgmb (approved):

qjojo (approved):

PR #767, opened 2024-07-22
Add 64-bit version of geqrf, geqr2

qjojo:
Adds 64-bit API support to geqr2 and geqrf along with their strided and batched variants.


library/src/lapack/roclapack_geqrf.cpp
```diff
@@ -48,12 +44,12 @@ rocblas_status rocsolver_geqrf_impl(rocblas_handle handle,
         return st;
 
     // working with unshifted arrays
-    rocblas_int shiftA = 0;
+    I shiftA = 0;
```
tfalders: Should be rocblas_stride

tfalders (changes requested):
Looking good so far, but I think you missed geqr2_batched.cpp, geqr2_strided_batched.cpp, geqrf_batched.cpp, and geqrf_ptr_batched.cpp.

EDIT: Also, it looks like you need to resolve a conflict with the PR I merged. Sorry about that.

qjojo:
Fixed some missing exports and added guards to allow building without the 64-bit API.

tfalders (approved):
Looks good!

jzuniga-amd (approved):
LGTM

EdDAzevedo (approved):

qjojo:
Since the failures are known and unrelated I will manually merge this.

PR #774, opened 2024-07-31
Added small-size larfg and larf kernels (limited architectures)

tfalders:
Same as https://github.com/ROCm/rocSOLVER/pull/759, but disabled on wave 32 architectures due to test failures in SYGVX.


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
     - GEQRF_64 (with batched and strided\_batched versions)
 
 ### Optimized
+- Improved performanced of LARFG, LARF, and downstream functions such as GEQR2 and GEQRF on some architectures
```
cgmb: We should provide enough information for users to identify if their GPUs will see improvements. I would explicitly say "CDNA architectures". Or maybe "wave64 architectures".

cgmb (approved):

library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -208,7 +208,8 @@ rocblas_status rocsolver_larfg_template(rocblas_handle handle,
     }
 
     // if n is small, use small-size kernel
-    if(n <= LARFG_SSKER_MAX_N)
+    // TODO: Some architectures have failures in sygvx with small-size kernels enabled, more investigation needed
+    if(n <= LARFG_SSKER_MAX_N && warpSize >= 64)
```
EdDAzevedo: You might check whether warpSize is always 64 on the host but the value of "warpSize" may be a different value inside device code for particular architecture. So just checking warpSize on the host may always have value 64 and cannot distinguish wave64 and wave32 GPU.

----------------

On my gfx1030 wave32 GPU in my Linux desktop, the output from the following is test code is 

on host, warpSize = 64
host_warpSize = 64, device_warpSize = 32




--------------------------

#include "hip/hip_runtime.h"
#include "hip/hip_runtime_api.h"

__global__ void kernel( int host_warpSize )
{
	constexpr auto device_warpSize = warpSize;
	printf("host_warpSize = %d, device_warpSize = %d\n",
			host_warpSize, device_warpSize );
}

int main()
{
	printf("on host, warpSize = %d\n", warpSize );

	int host_warpSize = warpSize;
	kernel<<< dim3(1,1,1), dim3(1,1,1), 0, 0 >>>(host_warpSize);

	return(0);
}

EdDAzevedo: Perhaps use hipGetDevice( &device ) then call hipGetDeviceProperties(&properties, device)  then get the length of wavefront from

wavefront_size = properties.warpSize,

qjojo (approved):

library/src/auxiliary/rocauxiliary_larfg.hpp
cgmb: > You might check whether warpSize is always 64 on the host

Good catch. I believe that is the case. Although, the property doesn't make much sense on the host. We should avoid using it there.

> the value of "warpSize" may be a different value inside device code for particular architecture. So just checking warpSize on the host may always have value 64 and cannot distinguish wave64 and wave32 GPU.

Also note that the warpSize property is a function in CUDA, rather than a constant like in HIP. We should be careful how we use it, as I wouldn't be too surprised if the compiler and runtime folks were to decide in the future to make it non-constant.

EdDAzevedo: I think it may make sense to get the correct warpSize on the host in case we need to optimize the launch configuration to use number of  threads that are multiple of warpSize or allocate arrays to have size to enforce alignment related to warpSize. 

We may need to query the GPU property to get this info.

tfalders: Now that https://github.com/ROCm/rocSOLVER/pull/775 has been merged in, the SYGVX error on gfx1101 may hopefully be resolved. I'll try removing this check to see if it's still needed.

tfalders: But it looks like @EdDAzevedo is right about warpSize. Even with this check in place, CI has failed with the same test.

library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -119,14 +119,17 @@ ROCSOLVER_KERNEL void set_taubeta(T* tauA,
 }
 
 template <typename T, typename I>
-void rocsolver_larfg_getMemorySize(const I n, const I batch_count, size_t* size_work, size_t* size_norms)
+rocblas_status rocsolver_larfg_getMemorySize(const I n,
+                                             const I batch_count,
+                                             size_t* size_work,
+                                             size_t* size_norms)
 {
```
EdDAzevedo, line 126: Can you provide more details about what was the compiler issue? 
I have a concern that this change may potentially cause a lot of conflicts with other unmerged branches.
Just a thought.

tfalders: The memory access fault came from the fact that I zero out the workspace when using the small-size kernel, but I forgot to disable this when I disabled the small-size kernel. Once I disabled it, the compiler error arose because the HIP_CHECK macro returns a rocblas_status, but the getMemorySize method returns void. The fix was to return a rocblas_status from getMemorySize.

tfalders: Also, it looks like size_norms and size_work will be undefined if HIP_CHECK triggers a return, so I should fix that too.

EdDAzevedo: You might consider setting size_norms and size_work to be zero first, rather than inside a if test, so that these are always initialized.

Is the coding convention is all the _getMemorySize() methods will consistently return a rocblas_status?

jzuniga-amd (approved):
LGTM

library/src/auxiliary/rocauxiliary_larfg.hpp
tfalders: Remind me to change it back to void once the SYGVX failure has been fixed. Alternatively, we can start returning a rocblas_status from all getMemorySize methods.

EdDAzevedo (approved):

PR #775, opened 2024-07-31
Fix synchronization issue in stein

tfalders:
As @jmachado-amd suspected, there is in fact a synchronization defect in STEIN. There may still be issues in STEBZ, but this is likely the primary culprit for our test failures in the X functions along with the initial eigenvector choices resolved by https://github.com/ROCm/rocSOLVER/pull/714.


qjojo (approved):

library/src/auxiliary/rocauxiliary_stein.hpp
```diff
@@ -238,6 +238,7 @@ __device__ void run_stein(const int tid,
                 nrm2<MAX_THDS, S>(tid, blksize, work, 1, sval2);
```
EdDAzevedo, line 238: Do we need a __syncthread() between iamax()  and nrm2()? 
Perhaps a __syncthread() can be added inside  iamax() and inside nrm2()?  

Perhaps the coding convention may be a __syncthread() before and after a parallel for loop?

EdDAzevedo (approved):

jzuniga-amd (approved):

PR #776, opened 2024-07-31
Add 64-bit api for potf2, potrf, potrs

qjojo:
Adds 64-bit variants of potf2, potrf, potrs, and their strided/batched variants.


EdDAzevedo (approved):

tfalders (changes requested):
I still need to look over the PR, but from an early glance it seems that you're returning a 32-bit info instead of 64-bit. Since m and n are 64-bit I think it's safest to have info be 64-bit as well. I'll prepare this change and open a PR on your local fork as soon as I can.

qjojo:
@tfalders I intended for all the infos to be I*, I just pushed a fix for the inconsistency in potf2. I believe other libraries leave the info param as int* even in 64-bit functions but that's something to worry about in the hipSOLVER layer.

tfalders:
> @tfalders I intended for all the infos to be I*, I just pushed a fix for the inconsistency in potf2. I believe other libraries leave the info param as int* even in 64-bit functions but that's something to worry about in the hipSOLVER layer.

Yeah, for getrf we added a getrf_info32 method that could be called from hipSOLVER since cuSOLVER expects a 32-bit info. I've opened a PR on your local fork that adds a similar method for potrf.

clients/common/lapack/testing_potf2_potrf.hpp
tfalders, line 1: Whoops! I appear to have missed converting the rocblas_ints on lines 332, 336, 358, 362, 378, and 381 to type I. Doesn't seem to be a blocker, but it would be nice to change those over.

library/src/lapack/roclapack_potrf_info32.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+/*
+ * ===========================================================================
+ *    C wrapper
+ * ===========================================================================
+ */
+
+extern "C" {
+
+rocblas_status rocsolver_spotrf_info32(rocblas_handle handle,
```
tfalders: Looks like I also forgot to add ROCSOLVER_EXPORT to these methods. This will be needed to call them from hipSOLVER.

tfalders (approved):
Looks good! I appear to have missed a couple things in my review PR. Please take care of them when you're able.

jzuniga-amd (approved):

qjojo:
The pipeline failures are known and unrelatedd, squashing and merging.

PR #777, opened 2024-08-01
Fix dead link to .pdf that does not exist

amd-jnovotny:
This fixes a dead link in memory.rst that leads to an old PDF file that is no longer in the develop branch. I replaced the link with one to the Device Memory section of the rocBLAS programmers guide, which should have up-to-date information.


cgmb (approved):

cgmb:
It's nice to hear that the PDF has been replaced with editable sources.

jmachado-amd (approved):

PR #778, opened 2024-08-02
(WIP) Add functionality to print hashes and computed errors in tests

jmachado-amd:
rocSOLVER's tests currently output no useful information for passing tests, which makes it hard to ascertain when new code adversely affects the accuracy of methods while also hindering the investigation of bugs that may affect determinism.

This PR adds code to print the computed error in each test and hashing functions to allow printing hashes of input and output matrices, when desired.

### DIFF
--- a/clients/common/lapack/testing_sygvx_hegvx.hpp
+++ b/clients/common/lapack/testing_sygvx_hegvx.hpp
@@ -430,8 +430,11 @@ void sygvx_hegvx_getError(const rocblas_handle handle,
     // Compute input data hash (combine matrices A and B)
     //
     std::size_t input_hash = 0;
-    input_hash = hash_combine(input_hash, hA[0], lda * n * bc);
-    input_hash = hash_combine(input_hash, hB[0], ldb * n * bc);
+    for(rocblas_int b = 0; b < bc; ++b)
+    {
+        input_hash = hash_combine(input_hash, hA[0], lda * n);
+        input_hash = hash_combine(input_hash, hB[0], ldb * n);
+    }
 
     // execute computations
     // GPU lapack
--- a/clients/common/lapack/testing_sygvx_hegvx.hpp
+++ b/clients/common/lapack/testing_sygvx_hegvx.hpp
@@ -263,6 +263,8 @@ void sygvx_hegvx_initData(const rocblas_handle handle,
             // for testing purposes, we start with a reduced matrix M for the standard equivalent problem
             // with spectrum in a desired range (-20, 20). Then we construct the generalized pair
             // (A, B) from there.
+
+            [[maybe_unused]] volatile auto mptr = memset(hB[b], 0, n * ldb * sizeof(T));
             for(rocblas_int i = 0; i < n; i++)
             {
                 // scale matrices and set hA = M (symmetric/hermitian), hB = U (upper triangular)
@@ -424,6 +426,13 @@ void sygvx_hegvx_getError(const rocblas_handle handle,
     sygvx_hegvx_initData<true, true, T>(handle, itype, evect, n, dA, lda, stA, dB, ldb, stB, bc, hA,
                                         hB, A, B, true, singular);
 
+    //
+    // Compute input data hash (combine matrices A and B)
+    //
+    std::size_t input_hash = 0;
+    input_hash = hash_combine(input_hash, hA[0], lda * n * bc);
+    input_hash = hash_combine(input_hash, hB[0], ldb * n * bc);
+
     // execute computations
     // GPU lapack
     CHECK_ROCBLAS_ERROR(rocsolver_sygvx_hegvx(STRIDED, handle, itype, evect, erange, uplo, n,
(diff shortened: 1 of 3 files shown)


EdDAzevedo (approved):
The std::hash may be implementation dependent. One may consider another hash function such as the following (found using edge browser) to make this more portable across OS and compiler versions. Just a suggestion.

```cpp
unsigned hash(const char* s, unsigned salt) {
    unsigned h = salt;
    while (*s)
        h = h * 101 + static_cast<unsigned>(*s++);
    return h;
}
```

clients/common/misc/rocsolver_test.hpp
```diff
@@ -51,8 +51,49 @@ namespace fs = std::experimental::filesystem;
 #define USE_ROCBLAS_REALLOC_ON_DEMAND true
 
 #ifdef ROCSOLVER_CLIENTS_TEST
-#define ROCSOLVER_TEST_CHECK(T, max_error, tol) ASSERT_LE((max_error), (tol)*get_epsilon<T>())
+
+#ifdef ROCSOLVER_CLIENTS_TEST_PRINT_EXTRA_MESSAGES
+// Format output similarly as GTEST messages
+#define ANSI_CODE_GTEST_GREEN "\033[0;32m"
+#define ANSI_CODE_NORMAL_TERM "\033[0;0m"
```
cgmb, line 58: If we're going to print with colour, we should follow Google Test's lead by obeying [the gtest_color option](https://google.github.io/googletest/advanced.html#colored-terminal-output) and using TTY detection for the default. I'd suggest printing without colour for now, as that is a bit of a distraction from the main purpose of this change.

cgmb: Also, the fmtlib that we use in rocsolver has a color API. https://fmt.dev/11.0/api/#color-api

clients/common/misc/rocsolver_test.hpp
```diff
@@ -51,8 +51,49 @@ namespace fs = std::experimental::filesystem;
...
+// Macro ROCSOLVER_GTEST_MSG_PRINTER is also used to print hashes in tests
+#define ROCSOLVER_GTEST_MSG_PRINTER \
+    std::cout << ANSI_CODE_GTEST_GREEN << "[          ] " << ANSI_CODE_NORMAL_TERM
```
cgmb, line 61: Use fmt. We added it as a dependency to do atomic prints and avoid the iostreams interfaces where the printing of a line may be split into multiple function calls. It's not that important at the moment, but it's an invariant we inherited from rocfft / rocblas and I want to retain it in case we do ever run tests in parallel.

clients/common/misc/rocsolver_test.hpp
```diff
@@ -51,8 +51,49 @@ namespace fs = std::experimental::filesystem;
...
+        constexpr auto max_precision{std::numeric_limits<double>::max_digits10 + 1};    \
+        double tol_ = static_cast<double>(tol) * static_cast<double>(get_epsilon<T>()); \
+        double max_error_ = static_cast<double>(max_error);                             \
+        ROCSOLVER_GTEST_MSG_PRINTER                                                     \
+            << "Computed error: " << ROCSOLVER_STRNGFY(max_error) << " / "              \
+            << ROCSOLVER_STRNGFY(((tol)*get_epsilon<T>())) << " = "                     \
+            << std::setprecision(max_precision)                                         \
+            << ((max_error_ >= 0.) && (tol_ > 0.) ? max_error_ / tol_ : -1.)            \
+            << std::setprecision(default_precision) << std::endl                        \
+            << std::flush;                                                              \
```
cgmb, line 79: This would be so much nicer using the fmtlib interface.

clients/common/lapack/testing_sygvx_hegvx.hpp
```diff
@@ -263,6 +263,8 @@ void sygvx_hegvx_initData(const rocblas_handle handle,
             // for testing purposes, we start with a reduced matrix M for the standard equivalent problem
             // with spectrum in a desired range (-20, 20). Then we construct the generalized pair
             // (A, B) from there.
+
+            [[maybe_unused]] volatile auto mptr = memset(hB[b], 0, n * ldb * sizeof(T));
```
cgmb, line 267: I assume the `volatile auto mptr` is an attempt to ensure that memset is not optimized out, but I'm not sure that's guaranteed. Why is this needed?

clients/common/lapack/testing_sygvx_hegvx.hpp
```diff
@@ -472,6 +484,41 @@ void sygvx_hegvx_getError(const rocblas_handle handle,
...
+        }
+    }
+
+    //
+    // Print hashes
+    //
+    ROCSOLVER_GTEST_MSG_PRINTER << "Input matrix hash: " << input_hash << std::endl << std::flush;
+    ROCSOLVER_GTEST_MSG_PRINTER << "Rocsolver eigenvalues hash: " << rocsolver_eigenvalues_hash
+                                << std::endl
+                                << std::flush;
```
cgmb, line 514: `std::endl` is defined as a newline and a flush, so this is redundant. Although, again, I would suggest using fmt rather than iostreams.

cgmb (changes requested):
I would like to be able to control this behaviour at runtime. Could we control this with an environment variable, rather than a preprocessor flag?

PR #786, opened 2024-08-12
STEDC optimizations

jzuniga-amd:
Minor optimizations to STEDC kernels so that they can work with a larger number of sub-blocks (more than 8 levels). The tuning is still pending.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_stedc.hpp
+++ b/library/src/auxiliary/rocauxiliary_stedc.hpp
@@ -754,7 +754,8 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_divide_kernel(const ro
                                                                         const rocblas_stride strideD,
                                                                         S* EE,
                                                                         const rocblas_stride strideE,
-                                                                        rocblas_int* splitsA)
+                                                                        rocblas_int* splitsA,
+                                                                        rocblas_int levs = 0)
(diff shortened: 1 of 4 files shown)


jmachado-amd (commented):
I could not follow all of the changes, yet the overall picture is sound and the code looks good!

jmachado-amd (approved):

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1281,6 +1281,128 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+                            }
+                        }
+                    }
+                    else
+                    {
+                        for(int j = tidb; j < (dd - 1) / 2; j += hipBlockDim_x)
+                        {
+                            if(tmpd[2 * j + 1] > tmpd[2 * j + 2])
+                            {
+                                valf = tmpd[2 * j + 1];
```
EdDAzevedo: Minor suggestion: Perhaps consider a swap method, say something like
auto swap = [](auto &x, auto &y) {   auto const t = x; x = y; y = t; };

or the swap method in src/include/lib_device_helpers.hpp

EdDAzevedo (approved):
I don't fully understand the code but if it passes the google tests and got better performance then it is worth merging.

PR #788, opened 2024-08-12
Add helper methods to simplify writing and debugging tests (part 1)

jmachado-amd:
Currently, most input matrices used in rocSOLVER's tests are written elementwise (by hand) in a test-by-test basis. This commit adds foundational methods to simplify the creation of input matrices with desired properties, and also syntactic sugar to abstract the common matrix operations required to support this task. Moreover, it also adds code to allow writing test code in a higher level of abstraction, to improve their legibility and future maintenance.

This PR focus on utilities and symmetric matrix generation required to refactor SYEVX's tests. Future PRs will address:
    * immutable, special (Toeplitz, Wilkinson, Clement, etc.) matrix generation, glued matrices, and random matrices;
    * lazy-evaluation and caching of expensive computations;
    * a method to export matrices build as part of a test to an octave script (for debugging purposes).


clients/common/matrix_utils/host_matrix.hpp
```diff
@@ -0,0 +1,1274 @@
...
+
+    static auto Eye(I dim) -> HostMatrix<T_, I_>
+    {
+        return HostMatrix<T_, I_>::Eye(dim, dim);
+    }
+
+    static auto Ones(I nrows, I ncols) -> HostMatrix<T_, I_>
+    {
+        /* HostMatrix<T_, I_> O(std::max(nrows, I(1)), std::max(ncols, I(1))); */
+        HostMatrix<T_, I_> O(nrows, ncols);
```
EdDAzevedo: I would like to suggest not using "O" (Oh) as variable since depending on the terminal font or printer font (and reader's eye sight), it may look too much like "zero".

clients/common/matrix_utils/matrix_utils_detail.hpp
```diff
@@ -0,0 +1,182 @@
...
+        const int ncolsQ = ncolsX;
+        const int ldQ = nrowsQ;
+        cpu_geqrf<T>(nrowsQ, ncolsQ, Q, ldQ, tau.data(), work.data(), worksize);
+
+        // Copy upper triangular part of intermediate result Q into R
+        const int nrowsR = ncolsX;
+        const int ncolsR = ncolsX;
+        const int ldR = nrowsR;
+        {
+            auto volatile mptr = memset(R, 0, nrowsR * ncolsR * sizeof(T));
```
EdDAzevedo: Minor suggestion, consider using "sizeof(T) * nrowsR * ncolsR" to avoid potential 32bit integer overflow, assuming the compiler evaluates the expression from left to right. The concern is evaluation order of "nrowsR * ncolsR" may already cause overflow.   

I see  "mptr" exists only in the block scope so it is not obvious to me what is the purpose of declaring "auto volatile mptr".

clients/common/matrix_utils/host_matrix.hpp
```diff
@@ -0,0 +1,1274 @@
...
+
+        nrows_ = nrows;
+        ncols_ = ncols;
+
+        return true;
+    }
+
+    virtual T operator()(I i, I j) const override
+    {
+        return data_[i + ld_ * j];
```
EdDAzevedo: Is there a concern about 32bit overflow? Perhaps consider something like  "i + static_cast<int64_t>(ld_)*j" or internally store ld_ to be type int64_t?

clients/common/matrix_utils/host_matrix.hpp
jmachado-amd: I've changed the variable name.

clients/common/matrix_utils/matrix_utils_detail.hpp
jmachado-amd: I updated the code so that everything promotes to `std::size_t`. However, those methods are meant to be used only by classes and functions in the `matrix_utils` folder, which can be tuned to only call functions in the `detail` namespace with indices that don't overflow.

Edit: my apologies I missed the second point, `mptr` is there just for debugging (if `memset` is not omitted by the compiler, it returns the address of the pointer pointing to the memory that was set, which can be used as a sanity check while debugging) -- it can be safely ignored.

clients/common/matrix_utils/host_matrix.hpp
jmachado-amd: We can just set the template parameter `I` of class `HostMatrix` to `std::int64_t` and achieve the same result, however, I disagree with the idea. The purpose of those methods is to simplify our tests; if we write them to handle overflows graciously, we may end with test methods that work around the library bugs.

The idea is for those methods to detect and raise exceptions in the case of range errors (overflows included); first checking for overflows in the constructors (in the short term, but not necessarily in this PR), and adding extra checks later to a selection of methods (the details of which should be discussed).

TorreZuk: I may be out to lunch as not aware of your test framework design, but will just say in the rocblas test harness client code we now only do int64_t (or size_t) for all variables and test argument models to simplify it.  As the host code we use to test both API (int, int64_t). I can't think of a case where you want host side to ever overflow.    I don't follow the notion of raising range errors of overflows as even your 32bit API is supposed to support > 32bit indices.  I would like this concept explained further in case it would be of use to rocblas.

jzuniga-amd (approved):
LGTM.

clients/common/matrix_utils/host_matrix.hpp
EdDAzevedo: If the rocsolver test framework is using int64_t everywhere, we'll need to be careful with calling CPU lapack or CPU blas or other CPU libraries (say for sparse matrices)  that expect int32_t in the interface.   Just a thought.

jmachado-amd: Hi Torre, these new utilities are meant to start a conversation and allow us to review the design of our tests -- they are a work in progress. That said, it is not a given that our 32-bit API does support >32-bit indices and, as Ed pointed out, we also rely on external code that doesn't support 64-bit indices (out of the box).

Although our current tests and the typical usage of a library like rocSOLVER tends to fall within the limit of 32-bit indices (for time complexity reasons), we must start the conversation about what we plan to do with a more general situation now and update our code accordingly with time, instead of waiting and having to touch dozens of tests at the same time later when it eventually becomes required. For a concrete example, it would be nice to generalize our tests to a point in which we can easily import input matrices saved to a file (for any given method) and, at that time, having host code that can trigger common alarms (such as integer overflow) is handy.

clients/common/matrix_utils/eigenvalues_generator.hpp
```diff
@@ -32,152 +32,150 @@
...
+        case 3:
+            if(n == 1)
+            {
+                eigenvalues(0) = S(1);
+            }
+            else
+            {
+                for(I i = 0; i < n; ++i)
+                {
+                    eigenvalues(i) = std::pow(kappa, -i / (n - 1));
```
EdDAzevedo: Perhaps use -S(i)/S(n-1)  instead of -i/(n-1)   to guarantee there is no truncation with integer division.

clients/common/lapack/testing_syevx_heevx.hpp
```diff
@@ -235,7 +238,27 @@ void syevx_heevx_initData(const rocblas_handle handle,
                 if(i == n / 4 || i == n / 2 || i == n - 1 || i == n / 7 || i == n / 5 || i == n / 3)
                     hA[b][i + i * lda] *= -1;
             }
+#else
+        // New matrix initialization
+        using hMat = HostMatrix<T, rocblas_int>;
```
qjojo: This type name should be in TitleCase as elsewhere in rocsolver.

clients/common/lapack/testing_syevx_heevx.hpp
```diff
@@ -292,6 +315,8 @@ void syevx_heevx_getError(const rocblas_handle handle,
                           Ih& hinfoRes,
                           double* max_err)
 {
+    using hMat = HostMatrix<T, rocblas_int>;
```
qjojo: Same as above.

clients/common/matrix_utils/matrix_interface.hpp
```diff
@@ -0,0 +1,117 @@
...
+class MatrixInterface
+{
+public:
+    using T = T_;
+    using I = I_;
+    using S = decltype(std::real(T{}));
+
+    virtual ~MatrixInterface() = default;
+
+    virtual T const* data() const
```
qjojo, line 48: Should these be pure virtual instead of returning nullptr? Since nrows() and friends are pure virtual, I think all these functions should either return defaults for an empty matrix or be pure virtual, if you're going to say that this->data() = nullptr, then this->ncols() should = 0, or both should be pure virtual.

clients/common/matrix_utils/matrix_utils_detail.hpp
```diff
@@ -0,0 +1,196 @@
...
+        const int ncolsQ = ncolsX;
+        const int ldQ = nrowsQ;
+        cpu_geqrf<T>(nrowsQ, ncolsQ, Q, ldQ, tau.data(), work.data(), worksize);
+
+        // Copy upper triangular part of intermediate result Q into R
+        const int nrowsR = ncolsX;
+        const int ncolsR = ncolsX;
+        const int ldR = nrowsR;
+        {
+            auto volatile mptr = memset(R, 0, sizeof(T) * nrowsR * ncolsR);
```
qjojo, line 131: I don't think this needs to be volatile, is there something that might cause mptr to mutate in a way that's not detectable by the implementation? If not we should not mark it volatile so the optimizer can do as it pleases.

clients/common/matrix_utils/matrix_interface.hpp
```diff
@@ -0,0 +1,117 @@
...
+#pragma once
+
+#include <vector>
+
+namespace matxu
+{
+template <typename T_,
+          typename I_,
+          /* typename = typename std::enable_if<std::is_arithmetic<std::decay_t<T_>>::value>::type, */
+          typename = typename std::enable_if<std::is_integral<std::decay_t<I_>>::value
```
qjojo, line 37: I like the inclusion of this pseudo-concept, we should declare it in a standalone template so it can have a descriptive name and possibly be reused elsewhere.

qjojo (approved):

clients/common/matrix_utils/matrix_interface.hpp
jmachado-amd: We sure could use some of those elsewhere in the library as defensive programming. Personally, I would like to extend `is_arithmetic` to include types such as `rocblas_complex_num`, but I'll have to leave it for a future PR.

clients/common/matrix_utils/matrix_utils_detail.hpp
jmachado-amd: This variable is just for debugging purposes and is never used, this is why it was declared `volatile`, so that the compiler would not optimize it away. That said, there is little use for this type of construction in this code, they will be removed in the future.

clients/common/matrix_utils/matrix_interface.hpp
jmachado-amd: I see a difference here, `data()` returning `nullptr` is valid for all possible models of the `MatrixInterface` class, but I am not sure whether the other methods should return `0`. The issue is related to how we expect empty matrices to work (provided that we agree on allowing them to exist in the first place), and it is easier to discuss in the case of the `norm()` function.

On one hand, it feels natural to say that the norm of an empty matrix should be `0`, but this would also mean that if a test failed to initialize one of its matrices properly then it could pass (say, by propagating this 0 to the effective error bound) when it shouldn't. A better approach would be to make the norm of an empty matrix a _NaN_, but this comes with other issues. I intend to bring this conversation to the whole team so that we can make a design that everyone, at least, is aware of.

clients/common/lapack/testing_syevx_heevx.hpp
jmachado-amd: Changing it to `HMat` for the time being.

clients/common/lapack/testing_syevx_heevx.hpp
jmachado-amd: Done.

PR #794, opened 2024-08-14
Revert "Update changelog to mention the `stebz` synchronization defect (#752)"

jmachado-amd:
This reverts commit [1eb0f16](https://github.com/ROCm/rocSOLVER/commit/1eb0f161413b6a6f68fc549003b295b454073d25), as it was not cherry-picked into 6.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,11 +39,6 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Fixed
 - Fixed potential accuracy degradation in SYEVJ/HEEVJ for inputs with small eigenvalues.
 
-### Known Issues
-- A known issue in STEBZ can lead to errors in routines based on Bisection to compute eigenvalues for
-  symmetric/hermitian matrices (e.g., SYEVX/HEEVX and SYGVX/HEGVX), as well as singular values (e.g.,
-  BDSVDX and GESVDX).
-
 
 ## rocSOLVER 3.25.0 for ROCm 6.1.0
 ### Added


CHANGELOG.md
```diff
@@ -39,11 +39,6 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Fixed
 - Fixed potential accuracy degradation in SYEVJ/HEEVJ for inputs with small eigenvalues.
 
-### Known Issues
-- A known issue in STEBZ can lead to errors in routines based on Bisection to compute eigenvalues for
-  symmetric/hermitian matrices (e.g., SYEVX/HEEVX and SYGVX/HEGVX), as well as singular values (e.g.,
-  BDSVDX and GESVDX).
-
```
jzuniga-amd, line 46: We should move this to 6.3, right?
Also, is "Fixed synchronization issue in STEIN" really in 6.2? I don't see that commit in rocm-rel-6.2 branch. 
Maybe it also needs to be moved out to 6.3.

jmachado-amd: Yes, we should move this to 6.3 at some point; but the optimistic part of me says that we might still have time to tackle the issue for 6.3. If we go with the route of adding the comment to 6.3's changelog now, we should also follow Cory's suggestion, and mention when the issue was originally introduced, so that users don't assume that this was introduced in 6.3. 

As far as I know, Troy's fix for STEIN is part of 6.3 and there are no plans to cherry pick it for 6.2.1. It should be included in the changelog as well.

jzuniga-amd: Ok good. 
Then could you move the STEIN entry to 6.3?
For some reason it is currently under 6.2 in develop branch

jmachado-amd: I am probably missing something, I don't see any STEIN entry in 6.2, just the STEBZ entry that this revert is removing.

jzuniga-amd: oh, maybe you need to update your branch.
https://github.com/ROCm/rocSOLVER/blob/5a0abdb4e6fe55a12fce6539228487ebf3f52945/CHANGELOG.md?plain=1#L58

jmachado-amd: Ah, that explains it! But I prefer to open a new PR to deal with that and probably use it to update the changelog in other places as well, we can move the conversation there in a moment.

jzuniga-amd (approved):

CHANGELOG.md
jmachado-amd: The extra changes were added to PR #795.

qjojo (approved):

PR #799, opened 2024-08-20
Use L3 BLAS in LARFT

AGonzales-amd:
This PR introduces a potential optimization to the LARFT routine. The modification aims to reduce the size of the gemv computations and instead offloads the block part of the computation to a call to gemm. Additionally, in some cases the modified method performs worse than the original so the latter is dispatched instead.

Following are performance LARFT figures with configuration k=64, rocblas_forward_direction and rocblas_column_wise (config for QR factorization).
 - Single Precision
log_compare_slarftcf_m
 - Double Precision
log_compare_dlarftcf_m
 - Complex Single Precision
log_compare_clarftcf_m
 - Complex Double Precision
log_compare_zlarftcf_m

column_wise and forward_direction has similar performance characteristic while row_wise configurations show improvements for all data types, e.g.,
 - Single Precision
compare_slarftrf_m
This also shows that the gains with the row_wise configuration is more significant (~x18 vs ~x3 speedup).

Curiously, GEQRF shows weird performance with this modification. These figures were generated with square matrices.
 - Single Precision
compare_sgeqrf_m
 - Double Precision
compare_dgeqrf_m
 - Complex Single Precision
compare_cgeqrf_m
 - Complex Double Precision
compare_zgeqrf_m

Although there is mostly performance gains, there are some cases where performance degrades with the modification.


library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -4,6 +4,10 @@
  *     Univ. of Tennessee, Univ. of California Berkeley,
  *     Univ. of Colorado Denver and NAG Ltd..
  *     December 2016
+ * and
+ * Joffrain, Low, Quintana-Ortí, et al. (2006). Accumulating householder
+ * transformations, revisited.
```
tfalders: It would be nice if this line were indented.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -71,9 +76,29 @@ ROCSOLVER_KERNEL void set_triangular(const rocblas_int n,
...
-                    Fp[j + i * ldf] = -tp[i] * Vp[i + j * ldv];
+                {
+                    if(!inc)
+                    {
+                        Fp[j + i * ldf] = -tp[i] * Vp[i + j * ldv];
+                    }
+                    else
+                    {
+                        Fp[j + i * ldf] *= -tp[i];
+                        Fp[j + i * ldf] += -tp[i] * Vp[i + j * ldv];
```
tfalders: You've got two writes to global memory here, as well as two reads from `tp[i]`. You may want to cache `tp[i]` and `Fp[j + i * ldf] * -tp[i]` in separate local variables, which will hopefully speed up the kernel.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -269,13 +390,50 @@ rocblas_status rocsolver_larft_template(rocblas_handle handle,
     rocblas_fill uplo;
     rocblas_operation trans;
 
+    const bool call_l3 = larft_do_l3<T>(n, direct, storev) && n > k;
```
tfalders: I'm not a fan of these names. Maybe something like `use_gemm` and `larft_use_gemm`.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -269,13 +390,50 @@ rocblas_status rocsolver_larft_template(rocblas_handle handle,
...
+
+    const rocblas_int u1_n = call_l3 ? k : n;
+    const rocblas_int u2_n = call_l3 ? n - k : 0;
+
+    if(call_l3)
+    {
+        if(direct == rocblas_forward_direction && storev == rocblas_column_wise)
```
tfalders, line 399: A short comment explaining what the gemm is doing would be appreciated.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -161,6 +247,41 @@ ROCSOLVER_KERNEL void set_tau(const rocblas_int k, T* tau, const rocblas_stride
...
+                 const rocblas_storev storev)
+{   
+    I value[] = {LARFT_L3_DEFAULT};
+    I intervals[] = {LARFT_L3_INTERVALS_DEFAULT};
+    I max = LARFT_L3_NUM_INTERVALS_DEFAULT;
+
+    return value[get_index(intervals, max, dim)];
+}
+
+template <typename T, typename I, std::enable_if_t<std::is_same_v<T, rocblas_float_complex>, int> = 0>
```
tfalders, line 261: A short comment describing why this exception for float complex exists would be appreciated.

tfalders (commented):
I need to do a deep dive and remember how LARFT works, but I have some general comments right off the bat.

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: changed to reduce redundant reads/writes

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: done

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: added a comment

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: added a comment

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: done, please check if I did it correctly.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -4,6 +4,10 @@
  *     Univ. of Tennessee, Univ. of California Berkeley,
  *     Univ. of Colorado Denver and NAG Ltd..
  *     December 2016
+ * and
+ * Joffrain, Low, & Quintana-Ortí, et al. (2006). Accumulating householder
```
EdDAzevedo: Maybe to help the reader understand the code, just add a summary of the main idea. Maybe  state the  block Householder reflector  as something like  

 I - Y * T * Y', where T = inv(S), and  S + S' = U'*U,   T is upper triangular, so S is also upper triangular.
 
 Just a suggestion.

library/src/auxiliary/rocauxiliary_larft.hpp
EdDAzevedo: Do we need to perform  the index calculation as "j + i * int64_t(ldf)" and "i + j  * int64_t(ldv)" to avoid 32bit integer overflow?

AGonzales-amd: fixed using `idx2D`

qjojo (approved):

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -4,6 +4,10 @@
  *     Univ. of Tennessee, Univ. of California Berkeley,
  *     Univ. of Colorado Denver and NAG Ltd..
  *     December 2016
+ * and
+ *     Joffrain, Low, Quintana-Orti, et al. (2006). Accumulating householder
+ *     transformations, revisited.
+ *             ACM Transactions on Mathematical Software 32(2), p. 169-179.
```
tfalders: It would be nice to have this line indented to the same extent as the previous line, but this is just me being picky.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -51,7 +56,8 @@ ROCSOLVER_KERNEL void set_triangular(const rocblas_int n,
                                      const rocblas_int ldf,
                                      const rocblas_stride strideF,
                                      const rocblas_direct direct,
-                                     const rocblas_storev storev)
+                                     const rocblas_storev storev,
+                                     const bool inc)
```
tfalders: I would suggest renaming `inc`, since when I see it I think of "increment", which doesn't seem to be what it's doing.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -106,7 +150,8 @@ ROCSOLVER_KERNEL void set_triangular(const rocblas_int n,
...
```
tfalders: Same here

tfalders (approved):
Looks good! I have just a couple non-blocking comments. I'd also suggest updating the changelog.

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: done

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: done

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: done

PR #801, opened 2024-08-23
Add option to skip sparse on pull requests

cgmb:
The CI will build and test without rocSPARSE when the ci:no-sparse label is added to a pull request.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -9,7 +9,7 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false, boolean
     String hipClang = ''
     String debug = project.buildName.contains('Debug') ? '-g' : ''
     String centos = platform.jenkinsLabel.contains('centos') ? 'source scl_source enable devtoolset-7' : ''
-    String[] options = []
+    List<String> options = []
     Boolean withSparse = true
 
     if (env.BRANCH_NAME ==~ /PR-\d+/)
@@ -28,7 +28,7 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false, boolean
         }
     }
 
-    String[] getDeps = []
+    List<String> getDeps = []
     getDeps << auxiliary.getLibrary('rocBLAS', platform.jenkinsLabel, null, sameOrg)
     if (withSparse)
     {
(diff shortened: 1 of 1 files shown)


.jenkins/common.groovy
```diff
@@ -9,31 +9,39 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false, boolean
     String hipClang = ''
     String debug = project.buildName.contains('Debug') ? '-g' : ''
     String centos = platform.jenkinsLabel.contains('centos') ? 'source scl_source enable devtoolset-7' : ''
-    String noOptimizations = ''
+    String[] options = []
```
jmachado-amd: Small typo here? I think you mean to use `List<String> options = []` (and similarly `List<String> getDeps = []` below); lists have the left shift operator `<<` overloaded to append new elements out of the box, while array sizes should be immutable.

.jenkins/common.groovy
```diff
@@ -9,31 +9,39 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false, boolean
...
+        getDeps << auxiliary.getLibrary('rocSPARSE', platform.jenkinsLabel, null, sameOrg)
+    }
+    getDeps << auxiliary.getLibrary('rocPRIM', platform.jenkinsLabel, null, sameOrg)
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}
-                ${getRocBLAS}
-                ${getRocSPARSE}
-                ${getRocPRIM}
+                ${getDeps.join('\\n')}
```
jmachado-amd, line 43: This is just a heads-up in the case the builds keep failing. In my head `${getDeps.join('\\n')}` parses to something like `rocBLAS\nrocSPARSE\nrocPRIM`, and I don't see where the `\n`s are converted to new lines afterwards.

cgmb: It's nested within two strings. The `"""` string interprets the `\\` as `\` and the `'` interprets the `\n`.

jmachado-amd (approved):
Looks good to me!

tfalders (approved):

cgmb:
I'll do a quick check without the label and then merge.

PR #802, opened 2024-08-26
Changed .jenkins dependencies

amd-garydeng:


tfalders (approved):

.jenkins/debug.groovy
```diff
@@ -22,6 +22,8 @@ def runCI =
 
     // customize for project
     prj.paths.build_command = './install.sh -c'
+    prj.libraryDependencies = ['rocPRIM', 'hipBLAS-common', 'hipBLASLt', 'rocBLAS', 'rocSPARSE']
+
```
jmachado-amd, line 26: My apologies if I am missing something, but why set the (same) dependencies on all of the configurations instead of once and for all in file `common.groovy`?

amd-garydeng: This is how it's done for most libraries. However, the purpose is in case we need to test CUDA for hip libraries. The dependency lists are different.

cgmb:
We'll need to decide how to resolve the conflict between this change and #801.

.jenkins/debug.groovy
jmachado-amd: Thus, since rocSOLVER doesn't support CUDA, I understand that the repetition is not required. This is a minor point though, more importantly, this PR conflicts with #801.

jmachado-amd:
As far as I can tell, this PR is superseded by #801. Other than code style, the only difference I can spot is in the call to `getLibrary()` -- without knowing its definition I can't assess what the change from `null` to `'develop'` is accomplishing.

amd-garydeng:
#801 lgtm, the difference between null and develop is which branch the CI pipelines grabs the dependencies from. Null should be fine. Should I close this PR?

jmachado-amd:
Hi @amd-garydeng, please don't close it!

cgmb:
> without knowing its definition I can't assess what the change from `null` to `'develop'` is accomplishing.

It should be `null` to use the dependency from the corresponding branch. Otherwise, the math-ci will use dependencies from `develop` when testing rocSOLVER release branches. We changed it from develop to null in a cherry-pick for ROCm 5.2: e947ded181f4d77c06928dd9038989f76bb8fbd3.

jmachado-amd:
Thanks for clarifying!

cgmb:
Thanks for the help, @amd-garydeng. I incorporated this change into #801 and merged as 3830d1d877158373c4d3d09e34d5b9292de7241f.

PR #804, opened 2024-08-27
Add gfx1151 to default targets

amd-garydeng:


cgmb (approved):

cgmb:
Our debug builds are too big but that's already true on develop. We'll have to trim the gfx architectures included in our ASAN builds for ROCm 6.3. In ROCm 6.4, I expect that gfx1100, gfx1101, gfx1102 and gfx1151 will be replaced with gfx11-generic in rocSOLVER.

jmachado-amd (approved):

PR #810, opened 2024-08-29
Improve robustness of tests (part1)

jmachado-amd:
This PR adds the first part of a series of changes that are meant to improve the robustness of some eigensolvers' tests. It tackles the case in which a subset of the eigenvalues is requested.

The reference eigenvalues are still being computed with LAPACK; tests' inputs will be updated in a future PR.


clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -426,6 +454,12 @@ void sygvdx_hegvdx_getError(const rocblas_handle handle,
     *max_err = 0;
     for(rocblas_int b = 0; b < bc; ++b)
     {
+        // Capture failures where B is not positive definite (hInfo[b][0] > n),
+        // all other LAPACK failures abort the test.
+        if((hInfo[b][0] != 0) && (hInfo[b][0] <= n))
+        {
+            return;
+        }
```
tfalders: I would suggest moving this to after the comparison with hInfoRes. If B is not posdef we still want to ensure that our info matches LAPACK's.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -434,24 +468,30 @@ void sygvdx_hegvdx_getError(const rocblas_handle handle,
     // Check number of returned eigenvalues
     for(rocblas_int b = 0; b < bc; ++b)
     {
-        EXPECT_EQ(hNev[b][0], hNevRes[b][0]) << "where b = " << b;
-        if(hNev[b][0] != hNevRes[b][0])
+        auto hNevResMatch = clss[b](hW[b], hNev[b][0], hWRes[b], hNevRes[b][0], tols[b]);
+        /* clss[b].print_debug(); */
```
tfalders: This line can be removed

clients/common/misc/clss.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+    template <typename J, typename = typename std::enable_if<std::is_integral<std::decay_t<J>>::value>::type>
+    [[maybe_unused]] auto operator()(T* a, J size_a, T* b, J size_b, S tol)
+        -> /**! Size of subsequences */ I
+    {
+        return this->operator()(const_cast<T const*>(a), static_cast<I>(size_a),
+                                const_cast<T const*>(b), static_cast<I>(size_b), tol);
+    }
+
+    auto distance() -> S
```
tfalders, line 291: Personally, I prefer the traditional C style where the return type is first. Is there a specific reason that you've used the arrow notation throughout?

clients/common/misc/clss.hpp
```diff
@@ -0,0 +1,540 @@
...
+        std::lock_guard<std::mutex> lock(m_);
+
+        auto a = seq_a_.data();
+        auto b = seq_b_.data();
+
+        const auto default_precision{std::cout.precision()};
+        const auto digits
+            = static_cast<I>(tol_ > S(0) ? std::ceil(-std::min(std::log10(tol_), S(0))) + 2
+                                         : std::numeric_limits<T>::max_digits10);
+        std::cout << std::fixed << std::setprecision(digits);
```
tfalders: I believe there were concerns over thread safety when using multiple << operators with std::cout. Generally, we use fmt's print functionality or combine the multiple strings into one with fmt's format function.

You may also wish to take a look at how things are printed by common_host_helpers.hpp since that's not too different from what you're doing here.

clients/common/misc/clss.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+            std::cout << ":: :: a = {";
+            for(I i = 0; i < a_size; ++i)
+            {
+                std::cout << a[i];
+                if(i != a_size - 1)
+                {
+                    std::cout << ", ";
+                }
+                else
```
tfalders: This else statement is unnecessary. It can be removed.

clients/common/misc/clss.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+            std::cout << ":: :: b = {";
+            for(I i = 0; i < b_size; ++i)
+            {
+                std::cout << b[i];
+                if(i != b_size - 1)
+                {
+                    std::cout << ", ";
+                }
+                else
```
tfalders: Same here

tfalders (commented):
Here are some preliminary comments. In general, I don't fully understand what the closest_largest_subsequences class methods are doing, so additional code comments would be appreciated.

I should note that sygvdx still seems to be failing on the Windows build even after this change. Is that expected?

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: This is a fundamental design issue that this PR is trying to address: A disagreement between rocSOLVER and LAPACK does not imply that there is an error in rocSOLVER. 

The idea is to update this test again in a future PR to start with the eigenvalues instead of relying on LAPACK to compute them, rocSOLVER's `info` values will have to be checked directly. For the time being, this means ignoring tests in which LAPACK fails to converge, while using LAPACK for checks that we still don't have implemented (such as positiveness).

clients/common/misc/clss.hpp
jmachado-amd: And I prefer the functional style, where the return types appear last. The reason here is that it is much more legible to have verbose types and types with comments appear after the methods' names. (Generally speaking, the semantics of `auto` can be useful on their own, but those are not being used here.)

The methods with simple return types are using the functional style for consistency, but I don't have a strong opinion about this. If it is a major annoyance I can update the code towards a compromise.

clients/common/misc/clss.hpp
jmachado-amd: For context, modern C++ ostreams are thread safe, the concern is that they can produce interleaved output if accessed from different threads, while `libfmt`s `print` is by default atomic. However, `libfmt` doesn't fix the case of multi-threaded output in general because, even though each print is atomic, multiple prints can still appear out of order.

For the case of this piece of code: It is not meant to be used in rocSOLVER's output at all, this is only for debugging and for the unit tests of class `clss` that I have in a separate repository. Nonetheless, I did update it to print to a `stringstream` that can be subsequently directed to standard output with `libfmt` or anything else (similarly to how rocSOLVER's `options_description` parses its output).

jmachado-amd:
> Here are some preliminary comments. In general, I don't fully understand what the closest_largest_subsequences class methods are doing, so additional code comments would be appreciated.
> 
> I should note that sygvdx still seems to be failing on the Windows build even after this change. Is that expected?

Thanks for the comments! The two remaining test failures were due to accuracy issues unrelated to the original issues the PR was meant to tackle, but they are now fixed as well. I've included documentation, but I can prepare something more complete and share internally if you want me to clarify anything.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -206,6 +208,75 @@ void testing_sygvdx_hegvdx_bad_arg()
...
+// is defined, method `sygvdx_hegvdx_getError` will compute errors using the
+// legacy error bounds (for debugging purposes).
+//
+// Otherwise the new error bounds are always used.
+//
+static bool sygvdx_hegvdx_use_legacy_tests()
+{
+    bool status = false;
+    if(const char* env_var = std::getenv("ROCSOLVER_SYGVDX_HEGVDX_USE_LEGACY_TESTS");
+       env_var != nullptr)
```
cgmb: FYI, you can drop the `; env_var != nullptr` and it will have the same behaviour. Though, if you prefer to leave it using C++17 conditional initializers, that's fine too.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -206,6 +208,75 @@ void testing_sygvdx_hegvdx_bad_arg()
...
+//
+// a) If `ROCSOLVER_LAX_EXPERT_EIGENSOLVERS_TESTS` is defined, either as a
+// macro or as an environment variable, the test suite won't force the equality
+// of the number of computed eigenvalues with the number of reference
+// eigenvalues and will, instead, use a subset of the computed eigenvalues that
+// match the reference eigenvalues to the given tolerance; except
+//
+// b) If `ROCSOLVER_FULL_EXPERT_EIGENSOLVERS_TESTS` is defined as an
+// environment variable, then the tests will unconditionally check all
+// eigenvalues for their accuracy.
```
cgmb: These comments could be more concise, and it's probably not needed to describe the mechanics that can be understood by looking at the code just below. That said, I'm not sure it's worth spending the time editing them for length, and I do prefer to have comments to not having comments. So, nothing necessarily needs to be done here.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -206,6 +208,75 @@ void testing_sygvdx_hegvdx_bad_arg()
...
+//
+// The relaxed tests are intended as a means to decouple the computation of
+// error bounds of eigenvalues and eigenvectors, and allow all tests to pass in
+// the case that not all eigenvalues could be accurately computed, but all
+// accurate eigenvalues have accurate eigenvectors.  If eigenvectors are not
+// accurate the corresponding tests will fail both in full mode and in relaxed
+// mode.
+//
+// Note: the relaxed version of the tests is only supported when using the new
+// error bounds, see alseo function `sygvdx_hegvdx_use_legacy_tests()`.
```
cgmb: s/alseo/also/

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -206,6 +208,75 @@ void testing_sygvdx_hegvdx_bad_arg()
...
+
+//
+// The `sygvdx_hegvdx_getError` method default behaviour is to check if the
+// number of computed eigenvalues match the number of reference eigenvalues,
+// and then to check all computed eigenvalues for their accuracy, but this
+// behaviour can be relaxed.  This leads to two modes of operation: a relaxed
+// check, and the default (full) check; which can be controlled in the
+// following manner:
+//
+// a) If `ROCSOLVER_LAX_EXPERT_EIGENSOLVERS_TESTS` is defined, either as a
```
cgmb: I assume it's `EXPERT` because this option is "for experts only"? I don't think we want to have to start making those value judgements. All undocumented environment variables are going to be for experts only.

The variable is also easier to read as just `ROCSOLVER_LAX_EIGENSOLVER_TESTS`. Ditto for the other 'expert' variables.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -210,6 +208,37 @@ void testing_sygvdx_hegvdx_bad_arg()
     }
 }
 
+static bool test_for_equality_of_number_of_computed_eigenvalues()
+{
+    bool status = true;
+#if defined(ROCSOLVER_LAX_EXPERT_EIGENSOLVERS_TESTS)
+    status = false;
+#else
+    if(const char* env_var = std::getenv("ROCSOLVER_LAX_EXPERT_EIGENSOLVERS_TESTS");
```
EdDAzevedo: Maybe it is simpler to split the statements to separate out the declaration, say
{
const char * env_var = std::getenv(...);
if (env_var != nullptr) { status = false; }
}

Just a minor suggestion.

clients/common/misc/clss.hpp
```diff
@@ -0,0 +1,796 @@
...
+    // \param a:      pointer to first sequence, array of const T.
+    //
+    // \param size_a: number of elements in first sequence.
+    //
+    // \param b:      pointer to second sequence, array of const T.
+    //
+    // \param size_b: number of elements in second sequence.
+    //
+    // \return l^1 distance between subsequences (same value returned by method
+    // `clss::distance`).
```
tfalders: Do these operator() methods really return the same value as distance()? How can that be when distance() returns type S but operator() returns type I?

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -420,101 +562,222 @@ void sygvdx_hegvdx_getError(const rocblas_handle handle,
...
-            if(hInfo[b][0] == 0)
+            if(use_legacy_tests)
             {
-                err = norm_error('F', 1, hNev[b][0], 1, hW[b], hWRes[b]);
+                err = norm_error('F', 1, numMatchingEigs, 1, lapackEigs.data(), rocsolverEigs.data());
+                *max_err = err > *max_err ? err : *max_err;
+            }
+            else
+            {
+                /* err = clss[b].inf_norm() / std::max(norms[b], std::numeric_limits<S>::epsilon()); */
```
tfalders: Can this comment be removed?

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -420,101 +562,222 @@ void sygvdx_hegvdx_getError(const rocblas_handle handle,
...
 
                     // move B*x into hZRes
                     for(rocblas_int i = 0; i < n; i++)
-                        for(rocblas_int j = 0; j < hNev[b][0]; j++)
-                            hZRes[b][i + j * ldz] = hB[b][i + j * ldb];
+                        for(rocblas_int j = 0; j < numMatchingEigs; j++)
+                        {
+                            int jj = hWResIds[j]; // Id of rocSOLVER eigen-pair associated to j-th LAPACK eigen-pair
+                            hZRes[b][i + j * ldz] = hB[b][i + jj * ldb];
+                        }
```
tfalders: Minor style issue, but I like to bracket the outer loop if the inner loop is bracketed.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -420,101 +562,222 @@ void sygvdx_hegvdx_getError(const rocblas_handle handle,
...
                     {
-                        alpha = T(1) / hWRes[b][j];
-                        cpu_symv_hemv(uplo, n, alpha, A[b], lda, hB[b] + j * ldb, 1, beta,
+                        int jj = hWResIds[j]; // Id of rocSOLVER eigen-pair associated to j-th LAPACK eigen-pair
+                        alpha = T(1) / hWRes[b][jj];
+                        cpu_symv_hemv(uplo, n, alpha, A[b], lda, hB[b] + jj * ldb, 1, beta,
                                       hA[b] + j * lda, 1);
                     }
+                    // move hZRes
+                    for(rocblas_int i = 0; i < n; i++)
```
tfalders, line 682: Same here

tfalders (approved):
Looks good! I appreciate all the comments that have been added. It really helps to clarify what's going on.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: I know, it is a personal preference to try and make comparisons explicit instead of relying on implicit conversions to `bool`, but I would agree that sometimes this can create some clutter. I don't mind changing it, though.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: I agree that comments should not describe how the code works, but I do want to specify in this comment that `ROCSOLVER_LAX_EXPERT_EIGENSOLVERS_TESTS` can be either a macro or an environment variable, but `ROCSOLVER_FULL_EXPERT_EIGENSOLVERS_TESTS` must be defined as an environment variable.

It looks to me that the rest of the comment is just stating general behaviour, but I will try to improve this particular piece, nonetheless.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: The `EXPERT` is coming from LAPACK: For historical reasons `syevx`, the eigensolver using bisection (`stebz`) and inverse iteration (`stein`), is known as the expert driver to compute eigenvalues and eigenvectors -- because it allows for way more control than what is possible with the original, simple, driver `syev`.  Thus, all eigensolvers that use bisection and inverse iteration tend to be named with an `x` and are included in the expert driver family.

As those controls are meant to affect only tests of eigensolvers using the bisection (`stebz`) method, I included the `EXPERT` as part of their names.  However, I now see the possible misunderstanding that this choice can create. 

I would prefer to still have something to remind us that those variables target the expert drivers, but I will just remove the name `EXPERT` if I can't find a substitute I like. In the meantime, any suggestions?

clients/common/misc/clss.hpp
jmachado-amd: Ouch, those comments are wrong. The code eventually settled with the `operator()` methods returning the number of matching elements of the input sequences, but it seems that some comments still target the initial implementation.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
tfalders: We could perhaps replace `EXPERT` with `BISEC` (for bisection), but I'm also good with removing it entirely

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: Of course!

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: Done.

clients/common/misc/clss.hpp
jmachado-amd: The comments are now fixed.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: Updated.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: I ended up trimming the comment; the parts I wanted to keep didn't feel as important as they did before, so they are gone.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: I ended up just removing the `EXPERT`, but I am now feeling the urge to add the `BISEC` in its place. I will try to resist.

clients/common/lapack/testing_sygvdx_hegvdx.hpp
jmachado-amd: Updated.

clients/common/misc/clss.hpp
cgmb: My goal with solver log output was just to ensure that each line of output is atomic. You're right that lines from different threads may be interleaved, but even that relatively basic guarantee is a big step up in readability from having threads interleave mid-line.

We inherited a rather complex custom ostream from rocblas and rocfft that was designed for multi-threaded printing. I replaced it with fmt to avoid the need to maintain a logger with its own thread and mutexes. I could probably have just trained everyone to use stringstream appropriately, but I figured it would be easier to use fmt appropriately and transition to `std::format` and `std::print` once those become available.

I also had a long-term goal of introducing some more useful matrix views with formatting options, though it has never made it to the top of my priority pile. Increasingly, it seems like it might never. :(

cgmb (approved):

clients/common/misc/clss.hpp
jmachado-amd: Thanks for sharing the context behind those design choices! For what it is worth, given this knowledge, I would likely have made similar choices as you did; `gtest` itself does not provide proper output synchronization at the test level, and having to maintain a complex logging infrastructure just for the possibility that it might be useful in a distant future feels unnecessarily burdensome (even today).

I would also like to allow the tests to provide more useful information in the future, let's find a time to discuss what we can do together! :)

PR #814, opened 2024-09-04
Add gfx1151 support

amd-garydeng:
cherrypick gfx1151 commits into 6.3


jmachado-amd:
Hi @amd-garydeng, it looks like this is a duplicate of #812.

PR #815, opened 2024-09-04
Add option to disable rocBLAS in rocsolver_gemm/_trsm

AGonzales-amd:
This PR adds a build option to disable the use of rocBLAS gemm and trsm in the internal rocsolver_gemm and rocsolver_trsm blas functions respectively. In addition, the rocsolver_gemm function signature is updated to support arguments with mixed batched and strided pointers. The option is enabled by doing `./install.sh --cmake-arg -DDISBALE_ROCBLAS_IN_INTERNAL=ON`.


CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 option(ROCSOLVER_EMBED_FMT "Hide libfmt symbols" ON)
 option(OPTIMAL "Build specialized kernels for small matrix sizes" ON)
 option(ROCSOLVER_FIND_PACKAGE_LAPACK_CONFIG "Skip module mode search for LAPACK" ON)
+option(DISBALE_ROCBLAS_IN_INTERNAL "Disable the use of rocBLAS gemm and trsm in the internal rocSOLVER BLAS implementation." OFF)
```
cgmb: `DISABLE=ON` is always confusing, so let's use a positive phrasing and simplify the option name as much as we can. A lot of our options are prefixed with `BUILD`, so let's maybe try to do that too... particularly when we're controlling if something is built or not.

How about `BUILD_INTERNAL_BLAS=ON` / `BUILD_INTERNAL_BLAS=OFF`? The help text can probably be shortened to something like, "Use internal implementation of GEMM and TRSM".

library/src/CMakeLists.txt
```diff
@@ -445,6 +445,10 @@ if(OPTIMAL)
   target_compile_definitions(rocsolver PRIVATE OPTIMAL)
 endif()
 
+if(DISBALE_ROCBLAS_IN_INTERNAL)
+  target_compile_definitions(rocsolver PRIVATE DISBALE_ROCBLAS_IN_INTERNAL)
```
cgmb: I might suggest `USE_ROCBLAS_GEMM_TRSM` or `USE_ROCBLAS_GEMM USE_ROCBLAS_TRSM`.

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
```diff
@@ -677,7 +683,7 @@ I rocsolver_trsm_blksize(const I m, const I n)
         blk = size[get_index(intervalsM, M, m)][get_index(intervalsN, N, n)];
     }
 
-    if(blk == 1)
+    if(blk == 1 || (NO_ROCBLAS && blk == 0))
```
cgmb: What's the difference between the rocblas and rocsolver implementation that requires this?

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
```diff
@@ -31,6 +31,12 @@
 
 ROCSOLVER_BEGIN_NAMESPACE
 
+#ifdef DISBALE_ROCBLAS_IN_INTERNAL
+#define NO_ROCBLAS 1
```
cgmb: I would again suggest phrasing the variable in the positive sense. e.g., `#define ROCSOLVER_INTERNAL_TRSM 1`.

cgmb (commented):
Just some suggestions on naming.

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
AGonzales-amd: As I understand it, If `blk == 0` then rocBLAS is called normally, but if rocBLAS is disabled then we fallback to the rocsolver implementation. Therefore, we need to set `blk` to a non-zero value in this case.

CMakeLists.txt
jmachado-amd: +1 on using positive naming conventions. But, in what concerns the use of the `BUILD` prefix, I understand that our internal BLAS would still be built and used regardless of whether `BUILD_INTERNAL_BLAS` is set to `OFF` (`rocBLAS` is called only for a subset of the possible input configurations), which might be confusing.

I would suggest to explicitly state that the option to force the use of our internal BLAS implementation everywhere is only meant for debugging purposes, and I would also suggest using the typical `cmake` etiquette of prefixing options with the project name (`ROCSOLVER` in this case), so `ROCSOLVER_USE_INTERNAL_BLAS_FOR_DEBUG` would tick all boxes, but anything on those lines would work.

AGonzales-amd: Sorry for the delay, as suggested, I updated the option and macro names to:
 - `ROCSOLVER_USE_INTERNAL_BLAS_FOR_DEBUG`
 - `USE_INTERNAL_GEMM USE_INTERNAL_TRSM`
 - `ROCSOLVER_INTERNAL_TRSM`

library/src/CMakeLists.txt
AGonzales-amd: updated

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
AGonzales-amd: updated

CMakeLists.txt
cgmb: I'd mildly prefer `ROCSOLVER_USE_INTERNAL_BLAS` and just mention that its intended for debug in the help text describing the option. Mostly to keep it short, but also because these sorts of options are likely to find uses beyond just debugging.

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -182,10 +182,18 @@ rocblas_status rocsolver_gemm(rocblas_handle handle,
...
     if(inca == 1 && incb == 1 && incc == 1)
         return rocblasCall_gemm(handle, transA, transB, m, n, k, alpha, A, shiftA, lda, strideA, B,
                                 shiftB, ldb, strideB, beta, C, shiftC, ldc, strideC, batch_count,
                                 work);
+#endif
+
+    if constexpr (!rocblas_is_complex<T>)
+    {
+        transA = (transA == rocblas_operation_conjugate_transpose ? rocblas_operation_transpose : transA);
+        transB = (transB == rocblas_operation_conjugate_transpose ? rocblas_operation_transpose : transB);
```
cgmb: We should make a helper function for this at some point.

cgmb (approved):
I still had a minor suggestion or two, but it looks good to me.

CMakeLists.txt
jmachado-amd: I have to agree that Cory's suggestion: `ROCSOLVER_USE_INTERNAL_BLAS` is better than mine.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2112,9 +2113,9 @@ void local_gemm(rocblas_handle handle,
                             strideA, work, rocblas_fill_full);
 
     // temp = work*B
-    rocblasCall_gemm(handle, rocblas_operation_none, rocblas_operation_none, n, n, n, &one, work,
-                     shiftT, ldt, strideT, B, shiftT, ldt, strideT, &zero, temp, shiftT, ldt,
-                     strideT, batch_count, workArr);
+    rocsolver_gemm<false, BATCHED || STRIDED, S>(
```
jmachado-amd: Since `BATCHED || STRIDED` is not doing anything in `rocsolver_gemm` I would just have written `rocsolver_gemm<false, false, S>` to match the provided template instantiations, but I guess that there is no good approach here.

At some point we should revisit `rocsolver_gemm`'s signature.

jmachado-amd (commented):
This PR reads well and does what it is meant to do, thanks Angelo!

Yet, we should discuss the missing complex cases and the fact that the simple gemm kernel is not doing any check on whether its input matrices or batch size are too big for the `x`, `y`, `z` dimensions. Of course, those issues were not introduced here but they must be addressed for our internal blas to achieve its full usefulness.

Ps.: @AGonzales-amd, I think that Cory has requested some small changes as well.

library/src/auxiliary/rocauxiliary_stedc.hpp
AGonzales-amd: I set it to `BATCHED || STRIDED` because the caller passes strided batched arrays as arguments if the original is batched or strided batched. I just thought it made sense to do it this way in case the template parameters are used internally in the future.

CMakeLists.txt
AGonzales-amd: Sorry, I missed this. It's updated now.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: IMHO, I think it would be nice if the rocsolver_gemm (or a wrapper) has the interface that is the same as rocblasCall_gemm() to reduce the opportunity for inadvertently introduce a bug by mistake.  Just a thought.

AGonzales-amd: The two interfaces differ only with the two bool template parameters and they are not used internally atm. So I think there's only a very small chance of introducing a bug related to this.

jmachado-amd (approved):

library/src/lapack/roclapack_gebrd.hpp
```diff
@@ -159,17 +160,17 @@ rocblas_status rocsolver_gebrd_template(rocblas_handle handle,
...
-                         shiftA + idx2D(j + jb, j + jb, lda), lda, strideA, batch_count,
-                         (T**)work_workArr);
-
-        rocblasCall_gemm(handle, rocblas_operation_none, rocblas_operation_none, m - j - jb,
-                         n - j - jb, jb, &minone, X, shiftX + jb, ldx, strideX, A,
-                         shiftA + idx2D(j, j + jb, lda), lda, strideA, &one, A,
-                         shiftA + idx2D(j + jb, j + jb, lda), lda, strideA, batch_count,
-                         (T**)work_workArr);
+        rocsolver_gemm<BATCHED, STRIDED, T>(
+            handle, rocblas_operation_none, rocblas_operation_conjugate_transpose, m - j - jb,
```
tfalders: I don't think we should apply these changes to gemm calls that use or may use conjugate transpose. They can be switched over once the internal kernels for conjugate transpose are in place.

library/src/specialized/roclapack_gemm_specialized_kernels_c.cpp
```diff
@@ -33,15 +33,139 @@ ROCSOLVER_BEGIN_NAMESPACE
...
+                 rocblas_float_complex*,
+                 rocblas_float_complex*,
+                 rocblas_float_complex*);
+INSTANTIATE_GEMM(0,
+                 1,
+                 rocblas_float_complex,
+                 rocblas_int,
+                 rocblas_float_complex*,
+                 rocblas_float_complex*,
+                 rocblas_float_complex*);
```
tfalders: Removing the BATCHED and STRIDED template arguments will be nice since we can collapse these two together.

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -182,10 +182,20 @@ rocblas_status rocsolver_gemm(rocblas_handle handle,
...
                                 work);
+#endif
+
+    if constexpr(!rocblas_is_complex<T>)
+    {
+        transA = (transA == rocblas_operation_conjugate_transpose ? rocblas_operation_transpose
+                                                                  : transA);
+        transB = (transB == rocblas_operation_conjugate_transpose ? rocblas_operation_transpose
+                                                                  : transB);
+    }
```
tfalders: I would suggest removing this and instead adjusting the code below so that the function doesn't immediately return when T is a real type.

tfalders (commented):
Looks good so far, @AGonzales-amd! I think we might want to avoid using rocsolver_gemm for some functions until the conjugate transpose kernels are ready, as otherwise we may get test failures.

library/src/lapack/roclapack_gebrd.hpp
jmachado-amd: @tfalders my understanding is that @AGonzales-amd will work on updating the internal gemm to add the conjugate transpose cases immediately after this PR. Isn't it better to just leave the code as it is, and just proceed with updating gemm?

tfalders: I don't like to leave things that we know are broken in the code, but I suppose you do have a point. We could also just ask @AGonzales-amd to add the kernels with this PR.

AGonzales-amd: I tried removing the calls to rocsolver_gemm when conjugate transpose is used but some tests are still failing with the flag enabled because of rocsolver_trsm. I don't want to change the existing calls to rocsolver_trsm so I would be okay with including the kernels in this PR to resolve the failures.

library/src/specialized/roclapack_gemm_specialized_kernels_c.cpp
AGonzales-amd: fixed

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
AGonzales-amd: fixed

library/src/lapack/roclapack_gebrd.hpp
jmachado-amd: Would you mind clarifying what is missing in trsm?

AGonzales-amd: rocsolver_trsm uses rocsolver_gemm, so if rocblas is disabled and conjugate transpose gemm is used then rocsolver_trsm fails.

AGonzales-amd: I added the support for conjugate transpose in `rocsolver_gemm`. @tfalders you mentioned there were difficulties in supporting interleaved types. The changes I've made are quite simple so please check if I've missed something.

tfalders (approved):
Nice job! LGTM

PR #819, opened 2024-09-11
Prevent bdsqr_estimate from rounding t to zero

tfalders:
In `bdsqr_estimate`, the tolerance used to decide that elements of E have converged is modified by the value `t`, which changes as the loop traverses elements of D and E. Thanks to the order of operations used to calculate it, the updated `t` may be rounded down to zero in some cases, which brings the tolerance down to zero and prevents elements of E from being considered to have converged.

### DIFF
--- a/clients/common/lapack/testing_gesvd.hpp
+++ b/clients/common/lapack/testing_gesvd.hpp
@@ -206,7 +206,8 @@ void gesvd_initData(const rocblas_handle handle,
                     const rocblas_int bc,
                     Th& hA,
                     std::vector<T>& A,
-                    bool test = true)
+                    const bool test,
+                    const bool singular)
 {
     if(CPU)
     {
@@ -214,15 +215,29 @@ void gesvd_initData(const rocblas_handle handle,
 
         for(rocblas_int b = 0; b < bc; ++b)
         {
-            // scale A to avoid singularities
-            for(rocblas_int i = 0; i < m; i++)
+            if(!singular)
             {
-                for(rocblas_int j = 0; j < n; j++)
+                // scale A to avoid singularities
+                for(rocblas_int i = 0; i < m; i++)
                 {
-                    if(i == j)
-                        hA[b][i + j * lda] += 400;
-                    else
-                        hA[b][i + j * lda] -= 4;
+                    for(rocblas_int j = 0; j < n; j++)
+                    {
+                        if(i == j)
+                            hA[b][i + j * lda] += 400;
+                        else
+                            hA[b][i + j * lda] -= 4;
+                    }
+                }
+            }
+            else
+            {
+                // form a singular matrix consisting of all ones
+                for(rocblas_int i = 0; i < m; i++)
+                {
+                    for(rocblas_int j = 0; j < n; j++)
+                    {
+                        hA[b][i + j * lda] = 1;
+                    }
                 }
             }
 
@@ -289,7 +304,8 @@ void gesvd_getError(const rocblas_handle handle,
                     Ih& hinfo,
                     Ih& hinfoRes,
                     double* max_err,
-                    double* max_errv)
+                    double* max_errv,
+                    const bool singular)
 {
     using W = decltype(std::real(T{}));
 
@@ -300,7 +316,8 @@ void gesvd_getError(const rocblas_handle handle,
     std::vector<T> A(lda * n * bc);
 
     // input data initialization
-    gesvd_initData<true, true, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A);
+    gesvd_initData<true, true, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A, true,
+                                  singular);
 
     // execute computations:
     // complementary execution to compute all singular vectors if needed (always in-place to ensure
@@ -315,7 +332,8 @@ void gesvd_getError(const rocblas_handle handle,
     if(right_svect == rocblas_svect_none && left_svect != rocblas_svect_none)
         CHECK_HIP_ERROR(Vres.transfer_from(dVT));
 
-    gesvd_initData<false, true, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A);
+    gesvd_initData<false, true, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A, true,
+                                   singular);
 
     // CPU lapack
     for(rocblas_int b = 0; b < bc; ++b)
@@ -436,7 +454,8 @@ void gesvd_getPerfData(const rocblas_handle handle,
                        const rocblas_int hot_calls,
                        const int profile,
                        const bool profile_kernels,
-                       const bool perf)
+                       const bool perf,
+                       const bool singular)
 {
     using W = decltype(std::real(T{}));
 
@@ -448,7 +467,8 @@ void gesvd_getPerfData(const rocblas_handle handle,
 
     if(!perf)
     {
-        gesvd_initData<true, false, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A, 0);
+        gesvd_initData<true, false, T>(handle, left_svect, right_svect, m, n, dA, lda, bc, hA, A,
+                                       false, singular);
 
         // cpu-lapack performance (only if not in perf mode)
         *cpu_time_used = get_time_us_no_sync();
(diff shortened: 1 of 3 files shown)


EdDAzevedo (approved):

jmachado-amd (approved):

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -64,7 +64,7 @@ __device__ T bdsqr_estimate(const rocblas_int n, T* D, T* E, int t2b, T tol, int
             smin = -1;
             break;
         }
-        t = std::abs(D[jd]) * t / (t + std::abs(E[je]));
+        t = std::abs(D[jd]) / (t + std::abs(E[je])) * t;
```
cgmb: Could you add a comment to mention this? The importance of this order of operations is not obvious just by looking at the code.

cgmb (commented):
Is there a good way to add a test for this?

library/src/auxiliary/rocauxiliary_bdsqr.hpp
tfalders: Certainly, I can do that

tfalders:
> Is there a good way to add a test for this?

We can trigger the defect using a 15x15 matrix of all ones, but since our test matrices are currently all random, it might be somewhat difficult to add a test in an elegant way. Definitely possible, just... probably a bit hacky.

clients/common/lapack/testing_gesvd.hpp
```diff
@@ -206,23 +206,38 @@ void gesvd_initData(const rocblas_handle handle,
                     const rocblas_int bc,
                     Th& hA,
                     std::vector<T>& A,
-                    bool test = true)
+                    const bool test,
+                    const bool singular)
```
EdDAzevedo, line 210: Just a thought that (in future) we might want the test code  to generate  a rank-k matrix.  

The SVD of A may be viewed as   A = sigma(1) * U(1) * V(1)' + sigma(2) * U(2) * V(2)' + ... + sigma(k) * U(k) * V(k)', where  sigma(:) are the singular values, U(:) and V(:) are the orthogonal singular vectors.  

The matrix of all ones  is a special case for U = (1/sqrt(m)) * ones(m,1), sigma(1) = sqrt(m*n), V = (1/sqrt(n)) * ones(n,1).   

Perhaps it is possible to generalize by say generating   a random(n,k) vector for U, random(n,k) vector for V and form rank-k  matrix A = U * V'.  The other (n-k) left singular vectors just need to be orthogonal to span(U) or (n-k) right singular vectors orthogonal to span(V).     

If the U(:) and V(:) vectors are orthogonal and generate positive values for sigma(:), then these can be the pre-determined singular vectors.

Just a suggestion.

EdDAzevedo (approved):

clients/common/lapack/testing_gesvd.hpp
tfalders: Yeah, the test for GESVD is quite messy since we have to figure out what the singular vectors are from the random matrix. It would be nice to have pre-determined singular vectors, though I think there are cases when the singular vectors are not unique (i.e. when we have repeated singular values?) and I don't know a good way to work around that.

EdDAzevedo: Yes, for a rank-k matrix, there are  min(m,n) - k,  singular values that are zero.  Those singular vectors need only to be orthogonal to Span(U)  (and similarly Span(V)). We can consider checking  for orthogonality within the space  for exactly same singular values,  and orthorgonal across span of other singular vectors of distinct different singular values.

cgmb:
> We can trigger the defect using a 15x15 matrix of all ones, but since our test matrices are currently all random, it might be somewhat difficult to add a test in an elegant way. Definitely possible, just... probably a bit hacky.

I'd be fine with just the comment if we didn't have proper infrastructure for this, but it looks like you've found a reasonable approach.

cgmb (approved):

cgmb:
@tfalders, when you squash and merge, please edit the commit message to include the description you wrote for the PR ("In bdsqr_estimate..."). It's a great description and it should be included in the commit message to ensure that it is preserved in source control.

jmachado-amd:
@tfalders , the updated test looks good to me! I have a question regarding the original design of the algorithm, but it should not prevent you from merging this PR as is if it meant to be cherry-picked into 6.3.

@jzuniga-amd, now that I am having a second look at it, I am under the impression that method `bdsqr_estimate` is doing only half of the work that it was supposed to.

I understand that `t` is meant to be a lower bound for the smallest singular value of the sub-block, but the way I know to prove this (e.g., see https://netlib.org/lapack/lawnspdf/lawn03.pdf, Lemma 1, pgs. 5,6) requires estimating both the `l_1` and the `l_\infty` norms of the inverse of the sub-block matrix and this method is estimating either one or the other.

Would you mind sharing a reference to the original reference this code is based on?

clients/common/lapack/testing_gesvd.hpp
jmachado-amd: @tfalders, next time we meet, please let me know how you intend to use pre-determined singular vectors in `bdsqr`'s tests!

library/src/auxiliary/rocauxiliary_bdsqr.hpp
jmachado-amd: No need to change anything at this point, but there is an important asymmetry in those operations. `t` is, morally speaking, a lower bound to the smallest singular value of the matrix sub-block, thus we have to be concerned with it underflowing (which this PR is tackling) but not so much with it overflowing.

The generic solution, to write: `t = std::abs(D[jd]) * (t / (t + std::abs(E[je])));` is not supposed to add much more precision here.

jzuniga-amd:
> Lemma 1, pgs. 5,6) requires estimating both the `l_1` and the `l_\infty` norms of the inverse of the sub-block matrix and this method is estimating either one or the other.

Correct, `bdsqr_estimate` estimates either one or the other because it is called with a direction: top_to_botton (`l_1`) or bottom_to_top (`l_\inf`).

It is just a rough estimate. It would not hurt to take the minimum of `l_1` and `l_\inf` but it is not necessary. These estimates are used to calculate the shifts for the qr steps (and these steps always run in the indicated direction according to which element at the ends of the diagonal, d_1 or d_n, is larger), so after a few iterations you know which sequence, `l_1` or `l_\inf`, has larger/smaller value.

jzuniga-amd:
Along the same line of thought. The purpose of the loop in `bdsqr_estimate` is to update `t` with the relative effect of the elements in the sub-diagonal, i.e. `t/(t+e)`
So, the way it was before looks better (more clear what the purpose is). Maybe it just needs an extra pair of parenthesis to force the right order of operations. 

@tfalders would you mind trying 
`t = std::abs(D[jd]) * ( t / (t + std::abs(E[je])) );`
to see if it has the same effect ?

clients/common/lapack/testing_gesvd.hpp
tfalders: @jmachado-amd I have no plans at this time to use pre-determined singular vectors in the tests, as it would be a difficult change and rather outside the scope of this PR. I was just noting that it would be a welcome change if we could manage it.

EdDAzevedo:
> Along the same line of thought. The purpose of the loop in `bdsqr_estimate` is to update `t` with the relative effect of the elements in the sub-diagonal, i.e. `t/(t+e)` So, the way it was before looks better (more clear what the purpose is). Maybe it just needs an extra pair of parenthesis to force the right order of operations.
> 
> @tfalders would you mind trying `t = std::abs(D[jd]) * ( t / (t + std::abs(E[je])) );` to see if it has the same effect ?

One may consider adding a temporary variable (as a sequence point) to guarantee the order of evaluation, say,   something like  
auto const ratio = t / (t + std::abs(E[je]));  
t = std::abs(D[jd]) * ratio;

clients/common/lapack/testing_gesvd.hpp
jmachado-amd: Absolutely! I was just asking _how_ you intended to use them; I am not saying that you _should_ use them.

jmachado-amd (approved):

tfalders:
Test failure on gfx908 appears unrelated, so I will force the merge.

PR #821, opened 2024-09-13
Add gfx1151 target

amd-garydeng:


jmachado-amd (approved):

jmachado-amd:
@jzuniga-amd what is the practice regarding point releases, i.e., shall we update rocSOLVER's revision number for the 6.2.4 release?

tfalders (approved):

cgmb:
> shall we update rocSOLVER's revision number for the 6.2.4 release?

Yes. We should update the changelog and version number, as otherwise the release notes for the rocm-6.2.4 tag will say that there were no changes in rocSOLVER in that release. (Much like the [rocm-6.2.1](https://github.com/ROCm/rocSOLVER/releases/tag/rocm-6.2.1) tag does.)

If the commit hash changes between releases, there should be a new version number and a changelog entry.

CHANGELOG.md
```diff
@@ -27,6 +27,12 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 
 ### Fixed
 - Fixed potential accuracy degradation in SYEVJ/HEEVJ for inputs with small eigenvalues.
+- Fixed synchronization issue in STEIN.
```
cgmb: [This change was not in ROCm 6.2.0](https://github.com/ROCm/rocSOLVER/blob/rocm-6.2.0/library/src/auxiliary/rocauxiliary_stein.hpp#L240-L241) and should therefore not be in the ROCm 6.2.0 release notes. Please create a new section with the title rocSOLVER 3.26.2 for ROCm 6.2.4.

We should also add a changelog with rocSOLVER 3.26.1 for ROCm 6.2.1 that describes [the documentation changes](https://github.com/ROCm/rocSOLVER/compare/rocm-6.2.0...rocm-6.2.1) in that release and include the known issue that it was released with the version number 3.26.0.

_Every tag with a new commit hash needs a unique version number and a corresponding changelog entry._

cgmb (changes requested):

cgmb (approved):
Thanks. That looks good to me. There's no real way to fix the mistake in the 6.2.1 versioning, so we don't have to handle that as part of this change.

PR #826, opened 2024-09-30
Potential improvement to set/restore_diag in GEQR2

AGonzales-amd:
This PR aims to reduce the impact of `set_diag` and `restore_diag` kernels to the runtime of GEQR2 indicated by profiling. This is achieved by:
1. Combining `larfg` and `set_diag` to reduce the number of global memory reads and writes:
   - This is achieved by modifying `larfg` to write both the unit diagonal and non-unit diagonal values thus eliminating the call to `set_diag`.
2. Reduce kernel launch overhead of `set_diag` and `restore_diag`:
   - `set_diag` is explained above. Launch overhead of `restore_diag` is reduced by launching the kernel once to restore all diagonal values at the expense of additional memory footprint.

The following chart shows the speedup of `geqrf` with these changes on real single precision square matrices.
log_compare_sgeqrf_m

Note:
- I tried the suggestion of using `larfb` instead of `larf` but it performed worse due to increased global memory access. I got similar results with my attempt to modify `larf` to assume implicit unit diagonal.
- This is my attempt of a solution to this problem and I am open to try other suggestions.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_larfg.hpp
+++ b/library/src/auxiliary/rocauxiliary_larfg.hpp
@@ -38,8 +38,8 @@
 
 ROCSOLVER_BEGIN_NAMESPACE
 
-template <typename T, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
-__device__ void run_set_taubeta(T* tau, T* norms, T* alpha)
+template <bool IMPLICIT_UNIT, typename T, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
+__device__ void run_set_taubeta(T* tau, T* norms, T* alpha, T* beta)
 {
     if(norms[0] > 0)
     {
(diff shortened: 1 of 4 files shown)


library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -183,20 +219,23 @@ rocblas_status
...
-                                        const rocblas_stride shiftx,
-                                        const I incx,
-                                        const rocblas_stride stridex,
-                                        T* tau,
-                                        const rocblas_stride strideP,
-                                        const I batch_count,
-                                        T* work,
-                                        T* norms)
+template <typename T, typename I, typename U, typename UB, bool COMPLEX = rocblas_is_complex<T>>
+rocblas_status rocsolver_larfg_general_template(rocblas_handle handle,
```
tfalders: I believe you can still call this rocsolver_larfg_template, and would prefer that as well.

tfalders (approved):
LGTM! Just one change request

library/src/auxiliary/rocauxiliary_larfg.hpp
AGonzales-amd: Fixed

library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -38,9 +38,10 @@
...
 ROCSOLVER_BEGIN_NAMESPACE
 
-template <bool IMPLICIT_UNIT, typename T, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
+template <typename T, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
 __device__ void run_set_taubeta(T* tau, T* norms, T* alpha, T* beta)
 {
+    const auto ignore_beta = (beta == nullptr);
     if(norms[0] > 0)
     {
         T n = sqrt(norms[0] + alpha[0] * alpha[0]);
```
EdDAzevedo, line 47: Minor comment: the variable "n" may be misunderstood to be an integer. Perhaps rename to say "dnorm" or something else to avoid confusion.

qjojo (approved):

PR #764, opened 2024-07-13
Fix release staging/rocm rel 6.1 to build with libfmt 11.0.1

perestoronin:
Fix to build with libfmt 11.0.1

PR #828, opened 2024-10-03
Fix libfmt build errors

jmachado-amd:
This is a hotfix to allow rocSOLVER to build with newer versions of libfmt.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -617,7 +617,8 @@ try
     // print help message
     if(vm.count("help"))
     {
-        fmt::print("{}{}\n", help_str, desc);
+        /* fmt::print("{}{}\n", help_str, desc); */
+        std::cout << help_str << desc << std::endl;
         return 0;
     }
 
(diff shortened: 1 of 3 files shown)


clients/benchmarks/client.cpp
```diff
@@ -617,7 +617,8 @@ try
     // print help message
     if(vm.count("help"))
     {
-        fmt::print("{}{}\n", help_str, desc);
+        /* fmt::print("{}{}\n", help_str, desc); */
+        std::cout << help_str << desc << std::endl;
```
tfalders: Why have you replaced fmt::print with an std::cout print? The other files seem to be using fmt::print okay. Maybe you need to have stderr as the first argument?

tfalders (approved):
Just one comment

clients/benchmarks/client.cpp
jmachado-amd: This line fails to compile because `desc` is an object of class `options_description`, which lacks a formatter. Adding one might be simple, but without having time to match rocSOLVER's design (which I would have to investigate), I preferred to proceed with the smallest set of changes that fixed the build failures; thus, I followed the fix that appears in #764.

I can update it to use `fmt` instead of `std::cout`, but we should add a proper formatter in the future.

qjojo (approved):

PR #830, opened 2024-10-07
Fix libfmt build errors (#828)

jmachado-amd:
* Fix build errors with libfmt 10.2.1

* Use `fmt::print` instead of `std::cout`

(cherry picked from commit e5e935060622fcd15e6e80dad877662f2587571f)

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -617,7 +617,9 @@ try
     // print help message
     if(vm.count("help"))
     {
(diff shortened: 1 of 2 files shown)


tfalders (approved):
The change for d_vector isn't included?

jmachado-amd:
> The change for d_vector isn't included?

No, as it turns out, the changes to `d_vector` that broke the compilation were introduced in 6.3, not 6.2.

PR #831, opened 2024-10-07
Fix libfmt build errors (#828)

jmachado-amd:
* Fix build errors with libfmt 10.2.1

* Use `fmt::print` instead of `std::cout`

(cherry picked from commit e5e935060622fcd15e6e80dad877662f2587571f)


tfalders (approved):

cgmb (approved):
Thanks for handling this.

PR #833, opened 2024-10-09
Convert change log to new format

amd-jnovotny:
These changes convert the changelog into the new standard format with the required headings and bullet format. I've also made some minor copy edits.


jmachado-amd (approved):

tfalders (approved):

tfalders:
It looks like only 6.3 is using asterisks instead of dashes. It's not strictly necessary, but it would be nice for 6.1 and 6.2 to have the same format.

amd-jnovotny:
@tfalders : I didn't change that part but only made sure the headings aligned. Maybe I can change that in 0.0 only (here and in hipSOLVER) as that is the branch people are most likely to look at.

cgmb (approved):

PR #834, opened 2024-10-09
Convert change log to new format

amd-jnovotny:
These changes convert the changelog into the new standard format with the required headings and bullet format. I've also made some minor copy edits.


jmachado-amd (approved):
This looks good to me, but I wonder whether we should include the new supported architectures here as well.

tfalders (approved):

cgmb (approved):

PR #836, opened 2024-10-09
Investigate gesvd test failures in Windows using internal gemms

jmachado-amd:
I am seeing test failures in `gesvd` in Windows, when I link rocSOLVER to a version of rocBLAS compiled without Tensile (either if using rocBLAS's gemm or rocSOLVER's internal gemm). The purpose of this PR is to investigate this issue.


TorreZuk:
If you tell me the failing size to gemm, you or I can test locally to confirm the sizes pass on windows rocblas.

jmachado-amd:
Hi @TorreZuk, I don't think the issue I am assessing here is related to rocBLAS, as I can reproduce the errors in my workstation if I use rocSOLVER's internal gemm implementation. However, it's likely that there are other windows issues that might be stemming from rocblas, and I'll get in touch with you soon to chat about those.

EDIT: updated the PR description to clarify that the tests also fail with rocSOLVER's internal gemm.

PR #840, opened 2024-10-23
Fix static builds

qjojo:
Adds the preprocessor flag ROCSOLVER_STATIC_LIB to check if we're building as a static library or not, and uses it to prevent trying to use dlopen() and dlsym() if so.

### DIFF
--- a/library/src/refact/rocsolver_rfinfo.cpp
+++ b/library/src/refact/rocsolver_rfinfo.cpp
@@ -67,41 +67,42 @@ ROCSOLVER_BEGIN_NAMESPACE
 template <typename Fn>
 static bool load_function(void* handle, const char* symbol, Fn& fn)
 {
-#if defined(_WIN32) && !defined(ROCSOLVER_STATIC_LIB)
+#ifndef ROCSOLVER_STATIC_LIB
+#ifdef _WIN32
     fn = (Fn)(GetProcAddress((HMODULE)handle, symbol));
     bool err = !fn;
-#elif !defined(ROCSOLVER_STATIC_LIB)
+#else
     fn = (Fn)(dlsym(handle, symbol));
     char* err = dlerror(); // clear errors
 #ifndef NDEBUG
     if(err)
         fmt::print(stderr, "rocsolver: error loading {:s}: {:s}\n", symbol, err);
-#endif
-#else
+#endif /* NDEBUG */
     bool err = false;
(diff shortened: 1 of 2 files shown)


tfalders:
Do we need to target this to 6.3, or are we planning to do a cherry-pick?

qjojo:
I was going to cherry pick it but I can open a second PR against 6.3 if that's preferred.

library/src/refact/rocsolver_rfinfo.cpp
```diff
@@ -67,36 +67,41 @@ ROCSOLVER_BEGIN_NAMESPACE
...
 #ifndef NDEBUG
     if(!handle)
         fmt::print(stderr, "rocsolver: error loading librocsparse.so.1: {:s}\n", err);
-#endif
-#endif /* _WIN32 */
+#endif /* NDEBUG */
+#endif /* defined(_WIN32) && !defined(ROCSOLVER_BUILD_STATIC) */
+#if defined(ROCSOLVER_STATIC_LIB)
+    return false; // nothing to load, static builds don't support sparse
+#else
```
TorreZuk: Would be simpler to just start off with this section for both win/linux if you don't support sparse in static.

library/src/refact/rocsolver_rfinfo.cpp
```diff
@@ -67,36 +67,41 @@ ROCSOLVER_BEGIN_NAMESPACE
 template <typename Fn>
 static bool load_function(void* handle, const char* symbol, Fn& fn)
 {
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(ROCSOLVER_STATIC_LIB)
```
TorreZuk: This is overly complex IMO, is static just returning false?  Do this upfront (same as other function suggestion).

tfalders (approved):
Looks good! Though I think I would prefer if the macros looked more like this:

```
#ifndef ROCSOLVER_STATIC_LIBS
#ifdef _WIN32
...
#else
...
#endif /* _WIN32 */
#endif /* ROCSOLVER_STATIC_LIB */
```

TorreZuk (approved):
I think Troy also suggests trying to make it simpler.  Fine if it builds correctly in static docker that was failing.   Assume a static sparse library gets consumed if available.

jzuniga-amd (approved):

qjojo:
The test failures are the known issues with sygvdx that are unrelated to this fix. Merging.

PR #842, opened 2024-10-30
Fix static builds (#840)

TorreZuk:
* don't try to use dlsym or dlopen in static builds

(cherry picked from commit 1fa9dcf6177e60e28959889f2153d8dc292587ec)

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -436,9 +436,10 @@ if (BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)
   )
 endif( )
 
-# Following Boost conventions of prefixing 'lib' on static built libraries, across all platforms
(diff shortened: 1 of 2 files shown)


jzuniga-amd (approved):

tfalders (approved):

qjojo (approved):

jzuniga-amd:
<s>oh, I realized this is directed to `release/rocm-rel-6.3` did we branch already? or should this go to `release-staging/rocm-rel-6.3`?

If branching is done, then we have to correct `release/rocm-rel-6.3` branch first. Somebody created it from develop not from release-staging so it has the versions bumped for 6.4 and it does not have all hot fixes applied to release-staging</s>

Nevermind, release and release-staging branches look the same now.

PR #845, opened 2024-11-05
Add LASR API 

jzuniga-amd:
This PR adds LASR as a new routine to the public API.
LASR applies a sequence of plane rotations defined by their sine and cosine numbers to a given matrix.
(Following PRs will adapt existing methods that use LASR functionality to call the new API)

### DIFF
--- a/docs/reference/auxiliary.rst
+++ b/docs/reference/auxiliary.rst
@@ -13,6 +13,7 @@ The auxiliary functions are divided into the following categories:
 
 * :ref:`vecmat`. Some basic operations with vectors and matrices that are not part of the BLAS standard.
 * :ref:`householder`. Generation and application of Householder matrices.
+* :ref:`rotations`. Generation and application of Gives (plane) rotations.
 * :ref:`bidiag`. Computations specialized in bidiagonal matrices.
 * :ref:`tridiag`. Computations specialized in tridiagonal matrices.
 * :ref:`symmetric`. Computations specialized in symmetric matrices.
@@ -160,6 +161,30 @@ rocsolver_<type>larfb()
 .. doxygenfunction:: rocsolver_dlarfb
    :outline:
 .. doxygenfunction:: rocsolver_slarfb
+
+
+
+.. _rotations:
+
+Givens (plane) rotations
+==================================
+
+.. contents:: List of plane rotations functions
+   :local:
+   :backlinks: top
+
+.. _lasr:
+
+rocsolver_<type>lasr()
+---------------------------------------
+.. doxygenfunction:: rocsolver_zlasr
+   :outline:
+.. doxygenfunction:: rocsolver_clasr
+   :outline:
+.. doxygenfunction:: rocsolver_dlasr
+   :outline:
+.. doxygenfunction:: rocsolver_slasr
+
 
 
 
--- a/docs/reference/auxiliary.rst
+++ b/docs/reference/auxiliary.rst
@@ -13,7 +13,7 @@ The auxiliary functions are divided into the following categories:
 
 * :ref:`vecmat`. Some basic operations with vectors and matrices that are not part of the BLAS standard.
 * :ref:`householder`. Generation and application of Householder matrices.
-* :ref:`rotations`. Generation and application of Gives (plane) rotations.
+* :ref:`rotations`. Generation and application of Givens (plane) rotations.
 * :ref:`bidiag`. Computations specialized in bidiagonal matrices.
 * :ref:`tridiag`. Computations specialized in tridiagonal matrices.
 * :ref:`symmetric`. Computations specialized in symmetric matrices.
--- a/library/include/rocsolver/rocsolver-functions.h
+++ b/library/include/rocsolver/rocsolver-functions.h
@@ -765,6 +765,129 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfb(rocblas_handle handle,
 //! @}
 
 /*! @{
+    \brief LASR applies a sequence of Givens plane rotations, represented as a transformation P,
+    to a general m-by-n matrix A.
+
+    \details
+    The transformation P is applied in one of the following forms, depending on
+    the value of side:
+
+    \f[
+        \begin{array}{cl}
+        PA & \: \text{(No transpose from the left),}\\
+        AP^T & \: \text{(Transpose from the right).}
+        \end{array}
+    \f]
+
+    P is defined as the product of k plane rotations, with k = m - 1 when applied from the left, and
+    k = n - 1 when applied from the right, as follows:
+
+    \f[
+        \begin{array}{cl}
+        P = P(1)P(2)\cdots P(k) & \: \text{if direct indicates backward direction, or} \\
+        P = P(k)\cdots H(2)H(1) & \: \text{if direct indicates forward direction}
+        \end{array}
+    \f]
+
+    Each P(i) is defined by a Givens rotation
+
+    \f[
+        R(i) = \left[\begin{array}{cc}
+        c_i & s_i \\
+        -s_i & c_i
+        \end{array}\right],
+    \f]
+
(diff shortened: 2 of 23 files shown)


library/src/auxiliary/rocauxiliary_lasr.hpp
```diff
@@ -57,13 +57,9 @@ __host__ __device__ static void lasr_body(const rocblas_side side,
                                           I const tid,
                                           I const i_inc)
 {
-    constexpr bool use_reorder = true;
-
     auto c = [&](auto i) -> const S { return (C_[i - 1]); };
```
EdDAzevedo: Minor thought: I wonder whether providing more hints to compiler may be helpful. Say declaring
S const * const __restrict__ C_,
S const * const __restrict__ S_,
T           * const __restrict__ A_,

library/src/auxiliary/rocauxiliary_lasr.hpp
```diff
@@ -57,13 +57,9 @@ __host__ __device__ static void lasr_body(const rocblas_side side,
...
     auto s = [&](auto i) -> const S { return (S_[i - 1]); };
     auto A = [&](auto i, auto j) -> T& { return (A_[i - 1 + (j - 1) * lda]); };
```
EdDAzevedo: Minor suggestion: In line 443, perhaps wrap the batch index in a "for loop" to get around the 64K limit in thread blocks. 
Say something like

auto const bid_start = hipBlockIdx_z;  
auto const bid_inc = hipGridDim_z;
for(auto bid=bid_start; bid < batch_count; bid += bid_inc ) { ....

library/src/auxiliary/rocauxiliary_lasr.hpp
```diff
@@ -55,7 +55,7 @@ __host__ __device__ static void lasr_body(const rocblas_side side,
                                           T* A_,
                                           I const lda,
                                           I const tid,
-                                          I const i_inc)
+                                          I const t_inc)
 {
     auto c = [&](auto i) -> const S { return (C_[i - 1]); };
     auto s = [&](auto i) -> const S { return (S_[i - 1]); };
```
EdDAzevedo: Minor suggestion: The indexing for matrix A may need to use 64-bit integer arithmetic, say something like

 A_[ (i-1) + (j-1) * static_cast<int64_t>(lda) ]

EdDAzevedo (approved):
looks good to me.

docs/reference/auxiliary.rst
```diff
@@ -13,6 +13,7 @@ The auxiliary functions are divided into the following categories:
 
 * :ref:`vecmat`. Some basic operations with vectors and matrices that are not part of the BLAS standard.
 * :ref:`householder`. Generation and application of Householder matrices.
+* :ref:`rotations`. Generation and application of Gives (plane) rotations.
```
tfalders: Spelling on Givens

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -764,6 +764,129 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfb(rocblas_handle handle,
...
+        \end{array}
+    \f]
+
+    P is defined as the product of k plane rotations, with k = m - 1 when applied from the left, and
+    k = n - 1 when applied from the right, as follows:
+
+    \f[
+        \begin{array}{cl}
+        P = P(1)P(2)\cdots P(k) & \: \text{if direct indicates backward direction, or} \\
+        P = P(k)\cdots H(2)H(1) & \: \text{if direct indicates forward direction}
```
tfalders, line 788: H(2)H(1) -> P(2)P(1)

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -764,6 +764,129 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfb(rocblas_handle handle,
...
+        \end{array}
+    \f]
+
+    Each P(i) is defined by a Givens rotation
+
+    \f[
+        R(i) = \left[\begin{array}{cc}
```
tfalders, line 795: P(i)?

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -764,6 +764,129 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfb(rocblas_handle handle,
...
+        c_i & s_i \\
+        -s_i & c_i
+        \end{array}\right],
+    \f]
+
+    where the \f$c_i\f$ and \f$s_i\f$ are the corresponding sine and cosine factors.
```
tfalders, line 801: I'd suggest "cosine and sine" factors, to match the order of "c_i and s_i".

library/include/rocsolver/rocsolver-functions.h
```diff
@@ -764,6 +764,129 @@ ROCSOLVER_EXPORT rocblas_status rocsolver_zlarfb(rocblas_handle handle,
...
+
+    The rotations are performed on different planes depending on the value of pivot. If pivot is
+    variable, the rototion R(i) is performed on plane (i,i+1), i.e. P(i) appears as a rank-2
```
tfalders, line 804: Spelling on rotation

tfalders (approved):
LGTM! Found a few typographical errors but that's it.

As mentioned in our meeting, it would be nice to have the lasr_body function use 0-based indexing and standard array access. Also, don't forget to update the changelog.

library/src/auxiliary/rocauxiliary_lasr.hpp
jzuniga-amd: It does not make any difference in practice. Looking at the generated code I cannot spot any substantial difference either. In addition, applying `const` to the content of the arrays has given problems before (in other places of the library) when having internal calls from other functions. I think this kind of change should be done library-wise if we want to adopt it.

library/src/auxiliary/rocauxiliary_lasr.hpp
jzuniga-amd: done

library/include/rocsolver/rocsolver-functions.h
jzuniga-amd: this is correct. R(i) is the rotation. P(i) is the transformation that shows the rotation in the corresponding plane according to the value of pivot.

tfalders: I don't recall seeing R(i) elsewhere in the documentation for LASR. It kind of comes out of nowhere.

jzuniga-amd: which documentation?
It only appears here and it is not a typo

library/src/auxiliary/rocauxiliary_lasr.hpp
```diff
@@ -83,18 +79,18 @@ __host__ __device__ static void lasr_body(const rocblas_side side,
...
+            for(I i = 0; i <= (m - 2); i++)
             {
-                const auto ctemp = c(i);
-                const auto stemp = s(i);
-                const auto temp_hold = A(i + 1, j);
-                A(i, j) = stemp * temp_hold + ctemp * temp;
+                const auto ctemp = c[i];
+                const auto stemp = s[i];
+                const auto temp_hold = A[(i + 1) + j * lda];
+                A[i + j * lda] = stemp * temp_hold + ctemp * temp;
```
EdDAzevedo: Minor comment: Perhaps call idx2D() for index calculations, so   instead of 
A[ (i+1) + j * lda] 
we may use
A[ idx2D( (i+1), j, lda) ]
or
A[  (i+1) + j * static_cast<int64_t>(lda) ]
------------
If lambda expression is not acceptable, perhaps use  #define macros such as 
#define A(i,j)  A[ (i) + (j) * static_cast<int64_t>(lda) ]

jzuniga-amd: oops! I saw the other comment but forgot to implement the change.
It should be there now.

PR #847, opened 2024-11-06
Optimize Jacobi eigensolver (WIP)

EdDAzevedo:
This PR updates the kernels for Jacobi eigensolver (syevj) to improve the performance several kernels such as  offd_rotate() that apply the block rotation matrix "J" to off-diagonal blocks.  The offd_rotate()  kernel may account for nearly 90% of the execution time on the GPU. 

The PR converts some of the kernels to use the grid-stride or "for-loop" coding style so that the launch configuration can be modified. The offd_rotate() kernel is modified to  store the sub-matrix blocks of matrix "A" and rotation matrix "J" in LDS shared memory. The launch configuration is modified to encourate the reuse of matrix "J" already in LDS to update several sub-matrix blocks of matrix "A".

On the gfx1030, there is about a 35% performance improvement in rocsolver-bench  for n=4000,  with eigenvectors requested in single and double precisions.

For example

rocsolver-bench -f syevj  -n  4000 --evect V --precision s --perf 1 --iters 1

original time 9.9 sec, new time 6.5 sec

rocsolver-bench -f syevj  -n  4000 --evect V --precision d --perf 1 --iters 1

original time 19.1 sec, new time 13.1 sec

Note that the algorithmic block size is currently set to be 32 and this prevents the matrix J to be loaded in LDS for double_complex type.  Further improvements might be possible if the kernels  can be updated to accept a different algorithmic block size.


library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -838,6 +838,240 @@ ROCSOLVER_KERNEL void syevj_diag_rotate(const bool skip_block,
...
+    // -------------------------------------------------
+    // function to calculation the size of the i-th block
+    // -------------------------------------------------
+    auto bsize = [=](auto iblock) {
+        auto const nb_last = n - (blocks - 1) * nb_max;
+        bool const is_last_block = (iblock == (blocks - 1));
+        return ((is_last_block) ? nb_last : nb_max);
+    };
+
+    auto idx2D = [](auto i, auto j, auto ld) { return (i + j * static_cast<int64_t>(ld)); };
```
jzuniga-amd: I think adding this lambda to every single kernel that you write is too much repetition. 
This functionality is used everywhere and thus we have a method for it  (https://github.com/ROCm/rocSOLVER/blob/b29339365edc98a18493e684bdd3f5c1df65d870/library/src/include/lib_host_helpers.hpp#L46)
you may want to use it

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -838,6 +838,240 @@ ROCSOLVER_KERNEL void syevj_diag_rotate(const bool skip_block,
...
+    bool const use_Jsh = (total_bytes <= max_lds);
+
+    for(auto bid = bid_start; bid < batch_count; bid += bid_inc)
+    {
+        if(completed[bid + 1])
+            continue;
+
+        T* const A_ = load_ptr_batch<T>(AA, bid, shiftA, strideA);
+
+        auto A = [=](auto i, auto j) -> T& { return (A_[idx2D(i, j, lda)]); };
```
jzuniga-amd: Not a big deal but, as discussed before (with the LASR PR), I think the standard C/C++ array accessing using [ ] is preferred.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1475,36 +2327,92 @@ rocblas_status rocsolver_syevj_heevj_template(rocblas_handle handle,
     }
     else
     {
+        // use original algorithm for small problems
+        auto const n_threshold = 1024;
```
jzuniga-amd, line 2318: This is OK for now. If we keep this mechanism (in case a fully blocked formulation is not better), we would need to move this to `ideal_sizes` file and document it.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1016,22 +1250,356 @@ ROCSOLVER_KERNEL void syevj_offd_kernel(const rocblas_int blocks,
...
+        if(completed[bid + 1])
+            continue;
+
+        T* const A_ = load_ptr_batch<T>(AA, bid, shiftA, strideA);
+        auto A = [=](auto i, auto j) -> T& { return (A_[idx2D(i, j, lda)]); };
+
+        for(auto ipair = ipair_start; ipair < npairs; ipair += ipair_inc)
+        {
+            auto const iblock = min(top[ipair], bottom[ipair]);
+            auto const jblock = max(top[ipair], bottom[ipair]);
```
jzuniga-amd: std::max and std::min ???

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1097,6 +1665,283 @@ ROCSOLVER_KERNEL void syevj_offd_rotate(const bool skip_block,
...
+        T* const __restrict__ A_ = load_ptr_batch<T>(AA, bid, shiftA, strideA);
+
+        for(auto ipair = ipair_start; ipair < npairs; ipair += ipair_inc)
+        {
+            // ------------------------------------
+            // consider blocks in upper triangular
+            // want   (iblock < jblock)
+            // ------------------------------------
+            auto const iblock = min(top[ipair], bottom[ipair]);
+            auto const jblock = max(top[ipair], bottom[ipair]);
```
jzuniga-amd: std::max and std::min ???

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1016,22 +1250,356 @@ ROCSOLVER_KERNEL void syevj_offd_kernel(const rocblas_int blocks,
...
                                         T* JA,
                                         rocblas_int* top,
                                         rocblas_int* bottom,
-                                        rocblas_int* completed)
+                                        rocblas_int* completed,
+                                        rocblas_int batch_count)
+{
+    auto ceil = [](auto n, auto nb) { return ((n - 1) / nb + 1); };
+    auto max = [](auto x, auto y) { return ((x > y) ? x : y); };
+    auto min = [](auto x, auto y) { return ((x < y) ? x : y); };
```
jzuniga-amd: simply use std::max and std::min

jzuniga-amd (approved):
LGTM. 
I did not review the code thoroughly but it is given good improvements. I want to give a try to a fully blocked formulation and I will revisit the whole mechanism then.

qjojo (approved):

jzuniga-amd:
Oh, and we need an entry for the Changelog

amd-jnovotny (approved):
Change log looks good. Thanks for updating it!

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1475,36 +2320,92 @@ rocblas_status rocsolver_syevj_heevj_template(rocblas_handle handle,
...
     {
+        // use original algorithm for small problems
+        auto const n_threshold = 1024;
+
+        bool const use_offd_kernel_org = (n <= n_threshold);
+        bool const use_diag_rotate_org = (n <= n_threshold);
+        bool const use_offd_rotate_org = (n <= n_threshold);
+
         // *** USE BLOCKED KERNELS ***
+        auto ceil = [](auto n, auto nb) { return ((n - 1) / nb + 1); };
```
tfalders: This is defined in lib_host_helpers.hpp isn't it? You don't need a lambda function for that.
It might also be worth putting the __host__ and __device__ keywords on the ceil function so it can also be used on the device.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1475,36 +2320,92 @@ rocblas_status rocsolver_syevj_heevj_template(rocblas_handle handle,
...
+        dim3 gridOR_org(half_blocks, 2 * blocks, batch_count);
+        dim3 threadsOR_org(2 * BS2, BS2 / 2, 1);
+
+        dim3 gridOR_new(half_blocks, std::max(1, blocks / 4), batch_count);
+        dim3 threadsOR_new(BS2, BS2, 1);
+
+        dim3 gridOR = (use_offd_rotate_org) ? gridOR_org : gridOR_new;
+        dim3 threadsOR = (use_offd_rotate_org) ? threadsOR_org : threadsOR_new;
+
+        size_t const size_lds = 64 * 1024;
```
tfalders, line 2355: Ideally, this would be queried from the device.

library/src/lapack/roclapack_syevj_heevj.hpp
```diff
@@ -1533,52 +2434,120 @@ rocblas_status rocsolver_syevj_heevj_template(rocblas_handle handle,
...
+                                        completed, batch_count);
+            }
 
             // update eigenvectors
             if(ev)
                 ROCSOLVER_LAUNCH_KERNEL((syevj_diag_rotate<false, T, S>), gridDR, threadsDR,
-                                        lmemsizeDR, stream, false, n, A, shiftA, lda, strideA, J,
-                                        completed);
+                                        lmemsizeDR, stream, false, nb_max, n, A, shiftA, lda,
+                                        strideA, J, completed, batch_count);
```
tfalders: Do we not need to switch between syevj_diag_rotate_org and syevj_diag_rotate here?

tfalders (approved):
I have some non-critical comments. I'll probably do a pass later on to try and homogenize the coding style.

library/src/lapack/roclapack_syevj_heevj.hpp
EdDAzevedo: Is the plan to make  a HIP call to get max size of lds only once and have it stored in the handle (or somewhere) or the plan is  query the size  by different rocsolver functions?  

Maybe similar question to query the number of compute units (CU). Whether it should be already stored in the handle or each rocsolver function will make the HIP call to get this information?

library/src/lapack/roclapack_syevj_heevj.hpp
EdDAzevedo: the code is modified to switch as well.

library/src/lapack/roclapack_syevj_heevj.hpp
EdDAzevedo: the one-liner lambda for ceil()  is removed.

library/src/lapack/roclapack_syevj_heevj.hpp
tfalders: My concern with storing this information is that it appears to be possible for users to switch devices using HIP in between rocSOLVER functions calls, which would invalidate any information we've stored. Whether this is actually possible to do without breaking anything is something I haven't tested, but I tend to lean towards querying this data each time a rocSOLVER function is called because of it.

EdDAzevedo: I agree that there is a concern that  the amount of LDS can be modified at execution time. This feature to reconfigure the amount of (unified combined) LDS shared memory  and L1 cache is available on some  nvidia GPUs and might be available on future AMD GPUs.   Should there be a standard function to query the amount of LDS, say in lib_host_helper, which may be  similar to a standard idx2D() ?  Maybe  something like  rocsolve_get_lds()?

If the amount of LDS can be easily reconfigured,  perhaps it opens up the possibility of reconfiguring the amount of LDS to optimize the performance of rocBLAS and rocsolver kernel  routines?


We might consider passing in the amount of LDS in the launch configuration as part of the subroutine arguments to rocsolver kernels  since it may be difficult for the kernel at execution time  to know how much LDS is actually  available.       Just a thought.

PR #851, opened 2024-11-13
Add sparse QR factorization support

qjojo:
This adds the support for CSRLSVQR to rocSOLVER. I will push an update to this branch adding some boilerplate to the tests for the argchecks but this gets everything algorithm wise up for review.


clients/common/sparselapack/testing_lsvqr.hpp
```diff
@@ -0,0 +1,175 @@
...
+#include "common/misc/client_util.hpp"
+#include "common/misc/clientcommon.hpp"
+#include "common/misc/lapack_host_reference.hpp"
+#include "common/misc/norm.hpp"
+#include "common/misc/rocsolver.hpp"
+#include "common/misc/rocsolver_arguments.hpp"
+#include "common/misc/rocsolver_test.hpp"
+
+template <bool CPU, bool GPU, typename T, typename Td, typename Ud, typename Th, typename Uh>
+void lsvqr_initData(const rocblas_handle handle,
```
EdDAzevedo, line 39: minor comment that  you can ignore but I wonder what is the advantage of using bool CPU, bool GPU as template arguments vs  having separate single purpose routines, say   lsvqr_initData_cpu() and lsvqr_initData_gpu_from_cpu(). I have a concern that there may be unexpected incorrect uses, say,   lsvqr_initData< cpu=false, gpu = true >

tfalders:
I was under the impression that we would not be touching rocSOLVER in order to support sparse QR in hipSOLVER. It should be possible to put the implementation directly in hipSOLVER. If you would open up the hipSOLVER PR with the APIs and testing, I can help you move the implementation over.

library/src/sparselapack/rocsparselapack_csrlsvqr.hpp
```diff
@@ -4,27 +4,121 @@
...
+    scalars = mem[0];
+    work_x_temp = mem[1];
+    workArr_temp_arr = mem[2];
+    diag_trfac_invA = mem[3];
+    trfact_workTrmm_invA_arr = mem[4];
+    ipiv_savedB = mem[5];
+    if(size_scalars > 0)
+        init_scalars(handle, (T*)scalars);
+
+    ROCSOLVER_LAUNCH_KERNEL((copy_mat<T, T*>), dim3(n, 1, 1), dim3(32, 1, 1), 0, stream, n, 1, B,
```
EdDAzevedo, line 106: Minor comment: would it be feasible to use a library routine such as hipMemcpyAsync() since batch_count is 1?

clients/common/sparselapack/testing_lsvqr.hpp
tfalders: initData< cpu=false, gpu = true > is a valid use case. We use it when gathering performance data to call a function multiple times with the same data (by loading it from the CPU which has been initialized with initData< cpu=true, gpu = false >).

As for why we designed it this way, I think it was mostly a stylistic choice, so that initData is conceptually a single method.

library/src/sparselapack/rocsparselapack_csrlsvqr.hpp
```diff
@@ -4,27 +4,121 @@
...
+    st = rocsolver_gels_template<false, false, T>(
+        handle, trans, n, n, nrhs, (T*)denseA, shiftA, lda, strideA, (T*)bCopy, shiftB, ldb,
+        strideB, info, batch_count, (T*)scalars, (T*)work_x_temp, (T*)workArr_temp_arr,
+        (T*)diag_trfac_invA, (T**)trfact_workTrmm_invA_arr, (T*)ipiv_savedB, optim_mem);
+
+    if(st == rocblas_status_success)
+        ROCSOLVER_LAUNCH_KERNEL((copy_mat<T, T*>), dim3(n, 1, 1), dim3(32, 1, 1), 0, stream, n, 1,
+                                (T*)bCopy, shiftB, batch_count, strideB, X, shiftB, batch_count,
+                                strideB);
+
```
EdDAzevedo, line 118: Minor comment: just a reminder to verify that  temporary storage for dense matrix and storage needed for rocsolver are deallocated.

EdDAzevedo (approved):

clients/common/sparselapack/testing_lsvqr.hpp
EdDAzevedo: A similar thought is to consider the relative merits of   bool cpu, bool gpu as regular arguments to the function, vs template arguments. Perhaps for performance sensitive routines, we do need to provide more compile time hints to the compiler to generate efficient code (say problem size is a power of 2, or to encourage aggressive loop unrolling). Otherwise, there may not be a real advantage in using template arguments vs regular arguments but there may be a separate issue with redundant object codes.   Just a thought.

PR #852, opened 2024-11-13
(WIP) Fix Memory Model and Internal TRSM failure

AGonzales-amd:
This PR fixes the `checkin_misc_MEMORY_MODEL.user_managed` test failure when `ROCSOLVER_USE_INTERNAL_BLAS` is set and internal trsm is used for `rocsolver_dgetrf_strided_batched`.

### DIFF
--- a/clients/gtest/memory_model_gtest.cpp
+++ b/clients/gtest/memory_model_gtest.cpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2020-2023 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2020-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -204,7 +204,12 @@ TEST_F(checkin_misc_MEMORY_MODEL, user_managed)
 
     // 12. stop query; required size at the end of query (54MB)
     rocblas_stop_device_memory_size_query(handle, &size);
+#ifndef USE_INTERNAL_TRSM
     EXPECT_GT(size, 2000000);
+#else
+    // internal trsm does not use scratch memory
+    EXPECT_LT(size, 2000000);
+#endif
 
     // 13. device memory size should not change; it should still be 2MB
     rocblas_get_device_memory_size(handle, &size);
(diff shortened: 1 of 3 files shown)


clients/gtest/memory_model_gtest.cpp
```diff
@@ -204,15 +204,25 @@ TEST_F(checkin_misc_MEMORY_MODEL, user_managed)
 
     // 12. stop query; required size at the end of query (54MB)
     rocblas_stop_device_memory_size_query(handle, &size);
+#ifndef USE_INTERNAL_TRSM
     EXPECT_GT(size, 2000000);
+#else
+    // internal trsm does not use scratch memory
+    EXPECT_LT(size, 2000000);
+#endif
```
jmachado-amd, line 212: Hi @AGonzales-amd, those changes fix the failure, but I am afraid that they are not improving the test.  The real problem is that this test is making assumptions about the implementation of `getrf`, and this is brittle: Unless it is absolutely necessary, we should not allow tests to make any assumptions about internal implementations. Implementations can change and render those assumptions redundant (which is exactly what is happening when `getrf` uses our `TRSM`).

A more appropriate solution would be to completely remove any calls to `getrf` that expect it to allocate an implementation dependent amount of memory, and make sure that the test itself sets everything it needs, explicitly.

clients/gtest/memory_model_gtest.cpp
```diff
@@ -323,15 +343,25 @@ TEST_F(checkin_misc_MEMORY_MODEL, DISABLED_user_owned)
...
     EXPECT_EQ(size, 2000000);
 
     // 12. When executing getrf, device memory is not enough for success
     status = rocsolver_dgetrf_strided_batched(handle, m, n, dA, lda, stA, dP, stP, dinfo, bc);
+#ifndef USE_INTERNAL_TRSM
     EXPECT_EQ(status, rocblas_status_memory_error);
+#else
+    // internal trsm does not use scratch memory
+    EXPECT_EQ(status, rocblas_status_success);
+#endif
```
EdDAzevedo, line 364: Minor questions: Is there a rocsolver function so the user can know  say  100MB will be sufficient for the dgetrf rocsolver call for a certain matrix size, or this is just trial-and-error or keep doubling the size until it works?  Perhaps when using rocblas managed mode, there is a way to query the max memory pool size used? Perhaps the user can  use rocblas managed mode, then later  switch to user owned mode (but knowing the max pool size used in rocblas)  for higher performance?

qjojo (approved):

PR #854, opened 2024-11-18
Enable hashing output for sygvdx

qjojo:
This adds a --hash option to rocsolver-bench which will use a deterministic hash function on the outputs and print the result. This is for testing bitwise reproducibility across machines and runs.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -168,6 +168,7 @@ try
             "Validate GPU results with CPU? 0 = No, 1 = Yes.\n"
             "                           This will additionally print the relative error of the computations.\n"
             "                           ")
+
         ("hash",
          value<rocblas_int>(&argus.hash_check)->default_value(0),
             "Print hash of GPU results? 0 = No, 1 = Yes.\n"
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -172,7 +172,7 @@ try
         ("hash",
          value<rocblas_int>(&argus.hash_check)->default_value(0),
             "Print hash of GPU results? 0 = No, 1 = Yes.\n"
-            "                           Hash values are deterministic across implementations and executions.\n"
+            "                           Meant for checking reproducibility of computations.\n"
             "                           ")
 
         // size options
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -168,6 +168,11 @@ try
             "Validate GPU results with CPU? 0 = No, 1 = Yes.\n"
             "                           This will additionally print the relative error of the computations.\n"
             "                           ")
+        ("hash",
+         value<rocblas_int>(&argus.hash_check)->default_value(0),
+            "Print hash of GPU results? 0 = No, 1 = Yes.\n"
+            "                           Hash values are deterministic across implementations and executions.\n"
+            "                           ")
 
         // size options
         ("k",
--- a/clients/common/lapack/testing_sygvdx_hegvdx.hpp
+++ b/clients/common/lapack/testing_sygvdx_hegvdx.hpp
@@ -376,7 +376,11 @@ void sygvdx_hegvdx_getError(const rocblas_handle handle,
                             Vh& hInfo,
                             Vh& hInfoRes,
                             double* max_err,
-                            const bool singular)
+                            const bool singular,
+                            size_t& hashA,
+                            size_t& hashB,
+                            size_t& hashW,
+                            size_t& hashZ)
 {
     constexpr bool COMPLEX = rocblas_is_complex<T>;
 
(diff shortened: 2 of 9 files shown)


clients/benchmarks/client.cpp
```diff
@@ -168,6 +168,11 @@ try
             "Validate GPU results with CPU? 0 = No, 1 = Yes.\n"
             "                           This will additionally print the relative error of the computations.\n"
             "                           ")
+        ("hash",
```
tfalders, line 172: A newline between hash and the previous argument would be appreciated

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -831,6 +832,12 @@ void testing_sygvdx_hegvdx(Arguments& argus)
                 dNev, dW, stW, dZ, ldz, stZ, dInfo, bc, hA, hB, hNev, hW, hZ, hInfo, &gpu_time_used,
                 &cpu_time_used, hot_calls, argus.profile, argus.profile_kernels, argus.perf,
                 argus.singular);
+
+        if(argus.hash_check)
```
tfalders: Correct me if I'm wrong, but won't getPerfData overwrite hA before we reach this point? You might want to put this directly after getError so that the output matrices aren't overwritten.

tfalders (changes requested):

tfalders (approved):

clients/common/lapack/testing_sygvdx_hegvdx.hpp
```diff
@@ -818,12 +819,18 @@ void testing_sygvdx_hegvdx(Arguments& argus)
...
                                                dB, ldb, stB, vl, vu, il, iu, dNev, dW, stW, dZ, ldz,
                                                stZ, dInfo, bc, hA, hB, hNev, hNevRes, hW, hWRes, hZ,
                                                hZRes, hInfo, hInfoRes, &max_error, argus.singular);
 
+        if(argus.hash_check)
+        {
+            hashA = deterministic_hash(hA);
+            hashZ = deterministic_hash(hZRes);
+        }
+
```
jmachado-amd: I am unsure whether this is accomplishing what it is supposed to, but I understand the necessity of moving forward and fixing possible issues later.

For the hash of the input (`hashA`), lapacks' `sygvx` always destroys the contents of matrix `hA`, and the current implementation of `sygvdx_hegvdx_getError` also uses `hA` as part of the computation of the error. I think you want to use the matrix `hA` as produced by `sygvdx_hegvdx_initData`, which will require some changes to this code: maybe call the `initData` method again here? Moreover, for rocsolver's `sygvx` and `sygvdx` tests you should hash both of the original matrices `hA` and `hB`, as `sygvdx_hegvdx_initData` uses other lapack methods to produce `hB` and this yields observable changes to rocsolver's output.

Similarly, for the hash of the output (`hashZ`), it should include the eigenvalues as well (matrix `hWRes`) instead of just the eigenvectors.

clients/common/misc/rocsolver_test.hpp
```diff
@@ -158,3 +160,55 @@ inline std::ostream& operator<<(std::ostream& os, printable_char x)
 
 // location of the sparse data directory for the re-factorization tests
 fs::path get_sparse_data_dir();
+
+/// Combines `seed` with the hash of `value`, following the spirit of
+/// `boost::hash_combine`.
+///
+/// Extends `std::hash` to combine the hashes of multiple values (e.g.,
+/// from an array). Values will be the same across implementations and executions.
```
jmachado-amd: This is not correct as written, `hash_combine` is deterministic but the values it produces can vary across implementations and even executions -- this is why it is important to provide hashes for both the input and the output.

If a _reproducible_ version is desired, then we will need more work.

clients/common/misc/rocsolver_test.hpp
```diff
@@ -158,3 +160,55 @@ inline std::ostream& operator<<(std::ostream& os, printable_char x)
...
+    }
+
+    return hash;
+}
+
+/// Wrapper for hash_combine
+template <typename T>
+std::size_t deterministic_hash(T& vector)
+{
+    return hash_combine(1, vector.data(), vector.size());
```
jmachado-amd: Very minor, if you have a chance, please cast `1` to `std::size_t` to prevent compiler warnings due to type mismatch.

clients/benchmarks/client.cpp
```diff
@@ -169,6 +169,12 @@ try
             "                           This will additionally print the relative error of the computations.\n"
             "                           ")
 
+        ("hash",
+         value<rocblas_int>(&argus.hash_check)->default_value(0),
+            "Print hash of GPU results? 0 = No, 1 = Yes.\n"
+            "                           Hash values are deterministic across implementations and executions.\n"
```
jmachado-amd: Hash values are deterministic, but they can change across implementations and executions.

jmachado-amd (commented):
Just a few comments here, so that we can discuss.

qjojo:
Added checks for getrf, potrf, syevx, sygvx, updated the checks on sygvdx, and updated comments.

jzuniga-amd (approved):
LGTM

qjojo:
CI failures are known issues with sygvdx unrelated to this PR, merging.

PR #862, opened 2024-11-30
Recursive QR factorization for tall skinny matrices

EdDAzevedo:
Initial implementation of the recursive QR factorization algorithm for tall skinny matrices that is motivated by the paper 

"Applying recursion to serial and parallel QR factorization" by Eric Elmroth and Fred Gustavson
August 2000IBM Journal of Research and Development 44(4):605 - 624
DOI:10.1147/rd.444.0605 https://www.researchgate.net/publication/224103218_Applying_recursion_to_serial_and_parallel_QR_factorization

The  recursive algorithm uses Level 3 BLAS operations for constructing the block Householder transformation 
H = eye -  Y * T  * Y', T is triangular

Other characteristics:
* There is a  simple implementation of GEMM for small  (m < 32) or (n < 32) that may  have better performance than rocBLAS GEMM
* If m == 1, or n == 1, the algorithm can call GEMV instead of GEMM when feasible to obtain better performance


EdDAzevedo:
Close pull request since code seems to be not ready.  Need to investigate failures  in gesvd.

PR #865, opened 2024-12-04
Hybrid execution of sterf

tfalders:
This is a revamp of https://github.com/ROCm/rocSOLVER/pull/462 using the new hybrid infrastructure that was introduced for GESVD. Here's a brief summary of changes:

- Added CPU execution path for STERF. Tests for STERF_HYBRID and SYEV_HYBRID have been added to verify correctness.
- Rearranged functions in lib_device_helpers, so that device functions are no longer present in the kernels section.
- is_device_pointer has been moved from a lambda function to a reusable function in lib_host_helpers.
- Created a new class, rocsolver_hybrid_array, to assist with memory allocation and data transfers to and from the device. I have revised rocsolver_bdsqr_host_batch_template to use this new functionality.

### DIFF
--- a/clients/gtest/lapack/syev_heev_gtest.cpp
+++ b/clients/gtest/lapack/syev_heev_gtest.cpp
@@ -81,6 +81,7 @@ Arguments syev_heev_setup_arguments(syev_heev_tuple tup)
     return arg;
 }
 
+template <rocblas_int MODE>
 class SYEV_HEEV : public ::TestWithParam<syev_heev_tuple>
 {
 protected:
@@ -93,6 +94,7 @@ protected:
     void run_tests()
     {
         Arguments arg = syev_heev_setup_arguments(GetParam());
+        arg.alg_mode = MODE;
 
         if(arg.peek<rocblas_int>("n") == 0 && arg.peek<char>("evect") == 'N'
            && arg.peek<char>("uplo") == 'L')
@@ -103,11 +105,19 @@ protected:
     }
 };
 
-class SYEV : public SYEV_HEEV
+class SYEV : public SYEV_HEEV<0>
 {
 };
 
-class HEEV : public SYEV_HEEV
+class HEEV : public SYEV_HEEV<0>
+{
+};
+
+class SYEV_HYBRID : public SYEV_HEEV<1>
+{
+};
+
+class HEEV_HYBRID : public SYEV_HEEV<1>
 {
 };
 
@@ -129,6 +139,26 @@ TEST_P(HEEV, __float_complex)
 }
 
 TEST_P(HEEV, __double_complex)
+{
+    run_tests<false, false, rocblas_double_complex>();
+}
+
+TEST_P(SYEV_HYBRID, __float)
+{
+    run_tests<false, false, float>();
+}
+
+TEST_P(SYEV_HYBRID, __double)
+{
+    run_tests<false, false, double>();
+}
+
+TEST_P(HEEV_HYBRID, __float_complex)
+{
+    run_tests<false, false, rocblas_float_complex>();
+}
+
+TEST_P(HEEV_HYBRID, __double_complex)
 {
     run_tests<false, false, rocblas_double_complex>();
 }
@@ -155,6 +185,26 @@ TEST_P(HEEV, batched__double_complex)
     run_tests<true, true, rocblas_double_complex>();
 }
 
+TEST_P(SYEV_HYBRID, batched__float)
+{
+    run_tests<true, true, float>();
+}
+
+TEST_P(SYEV_HYBRID, batched__double)
+{
+    run_tests<true, true, double>();
+}
+
+TEST_P(HEEV_HYBRID, batched__float_complex)
+{
+    run_tests<true, true, rocblas_float_complex>();
+}
+
+TEST_P(HEEV_HYBRID, batched__double_complex)
+{
+    run_tests<true, true, rocblas_double_complex>();
+}
+
 // strided_batched tests
 
 TEST_P(SYEV, strided_batched__float)
@@ -177,13 +227,49 @@ TEST_P(HEEV, strided_batched__double_complex)
     run_tests<false, true, rocblas_double_complex>();
 }
 
+TEST_P(SYEV_HYBRID, strided_batched__float)
+{
+    run_tests<false, true, float>();
+}
+
+TEST_P(SYEV_HYBRID, strided_batched__double)
+{
+    run_tests<false, true, double>();
+}
+
+TEST_P(HEEV_HYBRID, strided_batched__float_complex)
+{
+    run_tests<false, true, rocblas_float_complex>();
+}
+
+TEST_P(HEEV_HYBRID, strided_batched__double_complex)
+{
+    run_tests<false, true, rocblas_double_complex>();
+}
+
 // daily_lapack tests normal execution with medium to large sizes
 INSTANTIATE_TEST_SUITE_P(daily_lapack, SYEV, Combine(ValuesIn(large_size_range), ValuesIn(op_range)));
 
 INSTANTIATE_TEST_SUITE_P(daily_lapack, HEEV, Combine(ValuesIn(large_size_range), ValuesIn(op_range)));
+
+INSTANTIATE_TEST_SUITE_P(daily_lapack,
+                         SYEV_HYBRID,
+                         Combine(ValuesIn(large_size_range), ValuesIn(op_range)));
+
+INSTANTIATE_TEST_SUITE_P(daily_lapack,
+                         HEEV_HYBRID,
+                         Combine(ValuesIn(large_size_range), ValuesIn(op_range)));
 
 // checkin_lapack tests normal execution with small sizes, invalid sizes,
 // quick returns, and corner cases
 INSTANTIATE_TEST_SUITE_P(checkin_lapack, SYEV, Combine(ValuesIn(size_range), ValuesIn(op_range)));
 
 INSTANTIATE_TEST_SUITE_P(checkin_lapack, HEEV, Combine(ValuesIn(size_range), ValuesIn(op_range)));
+
+INSTANTIATE_TEST_SUITE_P(checkin_lapack,
+                         SYEV_HYBRID,
+                         Combine(ValuesIn(size_range), ValuesIn(op_range)));
+
+INSTANTIATE_TEST_SUITE_P(checkin_lapack,
+                         HEEV_HYBRID,
+                         Combine(ValuesIn(size_range), ValuesIn(op_range)));
--- a/library/src/include/lib_device_helpers.hpp
+++ b/library/src/include/lib_device_helpers.hpp
@@ -98,7 +98,7 @@ __device__ void
 /** FIND_MAX_TRIDIAG finds the element with the largest magnitude in the
     tridiagonal matrix **/
 template <typename T>
-__device__ T find_max_tridiag(const rocblas_int start, const rocblas_int end, T* D, T* E)
+__device__ __host__ T find_max_tridiag(const rocblas_int start, const rocblas_int end, T* D, T* E)
 {
     T anorm = abs(D[end]);
     for(int i = start; i < end; i++)
@@ -109,7 +109,8 @@ __device__ T find_max_tridiag(const rocblas_int start, const rocblas_int end, T*
 /** SCALE_TRIDIAG scales the elements of the tridiagonal matrix by a given
     scale factor **/
 template <typename T>
-__device__ void scale_tridiag(const rocblas_int start, const rocblas_int end, T* D, T* E, T scale)
+__device__ __host__ void
+    scale_tridiag(const rocblas_int start, const rocblas_int end, T* D, T* E, T scale)
 {
     D[end] *= scale;
     for(int i = start; i < end; i++)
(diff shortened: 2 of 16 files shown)


amd-jnovotny (approved):
changelog ok

library/src/include/lib_device_helpers.hpp
```diff
@@ -119,1122 +120,1122 @@ __device__ void scale_tridiag(const rocblas_int start, const rocblas_int end, T*
...
+    rocblas_int* m_mask;
+    bool m_negate;
+};
+
+/** COPY_MAT copies m-by-n array A into buffer if copymat_to_buffer, or buffer into A if copymat_from_buffer
+    An optional mask can be provided to limit the copy to selected matricies in the batch
+    If uplo = rocblas_fill_upper, only the upper triangular part is copied
+    If uplo = rocblas_fill_lower, only the lower triangular part is copied **/
+template <typename T, typename U, typename Mask = no_mask>
+ROCSOLVER_KERNEL void copy_mat(copymat_direction direction,
```
EdDAzevedo, line 625: Minor suggestion: if these is intended to be a  frequently reused library routine, perhaps we might consider a more general version  from (A,shiftA,lda,strideA) to (B,shiftB,ldb,strideB).  The special case of a linear "buffer" may be shiftB = 0, strideB = (m*n), ldb = m;  By using a more general interface, perhaps we don't need the extra enum for "direction". We just need to reverse the order of the "A" and "B" arguments. 

Perhaps consider a more suggestive name for Mask. It is not immediately obvious to me  whether we perform the copy if the mask is non-zero, or skip the copy if the mask is non-zero.        

Is there a convention to place action arguments before the data? Say in TRSM, we place the uplo, side, trans, diag, arguments before the data, instead of placing the diag or uplo at the end of subroutine argument list.

library/src/include/lib_device_helpers.hpp
```diff
@@ -119,1122 +120,1122 @@ __device__ void scale_tridiag(const rocblas_int start, const rocblas_int end, T*
...
-        if(a[i + i * lda] == 0)
+        if(full || (upper && j > i) || (lower && i > j) || (cdiag && i == j))
         {
-            rocblas_int _info_temp = _info;
-            while(_info_temp == 0 || _info_temp > i + 1)
-                _info_temp = atomicCAS(&_info, _info_temp, i + 1);
+            T* Ap = load_ptr_batch<T>(A, b, shiftA, strideA);
+            T* Bp = &buffer[b * strideB];
+
+            if(direction == copymat_to_buffer)
```
EdDAzevedo, line 658: I wonder whether this library routine can be further generalized to add a "trans" option to perform conjugate_transpose or transpose or none. Magma blas has a routine to perform efficient matrix transpose.   Just a suggestion.

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,249 @@
...
+#pragma once
+
+#include "common_host_helpers.hpp"
+#include "lib_host_helpers.hpp"
+#include "rocsolver/rocsolver.h"
+
+ROCSOLVER_BEGIN_NAMESPACE
+
+template <typename T, typename I, typename U>
+struct rocsolver_hybrid_array
```
EdDAzevedo: It would be nice if there is more documentation or description not of the internal implementation details but on the interface what it is intended to do and what functions or method are available.  Is it  keeping a data structure (strided array) on CPU and mirror it on GPU device, or vice versa? If so, perhaps the routines are trying to "sync"?   Should there be "async" and "noasync" versions since the name of the routine "_async" is suggestive there is a noasync version.  Are there calls to synchronize stream in the "_async" routines?   If these are not truyly "async" routines, perhaps just leave out "_async"?   I wonder whether the std::vector can be leveraged so there is  less worry about memory leaks.

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,249 @@
...
+                batch_array = nullptr;
+            else
+                batch_array = (T**)src_array;
+        }
+
+        return rocblas_status_success;
+    }
+    /* Copies data from val_array back to the device. Returns an error if initialized for pointers
+       only. */
+    rocblas_status push_to_device_async(hipStream_t stream)
```
EdDAzevedo: Minor suggestion: One may expect a "pop" when there is a "push".  Perhaps "copy_to_device" or "sync_to_device" is more suggestive?  Or perhaps just a  "copy" routine and accept a hipMemcpyKind argument for the direction  "host to device" or "device to host"?  Just a thought.

library/src/include/lib_device_helpers.hpp
tfalders: IIRC, this function has multiple overloads where B has fewer associated parameters than A. In these cases, specifying a direction is necessary since we can't easily swap A and B. It's not necessary for this particular overload, but I think we kept it for consistency.

I don't entirely remember how the mask works either, so we should definitely improve documentation at a minimum.

Conventionally, yes, those arguments are placed at the start, but when they have default values they can't be put before any arguments without default values. I seem to remember they were tacked on after the fact, and assigned default values to avoid breaking existing calls.

library/src/include/lib_device_helpers.hpp
tfalders: We have a copy_trans_mat function that does have this functionality.

library/src/include/rocsolver_hybrid_array.hpp
tfalders: Yeah, the documentation is rather sparse at the moment. I can take another look and try to improve it.

I added the _async suffix mostly to indicate to anyone using the function that a hipStreamSynchronize needs to be done before using the results of the function. I could have synchronized within the functions themselves, but I wanted to give us the option of queuing a number of hipMemcpys that can be sync'ed all at once for better performance.

library/src/include/rocsolver_hybrid_array.hpp
tfalders: Sure, I can change the name

clients/gtest/lapack/syev_heev_gtest.cpp
```diff
@@ -103,11 +105,19 @@ class SYEV_HEEV : public ::TestWithParam<syev_heev_tuple>
     }
 };
 
-class SYEV : public SYEV_HEEV
+class SYEV : public SYEV_HEEV<0>
```
EdDAzevedo, line 108: Minor: is there a convention to use all upper case for compile time constant or #define macros or constants? If so, perhaps consider using a mixed case?  Just a thought.

EdDAzevedo (approved):

clients/gtest/lapack/syev_heev_gtest.cpp
tfalders: I believe that changing the capitalization of the test class will also change the capitalization of the test suite output. I like the all caps text in the test output as it makes it very easy to pick out the function name.

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,263 @@
...
+    {
+    }
+    ~rocsolver_hybrid_array()
+    {
+        if(val_array)
+            free(val_array);
+        if(batch_array && (val_array || this->dim < 0))
+            free(batch_array);
+    }
+
```
qjojo: Since this class owns malloc-ed memory it might be a good idea to delete its implicit copy constructor and assignment operator to guard against a double free.

qjojo (approved):

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,272 @@
...
+   * Standard mode: Used to read device data onto the host, which may then be operated upon by host code.
+     A typical workflow will call init_async to allocate a host buffer and read the data into the buffer,
+     execute host code on the buffered data through the pointer provided by the [] operator, and then call
+     write_to_device_async to write the resultant data back to the device.
+
+   * Pointers only mode: Used to read device pointers from a batched array onto the host, so that device
+     kernels may be called on each individual batch instance. A typical workflow will call init_pointers_only
+     to allocate a host buffer for the device pointers, and then execute device kernels on the pointers
+     provided by the [] operator. */
+template <typename T, typename I, typename U>
```
jmachado-amd: Since this structs is using C functions to manage memory "with type" `T`, it is good practice to make sure that it [`is_standard_layout`](https://en.cppreference.com/w/cpp/types/is_standard_layout) and maybe consider adding checks to make sure that it is not a pointer or reference.

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    /* Used to read device pointers from a batched array for use on the host; no other data is read from the
+       device. */
+    rocblas_status init_pointers_only(U array, rocblas_stride stride, I batch_count, hipStream_t stream)
+    {
+        if(val_array)
+            free(val_array);
+        if(batch_array && (val_array || this->dim < 0))
+            free(batch_array);
```
jmachado-amd: This is problematic if a previous memory allocation fails and someone tries to use one of the `init_*` methods again. The common practice is to always follow a `free(ptr)` with `ptr = nullptr`. Moreover, since downstream users don't have access to those internal pointers, we should also `memset` the pointed memory to `0` before freeing it.

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+        bool constexpr is_strided = (std::is_same<U, T*>::value || std::is_same<U, T* const>::value);
+        bool is_device = is_device_pointer((void*)array);
+
+        if(is_device)
+        {
+            // allocate space on host for data from device
+            size_t dim_bytes = sizeof(T) * dim;
+            size_t val_bytes = sizeof(T) * dim * batch_count;
+            val_array = (T*)malloc(val_bytes);
```
jmachado-amd: A few comments here: Surprisingly enough, the spec does not define the behaviour of `malloc` when `val_bytes == 0`. Moreover, we should check for possible bad allocs and pass those errors downstream; and also `memset` this memory to `0` before it is used.

On a different topic: We probably also want those memory allocations to be aligned for performance. If you don't want to go with an in-house method to return aligned memory, we can go with [posix_menalign](https://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_memalign.html) (in Linux) and [_aligned_malloc](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=msvc-170) (in Windows).

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,272 @@
...
+    {
+        if(val_array)
+            free(val_array);
+        if(batch_array && (val_array || this->dim < 0))
+            free(batch_array);
+    }
+
+    /* Used to read device pointers from a batched array for use on the host; no other data is read from the
+       device. */
+    rocblas_status init_pointers_only(U array, rocblas_stride stride, I batch_count, hipStream_t stream)
```
jmachado-amd: Minor comment (code design): the behaviour of this struct when initialized with `init_pointers_only` is so different from the bahaviour yielded by `init_async` that it makes sense to break it apart into two different structs.  Maybe something to be pondered for the future?

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,272 @@
...
+     A typical workflow will call init_async to allocate a host buffer and read the data into the buffer,
+     execute host code on the buffered data through the pointer provided by the [] operator, and then call
+     write_to_device_async to write the resultant data back to the device.
+
+   * Pointers only mode: Used to read device pointers from a batched array onto the host, so that device
+     kernels may be called on each individual batch instance. A typical workflow will call init_pointers_only
+     to allocate a host buffer for the device pointers, and then execute device kernels on the pointers
+     provided by the [] operator. */
+template <typename T, typename I, typename U>
+struct rocsolver_hybrid_array
```
jmachado-amd: Minor comment (code design): I am not sure that this struct satisfies all of the semantics of an `array` (e.g., it is not stored on the stack), thus I would change its name to something more general like `rocsolver_hybrid_storage`.

library/src/include/rocsolver_hybrid_array.hpp
```diff
@@ -0,0 +1,272 @@
...
+{
+    I dim, batch_count;
+    rocblas_stride stride;
+
+    U src_array;
+    T** batch_array;
+    T* val_array;
```
jmachado-amd: Minor question (code design): what would be the use case in which we would want to access those members directly?

jmachado-amd (commented):
It looks good, thanks @tfalders!

The `rocsolver_hybrid_array` struct complements the implementation of the current hybrid methods well. I've just included a few comments there, let me know if you want me to clarify anything.

library/src/include/rocsolver_hybrid_array.hpp
jmachado-amd: It would be nicer to follow the [rule of 5](https://en.cppreference.com/w/cpp/language/rule_of_three) also explicitly `default` or `delete` the move constructor and the move assignment operator (those are implicitly deleted here, but you could also `default` them without any issues if this bodes well with your idea of how the struct is meant to be used).

EdDAzevedo: I wonder whether using thrust::host_vector<T> and thrust::device_vector<T> would potentially simplify the code.  This is in the spirit of "eating our own dog food" or "drinking our own champagne"  in re-using AMD  rocThrust   software.   Just a thought.

library/src/include/rocsolver_hybrid_array.hpp
tfalders: I don't believe there is a use case. Would you like me to mark them as private?

jmachado-amd: I pointed it out to start the conversation, I think that we will know better than to try and change those by hand; but it is always safer to constrain degrees of freedom that are not meant to be used.

Whether to keep or change the access type of those is a decision that I defer to you, I trust your judgement either way.

library/src/include/rocsolver_hybrid_storage.hpp
```diff
@@ -0,0 +1,350 @@
...
+        {
+            // allocate space on host for data from device
+            size_t dim_bytes = sizeof(T) * dim;
+            size_t val_bytes = sizeof(T) * dim * batch_count;
+#ifdef _WIN32
+            val_array = (T**)_aligned_malloc(val_bytes, sizeof(void*));
+            if(!val_array)
+                return rocblas_status_memory_error;
+#else
+            if(posix_memalign((void**)&val_array, sizeof(void*), val_bytes) != 0)
```
jmachado-amd, line 227: Small comment here: `sizeof(void*)` is just `uintptr_t`, so this is not really different from a `malloc`, more typical alignments would be `32` or `64`. But I see the point in leaving this as is and deciding the final value after benchmarking.

jmachado-amd (approved):
It looks good to me, thanks again @tfalders! I left a couple of comments but nothing that requires any update at this point in time -- unless you want to update anything, of course.

library/src/include/lib_host_helpers.hpp
```diff
@@ -170,6 +172,20 @@ static double imag_part(rocblas_complex_num<double> z)
...
 
+static bool is_device_pointer(void* ptr)
+{
+    hipPointerAttribute_t dev_attributes;
+    if(ptr == nullptr)
+        return false;
+
+    auto istat = hipPointerGetAttributes(&dev_attributes, ptr);
+    if(istat != hipSuccess)
+        fmt::print(stderr, "is_device_pointer: istat = {} {}\n", istat, hipGetErrorName(istat));
```
jmachado-amd: I've just noticed that this line is not compiling in Windows, likely because CI is using a version of libfmt that is newer than [10.0.0](https://github.com/fmtlib/fmt/blob/master/ChangeLog.md#1000---2023-05-09). Among many other things, that version deprecated the implicit conversion of enums, thus the simplest fix would be to update this line into something like:

`fmt::print(stderr, "is_device_pointer: istat = {} {}\n", static_cast<std::int32_t>(istat), hipGetErrorName(istat));`

Their documentation provides other options, but we should consider updating rocSOLVER to make those errors easier to catch going forward (of course, not in this PR). For the time being, please make sure to cast all inputs of `fmt::print` into basic types, or types defined in the `std` namespace.

PR #869, opened 2024-12-11
Reduced memory usage for SYEVD

tfalders:
This PR reduces the amount of memory required by SYEVD by about 1/4 in the real case and 2/5 in the complex case. It should result in a slight performance boost as well.

In STEDC, when evect is tridiagonal (which is how SYEVD always calls STEDC), the original code initializes C to the identity and then later uses local_gemm to compute C <- C*tempvect. But, since C is the identity, this is a do-nothing operation. I've changed it so that this case will skip the local_gemm and won't allocate tempvect (an n*n matrix), instead writing to C directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Removed
 ### Optimized
 
-* Reduced device memory requirements for STEDC, SYEVD/HEEVD, and SYGVD/HEGVD
+* Reduced the device memory requirements for STEDC, SYEVD/HEEVD, and SYGVD/HEGVD
 
 ### Resolved issues
 ### Known issues
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Changed
 ### Removed
 ### Optimized
+
+* Reduced device memory requirements for STEDC, SYEVD/HEEVD, and SYGVD/HEGVD
+
 ### Resolved issues
 ### Known issues
 ### Upcoming changes
@@ -22,7 +25,7 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 * Algorithm selection mechanism for hybrid computation
 * Hybrid computation support for existing routines:
     - BDSQR
-    - GESVD  
+    - GESVD
 
 ### Optimized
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Full documentation for rocSOLVER is available at the [rocSOLVER documentation](h
 ### Changed
 ### Removed
 ### Optimized
+
+* Reduced device memory requirements for STEDC, SYEVD/HEEVD, and SYGVD/HEGVD
```
amd-jnovotny:
```suggestion
* Reduced the device memory requirements for STEDC, SYEVD/HEEVD, and SYGVD/HEGVD
```

amd-jnovotny (approved):
Changelog looks good. One very minor editorial suggestion. Thanks!

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2192,23 +2192,16 @@ void rocsolver_stedc_getMemorySize(const rocblas_evect evect,
...
     else
     {
-        size_t s1, s2;
-
         // requirements for solver of small independent blocks
-        rocsolver_steqr_getMemorySize<T, S>(evect, n, batch_count, &s1);
+        rocsolver_steqr_getMemorySize<T, S>(evect, n, batch_count, size_work_stack);
 
         // extra requirements for original eigenvectors of small independent blocks
         *size_tempvect = (n * n) * batch_count * sizeof(S);
```
EdDAzevedo: Minor suggestion:  place the "sizeof(S)" first in the expression to encourage the use of 64bit arithmetic and to avoid 32bit integer overflow.  Say   something like  sizeof(S) * n * n * batch_count  instead of   n * n * batch_count * sizeof(S), just in case  "n * n * batch_count" can cause overflow.  Similar suggestion for other expressions involving "sizeof()".

qjojo (approved):

library/src/lapack/roclapack_syevd_heevd.hpp
```diff
@@ -86,12 +86,12 @@ void rocsolver_syevd_heevd_getMemorySize(const rocblas_evect evect,
     if(evect == rocblas_evect_original)
     {
         // extra requirements for computing eigenvalues and vectors (stedc)
-        rocsolver_stedc_getMemorySize<BATCHED, T, S>(rocblas_evect_tridiagonal, n, batch_count, &w12,
-                                                     &w22, &w31, size_tmpz, size_splits, &unused);
+        rocsolver_stedc_getMemorySize<BATCHED, T, S>(rocblas_evect_tridiagonal, n, batch_count, &w31,
+                                                     &w22, &w12, size_tmpz, size_splits, &unused);
```
jmachado-amd, line 90: It's not very fun to keep track of those `w*` variables.

jmachado-amd (approved):
It looks good!

PR #873, opened 2025-01-06
Fix Windows CI interaction with GitHub labels

jmachado-amd:
Windows CI may pass the `--ci_labels` flag to `rmake.py` and `rtest.py` when GitHub labels are selected, but this flag is not recognized by those Python scripts, which leads to runtime errors.  After this change is applied, `rmake.py` and `rtests.py` will silently ignore the `ci_labels` flag, thus preventing such errors.

Update: added similar code to handle upcoming `ci_gfx` flag.

### DIFF
--- a/rmake.py
+++ b/rmake.py
@@ -48,6 +48,8 @@ def parse_args():
                         help='Set GPU architectures to build for (optional)')
     parser.add_argument(      '--ci_labels', nargs='?', type=str, required=False, default="",
                         help='Semicolon (";") separated list of GitHub Labels with build options (not implemented, silently ignored)')
+    parser.add_argument(      '--ci_gfx', nargs='?', type=str, required=False, default="",
+                        help='Semicolon (";") separated list of gfx targets expected on test runs (not implemented, silently ignored)')
 
     return parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


EdDAzevedo (approved):
Accept temporary short-term fix for windows CI.  
A more long-term solution may be required later to correctly handle the labels.

tfalders (approved):
It would be nice if we could at least process the noOptimizations label, but I'll settle for getting Windows CI to work again.

TorreZuk:
You may as well add a no-op for newer coming feature (rmake and rtest):
general_opts.add_argument(      '--ci_gfx', type=str, required=False, default="",
                        help='Semi-colon seperated list of gfx targets expected on test runs (optional, e.g. "gfx1030;gfx1201")')

You were supposed to opt in for both these to be features so I notified windows CI team yesterday.

jmachado-amd:
> It would be nice if we could at least process the noOptimizations label, but I'll settle for getting Windows CI to work again.

It might be me, but I don't recall seeing Windows CI use our labels before the `ci_labels` flag was added.

jmachado-amd:
Thank you so much @TorreZuk!

TorreZuk:
The idea is you will get the ci_labels list and instead of groovy changes done by windows CI team you can adapt your build and test to utilize the label  list, and later gfx list.

jmachado-amd:
@tfalders, @EdDAzevedo: I've just added new code to handle the upcoming `ci_gfx` flag per Torre's comment. Again, those flags are being silently ignored for the time being, proper support will appear in future PRs.

PR #879, opened 2025-01-13
Fix tolerance for HEGVDX tests

jmachado-amd:
As part of PR #810, the tolerances for `sygvdx` and `hegvdx` tests were set too tight, which made a few of `hegvdx`' tests fail.

This PR changes the tolerance to allow all tests to pass on all architectures (some tests will only pass by a tiny margin).

### DIFF
--- a/clients/common/lapack/testing_sygvdx_hegvdx.hpp
+++ b/clients/common/lapack/testing_sygvdx_hegvdx.hpp
@@ -1108,9 +1108,9 @@ void testing_sygvdx_hegvdx(Arguments& argus)
     }
 
     // validate results for rocsolver-test
-    // using 4 * n * machine_precision as tolerance
+    // using 5 * n * machine_precision as tolerance
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, 4 * n);
+        ROCSOLVER_TEST_CHECK(T, max_error, 5 * n);
 
     // output results for rocsolver-bench
     if(argus.timing)


AGonzales-amd (approved):

AGonzales-amd:
Just as a reference for the future, how do we determine the size of the appropriate bounds when checking for error?

tfalders (approved):

jmachado-amd:
> Just as a reference for the future, how do we determine the size of the appropriate bounds when checking for error?

This is not entirely simple, as it depends on the problem and the algorithm that is used to solve it. For the typical numerical method in rocSOLVER, this often boils down to two steps: (1) finding an _a priori estimate_, which provides a bound on the output of the algorithm for all possible inputs; and (2) fine tuning the constants appearing on (1), which can be done empirically.

For a simple (related) example, consider the case of the QR eigen-decomposition which, for a given symmetric matrix A, computes its eigenvalues `a_i`. It is known ([Golub & Van Loan](https://epubs.siam.org/doi/book/10.1137/1.9781421407944), [Demmel](https://www.cambridge.org/gs/universitypress/subjects/mathematics/numerical-analysis/applied-numerical-linear-algebra), and others) that the computed eigenvalues of `A` are the same as the exact eigenvalues of a new symmetric matrix `B = A + E`, where `E` models the numerical error introduced by this eigensolver. Thus, denoting the eigenvalues of `B` by `b_i`, we have the following _a priori estimate_ ([Weyl's inequality](https://en.wikipedia.org/wiki/Weyl%27s_inequality) + an _a priori estimate_ that appears when we multiply `A` by unitary matrices):

(\*) `|b_i - a_i| < C(n) ||A||_2`,

where `||A||_2` is the [l^2 operator norm](https://en.wikipedia.org/wiki/Operator_norm) of `A`, and `C(n)` is a constant that depends on the method and the dimension of `A`, but not on the entries of `A`. If you look on the literature you will find `C(n) = C * n` or (the stricter) `C(n) = C * sqrt(n)`, and when you plug this in Eq. (\*), you'll see that all tests of the QR eigensolver will yield eigenvalues that satisfy (\*) for small values of `C`.

For the sake of the example, we could just run the algorithm with several matrices, record the smallest `C` that satisfy (\*) with `C(n) = C * n` for all of those runs, and set this `C` as starting point for our tests.

PR #880, opened 2025-01-13
Add GEMM device function

AGonzales-amd:
This adds a gemm device function which is callable in other kernels. The function is designed to be called by an entire wavefront and to compute a block of the output matrix. Therefore, problems can be decomposed into chunks that are operated on by individual wavefronts. Currently, it is used to implement an alternative rocsolver_gemm kernel.
 - Uses mfma instructions
 - Limited to `__gfx90a__`, `__gfx940__`, `__gfx941__`, and `__gfx942__` architectures
 - Supports both complex data types and transpose matrix operations

### DIFF
--- a/clients/common/unit/testing_gemm.hpp
+++ b/clients/common/unit/testing_gemm.hpp
@@ -1,0 +1,424 @@
+/* **************************************************************************
+ * Copyright (C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ * *************************************************************************/
+
+#pragma once
+
+#include "common/misc/client_util.hpp"
+#include "common/misc/clientcommon.hpp"
+#include "common/misc/lapack_host_reference.hpp"
+#include "common/misc/norm.hpp"
+#include "common/misc/rocsolver.hpp"
+#include "common/misc/rocsolver_arguments.hpp"
+#include "common/misc/rocsolver_test.hpp"
+
+#include "rocblas_utility.hpp"
+#include "rocsolver_run_specialized_kernels.hpp"
+
+template <bool CPU, bool GPU, typename T, typename I, typename Td, typename Th>
+void gemm_initData(const rocblas_handle handle, Td& dA, Th& hA, Td& dB, Th& hB, Td& dC, Th& hC)
+{
+    if(CPU)
+    {
+        rocblas_init<T>(hA, true);
+        rocblas_init<T>(hB, true);
+        rocblas_init<T>(hC, true);
+    }
+
+    if(GPU)
+    {
+        // now copy to the GPU
+        CHECK_HIP_ERROR(dA.transfer_from(hA));
+        CHECK_HIP_ERROR(dB.transfer_from(hB));
+        CHECK_HIP_ERROR(dC.transfer_from(hC));
+    }
+}
+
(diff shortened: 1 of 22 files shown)


library/src/specialized/roclapack_gemm_device_functions.hpp
```diff
@@ -0,0 +1,258 @@
...
+// Run with warpSize sized block
+template <typename T, typename I>
+__device__ void gemm_16x16xp(rocblas_operation transA,
+                                   rocblas_operation transB,
+                                   I m,
+                                   I n,
+                                   I p,
+                                   T alpha,
+                                   const T *A,
+                                   I inc_A,
```
EdDAzevedo: Is the inc_A, inc_B, ... etc the same as "stride" like strideA? If not, perhaps some description would be helpful.  Just a suggestion.

EdDAzevedo (approved):
It would be nice if there are tests scripts or updates to rocsolver-bench and/or rocsolver-test to check for correctness (and perhaps performance).  Note check for m,n,k to be nb, nb+1, nb-1  corner cases as well.

Will other BLAS operations, say TRMM, or SYR2K,   also take advantage of the new GEMM? One may consider a conceptually recursive formulation so that it can use GEMM.

library/src/specialized/roclapack_gemm_device_functions.hpp
AGonzales-amd: Added function documentation, thanks for the suggestion

jmachado-amd:
Hi @AGonzales-amd, I second Ed's suggestion: you should update `rocsolver-test` and `-bench` clients to support the internal gemm.  I can help you with the tests if you want (this would also be a good opportunity to provide a concrete answer to the question you asked in #879).

AGonzales-amd:
> Hi @AGonzales-amd, I second Ed's suggestion: you should update `rocsolver-test` and `-bench` clients to support the internal gemm. I can help you with the tests if you want (this would also be a good opportunity to provide a concrete answer to the question you asked in #879).

Thanks @jmachado-amd, I could use your help. I did consider updating the client programs but I had trouble exporting the function or making it visible in the clients.

AGonzales-amd:
Hi @jmachado-amd and @EdDAzevedo, the client programs have been updated to support internal gemm. One thing I'm not sure about is the tolerance for error checking.

jmachado-amd:
Hi @AGonzales-amd, as long as the input matrices are "small" to "medium" sized and have positive, relatively small integer entries, the current test tolerance will work just fine! Let me know if you want to generalize the tests or just understand how those bounds work, and I can explain the important bits of the theory to you.

On another topic, I see that there are many `gemm` tests failing in Windows, you probably want to have a look at that sooner rather than later.

clients/gtest/unit/gemm_gtest.cpp
```diff
@@ -0,0 +1,327 @@
...
+const vector<int64_t> large_n_size_range_64 = {1000};
+
+const vector<char> all_operations = {
+    'N',
+    'T',
+    'C',
+};
+
+const vector<double> alphas = {1.5};
+const vector<double> betas = {-2.0};
```
tfalders: I feel like alphas and betas can be combined into one vector.

clients/common/unit/testing_gemm.hpp
```diff
@@ -0,0 +1,425 @@
...
+
+template <bool BATCHED, bool STRIDED, typename T, typename I>
+void testing_gemm(Arguments& argus)
+{
+    // get arguments
+    rocblas_local_handle handle;
+    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_device);
+    I m = argus.get<I>("m");
+    I n = argus.get<I>("n");
+    I k = argus.get<I>("k");
```
tfalders: I'd suggest setting n and k to default to m

clients/common/unit/testing_gemm.hpp
```diff
@@ -0,0 +1,425 @@
...
+    I inca = argus.get<I>("inca", 1);
+    I incb = argus.get<I>("incb", 1);
+    I incc = argus.get<I>("incc", 1);
+
+    char tA = argus.get<char>("transA");
+    char tB = argus.get<char>("transB");
```
tfalders: I'd suggest setting tA and tB to default to 'N'

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
```diff
@@ -201,41 +235,79 @@ rocblas_status rocsolver_gemm(rocblas_handle handle,
...
-                                strideB, beta, C, shiftC, incc, ldc, strideC);
+        const auto warpSize = deviceProperties.warpSize;
+
+        // launch specialized kernel
+        const I numWarpsX = 4;
+        const I numWarpsY = 4;
+        const I blocksx = (m + (numWarpsX * 16 - 1)) / (numWarpsX * 16);
+        const I blocksy = (n + (numWarpsY * 16 - 1)) / (numWarpsY * 16);
+        dim3 grid(blocksx, blocksy, batch_count);
+        dim3 threads(numWarpsX * warpSize, numWarpsY, 1);
```
tfalders, line 259: This is a bit confusing for me. It looks like numWarpsY is actually the number of threads in y? What are numWarpsX * 16 and numWarpsY * 16 supposed to represent?

tfalders (approved):
The MFMA stuff is Greek to me, but everything else looks good. Just have some minor comments

library/src/specialized/roclapack_gemm_specialized_kernels.hpp
EdDAzevedo: Minor suggestion: it may be safer to query at execution time the actual value of "warpSize" and perhaps the max number of threads in a thread block.

Perhaps consider using a lambda  expression to calculate the ceiling function, say

auto ceil = [](auto n, auto b) { return( (n-1)/b + 1 ); };   

and use it as something like

const I blocksx = ceil( m,   (numWarpsX * 16) )

AGonzales-amd: @tfalders, numWarps variables represent how many warps are working in parallel for each dimension. In this case, a thread group is laid out as 4x4 warps:
| | | | |
|---|---|---|----|
|w|w|w|w|
|w|w|w|w|
|w|w|w|w|
|w|w|w|w|

Each warp works on a 16x16 block of the output matrix so the m and n dimensions are ceil divided by numWarps * 16.

@EdDAzevedo I think a ceil function would be valuable and should be available everywhere. I think we should add this separately to change all ceil operations to a function call.

clients/common/unit/testing_gemm.hpp
AGonzales-amd: done

clients/common/unit/testing_gemm.hpp
AGonzales-amd: done

clients/gtest/unit/gemm_gtest.cpp
AGonzales-amd: done

tfalders:
Test failure in SYGVDX likely unrelated. Forcing the merge.

PR #883, opened 2025-01-16
use rocjenkins xml instead of hardcoded

amd-garydeng:
Math-ci would like to be able to set what node static analysis job falls on.


samjwu (approved):

jmachado-amd (approved):
Thanks @amd-garydeng!

PR #886, opened 2025-01-23
Stedc merge optimization

jzuniga-amd:


library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1566,6 +1572,336 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+            // find degree of secular equation
+            rocblas_int dd = 0;
+            for(int i = 0; i < sz; ++i)
+            {
+                if(mask[i] == 1)
+                    dd++;
+            }
+
+/*            // Order the elements in tmpd and zz using a simple parallel selection/bubble sort.
+            // This will allow us to find initial intervals for eigenvalue guesses
```
EdDAzevedo: Just a reminder that if the bubble sort is a performance bottleneck then perhaps  consider using the quick sort routine.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2053,9 +2445,9 @@ void local_gemm(rocblas_handle handle,
     // Execute A*B -> temp -> A
 
     // everything must be executed with scalars on the host
-    rocblas_pointer_mode old_mode;
+/*    rocblas_pointer_mode old_mode;
     rocblas_get_pointer_mode(handle, &old_mode);
-    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+    rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);*/
     S one = 1.0;
     S zero = 0.0;
```
EdDAzevedo, line 2094: Perhaps there can be a separate kernel for performing the combined copy's and gemm's for computing "A = A * op(B)" or "A = op(B) * A".  This combined kernel might also be more cache friendly.   If the matrices are sufficiently small, one may store each batch entry for A, B in LDS for computations and write out results immediately to device memory.     Just a thought.

EdDAzevedo (approved):

jmachado-amd:
@jzuniga-amd, it appears that some `stedc` tests are failing; I'm running them again.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1742,74 +1749,106 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
-                    temps[i + (p2 + j) * n] = valf;
-                }
-                inrms[tidb] = nrm;
-                __syncthreads();
-
-                // reduction (for the norms)
-                for(int r = dim / 2; r > 0; r /= 2)
-                {
-                    if(tidb < r)
+                    // compute vectors of rank-1 perturbed system and their norms
```
jmachado-amd, line 1763: It appears that, when `USEGEMM` is set, this part of the code only updates the non-deflated entries of `tempgemm` (through the `putvec` pointer). However, the subsequent `gemm` will use the whole `n * n` first part of `tempgemm` which could lead to issues if the memory pointed by `tempgemm` contains `NaN`s.

`tempgemm` should be initialized before those kernels are run for the first time, to prevent spurious `NaN`s from propagating into the computed eigenvectors.

jmachado-amd (approved):
Looks good to me!

jzuniga-amd:
I just realized that the initialization of `tempgemm` in the `stedc_impl` function will have effect only for calls to the stedc C API.  This effectively corrects the CI errors with stedc, however, the problem could potentially still exist for other functions that call stedc internally (like the eigensolvers syevd, syevdj, etc.) 
So, even if we don't have CI failures for other functions, I'm moving the `tempgemm` array initialization to the `stedc_template` function... just to be sure.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2367,9 +2404,11 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
...
 
         // launch merge for level k
-        /** TODO: using max number of levels for now. Kernels return immediately when passing
-            the actual number of levels in the split block. We should explore if synchronizing
-            to copy back the actual number of levels makes any difference **/
+        // TODO: using max number of levels for now. Kernels return immediately when surpassing
+        // the actual number of levels in the split block. We should explore if synchronizing
+        // to copy back the actual number of levels makes any difference.
+        // TODO: the code computing the context of the level (first part of each kernel) could be
+        // resused.
```
tfalders: Spelling on reused

tfalders (approved):
LGTM! I think it might need a pass through clang format though. The indentation on line 2485 is incorrect.

PR #897, opened 2025-02-21
Use new LASR in STEQR and BDSQR and parallelize STEQR

tfalders:


jzuniga-amd:
Looks good. Adding a comment below to trigger the performance tracking system (PTS) so that we can verify performance improvements.

jzuniga-amd:
%PTS
gesvd

EdDAzevedo (approved):

EdDAzevedo (approved):

library/src/include/lib_device_helpers.hpp
```diff
@@ -109,11 +109,16 @@ __device__ __host__ T find_max_tridiag(const rocblas_int start, const rocblas_in
...
+                                       const rocblas_int end,
+                                       T* D,
+                                       T* E,
+                                       T scale,
+                                       const rocblas_int tid = 0,
+                                       const rocblas_int tid_inc = 1)
 {
     D[end] *= scale;
-    for(int i = start; i < end; i++)
+    for(int i = tid; i < end; i += tid_inc)
```
jzuniga-amd: and how is `start` taken into consideration now?
The old code scales elements in [start, end] 
This new multithreaded code scales elements in [first_tid, end]... what if the first thread index is not equal to `start`?

library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -62,126 +59,148 @@ __device__ void run_steqr(const rocblas_int n,
...
         if(lend == l)
             continue;
 
         // Scale submatrix
-        anorm = find_max_tridiag(l, lend, D, E);
         if(anorm == 0)
             continue;
         else if(anorm > ssfmax)
-            scale_tridiag(l, lend, D, E, anorm / ssfmax);
+            scale_tridiag(lsv, lendsv, D, E, anorm / ssfmax, tid, tid_inc);
```
jzuniga-amd, line 113: these `scale_tridiag` calls are being done by all threads (from `tid=0` to `tid=63`) to scale elements in `[lsv, lendsv]`
I don't understand what happened when `lsv != 0` ?
(See the comments about `scale_tridiag` above)

library/src/include/lib_device_helpers.hpp
tfalders: Whoops! Nice catch! I was supposed to add `start` to `tid`

library/src/auxiliary/rocauxiliary_steqr.hpp
EdDAzevedo: another idea is to consider using  cooperative groups and partition the threads in the  thread block (cg::this_thread_block() )   into smaller subsets of threads. One can then  pass in this thread group  "g"  to use  g.thread_rank() and g.size() and g.sync()   methods. 
Just a suggestion.

jzuniga-amd (approved):
LGTM

jmachado-amd (approved):

PR #898, opened 2025-02-25
add rocm-examples

amd-garydeng:
Math-CI has a request to add ROCM-Examples to precheckin testing.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -52,7 +52,7 @@ def runCompileCommand(platform, project, jobName, boolean sameOrg=false, boolean
     platform.runCommand(this, command)
 }
 
-def runTestCommand (platform, project, gfilter)
+def runTestCommand (platform, project, gfilter, boolean rocmExamples=false)
 {
     String buildType = project.buildName.contains('Debug') ? 'debug' : 'release'
     String hmmTestCommand = platform.jenkinsLabel.contains('gfx90a') ? 'HSA_XNACK=1 ./rocsolver-test --gtest_filter=*MANAGED_MALLOC* || true' : ''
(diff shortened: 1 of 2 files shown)


.jenkins/common.groovy
```diff
@@ -71,6 +71,33 @@ def runTestCommand (platform, project, gfilter)
...
+            buildString += "sudo dpkg -i *.deb"
+        }
+        else {
+            buildString += "sudo rpm -i *.rpm"
+        }
+        testCommand = """#!/usr/bin/env bash
+                    set -ex
+                    cd ${project.paths.project_build_prefix}/build/release/package
+                    ${buildString}
+                    cd ../../..
```
cgmb, line 87: I'd suggest using `pushd` and `popd`, rather than worrying about nesting levels for `cd`.

.jenkins/common.groovy
```diff
@@ -71,6 +71,33 @@ def runTestCommand (platform, project, gfilter)
...
+                    testDirs=("Libraries/rocSOLVER")
+                    git clone https://github.com/ROCm/rocm-examples.git
+                    rocm_examples_dir=\$(readlink -f rocm-examples)
+                    for testDir in \${testDirs[@]}; do
+                        cd \${rocm_examples_dir}/\${testDir}
+                        cmake -S . -B build
+                        cmake --build build
+                        cd ./build
+                        ctest --output-on-failure
```
cgmb, line 96: Would `ctest --output-on-failure --test-dir build` work? That seems consistent with the previous configure and build commands that specify the directory.

.jenkins/common.groovy
```diff
@@ -71,6 +71,33 @@ def runTestCommand (platform, project, gfilter)
...
+        testCommand = """#!/usr/bin/env bash
+                    set -ex
+                    cd ${project.paths.project_build_prefix}/build/release/package
+                    ${buildString}
+                    cd ../../..
+                    testDirs=("Libraries/rocSOLVER")
+                    git clone https://github.com/ROCm/rocm-examples.git
+                    rocm_examples_dir=\$(readlink -f rocm-examples)
+                    for testDir in \${testDirs[@]}; do
+                        cd \${rocm_examples_dir}/\${testDir}
```
cgmb, line 92: It's perhaps a matter of taste, but since you don't seem to have been familiar with `pushd` and `popd`, perhaps I thought I might suggest that this would also be a good place to apply them? Then you don't need to use readlink to get the absolute path. You can just `cd rocm-examples`, and `pushd ${testDir}` at the start of the `for`, then `popd` at the end of it.

Although, given that testDirs is a single directory perhaps we don't even need that? We could just omit all the complexity related to handling multiple directories and just cd into `rocm-examples/Libraries/rocSOLVER` and run the build and test commands.

.jenkins/precheckin.groovy
```diff
@@ -39,7 +39,7 @@ def runCI =
         platform, project->
 
         def gfilter = 'checkin*'
-        commonGroovy.runTestCommand(platform, project, gfilter)
+        commonGroovy.runTestCommand(platform, project, gfilter, true)
```
cgmb, line 42: It would be nice if groovy had a way to make this more readable. Just seeing this call, you need to either memorize what the arguments to `runTestCommand` are, or you need to look them up, in order to understand what `true` refers to.

cgmb (commented):
Thanks for the tests! Checking that examples build and run is certainly a useful thing. I'm not deeply involved in rocSOLVER anymore, so I'll leave it to others to approve the underlying concepts. It's all just nitpicks from me.

ammallya:

PR #899, opened 2025-02-25
Implement `gesdd`

jmachado-amd:
This PR adds a minimal implementation of `gesdd`, using `syevd` as a backend. A complete implementation, using `bdsdc`, will appear in the future.


clients/common/lapack/testing_gesdd.hpp
```diff
@@ -847,9 +847,9 @@ void testing_gesdd(Arguments& argus)
     // using 2 * min(m, n) * machine_precision as tolerance
```
EdDAzevedo: Is the expression in the comment still correct or it needs to be updated? Just checking.

library/src/lapack/roclapack_gesdd.cpp
```diff
@@ -77,42 +72,52 @@ rocblas_status rocsolver_gesdd_impl(rocblas_handle handle,
...
+    scalars = mem[2];
+    work1 = mem[3];
+    work2 = mem[4];
+    work3 = mem[5];
+    work4 = mem[6];
+    work5_ipiv = mem[7];
+    splits = mem[8];
+    tmptau_W = mem[9];
+    tau = mem[10];
+    workArr = mem[11];
```
EdDAzevedo, line 110: Just double checking whether mem[11] should be  reused.

Perhaps later, we might consider allocating just one big array and partition from "mem[0]", say   something like  std::byte *pfree = (std::byte *) mem[0];     VUtmp = pfree; pfree += size_VUtmp; UVtmpZ = pfree; pfree += size_UVtmpZ;  .... etc

library/src/lapack/roclapack_gesdd.cpp
```diff
@@ -132,10 +137,6 @@ rocblas_status rocsolver_sgesdd(rocblas_handle handle,
                                  const rocblas_int n,
                                  float* A,
                                  const rocblas_int lda,
-                                 const float abstol,
-                                 float* residual,
-                                 const rocblas_int max_sweeps,
-                                 rocblas_int* n_sweeps,
                                  float* S,
```
EdDAzevedo: Minor comment: Should there be "const" attribute also for the arrays, say  "float * const A" or "float * const S"?

EdDAzevedo (approved):
I wonder perhaps whether  these different algorithms  (say using Jacobi or forming A'*A or using bidiagonal form)  each have advantages and disadvantages and  can be retained but selected as an algorithm option by the application?   Just a thought.

library/src/lapack/roclapack_gesdd.hpp
```diff
@@ -0,0 +1,507 @@
...
+    for(b = bid; b < batch_count; b += gridDim.y)
+    {
+        // array pointers
+        S* Sigma = SS + b * strideS;
+        T* D = R + b * strideR;
+        T* U = Q + b * strideQ;
+
+        for(j = tid; j < n; j += gridDim.x * blockDim.x)
+        {
+            for(k = 0; k < n; k++)
```
EdDAzevedo, line 75: I wonder whether the code can use more threads or more parallelism, say batch index  b in z-dimension, j in x-dimension (as inner loop) , k in y-dimension.  Just a thought.

clients/common/lapack/testing_gesdd.hpp
jmachado-amd: The expression in the comment matches the code, you are probably looking at an old commit.

library/src/lapack/roclapack_gesdd.cpp
jmachado-amd: Although we do need to improve memory usage in rocSOLVER going forward, `mem[11]` is only used for `workArr`.

library/src/lapack/roclapack_gesdd.hpp
```diff
@@ -0,0 +1,507 @@
...
+                                   size_t* size_work4,
+                                   size_t* size_work5_ipiv,
+                                   size_t* size_splits,
+                                   size_t* size_tmptau_W,
+                                   size_t* size_tau,
+                                   size_t* size_workArr,
+                                   size_t* size_workArr2)
+{
+    // If quick return, set workspace to zero
+    if(n == 0 || m == 0 || batch_count == 0)
```
EdDAzevedo, line 240: minor comment: perhaps set all sizes to zero first, then check if (n==0) and return. This is just defensive programming to make sure all the size variables are initialized, no undefined values.  Just a minor suggestion.

library/src/lapack/roclapack_gesdd.hpp
```diff
@@ -0,0 +1,507 @@
...
+    *size_scalars = std::max({a1, a2});
+    *size_work1 = std::max({b1});
+    *size_work2 = std::max({c1, b2, b3});
+    *size_work3 = std::max({d1, c2, c3});
+    *size_work4 = std::max({d2, d3});
+    *size_work5_ipiv = std::max({e2});
+    *size_splits = std::max({f1});
+    *size_tmptau_W = std::max({g1});
+    *size_tau = std::max({h1});
+    *size_workArr = sizeof(T) * std::max({m, n}) * std::max({m, n}) * batch_count;
```
EdDAzevedo: Just double checking whether the size_workArr   is max(m,n)^2 * batch_count. It might be a very big number.

library/src/lapack/roclapack_gesdd.cpp
jmachado-amd: Not sure about which code this is targeting, but in `roclapack_gesdd.cpp` there is no batched code, thus there is no need for `T* const A[]` or similar constructions (where `T` is an arbitrary type) -- if this is what you are asking.

library/src/lapack/roclapack_gesdd.hpp
jmachado-amd: Yes, this looks like a typo, and I'll check it later. Thanks for catching that!

amd-jnovotny:
@jmachado-amd : Would this feature require a changelog update? If you're planning on adding that as a separate PR, then no worries.

library/src/lapack/roclapack_gesdd.hpp
jmachado-amd: As it turns out, that code was semantically incorrect: it was supposed to reserve space for the off-diagonal elements of the input matrix, as required by `stedc`. Thanks again for catching it!

amd-jnovotny (approved):

library/src/lapack/roclapack_gesdd.hpp
jmachado-amd: Done.

library/src/lapack/roclapack_gesdd.hpp
jmachado-amd: It can be improved, but this is intended as an intermediate step and not the final algorithm.

jmachado-amd:
@EdDAzevedo, my opinion is that svd algorithms based on creating the explicit products `A A^*` or `A^* A` are always an intermediate step and should not stay in the long run. However, I can't ignore the fact that they can be quite efficient sometimes, and some users will prefer to have the option of selecting those during runtime.

library/src/lapack/roclapack_gesdd.hpp
```diff
@@ -0,0 +1,510 @@
...
+        // Extra requirements for temporary V & U storage
+        *size_VUtmp = sizeof(T) * n * n * batch_count;
+        if(!leftv)
+            size_UVtmp = sizeof(T) * m * n * batch_count;
+    }
+    else
+    {
+        // Requirements for Divide-and-Conquer eigensolver
+        rocsolver_syevd_heevd_getMemorySize<BATCHED, T, SS>(
+            rocblas_evect_original, rocblas_fill_upper, n, batch_count, &a1, &b1, &c1, &d1, &e1,
```
jzuniga-amd: shouldn't the eigensolver be applied to a m-by-m matrix when m < n ??
It may save some memory

jmachado-amd: Yes, this is a typo -- many thanks for catching it!

For what it is worth, I've reread the code and could not find other typos; but I am not sure whether I'd be able to actually find any of those at this point.

jzuniga-amd (approved):
LGTM

PR #900, opened 2025-02-26
Remove support for gfx940, gfx941

qjojo:
Removes a switch needed for gfx940 and gfx941 as ROCm does not support those targets anymore.

### DIFF
--- a/library/src/specialized/roclapack_trsm_specialized_kernels.hpp
+++ b/library/src/specialized/roclapack_trsm_specialized_kernels.hpp
@@ -859,14 +859,6 @@ rocblas_status rocsolver_trsm_lower(rocblas_handle handle,
     }
 #endif
 
-    // TODO: Some architectures require synchronization between rocSOLVER and rocBLAS kernels; more investigation needed
-    int device;
-    HIP_CHECK(hipGetDevice(&device));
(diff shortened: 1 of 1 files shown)


library/src/specialized/roclapack_trsm_specialized_kernels.hpp
```diff
@@ -866,8 +866,6 @@ rocblas_status rocsolver_trsm_lower(rocblas_handle handle,
     HIP_CHECK(hipGetDeviceProperties(&deviceProperties, device));
     std::string deviceFullString(deviceProperties.gcnArchName);
     std::string deviceString = deviceFullString.substr(0, deviceFullString.find(":"));
```
tfalders: I believe you can remove all of this as well (and the 4 lines above), since it was only needed for the switch

library/src/specialized/roclapack_trsm_specialized_kernels.hpp
```diff
@@ -1151,8 +1137,6 @@ rocblas_status rocsolver_trsm_upper(rocblas_handle handle,
...
```
tfalders: Same here

tfalders (approved):

jmachado-amd (approved):
Looks good to me!

PR #901, opened 2025-02-26
Use stedc instead of sterf in syevd

tfalders:
Performance numbers reveal that running syevd with stedc is orders of magnitude faster than with sterf, so this PR changes syevd to use stedc instead of sterf.

To preserve access to the numerical properties of sterf, syevd will still use sterf if hybrid sterf is enabled, since the performance difference there is much less.

### DIFF
--- a/clients/common/lapack/testing_syevd_heevd.hpp
+++ b/clients/common/lapack/testing_syevd_heevd.hpp
@@ -422,15 +422,15 @@ void testing_syevd_heevd(Arguments& argus)
     rocblas_int bc = argus.batch_count;
     rocblas_int hot_calls = argus.iters;
 
-    EXPECT_ROCBLAS_STATUS(
-        rocsolver_set_alg_mode(handle, rocsolver_function_sterf, rocsolver_alg_mode_hybrid),
-        rocblas_status_success);
+    // EXPECT_ROCBLAS_STATUS(
+    //     rocsolver_set_alg_mode(handle, rocsolver_function_sterf, rocsolver_alg_mode_hybrid),
+    //     rocblas_status_success);
 
-    rocsolver_alg_mode alg_mode;
-    EXPECT_ROCBLAS_STATUS(rocsolver_get_alg_mode(handle, rocsolver_function_sterf, &alg_mode),
-                          rocblas_status_success);
+    // rocsolver_alg_mode alg_mode;
+    // EXPECT_ROCBLAS_STATUS(rocsolver_get_alg_mode(handle, rocsolver_function_sterf, &alg_mode),
+    //                       rocblas_status_success);
 
-    EXPECT_EQ(alg_mode, rocsolver_alg_mode_hybrid);
+    // EXPECT_EQ(alg_mode, rocsolver_alg_mode_hybrid);
 
     // check non-supported values
     if(uplo == rocblas_fill_full || evect == rocblas_evect_tridiagonal)
--- a/clients/common/lapack/testing_syevd_heevd.hpp
+++ b/clients/common/lapack/testing_syevd_heevd.hpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2021-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2021-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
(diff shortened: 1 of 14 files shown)


clients/common/lapack/testing_syevd_heevd.hpp
```diff
@@ -422,6 +422,19 @@ void testing_syevd_heevd(Arguments& argus)
...
+    {
+        EXPECT_ROCBLAS_STATUS(
+            rocsolver_set_alg_mode(handle, rocsolver_function_sterf, rocsolver_alg_mode_hybrid),
+            rocblas_status_success);
+
+        rocsolver_alg_mode alg_mode;
+        EXPECT_ROCBLAS_STATUS(rocsolver_get_alg_mode(handle, rocsolver_function_sterf, &alg_mode),
+                              rocblas_status_success);
+
+        EXPECT_EQ(alg_mode, rocsolver_alg_mode_hybrid);
```
EdDAzevedo, line 435: Just double checking whether if alg_mode is set, it must be hybrid mode.

clients/gtest/lapack/syevd_heevd_gtest.cpp
```diff
@@ -103,11 +105,19 @@ class SYEVD_HEEVD : public ::TestWithParam<syevd_heevd_tuple>
     }
 };
 
-class SYEVD : public SYEVD_HEEVD
+class SYEVD : public SYEVD_HEEVD<0>
```
EdDAzevedo, line 108: Just double checking whether "MODE" is an enum value such as rocsolver_alg_mode_hybrid. If so, perhaps it might be clearer to use the enum name.               Just a thought.

library/src/lapack/roclapack_syevd_heevd.hpp
```diff
@@ -184,7 +177,7 @@ rocblas_status rocsolver_syevd_heevd_template(rocblas_handle handle,
                                             strideE, tau, n, batch_count, scalars, (T*)work1,
                                             (T*)work2, tmptau_W, workArr);
 
-    if(evect != rocblas_evect_original)
+    if(sterf_mode == rocsolver_alg_mode_hybrid && evect != rocblas_evect_original)
```
EdDAzevedo, line 192: Minor comment: do we need more enum option  values for the new algorithm?  Just a thought.

EdDAzevedo (approved):

clients/common/lapack/testing_syevd_heevd.hpp
tfalders: Currently, alg_mode only has two valid values - hybrid (1) or not hybrid (0) - but you make a good point. To future proof it, I should probably check that it is actually equal to 1.

clients/gtest/lapack/syevd_heevd_gtest.cpp
tfalders: Hmmmm, it might be worth revisiting. IIRC, I made it 0 and 1 instead of an enum because it comes into rocsolver-bench that way, and also because other arguments in rocsolver_arguments.hpp are int types. Which aren't particularly compelling reasons to keep it that way.

library/src/lapack/roclapack_syevd_heevd.hpp
```diff
@@ -83,26 +83,15 @@ void rocsolver_syevd_heevd_getMemorySize(const rocblas_evect evect,
     rocsolver_sytrd_hetrd_getMemorySize<BATCHED, T>(n, batch_count, size_scalars, &w11, &w21, &t1,
                                                     &unused);
 
+    // extra requirements for computing eigenvalues and vectors (stedc)
+    rocsolver_stedc_getMemorySize<BATCHED, T, S>(rocblas_evect_tridiagonal, n, batch_count, &w31,
+                                                 &w22, &w12, size_tmpz, size_splits, &unused);
```
jzuniga-amd: I think we don't need these requirements when `hybrid && no-vectors`
(And I guess sterf_hybrid does not need workspace, correct?)
Maybe we don't save anything because of the max values below, but it would be worth checking.

library/src/lapack/roclapack_syevd_heevd.hpp
```diff
@@ -200,16 +193,19 @@ rocblas_status rocsolver_syevd_heevd_template(rocblas_handle handle,
...
-            handle, rocblas_side_left, uplo, rocblas_operation_none, n, n, A, shiftA, lda, strideA,
-            tau, n, tmptau_W, 0, ldw, strideW, batch_count, scalars, (T*)work2, (T*)work1,
-            (T*)work3, workArr);
-
-        // copy matrix product into A
-        const rocblas_int copyblocks = (n - 1) / BS2 + 1;
-        ROCSOLVER_LAUNCH_KERNEL(copy_mat<T>, dim3(copyblocks, copyblocks, batch_count),
-                                dim3(BS2, BS2), 0, stream, n, n, tmptau_W, 0, ldw, strideW, A,
-                                shiftA, lda, strideA);
+        if(evect == rocblas_evect_original)
```
jzuniga-amd, line 210: For future reference, it would be good to add a comment somewhere in here saying that divide&conquer `stedc` is used even if no vectors are required

library/src/lapack/roclapack_syevd_heevd.hpp
tfalders: I left it in mostly because the other hybrid methods also take the same amount of workspace as if they were running in GPU-only mode. I'll go ahead and change that.

tfalders: Oh, I see now. The getMemorySize functions don't take in the rocblas_handle, which means I can't check if we're in hybrid mode or not. Updating them will touch many other files, so it might be worth leaving that for a separate PR. EDIT: For sterf, anyway. It shouldn't be much trouble to fix it for syevd.

library/src/lapack/roclapack_syevd_heevd.hpp
tfalders: Done

jzuniga-amd (approved):
LGTM

jmachado-amd (approved):

PR #908, opened 2025-03-12
Optimize latrd reduction for sytrd

qjojo:
Reduces overhead from consecutive gemv calls in latrd by manually computing the column reduction in a bespoke kernel. On my machine this speeds up latrd by about 30%. There is one more gemv and some other odds and ends that can be moved into the reduction kernel but in the interest of deadlines I want to get this going through CI now.


qjojo:
Just realized the upper kernel has the wrong indices, will push a fix asap

qjojo:
Rejecting in favor of https://github.com/ROCm/rocSOLVER/pull/914

PR #909, opened 2025-03-12
Use gfx11-generic targets

cgmb:
The gfx11-generic build target would reduce build times and binary size by replacing four ISAs with one.


cgmb:
There is a bug in `rocm_check_target_ids`, so making the target optional would require https://github.com/ROCm/rocm-cmake/pull/265.

cgmb:
The gfx11-generic target requires `-mcode-object-version=6` for amdclang++ older than ROCm 6.4. Using COV6 with versions of clang from before [this December commit](https://github.com/ROCm/llvm-project/commit/d3eebcd1eab4b6694772a901d6d01c032874e0ca) will result in a warning that "code object v6 is still in development and not ready for production use yet."

cgmb:
The [User Guide for AMDGPU Backend](https://llvm.org/docs/AMDGPUUsage.html) has some excellent information about the generic targets. For gfx12-generic, there should be no drawbacks. For gfx11-generic, there are some instruction restrictions that may or may not affect performance.

I'm happy to see the tests passing, but I think I will split this into two PRs because the gfx12-generic update is a no-brainer and the gfx11-generic update may require a bit of consideration.

ammallya:

PR #911, opened 2025-03-13
Fix tests in debug build

AGonzales-amd:
Proposed changes:
 - Increase debug mode test timeout limit from 45min to 5h.
 - check for NaN before comparing values in device functions.
 - check for nullptr before shifting pointer in bdsqr.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_bdsqr.hpp
+++ b/library/src/auxiliary/rocauxiliary_bdsqr.hpp
@@ -90,10 +90,13 @@ __device__ void bdsqr_QRstep(const rocblas_int tid,
                              S* D,
                              S* E,
                              T* V,
+                             const rocblas_int shiftV,
                              const rocblas_int ldv,
                              T* U,
+                             const rocblas_int shiftU,
                              const rocblas_int ldu,
                              T* C,
+                             const rocblas_int shiftC,
                              const rocblas_int ldc,
                              const S sh,
                              S* rots)
@@ -176,18 +179,18 @@ __device__ void bdsqr_QRstep(const rocblas_int tid,
     rocblas_direct direc = (t2b ? rocblas_forward_direction : rocblas_backward_direction);
     if(V && nv)
     {
-        run_lasr(rocblas_side_left, rocblas_pivot_variable, direc, n, nv, rots, rots + n, V, ldv,
+        run_lasr(rocblas_side_left, rocblas_pivot_variable, direc, n, nv, rots, rots + n, V + shiftV, ldv,
                  tid, tid_inc);
     }
     if(U && nu)
     {
         run_lasr(rocblas_side_right, rocblas_pivot_variable, direc, nu, n, rots + nr, rots + nr + n,
-                 U, ldu, tid, tid_inc);
+                 U + shiftU, ldu, tid, tid_inc);
     }
     if(C && nc)
     {
         run_lasr(rocblas_side_left, rocblas_pivot_variable, direc, n, nc, rots + nr, rots + nr + n,
-                 C, ldc, tid, tid_inc);
+                 C + shiftC, ldc, tid, tid_inc);
     }
 }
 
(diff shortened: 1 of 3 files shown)


library/src/include/lib_device_helpers.hpp
```diff
@@ -287,6 +291,10 @@ __device__ static void shell_sort_descending(const I n, S* a, I* map = nullptr)
     {
         for(auto k = k_start; k < (n - 1); k += k_inc)
         {
+            // if NaN
+            if(a[k] != a[k])
```
EdDAzevedo: Is there something like  "std::isnan()" function or perhaps rocblas has it as a utility?   I have a concern the compiler might be too smart and optimize this away.

Should  we expect a NaN anyway?  If not, perhaps the problem is elsewhere?

library/src/include/lib_device_helpers.hpp
```diff
@@ -203,6 +203,10 @@ __device__ static void shell_sort_ascending(const I n, S* a, I* map = nullptr)
...
+                continue;
+
```
EdDAzevedo: perhaps use something less intrusive like  

assert(  isnan(a[k]) || isnan(a[k+1]) || (a[k] <= a[k+1]) );

library/src/auxiliary/rocauxiliary_bdsqr.hpp
```diff
@@ -604,8 +607,8 @@ ROCSOLVER_KERNEL void bdsqr_compute(const rocblas_int n,
             if(tid == 0)
                 splits[4 * sid] = (t2b ? 1 : -1);
 
-            bdsqr_QRstep(tid, tid_inc, t2b, k - i + 1, nv, nu, nc, D + i, E + i, V + i, ldv,
-                         U + i * ldu, ldu, C + i, ldc, smin, rots + incW * i);
+            bdsqr_QRstep(tid, tid_inc, t2b, k - i + 1, nv, nu, nc, D + i, E + i, V, i, ldv,
+                         U, i * ldu, ldu, C, i, ldc, smin, rots + incW * i);
```
EdDAzevedo: do we need the shift variables  to be 64bit variable (say type rocblas_stride)  and compute in 64bit, say i * static_cast<int64_t>(ldu)  ?   Just a thought.

library/src/include/lib_device_helpers.hpp
```diff
@@ -204,7 +204,7 @@ __device__ static void shell_sort_ascending(const I n, S* a, I* map = nullptr)
         for(auto k = k_start; k < (n - 1); k += k_inc)
         {
             // if NaN
-            if(a[k] != a[k])
+            if(std::isnan(a[k]) || std::isnan(a[k + 1]))
```
EdDAzevedo: I have a concern that this may have an impact even for non-debug build.   Perhaps consider incorporating the isnan() condition in the assert() with something like

assert( isnan(a[k]) || isnan(a[k+1]) || (a[k] <= a[k+1]) );

library/src/include/lib_device_helpers.hpp
```diff
@@ -292,7 +292,7 @@ __device__ static void shell_sort_descending(const I n, S* a, I* map = nullptr)
...
```
EdDAzevedo: Perhaps  consider something like the following so there is not impact for non-debug build

assert( isnan(a[k]) || isnan(a[k+1]) || (a[k] >= a[k+1] ) );

library/src/include/lib_device_helpers.hpp
AGonzales-amd: This code snippet is within an NDEBUG macro check, so it shouldn't affect non-debug builds.

library/src/include/lib_device_helpers.hpp
AGonzales-amd: done

library/src/auxiliary/rocauxiliary_bdsqr.hpp
AGonzales-amd: fixed

EdDAzevedo (approved):

tfalders (approved):

tfalders:
CI failures seem to be unrelated. Forcing the merge.

PR #914, opened 2025-03-14
Sytrd optimizations

jzuniga-amd:
Optimizations:
+ use general methods (gemm and gemv) instead of symmetric (syr2k and symv) 
+ use internal gemv-like kernels for the updates and computing W
+ some tuning for kernel configurations, modes, etc.
+ code is organized and ready for further optimizations of the internal gemv-like kernels

(I still need to add the dual kernels for the uplo = rocblas_fill_upper case.
I'll update the PR with them asap)

### DIFF
--- a/library/src/auxiliary/rocauxiliary_latrd.cpp
+++ b/library/src/auxiliary/rocauxiliary_latrd.cpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2019-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2019-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -91,9 +91,9 @@ rocblas_status rocsolver_latrd_impl(rocblas_handle handle,
         init_scalars(handle, (T*)scalars);
 
     // execution
-    return rocsolver_latrd_forsytrd_template<T>(handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau,
-                                       strideP, W, shiftW, ldw, strideW, batch_count, (T*)scalars,
-                                       (T*)work, (T*)norms, (T**)workArr);
+    return rocsolver_latrd_forsytrd_template<T>(
+        handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau, strideP, W, shiftW, ldw,
+        strideW, batch_count, (T*)scalars, (T*)work, (T*)norms, (T**)workArr);
 }
 
 ROCSOLVER_END_NAMESPACE
--- a/library/src/auxiliary/rocauxiliary_latrd.cpp
+++ b/library/src/auxiliary/rocauxiliary_latrd.cpp
@@ -91,7 +91,7 @@ rocblas_status rocsolver_latrd_impl(rocblas_handle handle,
         init_scalars(handle, (T*)scalars);
 
     // execution
-    return rocsolver_latrd_template<T>(handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau,
+    return rocsolver_latrd_forsytrd_template<T>(handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau,
                                        strideP, W, shiftW, ldw, strideW, batch_count, (T*)scalars,
                                        (T*)work, (T*)norms, (T**)workArr);
 }
--- a/library/src/auxiliary/rocauxiliary_latrd.cpp
+++ b/library/src/auxiliary/rocauxiliary_latrd.cpp
@@ -91,9 +91,9 @@ rocblas_status rocsolver_latrd_impl(rocblas_handle handle,
         init_scalars(handle, (T*)scalars);
 
     // execution
-    return rocsolver_latrd_forsytrd_template<T>(
-        handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau, strideP, W, shiftW, ldw,
-        strideW, batch_count, (T*)scalars, (T*)work, (T*)norms, (T**)workArr);
+    return rocsolver_latrd_template<T>(handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau,
+                                       strideP, W, shiftW, ldw, strideW, batch_count, (T*)scalars,
+                                       (T*)work, (T*)norms, (T**)workArr);
 }
 
 ROCSOLVER_END_NAMESPACE
--- a/library/src/auxiliary/rocauxiliary_latrd.hpp
+++ b/library/src/auxiliary/rocauxiliary_latrd.hpp
@@ -1,4 +1,4 @@
-/************************************************************************ 
+/************************************************************************
  * Derived from the BSD3-licensed
  * LAPACK routine (version 3.7.1) --
  *     Univ. of Tennessee, Univ. of California Berkeley,
@@ -407,7 +407,7 @@ ROCSOLVER_KERNEL void latrd_reduce_kernel(const rocblas_fill uplo,
     {
         i = ii * totalthsr + idr;
         val = 0;
-        
+
         it = (i < c) ? i : i - c;
         y = (i < c) ? y1 : y2;
 
@@ -545,7 +545,7 @@ ROCSOLVER_KERNEL void latrd_upper_updateA_kernel(const rocblas_int mm,
             __syncthreads();
         }
 
-        // write results 
+        // write results
         if(tidc == 0 && i < m)
             y[i] = ac;
     }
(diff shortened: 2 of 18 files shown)


library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -39,6 +39,133 @@
...
+
+            // operation for all rows
+            if(i < m && j < n)
+                ac-= A1[i + j * lda1] * sx1[j] + A2[i + j * lda2] * sx2[j];
+        }
+        acs[tidr + tidc * threadsr] = ac;
+        __syncthreads();
+
+        // group reduction
+        for(int r = threadsc / 2; r > 0; r /= 2)
```
EdDAzevedo: Maybe consider shfl_down() intrinsic for sum reduction but may be limited to a single warp.   Just a thought.

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -247,15 +251,1194 @@ rocblas_status rocsolver_latrd_template(rocblas_handle handle,
...
+    rocblas_set_pointer_mode(handle, old_mode);
+    return rocblas_status_success;
+}
+
+
+/***************** Kernels/Device functions *******************************************/
+/**************************************************************************************/
+
+template <typename T, typename U>
+ROCSOLVER_KERNEL void reduce_kernel(const rocblas_int m,
```
tfalders: I'd suggest prefixing the kernel name with latrd_, so we don't get a conflict if another function gets its own reduce_kernel.

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -247,15 +251,1194 @@ rocblas_status rocsolver_latrd_template(rocblas_handle handle,
...
+{
+    rocsolver_latrd_mode_symv,      // uses internal symv
+    rocsolver_latrd_mode_gemv_in,   // uses internal gemv
+    rocsolver_latrd_mode_gemv_out   // uses external gemv
+} rocsolver_latrd_mode;
+
+// Method to determine the mode depending on n and k
+// TODO: fine tuning may be required
+template<typename T>
+rocsolver_latrd_mode get_mode(const rocblas_int n, const rocblas_int k)
```
tfalders: Method name is too generic. I'd suggest prefixing with latrd_

Same for the get_config methods below.

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -247,15 +251,1194 @@ rocblas_status rocsolver_latrd_template(rocblas_handle handle,
...
+    T* dac = dacA + bid * strideD;
+
+    // rpgr is the number of rounds a group should run
+    // to cover all the rows
+    int ngrp = (m - 1) / threadsr + 1;
+    int rpgr = (ngrp - 1) / groupsr + 1;
+    int i, it;
+
+    // Registers/LDS:
+    extern __shared__ double smem[];
```
EdDAzevedo: Perhaps add a comment about how much LDS will be required.

library/src/auxiliary/rocauxiliary_latrd.hpp
EdDAzevedo: Perhaps also add "static" to hide the external symbol.  Just a thought.

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -247,15 +251,1194 @@ rocblas_status rocsolver_latrd_template(rocblas_handle handle,
...
+    ROCSOLVER_ENTER("latrd", "uplo:", uplo, "n:", n, "k:", k, "shiftA:", shiftA, "lda:", lda,
+                    "shiftW:", shiftW, "ldw:", ldw, "bc:", batch_count);
+
+    // quick return
+    if(n == 0 || k == 0 || batch_count == 0)
+        return rocblas_status_success;
+
+    hipStream_t stream;
+    rocblas_get_stream(handle, &stream);
+
```
EdDAzevedo, line 1953: Perhaps add code to explicitly set (and later restore) the pointer mode to be rocblas_pointer_mode_device as it is done in routine "latrd_template".   

This can ensure correct access to "*alpha" and "*beta" if rocblas gemv or rocblas symv is called on GPU without xnack.

library/src/auxiliary/rocauxiliary_latrd.hpp
jzuniga-amd: trying in the next optimization round

library/src/auxiliary/rocauxiliary_latrd.hpp
jzuniga-amd: good catch!!
I added the switch to _mode_device permanently in sytrd before calling latrd_forsytrd_template; this way we don't have to switch every call. 
(sytrd can now be always in _mode_device as syr2k is no longer used).

EdDAzevedo (approved):

tfalders:
Apparently you've got a merge conflict that needs to be resolved. I've also noticed that most of the files haven't had their copyright year updated.

jzuniga-amd:
It should be good now

qjojo (approved):

library/src/lapack/roclapack_sytrd_hetrd.hpp
```diff
@@ -201,16 +245,20 @@ rocblas_status rocsolver_sytrd_hetrd_template(rocblas_handle handle,
         while(j >= upkk)
         {
             // reduce columns j:j+k-1
-            rocsolver_latrd_template<T>(handle, uplo, j + k, k, A, shiftA, lda, strideA, E, strideE,
+            rocsolver_latrd_forsytrd_template<T>(handle, uplo, j + k, k, A, shiftA, lda, strideA, E, strideE,
                                         tau, strideP, tmptau_W, 0, ldw, strideW, batch_count,
                                         scalars, work, norms, workArr);
 
             // update unreduced block as a rank-2k update
```
tfalders: Should this comment be updated like the one for the previous case?

library/src/auxiliary/rocauxiliary_latrd.cpp
```diff
@@ -91,7 +91,7 @@ rocblas_status rocsolver_latrd_impl(rocblas_handle handle,
         init_scalars(handle, (T*)scalars);
 
     // execution
-    return rocsolver_latrd_template<T>(handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau,
+    return rocsolver_latrd_forsytrd_template<T>(handle, uplo, n, k, A, shiftA, lda, strideA, E, strideE, tau,
```
tfalders: Isn't the public API supposed to call the original code? If using the forsytrd variant, don't you also need to call the forsytrd variant of getMemorySize?

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -352,4 +349,1804 @@ rocblas_status rocsolver_latrd_template(rocblas_handle handle,
...
+                                    U yA,
+                                    const rocblas_int shiftY,
+                                    const rocblas_int ldy,
+                                    const rocblas_stride strideY,
+                                    T* workA,
+                                    const rocblas_stride strideblk)
+{
+    int bid = hipBlockIdx_z;
+    int bidr = hipBlockIdx_x;
+    int bidc = 0;
```
tfalders: I think here and for groupsc you can get the same results by using the HIP built-ins. It might be safer to use them for future-proofing.

jmachado-amd: I second this, at some (not too distant) point in the future we will want to remove all of those deprecated macros from our kernels (I mean `hipBlockIdx_x`, etc.).

library/src/auxiliary/rocauxiliary_latrd.cpp
tfalders: Seeing some test failures in CI for latrd. Guessing it's because the forsyrd variant of getMemorySize isn't being called.

tfalders (approved):
Looks good! I've left a few comments to be addressed. One of them should hopefully resolve the CI failures.

library/src/auxiliary/rocauxiliary_latrd.cpp
jmachado-amd: I think the issue is that the public API is meant to call `rocsolver_latrd_template`, as per your original comment.

library/src/lapack/roclapack_sytrd_hetrd.hpp
jzuniga-amd: done

library/src/auxiliary/rocauxiliary_latrd.cpp
jzuniga-amd: Yeah, it should be calling the original variant. I forgot to change it after I finished debugging the kernels for the upper case. It is corrected now.

library/src/auxiliary/rocauxiliary_latrd.hpp
jzuniga-amd: I think Troy meant to do `bidc = hipBlockIdx_y;`
But, yeah, also at some point we will need to replace all the `hipBlockIdx` by `BlockIdx` and so on in the entire library...

jmachado-amd (approved):
Looks good to me!

PR #917, opened 2025-03-20
Merge back rocm-rel-6.4 into develop

jzuniga-amd:


tfalders:
Is this needed? The Files Changed tab shows no changes.

jzuniga-amd:
Yeah, I wanted to check that we have all the 6.4 hot-fixes in develop. 
If we merge we'll add the 6.4 extra commits to the history, that's all. I may have already forgotten to do this for other releases so no big deal, I'm ok with merging it or not.

PR #924, opened 2025-04-01
LU factorization (GETRF) to use  64bit arithmetic for index address calculations

EdDAzevedo:
Minor change to getrf and getf2 to use 64bit arithmetic for index address calculations.  

This can allow large problems (say n=60,000) to be solved using 32bit rocblas_int version of rocsolver library. 

Similar changes may be applicable to Cholesky and QR factorization to solve large problems.


EdDAzevedo:
code not ready, still under development

PR #928, opened 2025-04-02
Fix CMAKE_INSTALL_PREFIX init behavior on Win32.

ScottTodd:
Per the guidance at https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html, `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` can be "used by project code to change the default without overriding a user-provided value".

This allows users and integration projects like https://github.com/ROCm/TheRock to set an install path of their choosing while still defaulting to this project's preferred value if the value is not already set. For more context, see https://github.com/ROCm/TheRock/pull/340.


cgmb (approved):
Thanks. That's a good improvement.

tfalders (approved):

ScottTodd:
Thanks for the reviews! Should I merge this myself, or would one of the project maintainers like to do that?

ScottTodd:
Same comment for https://github.com/ROCm/hipSOLVER/pull/386, which it looks like is awaiting one more review

>     At least 2 approving reviews are required by reviewers with write access.

PR #931, opened 2025-04-14
Replace deprecated GTest::GTest with GTest::gtest.

### LINKED ISSUE #419: [rocSOLVER] Replace unsupported GTest::GTest with GTest::gtest.
Upstream patch: https://github.com/ROCm/rocSOLVER/pull/931

stellaraccident:
Fixes https://github.com/ROCm/TheRock/issues/419

### DIFF
--- a/clients/extras/CMakeLists.txt
+++ b/clients/extras/CMakeLists.txt
@@ -46,7 +46,7 @@ if(UNIX AND BUILD_SHARED_LIBS)
       ROCSOLVER_LIB_NAME="$<TARGET_FILE:rocsolver>"
     )
     target_link_libraries(test-rocsolver-dlopen PRIVATE
-      $<IF:$<TARGET_EXISTS:GTest::gtest>, GTest::gtest, GTest::GTest>
+      $<IF:$<TARGET_EXISTS:GTest::gtest>,GTest::gtest,GTest::GTest>
       ${CMAKE_DL_LIBS}
     )
 
--- a/clients/extras/CMakeLists.txt
+++ b/clients/extras/CMakeLists.txt
@@ -45,8 +45,16 @@ if(UNIX AND BUILD_SHARED_LIBS)
     target_compile_definitions(test-rocsolver-dlopen PRIVATE
       ROCSOLVER_LIB_NAME="$<TARGET_FILE:rocsolver>"
     )
+    if(TARGET GTest::gtest)
+      target_link_libraries(test-rocsolver-dlopen PRIVATE
+        GTest::gtest
+      )
+    else()
+      target_link_libraries(test-rocsolver-dlopen PRIVATE
+        GTest::GTest
+      )
+    endif()
     target_link_libraries(test-rocsolver-dlopen PRIVATE
-      GTest::gtest
       ${CMAKE_DL_LIBS}
     )
 
--- a/clients/extras/CMakeLists.txt
+++ b/clients/extras/CMakeLists.txt
@@ -45,16 +45,8 @@ if(UNIX AND BUILD_SHARED_LIBS)
     target_compile_definitions(test-rocsolver-dlopen PRIVATE
       ROCSOLVER_LIB_NAME="$<TARGET_FILE:rocsolver>"
     )
-    if(TARGET GTest::gtest)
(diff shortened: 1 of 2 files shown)


cgmb (commented):
We should switch because the old target is deprecated, but this requires a minimum version of CMake 3.20. If you don't mind, I'll tweak this to prefer `GTest::gtest` but fall back to `GTest::GTest`. I think there's still folks building solver on Ubuntu 22.04 using CMake 3.16. They'll eventually need to upgrade, but this is easy enough to accommodate.

stellaraccident:
> We should switch because the old target is deprecated, but this requires a minimum version of CMake 3.20. If you don't mind, I'll tweak this to prefer `GTest::gtest` but fall back to `GTest::GTest`. I think there's still folks building solver on Ubuntu 22.04 using CMake 3.16. They'll eventually need to upgrade, but this is easy enough to accommodate.

Perfectly fine. About half of the projects are doing that fallback. Just when patching the other half, I didn't want to take the time without a need.

cgmb (approved):
This could also be done with the `$<TARGET_EXISTS>` [generator expression](https://cmake.org/cmake/help/v4.0/manual/cmake-generator-expressions.7.html#target-dependent-expressions) without adding any extra lines of code. I figured if/else might be easier to read and understand, but I'm not particularly attached to either method.

cgmb:
The generator expression alternative would replace `GTest::GTest` with `$<IF:$<TARGET_EXISTS:GTest::gtest>, GTest::gtest, GTest::GTest>`, making this a +2/-2 patch, like the original change. Anyway, I'm fine with whatever, so I'm setting this to auto-merge once it has sufficient reviews.

stellaraccident:
> The generator expression alternative would replace `GTest::GTest` with `$<IF:$<TARGET_EXISTS:GTest::gtest>, GTest::gtest, GTest::GTest>`, making this a +2/-2 patch, like the original change. Anyway, I'm fine with whatever, so I'm setting this to auto-merge once it has sufficient reviews.

The generator is concise, which is good. If we have to do it (i.e. support <3.20), that seems like the least bad way to do it.

TorreZuk:
$<TARGET_EXISTS:tgt>[¶](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_EXISTS)
Added in version 3.12.

cgmb:
> The generator is concise, which is good. If we have to do it (i.e. support <3.20), that seems like the least bad way to do it.

Sure. I waffled on this and made the changes yesterday, but paused at the push stage only because using a generator adds subtleties that I don't entirely understand to `install(EXPORT)`. However, given that the targets using GTest aren't exported anyway, it doesn't matter.

I will push the update later today.

> Added in version 3.12.

We're requiring at 3.13 anyway, so that should be fine.

cgmb:
Hmmm... the generator method seems to somehow accidentally be depending on behaviour newer than CMake 3.22. It works on CMake 3.25 and later (which is why it passed on my workstation and the math-ci). I'll take a look later.

TorreZuk:
Old form that is more robust probably better than using newer features just to save a couple lines.

cgmb:
It looks like removing the spaces fixed everything. The math-ci returned an UNSTABLE overall status, but every step of the build and test passed, so I'm presuming that's an infrastructure issue of some kind...

stellaraccident:
> It looks like removing the spaces fixed everything. The math-ci returned an UNSTABLE overall status, but every step of the build and test passed, so I'm presuming that's an infrastructure issue of some kind...

Oh yeah, spaces in generators don't mix well. Remember that they get substituted in the post evaluation phase by recursive resolution: all tokenization and syntax had been done, so what is left is literal.

You can write some terrifyingly long line generators... But this is pretty simple.

cgmb:
> The math-ci returned an UNSTABLE overall status, but every step of the build and test passed, so I'm presuming that's an infrastructure issue of some kind.

Our CI team confirmed this was some sort of issue communicating the build status to GitHub, and not an actual failure in the PR. I'm going to put this back for auto-merge..

tfalders (approved):

PR #932, opened 2025-04-15
update the libgfortran pkg dependencies for almalinux

srekolam:
removes the hardcoding of the libgfortran4 inside the CMakeLists.txt


CMakeLists.txt
```diff
@@ -238,15 +238,9 @@ if(BUILD_CLIENTS_TESTS OR BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_SAMPLES)
     string(TOLOWER "${CLIENTS_OS}" CLIENTS_OS)
     rocm_read_os_release(CLIENTS_OS_VERSION VERSION_ID)
   endif()
-  set(GFORTRAN_RPM "libgfortran4")
-  set(GFORTRAN_DEB "libgfortran4")
-  if(CLIENTS_OS STREQUAL "centos" OR CLIENTS_OS STREQUAL "rhel")
```
cgmb, line 243: I'm pretty sure that almalinux should just be the same as centos or rhel. We really should just be adding `OR CLIENTS_OS STREQUAL "almalinux"`

cgmb (changes requested):

CMakeLists.txt
pramenku: This PR might break SLES 15. Refer https://github.com/ROCm/hipBLAS/pull/994#discussion_r2050988827

pramenku: > I'm pretty sure that almalinux should just be the same as centos or rhel. We really should just be adding `OR CLIENTS_OS STREQUAL "almalinux"`

@cgmb We should clean up the code too. 
 What ever default earlier is kept is also not correct for majority of the OS. Please refer https://github.com/ROCm/hipBLAS/pull/994#discussion_r2050988827
 
Even what I have suggested will go with lower risk. libgfortran4 is for SLES and RHEL7. RHEL7 is no more ROCM supported so we all should not worry for RH7. SLES is only what we are keeping libgfortran4 so always good to keep default for major supported OS and keep If condition for few distros like SLES is here.

We should go generic with below which supports majority distros which ROCm supported as of now.
set(GFORTRAN_RPM "libgfortran")
set(GFORTRAN_DEB "libgfortran5")

For RHEL7/SLES 15, we can keep If condition to check for "libgfortran4" like we did for Azure Linux seperate.

cgmb (approved):

srekolam:
Can you please see if this PR can be merged.

cgmb:
I think this should probably be targeted at `develop`. @jzuniga-amd, can you confirm?

tfalders (approved):

cgmb:
There are failures on MI100 (SLES) and MI200 (Ubuntu) in SYEVDX, but they are clearly unrelated.

cgmb:
If you need these changes on another branch, please feel free to `git cherry-pick -x a63d421` into the new branch and open a PR.

PR #935, opened 2025-04-21
Find NaNs in debug mode (do not merge)

AGonzales-amd:


clients/gtest/lapack/sygvd_hegvd_gtest.cpp
```diff
@@ -116,92 +116,92 @@ class SYGVD_HEGVD : public ::TestWithParam<sygvd_tuple>
...
-                         Combine(ValuesIn(matrix_size_range), ValuesIn(type_range)));
-
-INSTANTIATE_TEST_SUITE_P(daily_lapack,
-                         HEGVD,
-                         Combine(ValuesIn(large_matrix_size_range), ValuesIn(type_range)));
-
-INSTANTIATE_TEST_SUITE_P(checkin_lapack,
-                         HEGVD,
-                         Combine(ValuesIn(matrix_size_range), ValuesIn(type_range)));
+// class SYGVD : public SYGVD_HEGVD
```
EdDAzevedo, line 119: If you need to disable or comment out a lot of code, perhaps you might consider something like
#if (0)
... lots of code ...
#endif
so in future you can easily re-activate the disabled code. 
Just a suggestion.

PR #938, opened 2025-04-30
Hybrid support for STEQR

tfalders:


amd-jnovotny (approved):
Documentation OK

tfalders:
> Documentation OK

Oh, that reminds me, I need to update the changelog.

amd-jnovotny (approved):

clients/common/lapack/testing_syev_heev.hpp
```diff
@@ -401,14 +401,17 @@ void testing_syev_heev(Arguments& argus)
     if(argus.alg_mode == 1)
```
EdDAzevedo, line 401: Minor: should it test against an enum value instead of using "1"? Just checking.

clients/common/lapack/testing_syev_heev.hpp
```diff
@@ -401,14 +401,17 @@ void testing_syev_heev(Arguments& argus)
...
-        rocsolver_alg_mode alg_mode;
-        EXPECT_ROCBLAS_STATUS(rocsolver_get_alg_mode(handle, rocsolver_function_sterf, &alg_mode),
+        rocsolver_alg_mode sterf_mode, steqr_mode;
+        EXPECT_ROCBLAS_STATUS(rocsolver_get_alg_mode(handle, rocsolver_function_sterf, &sterf_mode),
+                              rocblas_status_success);
+        EXPECT_ROCBLAS_STATUS(rocsolver_get_alg_mode(handle, rocsolver_function_steqr, &steqr_mode),
                               rocblas_status_success);
 
-        EXPECT_EQ(alg_mode, rocsolver_alg_mode_hybrid);
+        EXPECT_EQ(sterf_mode, rocsolver_alg_mode_hybrid);
```
EdDAzevedo, line 413: Minor: Can the modes for sterf_mode, steqr_mode, and bdsqr_mode be set independently?

Should sterf_mode always equals to steqr_mode (and equal to bdsqr_mode)?  If so, perhaps use a single variable, say  "sterf_steqr_mode" ?  Should the code also check for correctness that  this matching of modes (sterf_mode == steqr_mode)  at execution time?

Maybe the comments or documentation clarify this.

Just a suggestion.

library/src/auxiliary/rocauxiliary_steqr.hpp
```diff
@@ -37,9 +37,333 @@
...
+                scale_tridiag(lsv, lendsv, D, E, ssfmin / anorm);
+        }
+
+        // Check for convergence
+        for(int i = 0; i < n - 1; i++)
+            if(E[i] != 0)
+                *info = *info + 1;
+
+        // Sort eigenvalues and eigenvectors by selection sort
+        if(ordered)
```
EdDAzevedo, line 329: Minor: I wonder whether the shell sort routine can be used, especially when n is not small.  

Another thought is to perform sorting with a permutation vector, then apply permutation from this  vector to rearrange data on matrix.  This might reduce the amount of data movement.

Just a suggestion.

EdDAzevedo (approved):

clients/common/lapack/testing_syev_heev.hpp
tfalders: Yes, the intent is that users may set these independently.

It might be worthwhile to add a mechanism to set them all at once, but I haven't looked into that yet.

jmachado-amd (approved):

PR #943, opened 2025-05-13
Improve larft performance

tfalders:


qjojo (approved):

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -161,6 +165,200 @@ ROCSOLVER_KERNEL void set_tau(const rocblas_int k, T* tau, const rocblas_stride
...
+    // shared memory setup
+    extern __shared__ double lmem[];
+    T* work = reinterpret_cast<T*>(lmem);
+    T* F = work + k;
+    rocblas_int ldf = k;
+
+    // copy F to shared memory
+    for(rocblas_int i = tid; i < k; i += tid_inc)
+        for(rocblas_int j = i; j < k; j++)
+            F[i + j * ldf] = Ftemp[i + j * ldfA];
```
EdDAzevedo, line 201: Question: do we need to use 64bit arithmetic in expressions such as "i + j * ldf"?  

Perhaps consider something like  static_cast<int64_t>(ldf) and static_cast<int64_t>(ldfA) ?

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -229,6 +427,247 @@ rocblas_status rocsolver_larft_argCheck(rocblas_handle handle,
...
+                                       T** workArr,
+                                       size_t sharedMemPerBlock)
+{
+    if(k < 2 || n < 2)
+        return rocblas_status_success;
+
+    hipStream_t stream;
+    rocblas_get_stream(handle, &stream);
+
+    size_t lmemsize = sizeof(T) * (k + 1) * k;
```
EdDAzevedo, line 457: Suggestion: perhaps add an assertion that lmemsize is within limit, say 64KBytes?

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -229,6 +427,247 @@ rocblas_status rocsolver_larft_argCheck(rocblas_handle handle,
...
+    if(k <= LARFT_SWITCHSIZE && lmemsize <= sharedMemPerBlock)
+    {
+        ROCSOLVER_LAUNCH_KERNEL(larft_kernel_forward, dim3(1, batch_count),
+                                dim3(LARFT_SWITCHSIZE, 1), lmemsize, stream, storev, n, k, V,
+                                shiftV, ldv, strideV, tau, strideT, F, shiftF, ldf, strideF);
+        return rocblas_status_success;
+    }
+
+    rocblas_int blocks = (k - 1) / BS2 + 1;
+    rocblas_int l = k / 2;
```
EdDAzevedo, line 467: Minor suggestion:  consider avoid using letter "l" (el) as variable  since it may look like "1" (one).   Perhaps call it "el" or "L"  instead?

library/src/auxiliary/rocauxiliary_larft.hpp
tfalders: lmemsize is compared against sharedMemPerBlock, which is queried from the device

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -229,6 +427,247 @@ rocblas_status rocsolver_larft_argCheck(rocblas_handle handle,
...
+    T minone = -1;
+
+    if(storev == rocblas_column_wise) // QR
+    {
+        // F_1_1
+        larft_recursive_forward(handle, storev, n, l, V, shiftV, ldv, strideV, tau, strideT, F,
+                                shiftF, ldf, strideF, batch_count, scalars, work, workArr,
+                                sharedMemPerBlock);
+        // F_2_2
+        larft_recursive_forward(handle, storev, n - l, k - l, V, shiftV + (l * ldv + l), ldv,
```
EdDAzevedo, line 478: Minor comment: I wonder whether we should pick "el" so that n - el or k - el is a power of 2?  I think recursive routines in rocblas  try to split as  "smaller power of 2" of (n/2) instead of using (n/2) directly, since (n/2) may be  an odd number.  I think there is a routine in rocblas to compute "the smaller power of 2".

library/src/include/ideal_sizes.hpp
```diff
@@ -62,6 +62,12 @@
 #define LARFG_SSKER_MAX_N 2048
 #endif
 
+/******************************* larft ****************************************
+*******************************************************************************/
+#ifndef LARFT_SWITCHSIZE
+#define LARFT_SWITCHSIZE 64
```
EdDAzevedo, line 68: Minor comment: perhaps consider using something like the following that  depends on the type T to fit in 64KBytes of LDS, say 
  
#define LARFT_SWITCHSIZE(T)   ( (sizeof(T) == 16) ? 64 : (sizeof(T) == 4) ? 127 : (sizeof(T) == 8) ?  90 : 64)

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -161,6 +165,200 @@ ROCSOLVER_KERNEL void set_tau(const rocblas_int k, T* tau, const rocblas_stride
...
+                                           T* tauA,
+                                           const rocblas_stride strideT,
+                                           T* FA,
+                                           const rocblas_int shiftF,
+                                           const rocblas_int ldfA,
+                                           const rocblas_stride strideF)
+{
+    const rocblas_int bid = hipBlockIdx_y;
+    const rocblas_int tid = hipThreadIdx_x;
+    const rocblas_int tid_inc = hipBlockDim_x;
```
jmachado-amd, line 185: The macros `hipBlockIdx_*`, `hipThreadIdx_*`, etc. are deprecated. Only the built-in constants `threadIdx`, `blockIdx`, and friends are officially supported at this point in time.

tfalders:
Closing in favor of #944

PR #944, opened 2025-05-15
Optimize larft

AGonzales-amd:
Optimize larft to improve performance of downstream functions like ormtr


amd-jnovotny (approved):
Changelog OK

tfalders (approved):

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -65,30 +74,68 @@ ROCSOLVER_KERNEL void set_triangular(const rocblas_int n,
         Fp = F + b * strideF;
 
         if(j == i)
-            Fp[j + i * ldf] = tp[i];
+            Fp[idx2D(j, i, ldf)] = tp[i];
```
EdDAzevedo, line 77: Minor suggestion: One may consider renaming Fp to Fp_ in the function argument list,  then use a lambda something like 
auto Fp = [=](auto i,j) -> T& { return( Fp_[ idx2D(i,j,ldf) ]); };

then we can use something like   "Fp(j,i)".  This might make the code easier to read.   Similar approach for other arrays such as Vp.   

 Just a suggestion.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -161,6 +248,198 @@ ROCSOLVER_KERNEL void set_tau(const rocblas_int k, T* tau, const rocblas_stride
...
+    T* Ftemp = FA + bid * strideF;
+
+    // shared memory setup
+    extern __shared__ double lmem[];
+    T* work = reinterpret_cast<T*>(lmem);
+    T* F = work + k;
+    rocblas_int ldf = k;
+
+    // copy F to shared memory
+    for(rocblas_int i = tid; i < k; i += tid_inc)
```
EdDAzevedo, line 281: Minor suggestion: perhaps add an assertion to make sure the temporary arrays can fit in LDS shared memory.   Just a thought.

library/src/include/ideal_sizes.hpp
```diff
@@ -62,6 +62,12 @@
 #define LARFG_SSKER_MAX_N 2048
 #endif
 
+/******************************* larft ****************************************
+*******************************************************************************/
+#ifndef LARFT_SWITCHSIZE
+#define LARFT_SWITCHSIZE 64
```
EdDAzevedo, line 68: Minor: perhaps consider LARFT_SWITCHSIZE to depend on the type so that it will fit in LDS?  Maybe something like

#define LARFT_SWITCHSIZE(T)  ( sizeof(T) == 16) ? 63 : (sizeof(T) == 8) ? 91 : (sizeof(T) == 4) ? 127 : 64 )

Just a thought.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: LDS size is checked before kernel launch

tfalders:
Resolving conflict. I will merge after the next set of tests.

PR #945, opened 2025-05-21
Revisit the secular equations' solver

jmachado-amd:
This PR includes an initial implementation of the Gragg-Thorton-Warner cubic scheme (which is used to solve the secular equations).

### DIFF
--- a/library/src/include/lapack_device_functions.hpp
+++ b/library/src/include/lapack_device_functions.hpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2019-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2019-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -1188,32 +1188,23 @@ ROCSOLVER_KERNEL void scal_kernel(I const n, S const da, T* const x, I const inc
     }
 }
 
-template <typename S, typename Int>
-__device__ void slaed6(Int kniter,
-                       Int orgati_,
-                       S rho,
-                       S* d,
-                       S* z,
-                       S finit,
-                       S& tau,
-                       Int& info,
-                       S eps,
-                       S ssfmin,
-                       Int MAXIT = 50)
+template <typename S, typename I>
+__device__ void
+    slaed6(I kniter, bool orgati, S rho, S* d, S* z, S finit, S& tau, I& info, S eps, S ssfmin, I MAXIT = 50)
 {
-    auto ABS = [](auto x) -> auto
+    auto lam_abs = [](auto x) -> auto
     {
         return std::abs(x);
     };
-    auto SQRT = [](auto x) -> auto
+    auto lam_sqrt = [](auto x) -> auto
     {
         return std::sqrt(x);
     };
-    auto MAX = [](auto x, auto y, auto z) -> auto
+    auto lam_max = [](auto x, auto y, auto z) -> auto
     {
         return std::max(std::max(x, y), z);
     };
-    auto MIN = [](auto x, auto y) -> auto
+    auto lam_min = [](auto x, auto y) -> auto
     {
         return std::min(x, y);
     };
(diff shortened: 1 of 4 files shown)


library/src/include/lapack_device_functions.hpp
```diff
@@ -1188,4 +1188,1243 @@ ROCSOLVER_KERNEL void scal_kernel(I const n, S const da, T* const x, I const inc
...
+        ubd = S(0.);
+    }
+
+    niter = 1;
+    tau = S(0.);
+    if(kniter == 2)
+    {
+        if(orgati)
+        {
+            temp = (D(3) - D(2)) / S(2.);
```
EdDAzevedo, line 1271: Minor suggestion: I wonder if this is not performance critical,  whether it might be better to consistently use FP64 arithmetic even for float and complex float.

library/src/include/lapack_device_functions.hpp
```diff
@@ -1188,4 +1188,1243 @@ ROCSOLVER_KERNEL void scal_kernel(I const n, S const da, T* const x, I const inc
...
+            if(ABS(finit) <= ABS(temp))
+            {
+                tau = S(0.);
+            }
+        }
+    }
+    //
+    //     get machine parameters for possible scaling to avoid overflow
+    //
+    small1 = std::pow(S(2), (std::log(ssfmin) / std::log(S(2))) / S(3));
```
EdDAzevedo: Minor suggestion: I wonder whether these are pre-computed or available in std::numeric_limits

library/src/include/lapack_device_functions.hpp
```diff
@@ -1188,4 +1188,1243 @@ ROCSOLVER_KERNEL void scal_kernel(I const n, S const da, T* const x, I const inc
...
+            //
+            //        Scale up by power of radix nearest 1/SAFMIN**(1/3)
+            //
+            sclfac = sminv1;
+            sclinv = small1;
+        }
+        //
+        //        Scaling up safe because D, Z, tau scaled elsewhere to be O(1)
+        //
+        for(int I = 1; I <= 3; ++I)
```
EdDAzevedo: Minor: is there a coding convention when to use upper case for variables, say use lower case for "I", "J" in for loops?

library/src/include/lapack_device_functions.hpp
```diff
@@ -1188,4 +1188,1243 @@ ROCSOLVER_KERNEL void scal_kernel(I const n, S const da, T* const x, I const inc
...
+
+template <typename S, typename Int>
+__device__ void
+    slaed4(Int n, Int i, S* delta, S* z, S rho, S& dlam, Int& info, S eps, S ssfmin, Int MAXIT)
+{
+    auto ABS = [](auto x) -> auto
+    {
+        return std::abs(x);
+    };
+    auto POW2 = [](auto x) -> auto
```
EdDAzevedo: Minor: perhaps call this "SQUARE" since "POW2" might be misunderstood to mean to compute   "2**x" power of 2. 
Just a minor suggestion.

EdDAzevedo (approved):

qjojo (approved):

tfalders:
Static analysis is failing. Please do a format check.

library/src/include/lapack_device_functions.hpp
tfalders: Generally, we use upper case I to represent the integral type in 64-bit template functions. It would be preferable for this to be lower-case i.

library/src/include/lapack_device_functions.hpp
jmachado-amd: `ssfmin` is a parameter, it is allowed to vary from run to run (although we might not use it like this).

library/src/include/lapack_device_functions.hpp
jmachado-amd: Those methods have received a lot of attention towards numerical accuracy; using double precision is unnecessary at this point.

library/src/include/lapack_device_functions.hpp
```diff
@@ -1188,4 +1188,1212 @@ ROCSOLVER_KERNEL void scal_kernel(I const n, S const da, T* const x, I const inc
...
+    //     It is not hard to see that
+    //
+    //           1) Iterations will go up monotonically
+    //              if finit < 0;
+    //
+    //           2) Iterations will go down monotonically
+    //              if finit > 0.
+    //
+    iter = niter + 1;
+    for(int niter = iter; niter <= MAXIT; ++niter)
```
tfalders, line 1434: The loop is redeclaring `niter`?

tfalders (approved):

library/src/include/lapack_device_functions.hpp
EdDAzevedo: Minor suggestion: Unless you are trying to follow the original Fortran code, perhaps you might consider using "niter" to be the number of iterations, and "iter" to be the running index in for loop.  

Just a minor suggestion.

PR #949, opened 2025-05-26
Docs: Add environment variable reference page

adeljo-amd:
This is part of the environment variable initiative, where we aim to create a ROCm environment variable index page that contains the links to every component's environment variable page. Do let me know if there are any other environment variables besides the one that is listed, and whether the description/default values are correct.

### DIFF
--- a/docs/reference/env_variables.rst
+++ b/docs/reference/env_variables.rst
@@ -19,15 +19,13 @@ For more information, see :doc:`Use multi-level logging <../howto/logging>`.
 
 .. list-table::
     :header-rows: 1
-    :widths: 35,14,51
+    :widths: 70,30
 
     * - **Environment variable**
-      - **Default value**
       - **Value**
 
     * - | ``ROCSOLVER_LAYER``
         | Specifies which logging types, if any, are activated.
-      - Unset by default.
       - | Bitwise OR of zero or more bit masks:
         | 0: No logging (if not set)
         | 1: Trace logging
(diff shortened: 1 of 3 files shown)


docs/reference/env_variables.rst
```diff
@@ -0,0 +1,67 @@
...
+  :description: rocSOLVER environment variables reference
+  :keywords: rocSOLVER, ROCm, API, environment variables, environment, reference
+
+.. _rocsolver-environment-variables:
+
+********************************************************************
+rocSOLVER environment variables
+********************************************************************
+
+In this section, the reader can find all the important rocSOLVER environment variables,
```
amd-jnovotny:
```suggestion
This section describes the important rocSOLVER environment variables,
```

docs/reference/env_variables.rst
```diff
@@ -0,0 +1,67 @@
...
+which are grouped by functionality.
+
+Logging variables
+================================================================================
+
+The logging environment variables for rocSOLVER are collected in the following table.
+For more information, please check :doc:`Use multi-level logging <../howto/logging>`.
```
amd-jnovotny:
```suggestion
For more information, see :doc:`Use multi-level logging <../howto/logging>`.
```

docs/reference/env_variables.rst
```diff
@@ -0,0 +1,67 @@
...
+.. list-table::
+    :header-rows: 1
+    :widths: 35,14,51
+
+    * - **Environment variable**
+      - **Default value**
+      - **Value**
+
+    * - | ``ROCSOLVER_LAYER``
+        | Specifies which logging type(s) are activated.
```
amd-jnovotny:
```suggestion
        | Specifies which logging types, if any, are activated.
```

amd-jnovotny (approved):
Just a handful of minor suggestions.

docs/reference/env_variables.rst
adeljo-amd: Fixed

docs/reference/env_variables.rst
adeljo-amd: Fixed

docs/reference/env_variables.rst
adeljo-amd: Fixed

docs/reference/env_variables.rst
```diff
@@ -0,0 +1,67 @@
...
+
+The logging environment variables for rocSOLVER are collected in the following table.
+For more information, see :doc:`Use multi-level logging <../howto/logging>`.
+
+.. list-table::
+    :header-rows: 1
+    :widths: 35,14,51
+
+    * - **Environment variable**
+      - **Default value**
```
tfalders: I don't think "default value" really makes sense here since none of these variables have a value when they aren't set. Perhaps "fallback value" would make more sense?

docs/reference/env_variables.rst
```diff
@@ -0,0 +1,67 @@
...
+      - **Value**
+
+    * - | ``ROCSOLVER_LAYER``
+        | Specifies which logging types, if any, are activated.
+      - Unset by default.
```
tfalders: None of these variables are set by default, so it doesn't really make sense for this to be the only variable that is marked out as being "unset by default". I'd suggest putting "0 (no logging)" here instead.

docs/reference/env_variables.rst
adeljo-amd: Ah, if that's the case, we actually support a different format for that, where we remove the default value column. I just updated the table

docs/reference/env_variables.rst
adeljo-amd: Addressed in the reply I posted above

tfalders (approved):

PR #958, opened 2025-06-06
Add sy2sb

jzuniga-amd:
Tridiagonalization first stage with double detached blocking (nb = band size, k = block size for symmetric trailing matrix update)

I just need to enable complex case.


library/src/auxiliary/rocauxiliary_sy2sb_he2hb.hpp
```diff
@@ -0,0 +1,337 @@
...
+
+ #pragma once
+
+ #include "lapack/roclapack_gelqf.hpp"
+ #include "lapack/roclapack_geqrf.hpp"
+ #include "rocblas.hpp"
+ #include "rocsolver/rocsolver.h"
+
+ROCSOLVER_BEGIN_NAMESPACE
+
```
EdDAzevedo, line 36: Minor suggestion: it would be nice if there are more comments to document the internal workings or algorithms of the routines.

library/src/auxiliary/rocauxiliary_sy2sb_he2hb.hpp
```diff
@@ -0,0 +1,337 @@
...
+    T zero = 0;
+    T neghalf = -0.5;
+    T negone = -1;
+    S rone = 1;
+
+    rocblas_int ldt = k;
+    rocblas_stride strideT = k * k;
+    rocblas_int nk = n - nb;
+    rocblas_int ldacpy = nk;
+    rocblas_stride strideAcpy = nk * (nk + 1);
```
EdDAzevedo, line 215: minor comment: perhaps 64bit arithmetic may be required  for variables of type rocblas_stride (such as strideAcpy, or strideZ or setting up variable tau).    Maybe  something like 
rocblas_stride    strideAcpy =    static_cast<rocblas_stride>(nk) * (nk + 1 );

EdDAzevedo (approved):

tfalders:
Closing in favor of #965

PR #962, opened 2025-06-12
Add cmake option BUILD_PARALLEL_HIP_JOBS

trixirt:
Build speed is dependent on AMDGPU_TARGETS.
Use the  BUILD_PARALLEL_HIP_JOBS=ON option to use the -parallel-jobs compiler option if it is available and reduce the build time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,9 @@ option(WERROR "Treat warnings as errors" OFF)
 option(BUILD_COMPRESSED_DBG "Enable compressed debug symbols" ON)
 check_cxx_compiler_flag("--offload-compress" CXX_COMPILER_SUPPORTS_OFFLOAD_COMPRESS)
 cmake_dependent_option(BUILD_OFFLOAD_COMPRESS "Build with offload compression" ON CXX_COMPILER_SUPPORTS_OFFLOAD_COMPRESS OFF)
+check_cxx_compiler_flag("-parallel-jobs=4" CXX_COMPILER_SUPPORTS_PARALLEL_HIP_JOBS)
+cmake_dependent_option(BUILD_PARALLEL_HIP_JOBS "Build with parallel hip jobs" ON CXX_COMPILER_SUPPORTS_PARALLEL_HIP_JOBS OFF)
+
 
 cmake_dependent_option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY
   "Build with file/folder reorg backward compatibility enabled" OFF "NOT WIN32" OFF)
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -448,6 +448,10 @@ if(BUILD_OFFLOAD_COMPRESS)
   target_compile_options(rocsolver PRIVATE "--offload-compress")
 endif()
 
+if(BUILD_PARALLEL_HIP_JOBS)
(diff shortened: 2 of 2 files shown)


trixirt:
On a build with 4 targets and 32C/64T Threadripper build time reduced from 31min to 12min

CMakeLists.txt
```diff
@@ -150,6 +150,9 @@ option(WERROR "Treat warnings as errors" OFF)
 option(BUILD_COMPRESSED_DBG "Enable compressed debug symbols" ON)
 check_cxx_compiler_flag("--offload-compress" CXX_COMPILER_SUPPORTS_OFFLOAD_COMPRESS)
 cmake_dependent_option(BUILD_OFFLOAD_COMPRESS "Build with offload compression" ON CXX_COMPILER_SUPPORTS_OFFLOAD_COMPRESS OFF)
+check_cxx_compiler_flag("-parallel-jobs=4" CXX_COMPILER_SUPPORTS_PARALLEL_HIP_JOBS)
```
cgmb, line 153: If we do this check, we should probably use `HIP_CLANG_SUPPORTS_PARALLEL_JOBS` as that is the variable already used by some other component that rocSOLVER depends on.

library/src/CMakeLists.txt
```diff
@@ -448,6 +448,10 @@ if(BUILD_OFFLOAD_COMPRESS)
   target_compile_options(rocsolver PRIVATE "--offload-compress")
 endif()
 
+if(BUILD_PARALLEL_HIP_JOBS)
+  target_compile_options(rocsolver PRIVATE "-parallel-jobs=4")
```
cgmb, line 452: I'm not sure there's a reasonable default value that we can pick for this. The appropriate value depends on what resources are available on the build machine. Moreover, I know of existing users that are already setting `-parallel-jobs` via the `HIPCC_COMPILE_FLAGS_APPEND` or `CXXFLAGS` environment variables.

cgmb (commented):
IDK. I think what we're really missing is a proper CMake feature to handle this parallelism. Even aside from the fact that it's not available in upstream clang, the `-parallel-jobs` flag is not a great solution for build times. You end up with two thread/job pools (one for make and one for clang), which results in suboptimal parallelism as compared to if there were just one thread/job pool.

There are already environment variables you can set to enable parallel jobs for all builds if you want to use it (e.g., CXXFLAGS). I worry both about changing every library to enshrine a suboptional solution and that adding this flag might negatively impact existing users of `-parallel-jobs` that add the flag through those other mechanisms.

cgmb:
I've opened https://gitlab.kitware.com/cmake/cmake/-/issues/26997 with my suggestion for a better way to achieve this parallelism in the long term. In the short term, I would recommend just using `export CXXFLAGS="-parallel-jobs=4"`. I don't know of a principled way of choosing the value of `N` for `-parallel-jobs=N` within rocSOLVER's build code, so IMO we should be leaving that up to users.

trixirt:
This is the change used in fedora.
https://src.fedoraproject.org/rpms/rocsolver/c/f91f53912c5ddf0f6cd5a5695dbdc1051a2e470a?branch=rawhide
For the fedora build on the system mentioned above, there are 20 gpu targets.
The build time reduced from 145m to 45m.

ammallya:

PR #963, opened 2025-06-12
Don't use warpSize on host

ghost:
Usage of warpSize on host is removed from hip-runtime, this will cause rocSOLVER to be unable to compile with ROCm7.0. Instead, host needs to call hipDeviceGetAttribute to obtain device's warp size. This commit prepares rocSOLVER for the ROCm7.0 release.

### DIFF
--- a/common/include/rocblas_utility.hpp
+++ b/common/include/rocblas_utility.hpp
@@ -360,42 +360,42 @@ constexpr size_t rocblas_sizeof_datatype(rocblas_datatype type)
 
 // return rocblas_datatype from type
 template <typename> static constexpr rocblas_datatype rocblas_datatype_from_type     = rocblas_datatype_invalid;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_half>           = rocblas_datatype_f16_r;
-template <> static constexpr auto rocblas_datatype_from_type<float>                  = rocblas_datatype_f32_r;
-template <> static constexpr auto rocblas_datatype_from_type<double>                 = rocblas_datatype_f64_r;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_float_complex>  = rocblas_datatype_f32_c;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_double_complex> = rocblas_datatype_f64_c;
-template <> static constexpr auto rocblas_datatype_from_type<int8_t>                 = rocblas_datatype_i8_r;
-template <> static constexpr auto rocblas_datatype_from_type<uint8_t>                = rocblas_datatype_u8_r;
-template <> static constexpr auto rocblas_datatype_from_type<int32_t>                = rocblas_datatype_i32_r;
-template <> static constexpr auto rocblas_datatype_from_type<uint32_t>               = rocblas_datatype_u32_r;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_bfloat16>       = rocblas_datatype_bf16_r;
+template <> constexpr auto rocblas_datatype_from_type<rocblas_half>           = rocblas_datatype_f16_r;
+template <> constexpr auto rocblas_datatype_from_type<float>                  = rocblas_datatype_f32_r;
+template <> constexpr auto rocblas_datatype_from_type<double>                 = rocblas_datatype_f64_r;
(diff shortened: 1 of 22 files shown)


library/src/auxiliary/rocauxiliary_bdsqr_hybrid.hpp
```diff
@@ -52,7 +53,7 @@ static void swap_template(rocblas_handle handle,
                           I const incy,
                           hipStream_t stream)
 {
-    auto nthreads = warpSize * 2;
+    auto nthreads = get_device_warp_size() * 2;
```
EdDAzevedo, line 56: Minor suggestion: perhaps in order to make minimal changes to the existing code, one may use something  like

auto const warpSize = get_device_warp_size();

at the start of the routine, then the rest of the code can stay the same.  The compiler should be able to "do the right thing".

another idea may be  something like

#define warpSize ( get_device_warp_size() )


to have minimal changes to existing code.

Just a suggestion.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_bdsqr_hybrid.hpp
cgmb: That would conflict with the definition of warpSize on older versions of HIP. IDK if it would actually cause problems when building rocSOLVER with older versions of HIP, but it would certainly be confusing. IMO, we should avoid reusing the name.

cgmb:
I wonder if this fixes a bug. warpSize on the host was always defined as 64. IIRC, there was some technical limitation that forced it to be defined on the host despite there being no sensible value for it.

This patch will result in the warp size being 32 for users with RDNA GPUs, rather than being 64. That's probably what was expected when the code was originally written...

library/src/auxiliary/rocauxiliary_bdsqr_hybrid.hpp
EdDAzevedo: If the machine can change "mode" at execution time to use warp size to be 32 or 64, then "warpSize" cannot be a compile time constant.  The warp size need to  be correct to use warp level primitives.    Just a thought.

ghost: It is not advisable to mask away the warpSize on host, it will break any pre 7.0 code and is not explicit in nature it's a function call

ghost:
I added 2 new commits to resolve build issues, how to rerun the CI?

common/include/rocblas_utility.hpp
```diff
@@ -360,42 +360,42 @@ constexpr size_t rocblas_sizeof_datatype(rocblas_datatype type)
...
-template <> static constexpr auto rocblas_datatype_from_type<float>                  = rocblas_datatype_f32_r;
-template <> static constexpr auto rocblas_datatype_from_type<double>                 = rocblas_datatype_f64_r;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_float_complex>  = rocblas_datatype_f32_c;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_double_complex> = rocblas_datatype_f64_c;
-template <> static constexpr auto rocblas_datatype_from_type<int8_t>                 = rocblas_datatype_i8_r;
-template <> static constexpr auto rocblas_datatype_from_type<uint8_t>                = rocblas_datatype_u8_r;
-template <> static constexpr auto rocblas_datatype_from_type<int32_t>                = rocblas_datatype_i32_r;
-template <> static constexpr auto rocblas_datatype_from_type<uint32_t>               = rocblas_datatype_u32_r;
-template <> static constexpr auto rocblas_datatype_from_type<rocblas_bfloat16>       = rocblas_datatype_bf16_r;
+template <> constexpr auto rocblas_datatype_from_type<rocblas_half>           = rocblas_datatype_f16_r;
```
AGonzales-amd, line 363: Hi @brabrzak-amd, why is `static` removed from these template specializations. I think this change is causing the duplicate symbol linker errors.

ghost:
Closing for https://github.com/ROCm/rocSOLVER/pull/963

cgmb:
> Closing for #963

This is #963?

ghost:
Sorry, meant https://github.com/ROCm/rocSOLVER/pull/967

PR #965, opened 2025-06-13
Optimize latrd

tfalders:


library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -1179,6 +1179,177 @@ ROCSOLVER_KERNEL void latrd_lower_computeW_gemv_kernel(const rocblas_int mm,
...
+        a += tx;
+
+    a += it * size_t(ld);
+
+    T res = 0;
+
+    __shared__ T sdata[NB_X];
+
+    // partial sums
+    rocblas_int n_full = (n / NB_X) * NB_X;
```
EdDAzevedo, line 1367: Just checking  --- should this be a ceiling function, say n_full = ceil( n, NB_X) * NB_X,   where
auto ceil = [](auto n, auto b) { return( (n-1)/b + 1 ); };

If   (n < NB_X), do you expect n_full to be zero?

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -1179,6 +1179,177 @@ ROCSOLVER_KERNEL void latrd_lower_computeW_gemv_kernel(const rocblas_int mm,
...
+    rocblas_int bid = blockIdx.z;
+    rocblas_int tx  = threadIdx.x;
+    rocblas_int i = blockIdx.x;
+
+    T* A = load_ptr_batch<T>(AA, bid, shiftA, strideA);
+    T* W = load_ptr_batch<T>(WA, bid, shiftW, strideW);
+    T* y1 = workA + bid * strideblk;
+    T* y2 = load_ptr_batch<T>(yA, bid, shiftY, strideY);
+
+    int n = mm - c - 1;
```
EdDAzevedo, line 1343: minor: maybe add "typename I" as the integer type in the template argument list and consistently using "I" as the desired integer type? Say something like  (note perhaps shiftA , shiftW, shiftY,  should consistently be type rocblas_stride as well ?)

 template<int NB_X, typename T, typename I, typename Istride,  typename U> 
... const Istride shiftA,    const I lda,  const Istride strideA 
then     I n = mm - c - 1;

Just a suggestion.

library/src/auxiliary/rocauxiliary_latrd.hpp
```diff
@@ -1179,6 +1179,177 @@ ROCSOLVER_KERNEL void latrd_lower_computeW_gemv_kernel(const rocblas_int mm,
...
+            res += conj(a[j]) * x[tx + j];
+
+        if(tx + n_full < n)
+            res += conj(a[n_full]) * x[tx + n_full];
+
+        sdata[tx] = res;
+        __syncthreads();
+
+        // tree reduction of partial sums
+        for(int r = NB_X / 2; r > 0; r /= 2)
```
EdDAzevedo: Minor suggestion:  Perhaps also  test if NB_X is smaller than warp size to use  warp shuffle instructions?

EdDAzevedo (approved):

qjojo (approved):

library/src/auxiliary/rocauxiliary_latrd.hpp
AGonzales-amd: n_full is the floor div multiple (not sure what the correct term is) of n. The remaining columns are computed in lines 1240-1241

PR #967, opened 2025-06-17
Refactor warpSize as it is removed as compile time constant

satyanveshd:
Usage of warpSize as compile time constant on both host and device is removed from HIP runtime, this will cause rocSOLVER to be unable to compile with ROCm7.0. Instead, host needs to call hipDeviceGetAttribute to obtain device's warp size. And in device code, warpSize can only be used as a runtime variable.

This change prepares rocSOLVER for the ROCm7.0 release.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_larfg.hpp
+++ b/library/src/auxiliary/rocauxiliary_larfg.hpp
@@ -4,7 +4,7 @@
  *     Univ. of Tennessee, Univ. of California Berkeley,
  *     Univ. of Colorado Denver and NAG Ltd..
  *     November 2017
- * Copyright (C) 2019-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2019-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -38,8 +38,8 @@
 
 ROCSOLVER_BEGIN_NAMESPACE
 
-template <typename T, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
-__device__ void run_set_taubeta(T* tau, T* norms, T* alpha, T* beta)
+template <typename T, typename S, std::enable_if_t<!rocblas_is_complex<T>, int> = 0>
+__device__ void run_set_taubeta(T* tau, T* norms, T* alpha, S* beta)
 {
     const auto ignore_beta = (beta == nullptr);
     if(norms[0] > 0)
@@ -78,10 +78,9 @@ __device__ void run_set_taubeta(T* tau, T* norms, T* alpha, T* beta)
     }
 }
 
-template <typename T, std::enable_if_t<rocblas_is_complex<T>, int> = 0>
-__device__ void run_set_taubeta(T* tau, T* norms, T* alpha, T* beta)
+template <typename T, typename S, std::enable_if_t<rocblas_is_complex<T>, int> = 0>
+__device__ void run_set_taubeta(T* tau, T* norms, T* alpha, S* beta)
 {
-    using S = decltype(std::real(T{}));
     S r, rr, ri, ar, ai;
 
     ar = alpha[0].real();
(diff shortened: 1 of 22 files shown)


satyanveshd:
This PR contains the changes from https://github.com/ROCm/rocSOLVER/pull/963 along with few additional changes to fix the new failures due to the recent commits in the develop branch.

library/src/specialized/rocauxiliary_larfg_specialized_kernels.hpp
```diff
@@ -71,7 +71,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(MAX_THDS) larfg_kernel_small(const I n,
     S* b = beta ? load_ptr_batch<S>(beta, bid, shiftB, strideB) : nullptr;
 
     // shared variables
-    __shared__ T sval[MAX_THDS / warpSize];
+    __shared__ T sval[MAX_THDS / WarpSize];
```
EdDAzevedo, line 74: Minor question: will variable "WarpSize" be pre-defined to be correct in HIP  GPU kernel code?  Will the older "warpSize" still be supported?   Will this change be  compatible with cuda code (compile by cuda compiler)  and compiling with earlier version of rocm?   Just checking.

satyanveshd: "WarpSize" is not pre-defined in HIP.  I had defined it as a constant here library/src/auxiliary/rocauxiliary_larfg.hpp.  This is more like a workaround to avoid using the older warpSize in the Variable length arrays and to get the build failures fixed quickly for 7.0.  You can change this later and fix it differently if you think this confuses the users.

And the older "warpSize" is supported in device code only as a runtime variable and not as compile time constant. You can see that "warpSize" is still being used in this file still in the device code.  Changed this warpSize to WarpSize only at couple of places where it was being used as a compile time constant.

library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -36,6 +36,12 @@
 #include "rocsolver/rocsolver.h"
 #include "rocsolver_run_specialized_kernels.hpp"
 
+#if defined(__GFX9__)
```
EdDAzevedo, line 39: This part of code may be the critical part and I wonder whether this is fully correct or perhaps can be expanded to include other architectures?   Maybe it should be more widely used as well?   Just a thought.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_larfg.hpp
cgmb: How does the SPIR-V backend handle this?

AGonzales-amd (approved):

AGonzales-amd:
Hi @satyanveshd please fix the formatting to match clang format. I think the remaining CI errors are unrelated so we can merge once reformatted and CI is reran.

library/src/auxiliary/rocauxiliary_larfg.hpp
ghost: This works as a short term fix, especially now for ROCm 7.0 enablement. 

For the long term I suggest on the rocSOLVER side to refactor all current and future kernels that require warpsize to be constexpr to pass it as a template argument.

Please see https://github.com/ROCm/clr/issues/154#issuecomment-2877501208 for more info

satyanveshd: Yes, mentioned it as a quick fix to get the build ready for 7.0.  And already discussed offline with @EdDAzevedo about the similar suggestion and rocSOLVER team can revisit all the kernels and refactor the code accordingly.

satyanveshd: > How does the SPIR-V backend handle this?

AFAIK, for SPIRV, it becomes constant when the compiler backend gets to know the GPU it is currently on.

satyanveshd:
> Hi @satyanveshd please fix the formatting to match clang format. I think the remaining CI errors are unrelated so we can merge once reformatted and CI is reran.

Done. reformatted to match clang format and updated the PR.

library/src/auxiliary/rocauxiliary_larfg.hpp
cgmb: > For the long term I suggest on the rocSOLVER side to refactor all current and future kernels that require warpsize to be constexpr to pass it as a template argument.

LGTM. I think the only downside is that you might end up with some code bloat instantiating kernels for both wave64 and wave32 on hardware that can only do one or the other.

I suppose you can always optimize by using ifdefs to omit the unnecessary instantiations for architectures if you know the wave size at compile-time, while still being source-compatible with future hardware.

Thanks for the insights!

PR #968, opened 2025-06-17
Dc optim secular

jmachado-amd:
Merge updated code to solve secular equations.


library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1540,12 +1540,18 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
                     // 'tmpd' will be updated with the distances D - lambda_i.
                     // deflated values are not changed.
                     rocblas_int linfo;
+
+#if defined(ROCSOLVER_USE_REFERENCE_SECULAR_EQUATIONS_SOLVER)
```
EdDAzevedo, line 1544: The team might revisit later the general  issue on how to select different algorithm choices, whether determined in CMAKE compile time,  determined by environment variable at execution time, or perhaps explicitly in subroutine argument, or determined at execution time based on problem size or machine environment ....

EdDAzevedo (approved):

PR #973, opened 2025-06-19
Cherry-pick to release-staging/rocm-rel-7.0: Refactor warpSize as it is removed as compile time constant (#967)

AGonzales-amd:
This cherry-picks the patch for the removal of _AMDGCN_WAVEFRONT_SIZE & warpSize into the 7.0 staging branch.

### DIFF
--- a/clients/common/misc/rocblas_random.hpp
+++ b/clients/common/misc/rocblas_random.hpp
@@ -1,5 +1,5 @@
 /* **************************************************************************
- * Copyright (C) 2018-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2018-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -31,6 +31,7 @@
 #include "rocblas_math.hpp"
 #include <cinttypes>
 #include <random>
+#include <thread>
 #include <type_traits>
 
 /* ============================================================================================
--- a/library/src/auxiliary/rocauxiliary_larfg.hpp
+++ b/library/src/auxiliary/rocauxiliary_larfg.hpp
@@ -4,7 +4,7 @@
  *     Univ. of Tennessee, Univ. of California Berkeley,
  *     Univ. of Colorado Denver and NAG Ltd..
  *     November 2017
- * Copyright (C) 2019-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2019-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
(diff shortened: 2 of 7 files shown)


jmachado-amd (approved):

clients/common/misc/rocblas_random.hpp
```diff
@@ -31,6 +31,7 @@
 #include "rocblas_math.hpp"
 #include <cinttypes>
 #include <random>
+#include <thread>
```
EdDAzevedo, line 34: Is the code using explicit threading? Would using OpenMP be easier? I think clang compiler (and g++)  has support for OpenMP using the "-fopenmp" option.

#ifdef _OPENMP
#include <omp.h>
#endif

AGonzales-amd: Actually, not sure why this was added in the original PR. It's the only change in the file.

EdDAzevedo (approved):
I think we should later revisit this issue and consider adding the warpsize as a template argument for kernels that need compile time value of  warpsize for declaring the amount of LDS shared memory.

clients/common/misc/rocblas_random.hpp
TorreZuk: @AGonzales-amd   header includes were added to fix compilation errors with new hip headers reducing the number of includes they add beyond what was actually required. I would assume it is required here.

tfalders (approved):

library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -36,6 +36,12 @@
 #include "rocsolver/rocsolver.h"
 #include "rocsolver_run_specialized_kernels.hpp"
 
+#if defined(__GFX9__)
+__device__ static constexpr int WarpSize = 64;
+#else
+__device__ static constexpr int WarpSize = 32;
+#endif
+
```
tfalders, line 44: Do we know what this was added for? I don't see `WarpSize` being used in larfg.

AGonzales-amd: This is used in the small larfg kernels in rocauxiliary_larfg_specialized_kernels.hpp in develop which references `warpSize`. Should I remove this definition since that version of the kernel isn't in this release?

tfalders: I think we'll be wanting to get it into this release, so maybe it's worth keeping.

tfalders:
Should this be targeting release/rocm-rel-7.0 rather than release-staging/rocm-rel-7.0?

AGonzales-amd:
> Should this be targeting release/rocm-rel-7.0 rather than release-staging/rocm-rel-7.0?

Sorry I mistitled the PR, it should target staging

PR #977, opened 2025-06-26
Optimize LARF

AGonzales-amd:
This PR includes optimizations for `larf` to improve the performance of downstream functions like `geqrf`. The optimization is a dedicated kernel to handle larger problem sizes while taking advantage of shared memory for data reuse.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_larf.hpp
+++ b/library/src/auxiliary/rocauxiliary_larf.hpp
@@ -326,17 +326,24 @@ rocblas_status rocsolver_larf_template(rocblas_handle handle,
     const int lds_size = leftside ? (m + (NB / props.warpSize)) * sizeof(T)
                                   : (n + (NB / props.warpSize)) * sizeof(T);
 
-    if(lds_size <= props.sharedMemPerBlock && (n <= 1024 || m >= 2048))
+    if(lds_size <= props.sharedMemPerBlock)
     {
-        if(leftside)
+        // Launch larf kernel if tune parameters are met.
+        if(leftside && (n <= 1024 || m >= 2048))
+        {
             ROCSOLVER_LAUNCH_KERNEL((larf_left_kernel<NB>), dim3(1, n, batch_count), dim3(NB),
                                     lds_size, stream, m, n, x, shiftx, incx, stridex, alpha,
                                     stridep, A, shiftA, lda, stridea);
-        else
+            return rocblas_status_success;
+        }
+        // TODO: investigate right side tuning.
+        else if(!leftside && (m <= 1024 || n >= 2048))
+        {
             ROCSOLVER_LAUNCH_KERNEL((larf_right_kernel<NB>), dim3(1, m, batch_count), dim3(NB),
                                     lds_size, stream, m, n, x, shiftx, incx, stridex, alpha,
                                     stridep, A, shiftA, lda, stridea);
-        return rocblas_status_success;
+            return rocblas_status_success;
+        }
     }
 
     // everything must be executed with scalars on the device
--- a/library/src/auxiliary/rocauxiliary_larf.hpp
+++ b/library/src/auxiliary/rocauxiliary_larf.hpp
@@ -39,6 +39,12 @@
 
 ROCSOLVER_BEGIN_NAMESPACE
 
+/*
+*   LARF kernel for the left side case. Each work group of NB_X threads
+*   operates on a column of matrix A. (m + NB_X / warpSize) * sizeof(T)
+*   bytes of LDS memory is required. Grid dimensions = dim3(1, n, batch count)
+*   and block dimensions = dim3(NB_X).
+*/
 template <int NB_X, typename T, typename I, typename U>
 ROCSOLVER_KERNEL void __launch_bounds__(NB_X) larf_left_kernel(const I m,
                                                                const I n,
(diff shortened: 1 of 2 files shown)


library/src/auxiliary/rocauxiliary_larf.hpp
```diff
@@ -35,9 +35,80 @@
...
+    I start = (incX > 0 ? 0 : (m - 1) * -incX);
+
+    T res = 0;
+
+    extern __shared__ double smem[];
+    T* sdata = reinterpret_cast<T*>(smem);
+    T* xs = sdata + (NB_X / warpSize);
+
+    for(I i = tx; i < m; i += NB_X)
+        xs[i] = x[start + i * int64_t(incX)];
```
EdDAzevedo: Minor comment: should there be a check or assertion that m is sufficiently small to fit in LDS?

library/src/auxiliary/rocauxiliary_larf.hpp
```diff
@@ -160,13 +231,33 @@ rocblas_status rocsolver_larf_template(rocblas_handle handle,
...
+    HIP_CHECK(hipGetDevice(&device));
+    hipDeviceProp_t props;
+    HIP_CHECK(hipGetDeviceProperties(&props, device));
+
+    static constexpr int NB = 1024;
+    const int lds_size = (m + (NB / props.warpSize)) * sizeof(T);
+
+    // determine side
+    bool leftside = (side == rocblas_side_left);
+    if(leftside && lds_size <= props.sharedMemPerBlock && (n <= 1024 || m >= 2048))
```
EdDAzevedo: Minor: If m is smaller than 2048, then the larf kernel is not used? What if the kernel can be launched with fewer threads?  Maybe something like the following?   Just a thought.

if (m >= 2048) {   larf_left_kernel<NB>( grid, dim3(NB), .. ) ; }
else if (m >= 1024) { larf_left_kernel<NB/2>( grid, dim3(NB/2) ... ); }
else if (m >= 512) { larf_left_kernel<NB/4>( grid, dim3(NB/4), ... } 
else if (m >= 256) { larf_left_kernel<NB/8>( grid, dim3(NB/8)  }
else { larf_left_kernel<64>( grid, dim3(64>) ... };

library/src/auxiliary/rocauxiliary_larf.hpp
```diff
@@ -35,9 +35,80 @@
 #include "rocblas.hpp"
 #include "rocsolver/rocsolver.h"
 #include "rocsolver_run_specialized_kernels.hpp"
+#include <hip/hip_cooperative_groups.h>
 
 ROCSOLVER_BEGIN_NAMESPACE
 
+template <int NB_X, typename T, typename I, typename U>
```
EdDAzevedo, line 48: Minor and just my personal  view: I understand "real programmers don't comment code" but I  think it would be great (in general)  if there were more comments about the how the code works, or how the correctness depends on the launch configuration,  or the  assumptions or limitations on the sizes of m and n.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_larf.hpp
AGonzales-amd: These numbers are based on performance data. It's just minor tuning, outside of these bounds, the kernel performs worse than calling rocblas routines.

library/src/auxiliary/rocauxiliary_larf.hpp
AGonzales-amd: The kernel is only called when m can fit within LDS

jzuniga-amd (approved):
LGTM. We need to add the right version of the kernel as well (it will improve all the by-column orthogonal transformations, ie. gelqf, gerqf, and the bidiagonalization that works by rows and columns)
If it is not added in this PR, just add a TODO besides the larf_left_kernel so we don't forget.

AGonzales-amd:
> LGTM. We need to add the right version of the kernel as well (it will improve all the by-column orthogonal transformations, ie. gelqf, gerqf, and the bidiagonalization that works by rows and columns) If it is not added in this PR, just add a TODO besides the larf_left_kernel so we don't forget.

Added right side kernel but I have not tested its performance and has not been tuned.

library/src/auxiliary/rocauxiliary_larf.hpp
```diff
@@ -160,13 +313,38 @@ rocblas_status rocsolver_larf_template(rocblas_handle handle,
...
+    {
+        if(leftside)
+            ROCSOLVER_LAUNCH_KERNEL((larf_left_kernel<NB>), dim3(1, n, batch_count), dim3(NB),
+                                    lds_size, stream, m, n, x, shiftx, incx, stridex, alpha,
+                                    stridep, A, shiftA, lda, stridea);
+        else
+            ROCSOLVER_LAUNCH_KERNEL((larf_right_kernel<NB>), dim3(1, m, batch_count), dim3(NB),
+                                    lds_size, stream, m, n, x, shiftx, incx, stridex, alpha,
+                                    stridep, A, shiftA, lda, stridea);
+        return rocblas_status_success;
```
tfalders: Is it correct that m must be >= 2048? And do we not need to change the condition when switching between the left and right sides?

tfalders (approved):
Awesome! Looks good; just one comment.

library/src/auxiliary/rocauxiliary_larf.hpp
EdDAzevedo: Maybe a comment to explain the condition may be helpful. Perhaps the code is trying to detect whether it is a tall skinny matrix?

AGonzales-amd: fixed bounds for right side

tfalders:
CI failure unrelated. Forcing the merge.

Issue #929, opened 2025-04-05
[Feature]: Replace fmt:: with std::


trixirt:
### Suggestion Description

The fmt package is not available on all linux distros.
So the requirement here
https://github.com/ROCm/rocSOLVER/blob/develop/CMakeLists.txt#L67
This will fail for RHEL and OracleLinux

the fmt::format and fmt::print calls should be replaced with std::format and std::print

tcgu-amd:
Hey @trixirt, thank you for the suggestion! I am not sure if it will fail on RHEL and Oracle Linux, since FMT is available from epel repository. You can also build it from source https://github.com/fmtlib/fmt. Hope this helps. Thanks!

trixirt:
A requirement is not to use EPEL as EPEL is not supported, whereas RHEL's libstdc++ is supported.

tcgu-amd:
@trixirt, thanks for your input. However, since our software is cross-platform, switching away from fmt based solely on Red Hat's packaging isn't viable. The fmt library is widely adopted and actively maintained. Being from EPEL doesn't necessarily compromise its quality or our ability to support it. We will be happy to address any specific bugs/issues related to fmt if any rises. Hope this makes ense.

cgmb:
When I added the rocsolver dependency on fmtlib for `fmt::format` and `fmt::print` in #236, the `std::format` function was recently standardized but not yet implemented in the standard library and `std::print` was merely a proposal. One of the reasons why I chose fmtlib was that one day it could be replaced by the standard library and the fmtlib dependency could be dropped.

Until `std::print` is available on all the platforms we support, there would need to be a transition period in which rocsolver supports both `fmt::print` and `std::print`. We can probably alias them both to `rocsolver::print` or something like that.

Nobody from the rocSOLVER team has chimed in on this discussion, and I don't want to speak for them (as I'm no longer officially a rocSOLVER developer), but I think this would probably be a nice to have feature. I'd be happy to review any PRs submitted for this.

ammallya:
Imported to ROCm/rocm-libraries

PR #978, opened 2025-06-27
Optionally use std:: instead of fmt::

john00003:
This draft PR proposes a series of changes that are meant to address issue #929. An optional CMake option/preprocessor definition is used for conditional usage of the standard library in place of some fmt lib.

This feature does not impact the normal use of rocSOLVER, and does not add any mandatory dependencies, but instead acts as a solution for users on Linux distributions that no longer support fmt lib with the addition of formatted printing in C++20 and C++23.

To use this feature:
- use the `--no-fmt` flag for `install.sh`
- ensure that a version of libc++ or libstdc++ that supports formatted printing from C++20/C++23 is installed on your machine (this may also require you to pass --cmake-args=-DCMAKE_CXX_FLAGS="stdlib=libstdc++")


ammallya:

PR #980, opened 2025-06-30
Inverse triangular factor for GEQRF

AGonzales-amd:
This PR adds LARFT and LARFB functions which compute and operate on the inverse triangular factor. These are used in GEQRF (non-batched) to improve performance. The new functions are called from new overloads of existing templates; these overloads are used in performance critical areas.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_larfb.hpp
+++ b/library/src/auxiliary/rocauxiliary_larfb.hpp
@@ -339,30 +339,8 @@ rocblas_status rocsolver_larfb_template(rocblas_handle handle,
 
     // compute: trans(T) * (V1' * A1 + V2' * A2)
     //    or    (A1 * V1 + A2 * V2) * trans(T)
-    if(colwise && forward && leftside)
-    {
-        size_t size_work1, size_work2, size_work3, size_work4;
-        bool optim_mem;
-        rocsolver_trsm_mem<false, false, T>(side, transt, ldw, order, batch_count, &size_work1,
-                                            &size_work2, &size_work3, &size_work4, &optim_mem);
-        void *work1, *work2, *work3, *work4;
-        HIP_CHECK(hipMalloc(&work1, sizeof(size_work1)));
-        HIP_CHECK(hipMalloc(&work2, sizeof(size_work2)));
-        HIP_CHECK(hipMalloc(&work3, sizeof(size_work3)));
-        HIP_CHECK(hipMalloc(&work4, sizeof(size_work4)));
-        rocsolver_trsm_upper<false, false, T>(handle, side, transt, rocblas_diagonal_non_unit, ldw,
-                                              order, F, shiftF, ldf, strideF, tmptr, 0, ldw, strideW,
-                                              batch_count, optim_mem, work1, work2, work3, work4);
-        HIP_CHECK(hipFree(work1));
-        HIP_CHECK(hipFree(work2));
-        HIP_CHECK(hipFree(work3));
-        HIP_CHECK(hipFree(work4));
-    }
-    else
-    {
-        rocblasCall_trmm(handle, side, uploT, transt, rocblas_diagonal_non_unit, ldw, order, &one,
-                         0, F, shiftF, ldf, strideF, tmptr, 0, ldw, strideW, batch_count, workArr);
-    }
+    rocblasCall_trmm(handle, side, uploT, transt, rocblas_diagonal_non_unit, ldw, order, &one, 0, F,
+                     shiftF, ldf, strideF, tmptr, 0, ldw, strideW, batch_count, workArr);
 
     // compute: A2 - V2 * trans(T) * (V1' * A1 + V2' * A2)
     //    or    A2 - (A1 * V1 + A2 * V2) * trans(T) * V2'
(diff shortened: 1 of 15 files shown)


jzuniga-amd (approved):

library/src/auxiliary/rocauxiliary_larfb.hpp
```diff
@@ -375,4 +375,282 @@ rocblas_status rocsolver_larfb_template(rocblas_handle handle,
...
+                                           size_t* size_tmptr,
+                                           size_t* size_work1,
+                                           size_t* size_work2,
+                                           size_t* size_work3,
+                                           size_t* size_work4,
+                                           size_t* size_workArr,
+                                           bool* optim_mem)
+{
+    // if quick return, no workspace needed
+    if(m == 0 || n == 0 || batch_count == 0)
```
EdDAzevedo, line 394: Minor suggestion:   consider set all the sizes to be 0 first, then test if (m == 0). This is  defensive programming to make sure we don't have uninitialized values.

library/src/auxiliary/rocauxiliary_larfb.hpp
```diff
@@ -375,4 +375,282 @@ rocblas_status rocsolver_larfb_template(rocblas_handle handle,
...
+        order = n;
+        ldw = k;
+    }
+    else
+    {
+        order = k;
+        ldw = m;
+    }
+
+    rocsolver_trsm_mem<BATCHED, STRIDED, T>(side, transt, ldw, order, batch_count, size_work1,
```
EdDAzevedo: Minor: consider calling the new rocblas_internal_trsm_workspace_max_size_64()

to make sure the size is sufficient for all smaller cases.

library/src/auxiliary/rocauxiliary_larfb.hpp
```diff
@@ -375,4 +375,282 @@ rocblas_status rocsolver_larfb_template(rocblas_handle handle,
...
+    bool forward = (direct == rocblas_forward_direction);
+    bool leftside = (side == rocblas_side_left);
+    rocblas_operation transt
+        = (leftside && trans == rocblas_operation_transpose ? rocblas_operation_conjugate_transpose
+                                                            : trans);
+    rocblas_operation transp;
+    rocblas_fill uploV, uploT;
+    rocblas_int order, ldw;
+    rocblas_int shift1, shift2;
+    size_t offsetA1, offsetA2;
```
EdDAzevedo, line 507: should this be type rocblas_stride for the shift1, shift2 and offsetA1, offsetA2 variables?

AGonzales-amd: I just used what was in the original larfb function. It looks like we don't have a consistent type in rocsolver for this; I see both rocblas_int and rocblas_stride being used for shifts and offsets. We might want to consolidate this in the future.

tfalders: Originally we were using rocblas_int everywhere. I started moving them over to rocblas_stride in the 64-bit APIs.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -721,4 +721,200 @@ rocblas_status rocsolver_larft_template(rocblas_handle handle,
...
+            T* Bp = &buffer[b * strideB];
+
+            // copy buffer to A
+            Ap[i + j * lda] = Bp[i + j * ldb];
+        }
+    }
+}
+
+template <typename T>
+ROCSOLVER_KERNEL void set_diag(rocblas_int k,
```
tfalders: I believe there's already a kernel named set_diag in lib_device_helpers. I'd suggest changing the name so there isn't a conflict with the other kernel.

library/src/auxiliary/rocauxiliary_larfb.hpp
EdDAzevedo: My personal opinion is unless there are good reasons, the shift and stride should consistently both be rocblas_stride type. Otherwise, there may be a concern that the software is limited to  matrices with only 2^32 - 1 entries.

library/src/lapack/roclapack_geqrf.cpp
```diff
@@ -55,39 +55,46 @@ rocblas_status
...
+    work3 = mem[3];
+    work4 = mem[4];
+    Abyx_norms_trfact = mem[5];
+    diag_tmptr = mem[6];
+    workArr = mem[7];
     if(size_scalars > 0)
         init_scalars(handle, (T*)scalars);
 
     // execution
     return rocsolver_geqrf_template<false, false, T>(
```
tfalders, line 94: This is only applied to non-batched QR? Should we not also include batched and strided batched QR?

library/src/lapack/roclapack_syevd_heevd.cpp
```diff
@@ -73,19 +75,20 @@ rocblas_status rocsolver_syevd_heevd_impl(rocblas_handle handle,
...
+    tmpz = mem[5];
+    splits = mem[6];
+    tmptau_W = mem[7];
+    tau = mem[8];
+    workArr = mem[9];
     if(size_scalars > 0)
         init_scalars(handle, (T*)scalars);
 
     // execution
     return rocsolver_syevd_heevd_template<false, false, T>(
```
tfalders, line 88: Again, only applied to non-batched?

tfalders (approved):
Looks good! Some comments to be addressed.

As mentioned previously, I'm not a fan of all the redundant code, but I accept that managing all the workspace arrays is a pain. We should adopt a new workspace management approach ASAP.

library/src/auxiliary/rocauxiliary_larfb.hpp
tfalders: Yes, we should switch all of these to rocblas_stride at some point. Not in this PR though.

library/src/auxiliary/rocauxiliary_larft.hpp
AGonzales-amd: fixed

library/src/lapack/roclapack_geqrf.cpp
AGonzales-amd: updated non batched QR

library/src/lapack/roclapack_syevd_heevd.cpp
AGonzales-amd: updated non batched syevd

PR #982, opened 2025-07-02
Small performance improvements in stedc

pghysels:
Simplify bubble sort.
Use coalesced accesses.
Use binary search instead of loop.

For 
`./rocsolver-bench -f syevd -n 2048 --evect V --iters 3 --perf 1 -r s  --profile 10 --profile_kernels 1`
I get 21%  improvement for `stedc_mergeValues_kernel`, and 5% for `rocsolver_syevd_heevd_template`. (on MI300X)

### DIFF
--- a/library/src/auxiliary/rocauxiliary_stedc.hpp
+++ b/library/src/auxiliary/rocauxiliary_stedc.hpp
@@ -1463,33 +1463,15 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
 
             // Order the elements in tmpd and zz using a simple parallel selection/bubble sort.
             // This will allow us to find initial intervals for eigenvalue guesses
-            for(int i = 0; i < tsz; ++i)
+            for(int i = 0; i < dd; i++)
             {
-                if(i < dd)
+                for(int j = 2 * iam + i % 2; j < dd - 1; j += 2 * bdm)
                 {
-                    if(i % 2 == 0)
+                    if(tmpd[j] > tmpd[j + 1])
                     {
-                        for(int j = iam; j < dd / 2; j += bdm)
-                        {
-                            if(tmpd[2 * j] > tmpd[2 * j + 1])
-                            {
-                                swap(tmpd[2 * j], tmpd[2 * j + 1]);
-                                swap(zz[2 * j], zz[2 * j + 1]);
-                                swap(per[2 * j], per[2 * j + 1]);
-                            }
-                        }
-                    }
-                    else
-                    {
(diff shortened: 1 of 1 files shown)


jzuniga-amd (approved):
LGTM. Thanks @pghysels

tfalders (approved):
Awesome, thanks!

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1463,33 +1463,15 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
 
             // Order the elements in tmpd and zz using a simple parallel selection/bubble sort.
             // This will allow us to find initial intervals for eigenvalue guesses
-            for(int i = 0; i < tsz; ++i)
+            for(int i = 0; i < dd; i++)
             {
-                if(i < dd)
+                for(int j = 2 * iam + i % 2; j < dd - 1; j += 2 * bdm)
```
EdDAzevedo, line 1468: Minor suggestion: Perhaps the shell sort method can be explored as well? The non-recursive shellsort does not require extra storage.

https://en.wikipedia.org/wiki/Shellsort

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_stedc.hpp
pghysels: I tried it quickly but it seems to be a bit slower.
I can also try the rocPRIM sorting routines.

EdDAzevedo: In wikipedia page on shell sort, there is pseudo code with  the suggestion of using 

gaps = [701, 301, 132, 57, 23, 10, 4, 1]  # Ciura gap sequence

Just FYI

pghysels: Thanks.
Replacing
```
            for(int m = dd / 2; m > 0; m /= 2)
```
with
```
            for(int m : {701, 301, 132, 57, 23, 10, 4, 1})
```
works.
But now I'm not sure the rest is correct. It seems like there is one too many loops.
I will try to figure this out on Monday.

tfalders:
Since the performance improvement is already quite solid as-is, I think we should merge the original version of the code and experiment with different sorting methods offline or in a separate PR. @pghysels, would you mind rolling back the changes with shell sort so we can merge the improvement?

pghysels:
I removed the shell sort. 
Note that most of the performance improvements from this PR are not from the changes to the sorting code, but from the 2 other changes.

PR #988, opened 2025-07-15
Small size kernels for sytd2

tfalders:


library/src/lapack/roclapack_sytd2_hetd2.hpp
```diff
@@ -33,11 +33,278 @@
...
+            w[i] += sval[0] * x[i];
+        __syncthreads();
+
+        // ----- 3. apply the Householder reflector to A as a rank-2 update: A = A - v*w' - w*v' -----
+        // syr2
+        for(I i = tid; i < nn; i += MAX_THDS)
+        {
+            for(I jj = 0; jj <= i; jj++)
+            {
+                T* Atmp = A + (j + 1) + (j + 1) * lda;
```
EdDAzevedo: Is this expression that depends on "j" invariant in the "for(jj)" loop? If so, perhaps it can be hoisted outside the "for(i)" loop? Or perhaps  the Atmp expression depends on "jj" variable?   Just checking.

EdDAzevedo (approved):

jzuniga-amd (approved):

PR #992, opened 2025-07-17
Various fixes

tfalders:
Contains miscellaneous fixes and improvements. Namely:

1. Resolves warnings caused by template specializations labeled as static.
2. Adds the shift parameter to rocsolver_hybrid_storage (mainly for use with batched arrays).
3. Moves some kernels used by many eigensolvers to lib_device_helpers (now every eigensolver won't need to be recompiled when we modify syev).


common/include/rocblas_utility.hpp
```diff
@@ -360,44 +360,9 @@ constexpr size_t rocblas_sizeof_datatype(rocblas_datatype type)
 
 // return rocblas_datatype from type
 template <typename> static constexpr rocblas_datatype rocblas_datatype_from_type     = rocblas_datatype_invalid;
```
EdDAzevedo: Perhaps something like the following for argument  "(T *A)" and "(T )" 

template<typename T>
rocblas_datatype datatype_from_type( T *A ) {
  rocblas_datatype const a_type =     std::is_same<T,  __half >::value ? rocblas_datatype_f16_r :
                                                           std::Is_same<T, rocblas_half>::value ? rocblas_datatype_f16_r :
                                                           std::is_same<T, double>::value ? rocblas_datatype_f64_r: 
                                                           .... etc
                                                           rocblas_datatype_invalid;
                        return( a_type );
                      }

library/src/include/lib_device_helpers.hpp
```diff
@@ -1026,6 +1026,64 @@ ROCSOLVER_KERNEL void set_zero(const rocblas_int m,
...
+                                       const rocblas_stride strideD,
+                                       rocblas_int bc)
+{
+    int b = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+
+    if(b < bc)
+    {
+        T* A = load_ptr_batch<T>(AA, b, 0, strideA);
+        S* D = DD + b * strideD;
+        D[0] = A[0].real();
```
EdDAzevedo, line 1068: Minor: is there a preference in using A[0].real()  vs std::real( A[0] )?

EdDAzevedo (approved):

qjojo (approved):

PR #993, opened 2025-07-18
Add optimized version of slaed4 reference secular equation solver.

pghysels:
This new version of slaed4 does not update the poles, reducing data movement.


pghysels:
See also https://github.com/ROCm/rocSOLVER/pull/985 which switches to the reference solver.

I can provide performance data.

I did not try the new tests from https://github.com/ROCm/rocSOLVER/pull/986 yet.

PR #994, opened 2025-07-21
Add graphing and rocprofv3 to Python benchmarking scripts

john00003:
This PR adds two new features:
1) The ability to automatically graph the results from the existing benchmarks in the Python scripts
2) The ability to run rocprofv3 to generate and graph profiling data regarding which kernels the most time is spent in

This reduces the effort required to generate plots to visualize how run time is changing with input size.


ammallya:

PR #995, opened 2025-07-21
Use HIP events to instrument SYEVD performance

dcyca:
Use hipEventElapsedTime() to get GPU timing information for sections of SYEVD.  This is not intended to be merged, as the current implementation requires hipStreamSynchronize() and therefore isn't graph safe.  It also isn't very robust.


TorreZuk:
@rkamd just looking at similar changes in rocblas so added as "watcher"

ammallya:

PR #996, opened 2025-07-24
Leverage hipGRAPH to optimize GEQRF

qjojo:
This uses hipGRAPH to execute the kernels used for GEQRF to reduce launch overhead. GEQRF launches a lot of kernels internally so this should result in a speedup on all targets.

Uses host scalars for GEMM inside LARFT, a future optimization would be making LARFT always use host scalars to save some extra memcpys.


tfalders:
Rather than have a preprocessor define, I wonder if we should leverage the infrastructure we have for selecting hybrid execution and add graph execution as an option. We can discuss in the next technical meeting.

library/src/auxiliary/rocauxiliary_larft.hpp
```diff
@@ -557,34 +557,41 @@ rocblas_status rocsolver_larft_template(rocblas_handle handle,
     // SYRK/HERK can be used alternatively, but GEMM is currently more performant.
     if(use_gemm)
     {
+        rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
+        T alpha, beta;
```
EdDAzevedo: Minor: I thought T*scalars contain {-1.0, 0.0, 1.0};
Perhaps rename the original device scalars pointer to "d_scalars", and setup an equivalent array and pointers on host to preserve the original code. Maybe something like the following. Just a thought.

T const neg_one{-1}; 
T const zero{0}; 
T const one{1};
 T h_scalars[] = {  neg_one, zero, one };
 T  * const scalars = &(h_scalars[0]);

qjojo: Good suggestion, I'll include that as part of the future work to rearrange LARFT.

amd-jnovotny (approved):

clients/samples/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # ##########################################################################
-# Copyright (C) 2016-2022 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2016-2025 Advanced Micro Devices, Inc. All rights reserved.
```
AGonzales-amd, line 2: Only change to this file is the copyright header.

clients/gtest/lapack/geqr2_geqrf_gtest.cpp
```diff
@@ -240,6 +265,26 @@ TEST_P(GEQRF, batched__double_complex)
     run_tests<true, true, rocblas_double_complex>();
 }
 
+TEST_P(GEQRF_GRAPH, batched__float)
+{
```
AGonzales-amd, line 269: It's not unique to this case but why is batched case `true, true` in tests but `true, false` in library?

library/src/lapack/roclapack_geqrf.hpp
```diff
@@ -139,6 +139,19 @@ rocblas_status rocsolver_geqrf_template(rocblas_handle handle,
     I ldw = GEQxF_BLOCKSIZE;
     rocblas_stride strideW = rocblas_stride(ldw) * ldw;
 
+    // enable graph capture
```
AGonzales-amd, line 142: I added an overload of geqrf_template in develop branch which calls alternative larft and larfb functions. The overload should be updated too.

AGonzales-amd (changes requested):

clients/gtest/lapack/geqr2_geqrf_gtest.cpp
tfalders: When we added `geqrf_ptr_batched`, I think it was decided that the tests for `geqrf_strided_batched` would be `true, true` since `A` is batched but `ipiv` is strided batched, and `geqrf_ptr_batched` would be `true, false` since both `A` and `ipiv` are batched arrays. As for why it's `true, false` in the library, I believe that was necessary because the private rocBLAS API interpreted `true, false` as batched.

library/src/lapack/roclapack_geqrf.hpp
```diff
@@ -166,6 +179,20 @@ rocblas_status rocsolver_geqrf_template(rocblas_handle handle,
...
+    {
+        hipGraph_t graph;
+        HIP_CHECK(hipStreamEndCapture(graph_stream, &graph));
+        rocblas_set_stream(handle, stream);
+
+        hipGraphExec_t exec;
+        HIP_CHECK(hipGraphInstantiate(&exec, graph, nullptr, nullptr, 0));
+        HIP_CHECK(hipGraphDestroy(graph));
+        HIP_CHECK(hipGraphLaunch(exec, stream));
+        HIP_CHECK(hipGraphExecDestroy(exec));
```
john00003, line 192: For batched/strided_batched inputs, could a graph be reused between input matrices?

john00003 (approved):
After the comment from Angelo regarding the new geqrf overload is addressed, these changes seem great (and provide a great introduction to using hipGraph with rocSOLVER in the future!)

AGonzales-amd (approved):

ammallya:

PR #997, opened 2025-07-26
Speedup stedc_mergeValues_kernel

Kirpich30000:
The PR speeds up "merge Values" stage of the stedc for big cases (n > 1024) by splitting the stedc_mergeValues_kernel into multiple kernels and thus allowing to solve secular equations with full parallelism (original code had a limitation on the number of groups). Smaller merges are still done using nonsplitted stedc_mergeValues_kernel to avoid overhead.

The PR doesn't change algorithm or numerical properties of calculations and the output of the stedc is supposed to be bit exact to the previous implementation. However, only basic tests were done and PR might need additional testing and verification.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_stedc.hpp
+++ b/library/src/auxiliary/rocauxiliary_stedc.hpp
@@ -1288,6 +1288,154 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
     }
 }
 
+
+template <typename S>
+void __device__ inline sort_tmpd_zz(const rocblas_int dd,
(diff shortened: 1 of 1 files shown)


Kirpich30000:
Here are a few timing results for the whole syevd gathered using rocsolver-bench

Size | Base time, us | PR time, us | Speedup
-- | -- | -- | --
1024 | 57147 | 53369 | 1.07
2048 | 143276 | 116694 | 1.23
4096 | 468226 | 347890 | 1.35
8192 | 2166989 | 1650779 | 1.31

ammallya:

PR #998, opened 2025-07-28
Small size kernel for geqr2

AGonzales-amd:
This PR adds a kernel for small size geqr2 problems. Testing shows that it is ~2x faster for square matrices with size of m <= 64 for single precision case; non square matrices appear to run slower with the kernel so it's only enabled for the square case.


tfalders (approved):

tfalders:
Changelog should be updated

Edit: We should probably also mention the performance improvements for GEQRF coming from your other PRs

amd-jnovotny (approved):
Changelog OK

jmachado-amd (approved):

PR #999, opened 2025-07-28
Fix debug build error

AGonzales-amd:
Fix the debug build error caused by mfma device functions. Use new array struct for mfma accumulator instead of std::array for complex types

It's unusual that std::array can be used on the device in release build but not in debug.

### DIFF
--- a/library/src/include/roclapack_gemm_device_functions.hpp
+++ b/library/src/include/roclapack_gemm_device_functions.hpp
@@ -74,7 +74,29 @@ template <typename T>
 struct mfma_16x16x4
 {
     using RegT = T;
-    using AccT = std::array<T, 4>;
+
+    struct AccT
+    {
+        T data[4];
+
+        __host__ __device__ AccT(T val)
+            : data{val}
+        {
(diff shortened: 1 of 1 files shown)


library/src/include/roclapack_gemm_device_functions.hpp
```diff
@@ -74,7 +74,29 @@ template <typename T>
 struct mfma_16x16x4
 {
     using RegT = T;
-    using AccT = std::array<T, 4>;
+
+    struct AccT
```
EdDAzevedo, line 78: Minor: I wonder whether something simpler like  
typedef T AccT4];
and initialize a variable as
AccT   zarr4{ z0, z1, z2, z3 };
may be easier to understand and maintain.
   
  Just a thought.

EdDAzevedo (approved):

tfalders (approved):

jmachado-amd (approved):
Thanks @AGonzales-amd!

ammallya:

PR #1003, opened 2025-08-03
Stedc new split-blocks handling

jzuniga-amd:


library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1000,37 +206,29 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
-    rocblas_int mid = hipBlockIdx_x;
+    rocblas_int sid = hipBlockIdx_x;
     // thread id
     rocblas_int tidb = hipThreadIdx_x;
     rocblas_int tid, tx;
-    /* --------------------------------------------------- */
 
     // select batch instance to work with
-    /* --------------------------------------------------- */
     S* C;
```
EdDAzevedo, line 218: Perhaps use default value as "S* C = nullptr;" to make sure variable "C"  has a defined value. If CC is nullptr, then variable "C"  may have undefined value?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1041,269 +239,219 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+        sz = ns[tid];
+        for(int j = 1; j < bd; ++j)
+            sz += ns[tid + j];
+        // with this, all threads involved in a merge
+        // will point to the same row of C and the same off-diag element
+        S* ptz = (iam == 0) ? C + p2 - 1 + sz : C + p2;
+        S p = (iam == 0) ? 2 * E[p2 - 1 + sz] : 2 * E[p2 - 1];
+
+        // copy elements of z
+        for(int j = tx; j < sz; j += dim)
```
EdDAzevedo, line 281: Minor: I wonder whether we can use something like  "S const inv_sqrt2 = (S)  1.0/std::sqrt( 2.0);" to make sure the compiler is using a multiply instead of division if this is performance critical.

for(int j ... ) { 
z[p2 + j] = ptz[(p2 + j) * ldc] * inv_sqrt2;
}

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1635,221 +719,172 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
-                            }
-                            temps[pers[i - dd] + in + (p2 + j) * n]
-                                = vecs[i - dd - in + (p2 + j) * n] / nrm;
+                            if(idd[k] == 0)
+                                dd++;
                         }
-                        else
-                            temps[i + (p2 + j) * n] = 0;
+                        temps[pers[i - dd] + in + (p2 + j) * n]
+                            = vecs[i - dd - in + (p2 + j) * n] / nrm;
```
EdDAzevedo, line 826: Minor: perhaps consider "inv_nrm = 1.0/nrm" outside the for loop and perform multiply in body of  for loop?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1355,230 +495,182 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+            sz += ns[in + i];
+        in = ps[in];
+
+
+        // 1. Organize data with non-deflated values to prepare secular equation
+        // ----------------------------------------------------------------- 
+        // All threads of the group participating in the merge will work together
+        // to solve the correspondinbg secular eqn. Now 'iam' indexes those threads
+        iam = tidb;
+        bdm = hipBlockDim_x;
```
EdDAzevedo, line 544: Minor: is using all threads in the thread block this way the best use of resources? Would using a warp or sub-warp (using tiled partition of  cooperative groups)  be reasonable?  It is still possible to synchronize a group of threads. Just a thought.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -181,11 +129,10 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_divide_kernel(const ro
                                                                         rocblas_int* splitsA)
 {
     // threads and groups indices
-    rocblas_int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
-    rocblas_int numthds = hipGridDim_x * hipBlockDim_x;
+    rocblas_int bid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
```
EdDAzevedo, line 74: Minor: is the code  using a single thread for each batch matrix?  

I wonder whether the code can use more resources, say 1 warp or 1 thread block per batch matrix.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -421,269 +297,255 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
-                            for(int ii = 0; ii < n; ++ii)
-                            {
-                                valf = C[ii + base * ldc];
-                                valg = C[ii + top * ldc];
-                                C[ii + base * ldc] = valf * c - valg * s;
-                                C[ii + top * ldc] = valf * s + valg * c;
-                            }
+                            valf = C[ii + base * ldc];
+                            valg = C[ii + top * ldc];
+                            C[ii + base * ldc] = valf * c - valg * s;
```
EdDAzevedo, line 401: Minor: I wonder whether std::fma()  may be useful for higher accuracy and perhaps better performance in computing the 2x2 matrix multiplication.

tfalders (approved):
Looks good! Static analysis seems to be failing, though I'm not sure if we should address this since it may create conflicts with the upcoming PR. Perhaps we should do a format check on every file except the stedc header.

jmachado-amd (approved):

tfalders:
Forcing the merge. There are static analysis failures, but resolving them will probably create merge conflicts with upcoming PR so leaving as-is.

PR #1004, opened 2025-08-08
Recursive QR

EdDAzevedo:
Recursive implementation of  QR factorization for m by n matrices where m >= n. The algorithm is based on the paper "Applying recursion to serial and parallel QR factorization" by Erik Elmroth and Fred G. Gustavson, IBM Journal of Research and Development, August 2000.

One idea is the use of routine "rgeqr3" for factorizing block columns instead of using "geqr2" and "larft". The "T" upper triangular matrix for block Householder transform is explicit constructed and updated as part of the factorization. The block Householder transform has the form   H = eye - Y * T  * Y',   where Y is unit lower trapezoidal.

The QR factorization of m by n matrix   A(1:m, 1:n)  can be recursively performed by splitting the columns with  n1 = n/2.

Let the QR factorization of A(1:m, 1:n1)  yields  (Y1, R1, T1)   where R1 is the upper triangular "R" matrix and 
Q1 = eye - Y1 * T1 * Y1'.  Note Y1 is  m by n1  unit lower trapezoidal.

Let  the QR  factorization of   the  updated submatrix 

Ahat(1:m, (n1+1):n) <---  Q1' * A(1:m, (n1+1):n)  

yield   (Y2, R2, T2), then the final factorization is matrix Y = [Y1, Y2] is m by n unit  lower trapezoidal,   the upper triangular "R" matrix is

```
R = [R1       Ahat(1:n1,  (n1+1):n) ]
    [ 0          R2                 ]
```
       
and the upper triangular matrix for block Householder transform is
```
T = [T1       T3]
    [         T2]
```

where T3 = -T1 * (Y1' * Y2) * T2


ammallya:

PR #1005, opened 2025-08-08
Small-size sytd2 ignore imaginary values on the diagonal

tfalders:
Unlike previous versions of SYTRD, the new small-size SYTD2 kernels do not ignore imaginary values on the diagonal. This causing hipSOLVER tests to fail, since those tests do not explicitly remove the imaginary component from the diagonal.

### DIFF
--- a/library/src/lapack/roclapack_sytd2_hetd2.hpp
+++ b/library/src/lapack/roclapack_sytd2_hetd2.hpp
@@ -87,6 +87,10 @@ ROCSOLVER_KERNEL void __launch_bounds__(MAX_THDS)
         const auto tidy = tid / (MAX_THDS / 2);
         for(I j = tidy; j < n; j += 2)
         {
+            // ignore imaginary part of the diagonal
+            if(i == j)
+                a[i + j * n] = std::real(a[i + j * n]);
+            // copy lower triangle to upper triangle
             if(i < j)
                 a[i + j * n] = conj(a[j + i * n]);
         }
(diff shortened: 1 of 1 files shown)


library/src/lapack/roclapack_sytd2_hetd2.hpp
```diff
@@ -87,6 +87,10 @@ ROCSOLVER_KERNEL void __launch_bounds__(MAX_THDS)
         const auto tidy = tid / (MAX_THDS / 2);
         for(I j = tidy; j < n; j += 2)
         {
+            // ignore imaginary part of the diagonal
```
EdDAzevedo, line 90: Minor: is the code trying to view the one dimensional array of  MAX_THDS threads as  2 by (MAX_THDS/2) two dimensional array of threads? If so, perhaps there can be a comment. 

Another idea is to launch as   dim3(2, MAX_THDS/2,1) array of threads then compute " tid = tx + ty * blockDim.x;". Maybe this may make the code easier to read and understand?
---------------
With Fortran column-oriented storage  of index calculation as  "i + j * lda", one may also consider using the "for(i)" loop  for rows as the inner loop and "for(j)" over columns as the outer loop to encourage stride-1 contiguous read or write  access in the inner loop.  It may also be of benefit to organize the threads so that the threads of the same warp is has stride-1 access to device memory as the inner "for(i)" loop.

Perhaps consider launching as   
dim3( warp_size,  MAX_THDS/warp_size,1) then the loop may be something like the following. Just a suggestion.


 auto nx = blockDim.x;  auto ny = blockDim.y; 
 auto tx = threadIdx.x;  auto ty = threadIdx.y;
 for( auto j=ty; j < n; j += ny ) {
   for(auto i=tx; i < n; i += nx) {
     
       a[ i + j  * n ] = A(i + j * lda );
     }
     }
     // to use a one-dimensional array of threads 
     auto txy = tx  + ty * nx;  auto nthreads = nx * ny;

EdDAzevedo (approved):

jmachado-amd (approved):

PR #1006, opened 2025-08-12
Improved stedc

tfalders:


library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -50,6 +51,33 @@ ROCSOLVER_BEGIN_NAMESPACE
...
+    // 9     rocblas_int dcount[n];   // number of deflations
+    // 10    rocblas_int mns[n];      // merge sizes
+    // 11    rocblas_int mps[n];      // merge initial positions
+    // 12    rocblas_int dbg[n];      //
+    // 13    rocblas_int midd[n];     // sorted idd
+    // 14    rocblas_int dbg2[n];     //
+    // 15    rocblas_int dbg3[n];     //
+    // 16    rocblas_int dbg4[n];     //
+    // };
+    return 17 * n + 2;
```
EdDAzevedo: Minor: consider
   return  (n+1) + 1 + 16 * n;
   to reflect how this expression was calculated.
  --------------------- 
   Another approach is to have a variable with running partial sum say
    rocblas_int num_ints = 0;
       // 0     rocblas_int splits[n+1]; // positions where each block begings
       num_ints += (n+1);
    //       rocblas_int nb;          // total number of split blocks
    num_ints += 1;
    // 1     rocblas_int ns[n];       // the sub-blocks sizes
    num_ints += n;
    ... etc 
    in case in future, there are changes in the removal of arrays or there are  additional arrays, this may make it easier to calculate and ensure that  the correct value is returned.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+    rocblas_int* dds = szs + n;
+    // the rank-1 modification vectors in the merges
+    S* z = tmpzA + bid * (2 * n);
+    // temp for tolerance values used in deflation
+    // IMPORTANT: tols maps to the same memory as evs
+    S* tols = z + n;
+    // updated eigenvectors after merges
+    S* vecs = vecsA + bid * 2 * (n * n);
+    // temp values during the merges
+    S* temps = vecs + (n * n);
```
EdDAzevedo, line 428: Minor: may consider another variable, say 
auto pfree = vecs;
S * temps = pfree; pfree += n * n;
S* cc = pfree; pfree +=  n;
S* ss = pfree; pfree += n;   
...etc 

in case in future, some of the arrays are removed or added or the size changes  (say if array cc is modified to size n+1), using an extra variable "pfree' may simplify the change and ensure correctness independent of the order of allocation.       Just a minor suggestion.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+                eq += (dz == dzs[i]) && (bval[i] == d && (p + x) < sid);
+            }
+        }
+    }
+
+    int pos = lt + eq;
+    __shared__ int lds[STEDC_BDIM];
+    int bit = 1;
+    while (bit < STEDC_BDIM) {
+        lds[hipThreadIdx_x ^ bit] = pos;
```
EdDAzevedo, line 573: Minor: it would be nice to have a comment about  what is the while loop  trying to do and the action of  taking XOR of index.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+
+    if (hipThreadIdx_x == 0) {
+        map [pos+p] = sid;
+        md  [pos+p] = d;
+        midd[pos+p] = dz;
+    }
+
+    __syncthreads();
+    if (nan) {
+        md[pos + p] = NAN;
```
EdDAzevedo: Minor: it would be nice if there is a comment. Is it (on purpose) assigning a numerical NaN value into the array?  Perhaps the name of the variable "nan" and "NAN" are misleading?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+    rocblas_int* dbg    = splits + 12 * n + 2; 
+    rocblas_int* midd   = splits + 13 * n + 2;
+    rocblas_int* dbg2   = splits + 14 * n + 2; 
+    rocblas_int* dbg3   = splits + 15 * n + 2; 
+    rocblas_int* dbg4   = splits + 16 * n + 2; 
+
+    constexpr rocblas_int deflate_max_n = 32768 / sizeof(S);
+    __shared__ S   ldsD[deflate_max_n];
+    __shared__ int cand[deflate_max_n];
+    constexpr int L_F_BCAND_BIT = 0; // bit indicating base candidate
```
EdDAzevedo: Minor: personally, I find bit manipulations to be a little opaque to understand and error prone. It would be helpful if there can be some comments, say bit manipulation to avoid integer multiplication and division, or  trying to encode different pieces of information in the bit pattern to save space (in limited LDS memory), say bits 0:13 is lower bound, bits 14:27 is upper bound. If encoding information in some bits,  perhaps uint32_t  instead of int may be more appropriate.  Maybe using bit fields may also make the code clearer to let the compiler handle the bit manipulations?

Perhaps some #define macros might help to make the code clearer  as well.    Just a suggestion.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+            S baseval = lz[base];
+            for (int j = 0; j < cnt; j++) {
+                int top = base + j + 1;
+                S g = lz[top];
+                S f = baseval;
+                S c, s, rr;
+                lartg(f, g, c, s, rr);
+                baseval = rr;
+
+                if (1) {
```
EdDAzevedo: Minor: maybe a comment might help to explain the motivation or intent of using "if (1)". 
Maybe the original intent was   something like the following?

bool const is_valid = (0 <= top) && (top < n) &&  (0 <= lmap[top]) && (lmap[top]) < n);
if (is_valid) {
 idd[lmap[top]] = 0;   z[ lmap[top] ] = 0;  cc[ lmap[top] ] = c; ss[ lmap[top] ] = s; }

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+template <typename S>
+void __device__ inline sort_tmpd_zz(const rocblas_int dd,
+                                    const rocblas_int iam,
+                                    const rocblas_int bdm,
+                                    S* tmpd,
+                                    S* zz,
+                                    rocblas_int* per)
+{
+    // Order the elements in tmpd and zz using a simple parallel selection/bubble sort.
+    // This will allow us to find initial intervals for eigenvalue guesses
```
EdDAzevedo, line 1120: Minor: is this the implementation of the parallel odd/even bubble sort? 
https://en.wikipedia.org/wiki/Odd%E2%80%93even_sort

 If not, will it guarantee a sorted array?  Perhaps the comment may state this more explicitly.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1278,26 +2500,92 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
 
         // 3. merge phase
         //----------------
-        size_t lmemsize1 = sizeof(S) * 2 * STEDC_BDIM;
         size_t lmemsize3 = sizeof(S) * STEDC_BDIM;
         rocblas_int numgrps3 = ((n - 1) / blks + 1) * blks;
+        char* env_new_merge_values  = getenv("MERGE_VALUES_NEW");
```
EdDAzevedo: Minor: perhaps these environment variables will be modified in future?

EdDAzevedo (approved):
This is a very sophisticated (and complex) piece of code.  This is just a superficial review.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+            sz += ns[tid + j];
+        szfs[tid] = sz;
+        // with this, all threads involved in a merge
+        // will point to the same row of C and the same off-diag element
+        S* ptz = (iam == 0) ? C + p2 - 1 + sz : C + p2;
+        S p = (iam == 0) ? 2 * E[p2 - 1 + sz] : 2 * E[p2 - 1];
+
+        // copy elements of z
+        for(int j = tx; j < sz; j += dim)
+            z[p2 + j] = ptz[(p2 + j) * ldc] / sqrt(2);
```
EdDAzevedo, line 315: Minor: perhaps consider  precomputing  inverse of sqrt(2) to  using multiply (instead of divide) in for loop?  
auto const inv_sqrt2 = 1.0/std::sqrt(2.0);

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+        __syncthreads();
+
+        dim2 = dim;
+        while(dim2 > 0)
+        {
+            if(tidb < dim2)
+            {
+                valf = inrmsd[tidb + dim2];
+                valg = inrmsz[tidb + dim2];
+                maxd = valf > maxd ? valf : maxd;
```
EdDAzevedo, line 341: Minor: perhaps consider using std::max()?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+                        idd[top] = 0;
+                        // rotation to eliminate component in z
+                        S f, g, c, s, rr;
+                        g = z[top];
+                        f = z[base];
+                        lartg(f, g, c, s, rr);
+                        z[base] = rr;
+                        z[top] = 0;
+                        // update C with the rotation
+                        for(int ii = 0; ii < n; ++ii)
```
EdDAzevedo: Minor: I wonder whether the for(ii) loop to update C can be performed in parallel, especially if n is not small. Perhaps need to  launch with 2D array of threads in thread block or repartition the threads into  2D array?    The sync may wait for the slowest    thread.     Just a thought.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+            && std::abs(d - dn) <= tol  // within tolerance
+            && i != (n-1)               // isn't last 
+            ;
+        int tcandidate =
+            midd[i] && midd[prev]       // not yet deflated
+            && p == pp                  // in the same merge block
+            && std::abs(d - dp) <= tol  // within tolerance
+            && i > 0                    // isn't first
+            ;
+        cand[i] = (bcandidate << L_F_BCAND_BIT) + (tcandidate << L_F_TCAND_BIT);
```
EdDAzevedo, line 668: Is array "cand" involved with integer calculations or just holding two bool values (essentially 2 bits)? If it is just holding two bool values, how about explicitly   using two bool arrays, say cand_bcandidate[], and cand_tcandidate[]?  Would this avoid the bit manipulations to make the code easier to read? If space is very tight, then perhaps use array of bits and #define macros?  If the max size is known at compile time, perhaps std::bitset may be used?   Just a thought.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -231,238 +257,1512 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+
+
+    rocblas_int dgs = hipBlockIdx_x;
+    rocblas_int dcnt = dcounts[dgs];
+    if (dcnt)
+    {
+        rocblas_int base = map[dgs];
+        S* Cbase = C + base * ldc;
+
+        for (int chunk = 0; chunk < n_chunks; chunk++) {
```
EdDAzevedo: Minor: just checking whether there are opportunities for parallelism to use multiple threads? All the for loops seem to be serial without parallelism.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: Agreed. Fixed in the latest commit. The code related to splits and temps should ideally be refactored. The motivation in the new code is to minimize the changes and related risk and more or less to follow existing code.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: Code related to temps/vecs/splits ideally needs to be refactored, but not in this PR as it is already quite heavy. So using "existing" approach for now.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: done

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: done

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: removed, it was a debug related leftover

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: That peace is the old code and old comment wrapped into a function to avoid code duplication

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: The env variables are kind of a chiken-bits and in a sense a debug related code. If the functionality and performance are well tested and have no regressions, it should be removed. Not sure if we should keep it for a while or not. That is probably up to a code/project owner to decide. The way it is implemented now is a quick hack. If it is decided to have those for a while, env variables have to be properly renamed and the query have to be done in a better way.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: That is an old code untouched. Anyway the kernel (the way it is currently implemented) is not even close to be compute bound and I hope compiler is smart enough to do trivial optimization where applicable. So it will not buy us anything except making a code a bit less readable.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: an old code untouched

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: an old code untouched

Kirpich30000: Also that very part is now implemented in a better way in stedc_mergeRotate_kernel()

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: Those are two bool values. The bits have to be packed due a space constrains. A bitset could be a possibility, but needs to be tested (probably in a different PR)

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: It does use a parallelism as memory operations are async. And it should be more deterministic and cause less memory thrashing comparing to "multiple threads". The way it is written should be easier for compiler to make a proper loop unroll and keep data in registers.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: Would using uint32_t or uint64_t be better than type int for packing bits?  Maybe some comments  that  describe the bit packing or use  #define macro or lambda  to hide the bit manipulation  might  be helpful for future maintainers.   

 Just a suggestion.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -160,77 +190,1382 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_solve_kernel(const roc
...
+                bval[i] = D[p + x];
+                dzs[i] = idd[p + x];
+            }
+        }
+        for(int i = 0; i < regs; i++)
+        {
+            int x = chunk * chunk_width + i * hipBlockDim_x + hipThreadIdx_x;
+            if(x < sz)
+            {
+                nan += (bval[i] != bval[i]);
```
EdDAzevedo, line 551: Minor: consider using std::isnan() to make it clear what is the intent of the code, and avoid possibility of  incorrect compiler optimization

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: Minor: consider a more conventional loop to compute the reduction using LDS without the need for XOR of the index. I suspect the XOR has only a small impact on improving the performance. Perhaps the __syncthreads may be more expensive.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -160,77 +190,1382 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_solve_kernel(const roc
...
+        md[pos + p] = d;
+        midd[pos + p] = dz;
+    }
+
+    __syncthreads();
+    // The NAN fp value is unordered, so it is possible that with computed
+    // new positions it would be silently overwriten with non NAN value.
+    // Make sure we propagate NAN. It is likely to have more NANs in the output
+    // than in the input, but the following computations are doomed anyway.
+    if(nan)
```
EdDAzevedo, line 592: Minor: if there are  NaNs detected, should the code just abandon the calculations and return with an error status?   Just a thought.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: Minor: perhaps mention in comments how many packed 2-bits are stored in the 32bit signed integer "cand[]" array?  Say, if uint32_t were used, one may expect 16  copies of packed  2-bits per entry in "cand[]" array.  I think a bool array uses a byte for each bool value.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -160,77 +190,1382 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_solve_kernel(const roc
...
+                for(int i = 0; i < regs; i++)
+                {
+                    int x = chunk * chunk_width + i * hipBlockDim_x + hipThreadIdx_x;
+                    if(x < n)
+                    {
+                        tval[i] = Ctop[x];
+                    }
+                }
+
+                for(int i = 0; i < regs; i++)
```
EdDAzevedo, line 945: Minor: since regs is constexpr == 16, perhaps consider adding #pragma unroll   hint for compiler?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -160,77 +190,1382 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_solve_kernel(const roc
...
+            rocblas_int linfo;
+
+#if defined(ROCSOLVER_USE_REFERENCE_SECULAR_EQUATIONS_SOLVER)
+            linfo = slaed4(dd, cc, tmpd + j * n, zz, std::abs(p), ev[j]);
+#else
+            if(cc == dd - 1)
+                linfo = seq_solve_ext(dd, tmpd + j * n, zz, (p < 0 ? -p : p), ev + j, eps, ssfmin,
+                                      ssfmax);
+            else
+                linfo = seq_solve(dd, tmpd + j * n, zz, (p < 0 ? -p : p), cc, ev + j, eps, ssfmin,
```
EdDAzevedo, line 1217: Minor: perhaps use abs(p)?

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -299,170 +331,1335 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+        in = ps[in];
+
+        // first deflate zero components
+        for(int i = tidb; i < sz; i += hipBlockDim_x)
+        {
+            tx = in + i;
+            S g = z[tx];
+            if(abs(p * g) <= tol)
+                // deflated ev because component in z is zero
+                idd[tx] = 0;
```
EdDAzevedo, line 376: Minor: is this  a bool expression?

  bool const is_deflated_ev = ( std::abs( p * g ) <= tol);
  idd[tx] =   !is_deflated_ev;

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -299,170 +331,1335 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+    rocblas_int* mps = splits + 11 * n;
+    rocblas_int* cand = splits + 12 * n;
+    rocblas_int* midd = splits + 13 * n;
+
+    S d = D[sid];
+    rocblas_int sz = mns[sid];
+    rocblas_int p = mps[sid];
+    rocblas_int dz = idd[sid];
+
+    constexpr int regs = 8;
```
EdDAzevedo, line 525: Minor: consider consistently using rocblas_int, or add "typename I = rocblas_int" in template argument and consistently "I", maybe   something like 
   I nan = 0;  I lt = 0;   I eq = 0;

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -299,170 +331,1335 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+    rocblas_int* midd = splits + 13 * n;
+
+    constexpr int F_BCAND = 1 << L_F_BCAND_BIT;
+    constexpr int F_TCAND = 1 << L_F_TCAND_BIT;
+
+    // find deflate candidates
+    int i = hipThreadIdx_x + hipBlockDim_x * hipBlockIdx_x;
+    if(i < n)
+    {
+        int next = (i + 1) < n ? (i + 1) : i;
```
EdDAzevedo, line 648: Minor: perhaps the code is performing something like 

  auto const  next = std::min( n-1, i+1);
  auto const  prev = std::max( 0,  i-1);
  
 if so,    since 0, and n-1 are the bounds for the array,  the above code sequence may be easier to understand as limiting  the values of  (i-1), (i+1) to be within the bounds of the array?

EdDAzevedo (approved):

AGonzales-amd (commented):
Troy's approval

AGonzales-amd (approved):
Troy's approval

PR #1017, opened 2025-08-27
Improved stedc (reshuffle gemm)

tfalders:
Code refactor and improvements to gemm calls.

- Original code used monolithic gemms that cover the entire matrix. Breaking them down to only work on blocks improves performance.
- Eigenvalues are now sorted. They are sorted before deflation to make deflation code faster, and after deflations to group non deflated values together. That allows to simplify and speedup the rest of the code.
- There are redundant data movements of D/ev and vectors (the reason is to localize the changes and structurally mimic original processing to reduce risks of a big refactoring). Optimizing it should provide additional (but probably minor) performance boost and simplify code. But it is best to postpone it to a separate PR
- Performance of the secular eq solver was improved.
- Performance of the mergeVectors was significantly improved.

### DIFF
--- a/library/src/auxiliary/rocauxiliary_stedc.hpp
+++ b/library/src/auxiliary/rocauxiliary_stedc.hpp
@@ -44,6 +44,8 @@
 
 ROCSOLVER_BEGIN_NAMESPACE
 
+#define DEBUG_OUTPUT 1
+
 #define STEDC_BDIM 512 // Number of threads per thread-block used in main stedc kernels
 #define STEDC_SOLVE_BDIM 4  // Number of threads per thread-block used in solver kernel
 
@@ -558,20 +560,20 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
     int pos = lt + eq;
     __shared__ int lpos[STEDC_BDIM];
     __shared__ int ldd[STEDC_BDIM];
-    // Reduction (sum of pos across all lanes in a workgroup)
-    // on each iteration reduction is done within a subgroup of size of 2*bit
-    // by xoring corresponding bit of an address.
-    // The faster implementation should use dpp + a single trip through lds
-    // but keeping code simple for now.
-    int bit = 1;
-    while (bit < STEDC_BDIM) {
-        lpos[hipThreadIdx_x ^ bit] = pos;
-        ldd[hipThreadIdx_x ^ bit] = dd;
+
+    // reduction
+    lpos[hipThreadIdx_x] = pos;
+    ldd[hipThreadIdx_x] = dd;
+    for(int r = hipBlockDim_x / 2; r > 0; r /= 2)
+    {
         __syncthreads();
-        pos += lpos[hipThreadIdx_x];
-        dd += ldd[hipThreadIdx_x];
-        __syncthreads();
-        bit *= 2;
+        if(hipThreadIdx_x < r)
+        {
+            pos += lpos[hipThreadIdx_x + r];
+            dd += ldd[hipThreadIdx_x + r];
+            lpos[hipThreadIdx_x] = pos;
+            ldd[hipThreadIdx_x] = dd;
+        }
     }
 
     if (hipThreadIdx_x == 0) {
@@ -1309,19 +1311,19 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
             nrm += valf * valf;
             putvec[i + eid*n] = valf;
         }
-        inrms[tidb] = nrm;
-        __syncthreads();
 
         // reduction (for the norms)
+        inrms[tidb] = nrm;
         for(int r = dim / 2; r > 0; r /= 2)
         {
+            __syncthreads();
             if(tidb < r)
             {
                 nrm += inrms[tidb + r];
                 inrms[tidb] = nrm;
             }
-            __syncthreads();
         }
+        __syncthreads();
         nrm = sqrt(inrms[0]);
     }
 
@@ -1355,19 +1357,19 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
                 S temp = 0;
                 for(int kk = tidb; kk < dd; kk += dim)
                     temp += C[i + (p1 + kk) * ldc] * etmpd[kk + eid * n];
-                inrms[tidb] = temp;
-                __syncthreads();
 
                 // reduction
+                inrms[tidb] = temp;
                 for(int r = dim / 2; r > 0; r /= 2)
                 {
+                    __syncthreads();
                     if(tidb < r)
                     {
                         temp += inrms[tidb + r];
                         inrms[tidb] = temp;
                     }
-                    __syncthreads();
                 }
+                __syncthreads();
 
                 // result
                 if(tidb == 0)
@@ -1623,27 +1625,27 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_sort(const rocblas_int
         }
     }
 
+
     int pos = lt + eq;
-    __shared__ int lds[STEDC_BDIM];
-    // Reduction (sum of pos across all lanes in a workgroup)
-    // on each iteration reduction is done within a subgroup of size of 2*bit
-    // by xoring corresponding bit of an address.
-    // The faster implementation should use dpp + a single trip through lds
-    // but keeping code simple for now.
-    int bit = 1;
-    while(bit < STEDC_BDIM) {
-        lds[hipThreadIdx_x ^ bit] = pos;
+    // reduction
+    __shared__ int lpos[STEDC_BDIM];
+    lpos[hipThreadIdx_x] = pos;
+    for(int r = hipBlockDim_x / 2; r > 0; r /= 2)
+    {
         __syncthreads();
-        pos += lds[hipThreadIdx_x];
-        __syncthreads();
-        bit *= 2;
+        if(hipThreadIdx_x < r)
+        {
+            pos += lpos[hipThreadIdx_x + r];
+            lpos[hipThreadIdx_x] = pos;
+        }
     }
+    __syncthreads();
+    pos = lpos[0];
 
     if (hipThreadIdx_x == 0) {
         Dout[pos] = d;
     }
 
-    __syncthreads();
     // The NAN fp value is unordered, so it is possible that with computed
     // new positions it would be silently overwriten with non NAN value.
     // Make sure we propagate NAN. It is likely to have more NANs in the output
@@ -1690,11 +1692,13 @@ inline rocblas_int stedc_num_levels(const rocblas_int n)
     else
         levels = std::ceil(std::log2(n)) - 4;
 
+#if DEBUG_OUTPUT
     char* env_levs = std::getenv("LEVS");
     if(env_levs)
     {
         levels = std::atoi(env_levs);
     }
+#endif
 
     return levels;
 }
@@ -1837,7 +1841,6 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
     ROCSOLVER_ENTER("stedc", "evect:", evect, "n:", n, "shiftD:", shiftD, "shiftE:", shiftE,
                     "shiftC:", shiftC, "ldc:", ldc, "bc:", batch_count);
 
-#define DEBUG_OUTPUT 1
 #if DEBUG_OUTPUT
     static int global_cnt = 0;
     global_cnt++;
--- a/library/src/auxiliary/rocauxiliary_stedc.hpp
+++ b/library/src/auxiliary/rocauxiliary_stedc.hpp
@@ -64,69 +64,36 @@ __host__ __device__ inline rocblas_int get_splits_size(const rocblas_int n)
     // n - number of eigenvalues (matrix size)
     // m - number of merges on a current level
     // struct {
-    // 0     rocblas_int sort_tmp[n];         // temp buffer used for mapping in stedc_sort()
-    // 1     rocblas_int      ;               //
-    // 2     rocblas_int      ;               //
-    // 3     rocblas_int idd[n];              // if idd[i] = 0, the value in position i has been deflated  (aka mask)
-    // 4     rocblas_int ;             //
-    // 5     rocblas_int ;      //
-    // 6     rocblas_int ;      //
-    // 7     rocblas_int dds[m], _[n-m];      // degrees of secular equation
-    // 8     rocblas_int map[n];              // 
+    // 0     rocblas_int msz[m], _[n-m];      // size of each merge
+    // 1     rocblas_int mps[m], _[n-m];      // starting position of each merge
+    // 2     rocblas_int bsz[2*m], _[n-2*m];  // size of each block
+    // 3     rocblas_int bps[2*m], _[n-2*m];  // starting position of each block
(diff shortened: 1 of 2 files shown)


library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -40,10 +40,14 @@
...
 #include <algorithm>
 #include <cmath>
+#include <cstdlib>
 
 ROCSOLVER_BEGIN_NAMESPACE
 
+#define DEBUG_OUTPUT 1
+
 #define STEDC_BDIM 512 // Number of threads per thread-block used in main stedc kernels
+#define STEDC_SOLVE_BDIM 4 // Number of threads per thread-block used in solver kernel
```
EdDAzevedo, line 48: Minor: just double checking whether this is really the intended value of using only 4 threads per thread block.   Perhaps use at least one full warp per thread block?

Can  using cooperative groups and partition using 4 threads for  each sub-partition be considered?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -58,26 +62,182 @@ ROCSOLVER_BEGIN_NAMESPACE
...
+    // 10    rocblas_int map[n];              // original indices of a sorted values
+    // 11    rocblas_int cand[n];             // deflation candidate flags
+    // 12    rocblas_int dbg[n];              //
     // };
-    return 17 * n;
+    return 13 * n;
+}
+
+template <typename S>
+__host__ __device__ inline S* ptr_msz(rocblas_int n, S* splits)
```
EdDAzevedo, line 84: Minor: I wonder what is the advantage over calling the "_getMemorySize()" and adjust the "pfree" moving pointer to allocate storage.   

For example, is the order of allocation essential for correctness?  What if one of the array size is modified or array is  removed,  how easy and robust can the code be changed.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -182,43 +337,110 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_solve_kernel(const roc
...
+        - This kernel is to be called with 1 group in x and batch_count groups in y.
+        - Size of groups is set to STEDC_BDIM. **/
+ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_update_splits(const rocblas_int levs,
+                                                                        const rocblas_int k,
+                                                                        const rocblas_int n,
+                                                                        rocblas_int* splitsA)
+{
+    rocblas_int bid = hipBlockIdx_y;
+    rocblas_int* splits = splitsA + bid * get_splits_size(n);
+    rocblas_int* em = ptr_em(n, splits);
```
EdDAzevedo, line 372: Minor suggestion: maybe consider   something like the following

std::byte *  pfree = (std::byte *) splits;
rocblas_int * em = (rocblas_int *) pfree; pfree += size_em_bytes;
rocblas_int * msz = (rocblas_int *) pfree; pfree += size_msz_bytes;
rocblas_int * nsz = (rocblas_int* ) pfree; pfree += size_nsz_bytes; 

... etc.

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -591,7 +681,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
     // than in the input, but the following computations are doomed anyway.
     if(nan)
     {
-        md[sid] = NAN;
+        md[gid] = NAN;
```
EdDAzevedo, line 682: Minor: just double checking that there is something like 

auto const NAN = nan("");  

if computation is doomed anyway, should there be a "break" or "return" or something like "goto L9999" at the end?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -975,687 +1005,267 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+            if(cc == dd - 1)
+                linfo = seq_solve_ext(dd, etmpd + i * n, z + p1, std::abs(p), evs + i, eps, ssfmin,
+                                      ssfmax);
+            else
+                linfo = seq_solve(dd, etmpd + i * n, z + p1, std::abs(p), cc, evs + i, eps, ssfmin,
+                                  ssfmax);
+#endif
+
+            if(p < 0)
+                evs[i] *= -1;
```
EdDAzevedo, line 1252: Minor: is multiply by -1 better than setting 
evs[i] = -evs[i]?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1664,340 +1274,207 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
+    rocblas_int i = eid - p1;
+    if(i < dd)
     {
-        rocblas_int iam, sz, dim, p2;
-        S valf, valg;
+        S valf = 1;
+        for(int j = hipThreadIdx_x; j < dd; j += hipBlockDim_x)
+        {
+            S valg = etmpd[(p1 + j) * n + i];
+            valf *= ((p1 + i) == (p1 + j)) ? valg : valg / (D[p1 + i] - D[p1 + j]);
```
EdDAzevedo, line 1306: Minor: I wonder whether one can perform a check before doing the division, say something like 
auto const max_val = std::numeric_limits<T>::max();

// assume signs are all positive
auto val = (valg > (D[p1+i] - D[p1+j]) * max_val) ? max_val :   valg/( D[p1+i]-D[p1+j]);

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -1664,340 +1274,207 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
...
-          be in the matrix if its size is exact multiple of the number of sub-blocks 'blks'.
+        - Call this kernel with batch_count groups in y, and n groups in x.
           Each group works with a column. Groups are size STEDC_BDIM.
         - If a group has an id larger than the actual number of columns it will do nothing. **/
 template <bool USEGEMM, typename S>
 ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM)
-    stedc_mergeVectors_kernel(const rocblas_int levs,
-                              const rocblas_int blks,
-                              const rocblas_int k,
+    stedc_mergeVectors_kernel(const rocblas_int k,
```
EdDAzevedo, line 1340: Minor: I wonder whether USEGEMM need to be a template argument, or whether it can be a regular bool argument?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2164,6 +1603,57 @@ ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_copyC(const rocblas_in
...
+            int x = chunk * chunk_width + i * hipBlockDim_x + hipThreadIdx_x;
+            if(x < n)
+                dst[x] = bval[i];
+        }
+    }
+}
+
 /** STEDC_SORT sorts computed eigenvalues and eigenvectors in increasing order **/
 template <typename T, typename S, typename U1, typename U2>
 ROCSOLVER_KERNEL void __launch_bounds__(STEDC_BDIM) stedc_sort(const rocblas_int n,
```
EdDAzevedo, line 1657: Minor:  perhaps consider a separate kernel launch  to check for number of nan and lt and eq, rather than doing this inside a "sort" kernel?

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2417,6 +1907,91 @@ rocblas_status rocsolver_stedc_argCheck(rocblas_handle handle,
...
+    {
+        s += sz[i];
+    }
+    if(s != n)
+    {
+        std::cout << "ERROR: n(" << n << ") != s(" << s << ")\n";
+        std::exit(0);
+    }
+
+    bool is_blk_diag = true;
```
EdDAzevedo: Minor suggestion: perhaps consider adding OpenMP pragma directive if the computation is expensive.

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: Yes, it is intentional as the current implementation is not gpu friendly.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: This approach is far from ideal, but has advantages over the "pfree". With "pfree" you have to copy-past all definitions of all arrays into all kernels and they must be in the same order. It is very error prone and pretty much unmaintainable if one need to add or remove arrays from the splits struct. Also it makes usage and dependency analysis very hard. With the current approach one could just grep corresponding ptr_*() and see where exactly it is used. A call to ptr_*() also contains the name of a corresponding allocation, so it is easier to relate to a kernel argument. Moreover, if in future we would want each array to be aligned to a cacheline, it would be an easy change in ptr_*() and get_size*() only. And it would not work that well with "pfree".

As a better solution it should be wrapped into struct/class or all pointers should be defined on in a host side code and passed via independent kernel arguments (it is still could be a single allocation).

jmachado-amd (approved):

library/src/auxiliary/rocauxiliary_stedc.hpp
```diff
@@ -2621,15 +2193,72 @@ rocblas_status rocsolver_stedc_template(rocblas_handle handle,
...
+                            std::vector<S*> hABC(nbb * 3);
+                            for(size_t j = 0; j < nbb; ++j)
+                            {
+                                auto ps = b[j];
+                                hABC[j + 0 * nbb] = ps + ps * ldv + V;
+                                hABC[j + 1 * nbb] = ps + ps * n + ptr_etmpd(n, tempgemm);
+                                hABC[j + 2 * nbb] = ps + ps * n + ptr_vecs(n, tempgemm);
+                            }
+                            HIP_CHECK(hipMemcpy(workArr, hABC.data(), 3 * nbb * sizeof(S*),
+                                                hipMemcpyHostToDevice));
```
jmachado-amd: At some point this `hipMemcpy` will have to be updated into its async version.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: The NAN macro is defined in <cmath>. But agreed, that doesn't look like a standard lib thing, instead more like something that is defined who knows where and means who knows what (and could possibly be redefined in a third place). So using `std::numeric_limits<S>::quiet_NaN()` could be a better option.

I don't want to user "break" or "return" here as it will not make things any faster and the control flow is simplier without it (espesially it could be confusing in a presense of a `__syncthreads()`).

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: The "pfree" approach need not allocate in the same order. It will work if you allocate "msz" first, then allocate "em" or in any arbitrary order.   Even if say "size_em_bytes" is changed, the code requires changing only the value of "size_em_bytes" and the rest of the code will still work and stay the same.  

 This is just a suggestion for future code maintenance.  My concern is if there is explicit calculation of offsets (from same starting "splits"),  then any single change in size (or removal of array)   will require adjustment of all following offsets.  This can be error prone as well, especially if there are many allocations (say there are 50 arrays).      An error in calculating the offsets  may not be obvious just by inspecting the source code.

 The issue of alignment can be easily solved (if this is important) by using a lambda or #define macro for adjusting the size to be  say multiple of 128 bytes or the size of cache line.    Maybe something like
 
 auto ceil = [](auto n, auto b) { return(  (n-1)/b + 1 ); };

 auto adjust_for_alignment = [] ( size_t& size_work ) {     size_work = ceil( size_work, 128 ) * 128; };
 
 adjust_for_alignment( size_em_bytes );
 adjust_for_alignment( size_msz_bytes);  

 The use of "std::byte * pfree" allows allocation of different types, say arrays of "int" and arrays of "double_complex".   When all allocations are the same type  "rocblas_int", one can set "rocblas_int *pfree" as well.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: I don't mind either way. Both looks ok for me. If you think one is better readable than the other, than it is probably the way to go.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: It should be possible to fill the struct with pointers using gpu kernel and eliminate interaction with host and memcpy.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: I think original motivation was to make smaller binary code and that makes sense for me. However, as it is implemented it is not efficient (the USEGEMM=false path). And in general it would be quite hard to make it faster than a properly optimized gemm. I think it could be removed in future and code could be simplified in that place. However, due to a migration between different blas implementations that path could be used for debugging possible gemm related issues. So, we could keep it until everything is settled.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: That would require multiple additional kernel dispatches and storing results of "nan check" in memory somewhere (first check for nans, then call sort, then write nans because they could be overwritten by the sort). Checking nans during a sort is very cheap.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: That is debug related code. I would suggest to consider deleting everything under DEBUG_OUTPUT, maybe with exception for the "old_gemm". But that could be deleted as well if the new code is tested and works well on all important target platforms.

library/src/auxiliary/rocauxiliary_stedc.hpp
Kirpich30000: I would also recommend to format the ptr_*() definition code as in the original branch using one-liners and fully aligned. It is a way more readable and easily to check for possible problems with constants. If the formatting checks complain about it they should be disabled for that part of code.

library/src/auxiliary/rocauxiliary_stedc.hpp
EdDAzevedo: I think if ultimately  there will be two versions of the code, one with USGEMM=true, and one with USEGEMM=false, then I think passing in USEGEMM as function argument might actually lead to total combined smaller binary code sizes.

Kirpich30000:
@tfalders please add important notes to the PR description:

- Eigenvalues are now sorted. They are sorted before deflation to make deflation code faster, and after deflations to group non deflated values together. That allows to simplify and speedup the rest of the code.
- There are redundant data movements of D/ev and vectors (the reason is to localize the changes and structurally mimic original processing to reduce risks of a big refactoring). Optimizing it should provide additional (but probably minor) performance boost and simplify code. But it is best to postpone it to a separate PR
- Performance of the secular eq solver was improved
- Performance of the mergeVectors was significantly improved

PR #1019, opened 2025-09-03
Fix buffer overflow

AGonzales-amd:
Fix buffer overflow in `dot` device function which is causing ASAN test failures.

### DIFF
--- a/library/src/specialized/rocauxiliary_larf_specialized_kernels.hpp
+++ b/library/src/specialized/rocauxiliary_larf_specialized_kernels.hpp
@@ -118,7 +118,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(MAX_THDS)
     const T* tau = tauA + bid * strideP;
 
     // shared variables
-    __shared__ T sval[MAX_THDS];
+    __shared__ T sval[MAX_THDS / WarpSize];
     __shared__ T xs[LARF_SSKER_MAX_DIM];
 
     // load x into shared memory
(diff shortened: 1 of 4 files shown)


EdDAzevedo (approved):

library/src/specialized/rocauxiliary_larf_specialized_kernels.hpp
```diff
@@ -118,7 +118,7 @@ ROCSOLVER_KERNEL void __launch_bounds__(MAX_THDS)
     const T* tau = tauA + bid * strideP;
 
     // shared variables
-    __shared__ T sval[MAX_THDS];
+    __shared__ T sval[MAX_THDS / WarpSize];
```
tfalders, line 121: Is it possible to use warpSize here rather than WarpSize? At some point, I'd like to remove WarpSize from the library.

tfalders (approved):

library/src/specialized/rocauxiliary_larf_specialized_kernels.hpp
AGonzales-amd: warpSize can't be used since a constexpr value is needed to set the size of the static array.

Are there better alternatives to get the warp size in compile time?

EdDAzevedo: How about use dynamic size for __shared__ T sval[]   and pass in the necessary values as kernel and launch arguments?

EdDAzevedo (approved):

library/src/specialized/rocauxiliary_larf_specialized_kernels.hpp
tfalders: > warpSize can't be used since a constexpr value is needed to set the size of the static array.
> 
> Are there better alternatives to get the warp size in compile time?

Ah yes, that makes sense. I think the most obvious alternative is as Ed says, to pass in the shared mem size as a kernel argument. Another option is to divide by 32, and we'll just have some unused space when the warp size is 64.

PR #1020, opened 2025-09-09
Integrate CMake FetchContent module for external dependencies

john00003:
CMake introduced the FetchContent module in CMake Version 3.24 in 2022. This feature provides an alternative to find_package() for finding external dependencies for a project. 

The FetchContent_Declare statement allows you to specify a URL or Github repository to download and build external dependencies from if they are not already present on the machine. Furthermore, if the dependency is already installed on the machine, FetchContent_Declare can be configured to use find_package to search for the dependency before resorting to downloading and building it from source; this is what I have done by adding the FIND_PACKAGE_ARGS keyword to the FetchContent_Declare statement.

I'd like to draw attention to [this](https://github.com/john00003/rocSOLVER/blob/3dcb274e86ce90611b8aa2c5163dcec4b6eedb2b/clients/gtest/CMakeLists.txt#L195), as I am unsure how this might impact Windows builds when GoogleTest is built from source using FetchContent_Declare. Will the directory and .dll still be present? Hopefully CI will clarify.


AGonzales-amd:
Hey John, please help me understand, will this replace the install dependencies option in `install.sh`? Does FetchContent build and install the dependencies?

john00003:
> Hey John, please help me understand, will this replace the install dependencies option in `install.sh`? Does FetchContent build and install the dependencies?

Hey Angelo! I've configured it to first attempt to find an existing installation of the dependency on the machine (using the fetch_package CMake command). Then, if it cannot be found, FetchContent will indeed build dependencies from source.

Like you mentioned, this would mean that you no longer have to use the `./install.sh -d` command to install the external dependencies before building rocSOLVER, you can just let CMake handle this for you.

ammallya:

PR #1021, opened 2025-09-11
Split ideal sizes

john00003:
Modifications to ideal_sizes.hpp is a culprit for triggering large incremental recompilation. This is because the header file is #include'd in many other translation units, where the change to the header file will result in significant recompilation.

This PR makes a few changes:
1) changes the ideal_sizes.hpp file to declare extern const int variables instead of preprocessor macros for all variables that are:
    - not required to be compile time constants
    - not required to be used on the device
2) introduces an ideal_sizes.cpp file to define the const int variables with their values (same as the original values from their previous preprocessor macros)
3) changes library/src/CMakeLists.txt to include this file and link it to the rest of the target
4) refactors a few device kernels (in stedc, sytrf, larf_specialized_kernels) to allow for the host-side variable to be passed as a function argument (otherwise the variable is unusable on the host)

Now, changes to any of the variables in ideal_sizes.cpp only trigger a short recompilation.

Please check out the changes related to 4), as those are the most intrusive changes. Perhaps the refactored function headers could use the host side ideal size variable as a default argument?

These changes are meant to maintain the ability to override these ideal size parameters using CMake compile definitions.


library/src/common/ideal_sizes.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+#ifndef LARFT_SWITCHSIZE
+const int LARFT_SWITCHSIZE = 64;
+#endif
+
+/***************** geqr2/geqrf and geql2/geqlf ********************************
+*******************************************************************************/
+
+#ifndef GEQxF_BLOCKSIZE
+const int GEQxF_BLOCKSIZE = 64;
```
EdDAzevedo, line 70: Minor: Maybe  the block size can depend on the type as well, say to fit in LDS or VGPR or last level cache.

Maybe something like 
#define GEQF_BLOCKSIZE(T)  (  (sizeof(T) == 8) ? 90 : (sizeof(T) == 4) ? 128 : 64 )

EdDAzevedo (approved):

library/src/auxiliary/rocauxiliary_larfg.hpp
```diff
@@ -273,8 +273,11 @@ rocblas_status rocsolver_larfg_template(rocblas_handle handle,
     if(true)
```
EdDAzevedo: Minor: instead of "if (true)"  perhaps give a variable name to this to make the code more clear?  
Say maybe something like

bool constexpr use_small_size_kerne = true;
if (use_small_size_kernel) {
  ...
  }

john00003: Oops! This file was not meant to be changed. This was from reverting Troy's PR in order to build rocSOLVER without the updated rocBLAS changes.

library/src/common/ideal_sizes.cpp
john00003: That sounds good. I'm unfamiliar with this tuning parameter myself, but maybe this could be addressed in a PR after this? I will make note of this to assess in the future.

ammallya:

PR #1025, opened 2025-09-17
Allow install.sh to run on RHEL clones and Debian

qjojo:
This allows install.sh to run on RHEL compatible distros such as Rocky Linux and AlmaLinux, as well as upstream Debian to enable smoother development on those platforms. $ID_LIKE is defined in /etc/os-release and exists for this very purpose. This is safe in the case of Rocky/Alma because those distros are drop in replacements for RHEL, and for Debian because we don't rely on anything Ubuntu specific (our packages are in Debian and we list Debian in the support matrix in the ROCm docs).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -95,21 +95,44 @@ main=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 # This function is helpful for dockerfiles that do not have sudo installed, but the default user is root
 # true is a system command that completes successfully, function returns success
 # prereq: ${ID} must be defined before calling
-supported_distro( )
-{
-  if [ -z ${ID+foo} ]; then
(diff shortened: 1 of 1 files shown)


john00003:
In the repository there are a couple other places where Ubuntu is mentioned, like [here](https://github.com/qjojo/rocSOLVER/blob/b6820dc844d377ba0a126ab45f6bab6cfbcac9d9/CMakeLists.txt#L238) or [here](https://github.com/qjojo/rocSOLVER/blob/b6820dc844d377ba0a126ab45f6bab6cfbcac9d9/scripts/get-clang-format#L37). Do those Ubuntu-specific operations also work on Debian? If not, should we add a switch case to those places for Debian installs?

john00003 (approved):

ammallya:

